PR #346, opened 2022-05-31
Update spack resource environments and flags

douglasjacobsen:
This merge makes several updates to the spack module:
 - environments can be defined with a provided spack.yaml
 - Files that are written (spack.yaml or configs) will not expand
   environment variables anymore
 - Default version of spack is updated to v0.18.0
 - install_flags and concretize_flags can be passed in with
   a toolkit yaml

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/spack-gromacs.yaml
+++ b/community/examples/spack-gromacs.yaml
@@ -51,23 +51,24 @@ deployment_groups:
     settings:
       install_dir: /sw/spack
       spack_url: https://github.com/spack/spack
-      spack_ref: v0.17.1
+      spack_ref: v0.18.0
       log_file: /var/log/spack.log
       configs:
       - type: single-config
         scope: defaults
-        value: "config:build_stage:/sw/spack/spack-stage"
+        content: "config:build_stage:/sw/spack/spack-stage"
       - type: file
         scope: defaults
-        value: |
+        content: |
           modules:
-            tcl:
-              hash_length: 0
-              all:
-                conflict:
-                  - '{name}'
-              projections:
-                all: '{name}/{version}-{compiler.name}-{compiler.version}'
+            default:
+              tcl:
+                hash_length: 0
+                all:
+                  conflict:
+                    - '{name}'
+                projections:
+                  all: '{name}/{version}-{compiler.name}-{compiler.version}'
       compilers:
       - gcc@10.3.0 target=x86_64
       packages:
--- a/community/modules/scripts/spack-install/README.md
+++ b/community/modules/scripts/spack-install/README.md
@@ -36,7 +36,7 @@ see this module used in a full blueprint, see the [spack-gromacs.yaml] example.
     settings:
       install_dir: /sw/spack
       spack_url: https://github.com/spack/spack
-      spack_ref: v0.17.0
+      spack_ref: v0.18.0
       spack_cache_url:
       - mirror_name: 'gcs_cache'
         mirror_url: gs://example-buildcache/linux-centos7
@@ -48,13 +48,14 @@ see this module used in a full blueprint, see the [spack-gromacs.yaml] example.
         scope: defaults
         value: |
           modules:
-            tcl:
-              hash_length: 0
-              all:
-                conflict:
-                  - '{name}'
-              projections:
-                all: '{name}/{version}-{compiler.name}-{compiler.version}'
+            default:
+              tcl:
+                hash_length: 0
+                all:
+                  conflict:
+                    - '{name}'
+                projections:
+                  all: '{name}/{version}-{compiler.name}-{compiler.version}'
       compilers:
       - gcc@10.3.0 target=x86_64
       packages:
@@ -65,6 +66,29 @@ see this module used in a full blueprint, see the [spack-gromacs.yaml] example.
         - intel-mkl%gcc@10.3.0 target=skylake
         - intel-mpi@2018.4.274%gcc@10.3.0 target=skylake
         - fftw%intel@18.0.5 target=skylake ^intel-mpi@2018.4.274%intel@18.0.5 target=x86_64
+      - name: explicit-env
+        content: |
+          spack:
+            definitions:
+            - compilers:
+              - gcc@10.3.0
+            - mpis:
+              - intel-mpi@2018.4.274
+            - packages:
+              - intel-mkl
+            - mpi_packages:
+              - fftw
+            specs:
+            - matrix:
+              - - $packages
+              - - $%compilers
+            - matrix:
+              - - $mpis
+              - - $%compilers
+            - matrix:
+              - - $mpi_packages
+              - - $%compilers
+              - - $^mpis
 ```
 
 Following the above description of this module, it can be added to a Slurm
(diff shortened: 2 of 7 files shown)


community/modules/scripts/spack-install/README.md
```diff
@@ -48,23 +48,49 @@ see this module used in a full blueprint, see the [spack-gromacs.yaml] example.
...
       environments:
       - name: main-env
+        type: packages
         packages:
         - intel-mkl%gcc@10.3.0 target=skylake
         - intel-mpi@2018.4.274%gcc@10.3.0 target=skylake
         - fftw%intel@18.0.5 target=skylake ^intel-mpi@2018.4.274%intel@18.0.5 target=x86_64
+      - name: explicit-env
+        type: file
+        value: |
```
heyealex, line 72: Just an idea, let me know what you think:
We could have an environments directory in the spack module and have this "explicit env" point to that. That way, we can have the files run through the yaml linter and decrease the size of this example blueprint. We'd also be able to grow our example environments more naturally, and users could swap or add only a few lines in the example to build a different stack.

That would be a follow up of course, no action needed here.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -178,11 +190,32 @@ EOT
 
 variable "environments" {
   description = "Defines a spack environment to configure."
```
heyealex, line 192: Given this has gotten more complicated, the description should include more information on expected keys and values.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -178,11 +190,32 @@ EOT
...
+  }
+  validation {
+    condition = alltrue([
+      for e in var.environments : ((e["type"] == "packages" && contains(keys(e), "packages")) || e["type"] == "file")
+    ])
+    error_message = "Environments of type 'packages' require a packages list."
+  }
+  validation {
+    condition = alltrue([
+      for e in var.environments : (e["type"] == "packages" || (e["type"] == "file" && contains(keys(e), "value")))
```
heyealex, line 215: To keep this consistent with startup scripts, value should be replaced with "content"

community/modules/scripts/spack-install/templates/install_spack.tpl
```diff
@@ -87,26 +87,39 @@ spack compiler find --scope site >> ${LOG_FILE} 2>&1
...
+      rm -f ${INSTALL_DIR}/spack_env.yaml
+      spack env activate ${e.name}
+    } &>> ${LOG_FILE} 2>&1
+  %{endif ~}
 
-  echo "$PREFIX    Configuring spack environment ${e.name}"
-  %{for p in e.packages ~}
-    spack add ${p} >> ${LOG_FILE} 2>&1
-  %{endfor ~}
+  %{if e.type == "packages" ~}
```
heyealex, line 107: There should be a [test config](https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/tools/validate_configs/test_configs) to cover a lot of this so that at least the basic functionality is being verified in `make tests`. There is already a test config for [spack-cache](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/tools/validate_configs/test_configs/spack-buildcache.yaml), it might make sense to rename that and add different variants of environments to that file.

community/modules/scripts/spack-install/README.md
douglasjacobsen: Sure, so in this case you mean have something like an external-file option, or something. That would point to an external environment definition, right? I think that could be a good idea, and shouldn't be too hard to add next.

community/modules/scripts/spack-install/templates/install_spack.tpl
douglasjacobsen: I was working on adding this, and apparently you can't have `list(map(any))` where the map contains keys that are of type strings and lists.

So, this validate fails on the current version of environments, where they take a list of packages. Given this is already in the code, I didn't remove it. I figured after you're back we can discuss this aspect of it. For now, I added a test config for the new environment format.

community/modules/scripts/spack-install/variables.tf
douglasjacobsen: Thanks, I changed this, and how it's defined in the configs, to help align them with startup scripts.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -177,12 +189,41 @@ EOT
...
-    packages = list(string)
-  }))
+  description = <<EOT
+  Defines spack environments to configure, given as a list.
+  Each environment must define a name, and type.
+  Valid types for environments are 'file', and 'packages'.
+  Environments of type 'packages' must define a list named packages, containing
+  all of the packages to install in that environment.
+  Environments of type 'file' must define a string named 'content' containing the
+  contents of a spack.yaml file to create the environment from.
```
tpdownes, line 199: nit: although it ends up in `spack.yaml` I would rather say something like "contents of Spack YAML environment file" which adopts the headers you'd find in their Read The Docs

community/modules/scripts/spack-install/README.md
tpdownes: since this is discussing future work, we can track it separately.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -177,12 +189,41 @@ EOT
...
 variable "environments" {
-  description = "Defines a spack environment to configure."
-  default     = null
-  type = list(object({
-    name     = string
-    packages = list(string)
-  }))
+  description = <<EOT
+  Defines spack environments to configure, given as a list.
+  Each environment must define a name, and type.
```
tpdownes, line 194: Each environment must define fields "name" and "type" as described below.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -177,12 +189,41 @@ EOT
...
+  Valid types for environments are 'file', and 'packages'.
+  Environments of type 'packages' must define a list named packages, containing
+  all of the packages to install in that environment.
+  Environments of type 'file' must define a string named 'content' containing the
```
tpdownes, line 198: Environments of type 'file' must define a field named 'content' that contains...

community/examples/spack-gromacs.yaml
```diff
@@ -51,23 +51,24 @@ deployment_groups:
...
+        content: |
           modules:
-            tcl:
-              hash_length: 0
-              all:
-                conflict:
-                  - '{name}'
-              projections:
-                all: '{name}/{version}-{compiler.name}-{compiler.version}'
+            default:
```
tpdownes, line 64: Marking this down to point out duplicate instance of issue noted in README.

tpdownes (commented):
Please respond to comments about "breaking" behavior. Make changes in the descriptions as requested.

community/modules/scripts/spack-install/README.md
tpdownes, line 72: I think this may need to be updated to `content`

tpdownes:
After offline discussion, the path I recommend forward for least YAML-breaking changes is to break out the new environment-from-a-file feature as a new setting `environment_files` with behavior like:

- 3 fields: name, content, source. content and source should be mutually exclusive

Then `var.environments` can be left entirely alone such that existing configurations require no updates.

cboneti:
Having the ability to have a simple environment in a file is very valuable. I would not like to loose that. I also understand we need the ability to point to an external file when the environment gets more complex. Let's just make sure both directions are well supported.

douglasjacobsen:
@cboneti Just so you know, this PR was not intended to add external files. Personally, I find that functionality less useful than having it in the YAML (which is what this PR was intended to add). @heyealex mentioned he'd like to see the external functionality added later, and @tpdownes mentioned he'd rather see it now, with (effectively) a rework of the functionality that this PR was intended to add.

However, when writing this PR I realized that the existing environment functionality (a list of packages) is not really useful, and should be deprecated instead for the explicit syntax (since you can always put a list of packages in that syntax). In addition to it's utility, the testing suite fails to validate the existing schema, since (apparently) terraform does not allow you to have maps where the keys of the map are of different types (like the existing environment definitions).

Personally, I'd rather the name of the top level environment key remain the same, rather than naming it `environment_files` because in reality you are creating environments (though they can be seeded with files).

I don't currently have time to do the re-implementation and testing, so this PR will wait until @heyealex and I are both back, and can revisit the best way to proceed.

community/modules/scripts/spack-install/variables.tf
```diff
@@ -177,12 +189,38 @@ EOT
...
+    error_message = "All environments must have a name."
+  }
+
+  validation {
+    condition = alltrue([
+      for e in var.environments : (contains(keys(e), "packages") ? alltrue(([for p in e["packages"] : alltrue([tostring(p) == p])])) : true)
+    ])
+    error_message = "The packages attribute within environments is required to be a list of strings."
+  }
+
```
heyealex, line 216: nit: double space between validations

community/modules/scripts/spack-install/templates/.shellcheckrc
```diff
@@ -21,3 +21,8 @@ disable=SC2034
 # Disabled because string is not literal, variable is prefaced with $$ for template
 # https://github.com/koalaman/shellcheck/wiki/SC2157
 disable=SC2157
+
+
+# Disabled because parenthesis are used in terraform code
+# https://github.com/koalaman/shellcheck/wiki/SC1088
```
heyealex, line 27: This link needs to be updated: SC1088 -> SC1036

heyealex (approved):

PR #347, opened 2022-05-31
Upgrade to yaml.v3

nick-stroud:
The `UnmarshalStrict` api is no longer available in v3 but the same functionality is supported using a `Decoder` with `KnownFields(true)` set. 

Changes from `map[interface{}]interface{}` to `map[string]interface{}` were required to avoid panic. It seems that the decoder is providing more explicit type information for deeply nested values. 

New behavior on marshaling introduces additional white space in listed items (starting with `-`). This means that `expanded.yaml` files generated using `expand` have different formatting after this change. There is a good summary [here](https://github.com/kubernetes-sigs/yaml/issues/72).

Tested:
- Ran expand before and after on `community/examples/spack-gromacs.yaml` and observed diff (whitespace changes only)
- Tested unmarshal with unknown fields (strict behavior). Fails as expected.
- Tested unmarshal with missing fields. Fails as expected.

### Submission Checklist

  install`)


tpdownes (changes requested):
When I run `ghcp expand` on the PR branch and use the spack-gromacs example, I get output that looks like:

```yaml
blueprint_name: spack-gromacs
validators:
    - validator: test_project_exists
      inputs:
        project_id: ((var.project_id))
...
vars:
    deployment_name: spack-gromacs
    labels:
        ghpc_blueprint: spack-gromacs
        ghpc_deployment: spack-gromacs
    project_id: toolkit-demo-zero-e913
    region: us-central1
    zone: us-central1-c
```

Lists are indented by 4 characters and, for some reason, the labels map is also. Other maps are indented by 2 characters.

The status quo looks like this

```yaml
blueprint_name: spack-gromacs
validators:
- validator: test_project_exists
  inputs:
    project_id: ((var.project_id))
- validator: test_region_exists
  inputs:
    project_id: ((var.project_id))
    region: ((var.region))
- validator: test_zone_exists
  inputs:
    project_id: ((var.project_id))
    zone: ((var.zone))
- validator: test_zone_in_region
  inputs:
    project_id: ((var.project_id))
    region: ((var.region))
    zone: ((var.zone))
vars:
  deployment_name: spack-gromacs
  labels:
    ghpc_blueprint: spack-gromacs
    ghpc_deployment: spack-gromacs
  project_id: toolkit-demo-zero-e913
  region: us-central1
  zone: us-central1-c
  ```

I believe a few paths are reasonable from here:

1. `ghpc expand` should produce a "diff-less" YAML file compared to status quo
2. `ghpc expand` should produce something that passes our linting tests
3. we should alter the behavior of `ghpc expand` to something we can defend (and linting tests should be modified to match)

**Comment by tpdownes** [comment]
Since the full checks passed, I don't think they need to be re-run if you address the "superficial" concerns in the review.

**Review by tpdownes — DISMISSED** [review]
The final result passes `yamllint` in its default configuration, rather than ours which mimics common patterns seen in the wild. I am happy enough with that outcome subject to all the upstream library problems you note.

File: pkg/config/config_test.go [hunk 1]
```diff
@@ -319,23 +319,38 @@ func (s *MySuite) TestNewBlueprint(c *C) {
 	dc := getDeploymentConfigForTest()
 	outFile := filepath.Join(tmpTestDir, "out_TestNewBlueprint.yaml")
 	dc.ExportBlueprint(outFile)
-	newDC := NewDeploymentConfig(outFile)
+	newDC, _ := NewDeploymentConfig(outFile)
```
heyealex, line 322: You should add a `c.Assert(err, IsNil)` here and anywhere else that's now returning an error.

pkg/config/config_test.go
```diff
@@ -319,23 +319,38 @@ func (s *MySuite) TestNewBlueprint(c *C) {
...
-		len(obtainedBlueprint.Vars["labels"].(map[interface{}]interface{})),
+		len(obtainedBlueprint.Vars["labels"].(map[string]interface{})),
 		Equals,
 		len(expectedSimpleBlueprint.Vars["labels"].(map[string]interface{})),
 	)
 	c.Assert(obtainedBlueprint.DeploymentGroups[0].Modules[0].ID,
 		Equals, expectedSimpleBlueprint.DeploymentGroups[0].Modules[0].ID)
 }
 
+func (s *MySuite) TestImportBlueprintExtraFieldThrowsError(c *C) {
```
heyealex, line 339: nit: Naming of variations of a test function should be separated by an underscore, as it's done elsewhere:
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/pkg/modulereader/resreader_test.go#L134

So this should be `TestImportBlueprint_ExtraFieldThrowsError`

nick-stroud: Good call. I actually went with `TestImportBlueprint_ExtraField_ThrowsError` in the pattern of `<method>_<state>_<outcome>`

tpdownes (approved):

PR #349, opened 2022-06-02
Add support for Slurm to `use` the `startup_script` module

nick-stroud:
Allows for passing of startup_script to Slurm V4 using the `use` keyword. Modules only fall back to this behavior if the other startup script variables are not explicitly set (ex: `compute_startup_script`)

The controller module takes startup scripts for both compute and controller. I elected to not fallback to the `startup_script` variable unless neither were set. This is to avoid confusion where startup_script is being only partially overwritten. 

Tested:
- Ran `ghpc create community/examples/spack-gromacs.yaml --vars project_id=test` and verified in generated terraform that in `main.tf`, `slurm_login` contained `startup_script  = module.spack-startup.startup_script`.
- Ran integration tests that contains spack-growmacs example.

### Submission Checklist

  install`)


tpdownes (approved):
Please wait for the full integration checks to go through.

PR #352, opened 2022-06-02
Support extra args for ansible playbooks

tpdownes:
A common pattern in Ansible is to supply the `--extra-vars` argument or other arguments to `ansible-playbook`. See, for example, our [own invocation in integration tests](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/b0a5f6f1ef6298ccda812e9c332be7a195e1f117/tools/cloud-build/daily-tests/integration-group-1.yaml#L61-L62).

We do not presently support this in Runners. This PR adds that support. It will be useful for use cases such as rendering a Terraform value into an Ansible variable:

```hcl
  role_example = {
    "type"        = "ansible-local"
    "content"     = file("${path.module}/files/example_role.yml")
    "destination" = "example_role.yml"
    "args"        = "-e \"password_id=${google_secret_manager_secret.pool_password.secret_id}\""
  }
```

### Submission Checklist

  install`)

### DIFF
--- a/modules/scripts/startup-script/templates/startup-script-custom.tpl
+++ b/modules/scripts/startup-script/templates/startup-script-custom.tpl
@@ -6,7 +6,7 @@ stdlib::run_playbook() {
     "Please install ansible before running ansible-local runners."
     exit 1
   fi
-  /usr/bin/ansible-playbook --connection=local --inventory=localhost, --limit localhost $1
+  /usr/bin/ansible-playbook --connection=local --inventory=localhost, --limit localhost $1 $2
 }
 
 stdlib::runner() {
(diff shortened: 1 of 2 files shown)


tpdownes:
Note that all build checks have passed.

modules/scripts/startup-script/templates/startup-script-custom.tpl
```diff
@@ -6,7 +6,7 @@ stdlib::run_playbook() {
     "Please install ansible before running ansible-local runners."
     exit 1
   fi
-  /usr/bin/ansible-playbook --connection=local --inventory=localhost, --limit localhost $1
+  /usr/bin/ansible-playbook --connection=local --inventory=localhost, --limit localhost $1 $2
```
heyealex, line 9: I did some testing, and I believe this will allow further `--limit ...` flags, which would allow this to run non-locally. Can a validation be added to ensure `--connection=.*`, `--inventory=.*` or `--limit .*` are not included in args for ansible?

I'm also open to allowing that since it's already possible to do that via a shell script as is.

tpdownes: How do you feel about addressing this in README for now? I see a note about args only applying to `shell` that needs to be updated anyways.

heyealex: Sure, that seems reasonable.

tpdownes: (I will squash these commits upon merge, FYI)

heyealex (approved):

PR #355, opened 2022-06-02
Improve use of service account scopes in Packer module

tpdownes:
The default scopes granted to the service account that is attached to the short-lived Packer VM are:

```hcl
[
  "https://www.googleapis.com/auth/userinfo.email",
  "https://www.googleapis.com/auth/compute",
  "https://www.googleapis.com/auth/devstorage.full_control"
]
```

This PR eliminates an accidentally unused input variable and set a new default:

```hcl
[
    "https://www.googleapis.com/auth/userinfo.email",
    "https://www.googleapis.com/auth/cloud-platform.read-only",
    "https://www.googleapis.com/auth/logging.write",
  ]
```

Reasoning:

1. `devstorage.full_control` is an awful lot
2. Broad read-only access (limited by IAM bindings of the service account) seems reasonable
3. Being able to write to Cloud Logging is _highly_ valuable for debugging because the startup-script stdout is not printed to the console running Packer while it is typically written to Cloud Logging

### Submission Checklist

  install`)


cboneti:
I would also recommend documenting the scopes (if not already done), and documenting how one could use Cloud Logging to debug the progress of the startup script (if not already done)

modules/packer/custom-image/variables.pkr.hcl
```diff
@@ -98,13 +98,18 @@ variable "service_account_email" {
   default     = null
 }
 
-variable "service_account_scopes" {
+variable "scopes" {
```
nick-stroud, line 101: This is a breaking change. I assume you changed the name so that it would match the google compute. I earnest I don't think we can make this change without rolling version number to 2.0 on our next release. My suggestion would be to leave the name what it was, even though there is a mismatch. Mismatch aside it is still a reasonable name for the variable. 

If you think it is critical to have symmetry with the `googlecompute` API then let's talk change management offline.

tpdownes: Although I realize you are focusing on the variable name, I think one could reasonably describing the set of choices as a "fixing" change since the status quo is to do nothing with the supplied values. If there were a good Packer equivalent to `tflint` this would have been caught during its introduction. We can discuss further offline to handle this.

nick-stroud: I agree with you in this specific case as it is impossible that users could be negatively impacted by this change as it was never working to start.

nick-stroud (approved):
LGTM. I will let you decide how you would like to address the other comment.

tpdownes:
@nick-stroud while addressing https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/355#issuecomment-1145810455 in the README I decided to also add `var.labels` (entirely new so not breaking) which will integrate Packer partially into Toolkit label management.

I appreciate a new scan.

nick-stroud (approved):

PR #361, opened 2022-06-03
Add HTCondor Pool blueprint (experimental)

tpdownes:
This PR is dependent upon #359 and #360 and validation tests will fail until they are accepted and this branch is rebased. Nevertheless, the example itself is instructive for how #359 and #360 will be used.

### Submission Checklist

  install`)

### DIFF
--- a/tools/validate_configs/test_configs/htcondor-pool.yaml
+++ b/tools/validate_configs/test_configs/htcondor-pool.yaml
@@ -1,0 +1,129 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+blueprint_name: htcondor-pool
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: htcondor-001
+  region: us-central1
+  zone: us-central1-c
+  htcondor_image_family: htcondor-9x
+  network_name: htcondor-pool
+  subnetwork_name: htcondor-pool-usc1
+
+deployment_groups:
+- group: htcondor-env
+  modules:
+  - source: modules/network/vpc
+    kind: terraform
+    id: network1
+    outputs:
+    - network_name
+
+  - source: community/modules/scripts/htcondor-install
+    kind: terraform
+    id: htcondor_install
+
+  - source: community/modules/project/service-enablement
+    kind: terraform
+    id: htcondor_services
+    use:
+    - htcondor_install
+
+  - source: modules/scripts/startup-script
+    kind: terraform
+    id: htcondor_install_scripts
+    settings:
+      runners:
+      - type: shell
+        source: modules/startup-script/examples/install_ansible.sh
+        destination: install_ansible.sh
+      - $(htcondor_install.install_htcondor_runner)
+    outputs:
+    - startup_script
+
+- group: packer
+  modules:
+  - source: modules/packer/custom-image
+    kind: packer
+    id: custom-image
+    settings:
+      image_family: $(vars.htcondor_image_family)
+
+- group: pool
+  modules:
+  - source: modules/network/pre-existing-vpc
+    kind: terraform
+    id: cluster_network
+
+  - source: community/modules/scheduler/htcondor-configure
+    kind: terraform
+    id: htcondor_configure
+
+  - source: modules/scripts/startup-script
+    kind: terraform
+    id: htcondor_configure_central_manager
+    settings:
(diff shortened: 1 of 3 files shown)


tpdownes:
@cboneti this should be reviewed prior to #362, however, you can look at the two PRs together on this branch: 

https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/htcondor

tools/validate_configs/test_configs/htcondor-pool.yaml
```diff
@@ -0,0 +1,117 @@
...
+    id: htcondor_install_scripts
+    settings:
+      runners:
+      - type: shell
+        source: modules/startup-script/examples/install_ansible.sh
+        destination: install_ansible.sh
+      - $(htcondor_install.install_htcondor_runner)
+    outputs:
+    - startup_script
+- group: packer
```
nick-stroud, line 54: nit: I prefer to put whitespace between the groups, similar to what is in image-builder.yaml.

tools/validate_configs/test_configs/htcondor-pool.yaml
```diff
@@ -0,0 +1,117 @@
...
+  modules:
+  - source: modules/packer/custom-image
```
nick-stroud, line 56: My understanding: you are building htcondor install into a custom image. Is this required or just an optimization. It might make the example considerably clearer to just do the htcondor install on boot even if it is a little slower, and then we can always point people to the packer example if they want to optimize. 

Especially since the example does not actually work without the special passing of startup script between group 1 and 2, which you probably don't want to document again.

nick-stroud, line 56: My understanding is that this will use the default hpc image as a starting point and then will spit out an image named htcondor-9x after build. I do not like the implicit passing of image_family to this module and would prefer we do it similarly to how it is done in image-builder.yaml, by explicitly passing the output image name (leaving input as implicit default is fine though).

tools/validate_configs/test_configs/htcondor-pool.yaml
tpdownes: I do not prefer the degree to which YAML allows for empty lines, but I will match the rest of our examples in style.

tools/validate_configs/test_configs/htcondor-pool.yaml
tpdownes: This is a test configuration, not an example. Its intent is to be the basis for an integration test that builds an image and provisions an HTCondor pool from it.

I agree that a user-facing example faces the challenge you describe. When full HTCondor functionality is ready, I will produce a suitably simple example.

tpdownes: `image-builder.yaml` doesn't do precisely what you describe. The produced image name is never used (or output) anywhere. However, the example it does not rely on the invisible propagation of `image_family` in several locations as this does. I think what you are proposing is to name the global variable anything except `image_family` and explicitly pass it?

tools/validate_configs/test_configs/htcondor-pool.yaml
tpdownes: Done

tools/validate_configs/test_configs/htcondor-pool.yaml
nick-stroud: To clarify _"by explicitly passing the output image name"_, I should have said:

> by explicitly passing the output image name **to the packer module**. 
---
> I think what you are proposing is to name the global variable anything except image_family and explicitly pass it?

Yes, exactly this.

nick-stroud: 🤷‍♂️ It seems like the eventual integration test will also be simpler if it were laid out as I describe, but your call. Feel free to close.

tools/validate_configs/test_configs/htcondor-pool.yaml
nick-stroud: You did go a step farther than I was asking. The request was to put white space between the 3 groups, not every module, which would match how image-builder.yaml is. But this is a nit so feel free to close as you see fit.

tpdownes: I decided to interpret the nit as "match the style of files under community/examples" since an example like this might eventually land there. I don't like how wild west YAML can be with respect to whitespace.

tools/validate_configs/test_configs/htcondor-pool.yaml
tpdownes: I think building an image tests a production-like deployment and is less prone to transient failures (yum install must succeed only once vs O(N) times).

tpdownes: I also meant to add that your comment made me take a look back at the `htcondor-install` module. I need to double check that it is idempotent with respect to firewall rules and multiple boot cycles. I consider that a separate task.

nick-stroud: In this case consider toolkit developers the user. If it make it clearer to read for people debugging the test there is still value in that.

nick-stroud (approved):

tools/validate_configs/test_configs/htcondor-pool.yaml
tpdownes: Task created

PR #364, opened 2022-06-10
Reliably detect when nodes fail to be scaled in

tpdownes:
This PR makes the following changes:

- for the high IO example, reduce the `suspend_time` to 60 seconds to more rapidly scale in nodes
- for all integration tests that rely on `test-mounts-and-partitions.yml` (all Slurm tests) ensure that all nodes enter the `POWERED_DOWN` state within 10 minutes of the last job submission; if they do not, the integration test fails

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
@@ -15,29 +15,42 @@
 ---
 
 - name: Get partition info
-  command: sinfo --format='%P' --noheader
+  ansible.builtin.command: sinfo --format='%P' --noheader
+  changed_when: False
   register: partition_output
+- name: Clount Slurm nodes
+  ansible.builtin.shell:
+    sinfo -t 'IDLE&POWERED_DOWN' --noheader --format "%n"
+  args:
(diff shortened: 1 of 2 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
```diff
@@ -15,29 +15,42 @@
 ---
 
 - name: Get partition info
-  command: sinfo --format='%P' --noheader
+  ansible.builtin.command: sinfo --format='%P' --noheader
+  changed_when: False
```
cboneti, line 19: just out of curiosity, are you defining changed_when: false to make sure this is always executed?

cboneti (approved):
This looks great. Thanks!

tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
tpdownes: I am defining `changed_when` to False to indicate that, in fact, this command never changes anything. In the output of Ansible playbook it will always report OK. I do not believe that this impacts the default value of `failed_when`.

PR #365, opened 2022-06-10
Make Slurm integration tests more strict

tpdownes:
This modification will make all Slurm integration tests fail if "ERROR:"
appears in the autoscaler logs (resume.log or suspend.log)

### Submission Checklist

  install`)


cboneti (commented):
This appears to be handled after cleanup, which is good. 
Is this able to gracefully handle errors that occur before the resume_output or suspend_output are retrieved? (and so these variables do not exist?) What is the behavior for early failures?

cboneti (approved):
Since 1) this is an important improvement and 2) the potential problem is not severe and somewhat hypothetical, let's merge this.

PR #366, opened 2022-06-13
Fix rare failure modes of monitoring test

tpdownes:
Address 3 intermittent issues with integration test 3:

- timeout on startup-script completion (change 5 minutes to 10 minutes)
- failure to create an Ansible tmp directory under /home when /home is mounted over NFS during startup-script
- rare occasions when ghpc binary is over-written during execution because of multiple invocations of `create_deployment.sh` in a single build

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/create_deployment.sh
+++ b/tools/cloud-build/daily-tests/create_deployment.sh
@@ -20,6 +20,8 @@ BLUEPRINT_DIR=${BLUEPRINT_DIR:-blueprint}
 DEPLOYMENT_NAME=${DEPLOYMENT_NAME:-missing-deployment-name}
 NETWORK=${NETWORK:-missing-network-name}
 MAX_NODES=${MAX_NODES:-2}
+ALWAYS_RECOMPILE=${ALWAYS_RECOMPILE:-yes}
+
 echo "Creating blueprint from ${EXAMPLE_YAML} in project ${PROJECT}"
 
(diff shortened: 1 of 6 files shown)


tools/cloud-build/daily-tests/create_deployment.sh
```diff
@@ -39,7 +41,12 @@ cd "$ROOT_DIR" ||
 		echo "*** ERROR: failed to access root directory ${ROOT_DIR} when creating blueprint"
 		exit 1
 	}
-make
+
+if [[ $ALWAYS_RECOMPILE != "no" || ! -f ghpc ]]; then
```
nick-stroud, line 45: nit: is it possible to pass `$ALWAYS_RECOMPILE` as a bool instead of a string?

nick-stroud (approved):

tools/cloud-build/daily-tests/create_deployment.sh
tpdownes: I gave thought to using a bash boolean but didn't want to spend time handling edge cases like invocation at shell (which shell?) and conversion of YAML value in playbook to bash environment.

tpdownes:
For posterity: the Group 1 check failed for a reason unrelated to this commit (Slurm appears not to have started up before `sinfo` command is invoked to list partitions)

nick-stroud (approved):

PR #378, opened 2022-06-28
Add slurm-gcp v5 controller module

heyealex:
### Submission Checklist

  install`)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
```diff
@@ -0,0 +1,92 @@
...
+  ghpc_startup_script_compute = [{
+    filename = "ghpc_startup.sh"
+    content  = var.compute_startup_script
+  }]
+}
+
+module "slurm_controller_instance" {
+  source = "github.com/SchedMD/slurm-gcp.git//terraform/slurm_cluster/modules/slurm_controller_instance?ref=v5.0.2"
+
+  access_config                = var.access_config
```
tpdownes, line 31: I realize that there is some value in clustering alike configuration settings here (and below). But, at same time, I think it might make sense just to go fully alphabetical.

I am trying to look for duplicate settings in the template and in the instance and understand if duplication is a problem.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+output "controller_instance_id" {
+  description = "Self link of the controller node"
+  value       = module.slurm_controller_instance.slurm_controller_instance.instances_details[0].id
+}
+
+output "controller_instance" {
+  description = "Self link of the controller node"
```
tpdownes, line 23: The description is repeated

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "controller_instance_id" {
+  description = "Self link of the controller node"
+  value       = module.slurm_controller_instance.slurm_controller_instance.instances_details[0].id
```
tpdownes, line 19: I do not believe that this is, in fact, the self_link of the instance.

https://github.com/SchedMD/slurm-gcp/blob/master/terraform/slurm_cluster/modules/_slurm_instance/outputs.tf#L23

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
```diff
@@ -0,0 +1,488 @@
...
+  type        = string
+  description = "Project ID to create resources in."
+}
+
+variable "slurm_cluster_name" {
+  type        = string
+  description = "Cluster name, used for resource naming and slurm accounting."
+
+  validation {
+    condition     = can(regex("(^[a-z][a-z0-9]*$)", var.slurm_cluster_name))
```
tpdownes, line 33: it seems to me that if you are going to do any validation, that you should match the [exact regex used by SchedMD](https://github.com/SchedMD/slurm-gcp/blob/master/terraform/slurm_cluster/modules/slurm_partition/variables.tf#L26). I find it a bit odd that they do not allow dashes, but it's fine and they are imposing a 10-character limit on the whole string that you are not.

You should be aware that Terraform will perform the regex check in the wrapper and in the other module so the value here isn't enormous relative to the potential for a conflict.

I also would have the error message say something in "English" rather than "regex". e.g. "the cluster name can be no longer than 10 lowercase alphanumeric characters and must begin with a letter."

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/versions.tf
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+terraform {
+  required_version = ">= 0.14.0"
```
tpdownes, line 18: - Did you pick 0.14 for a reason?

I'm not sure whether we should be adding a provider and/or provider_meta block here given that there are 0 resources being directly managed in this module.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -0,0 +1,141 @@
...
+  - network1
+  - homefs
+  - compute_partition
+```
+
+This creates a controller node connected to the primary subnetwork of
+`network1`. The controller will also have the `homefs` file system
+mounted and manage one partition, both declared in the `use`
+field. For more context see the
+[hpc-cluster-slurm-vt example](../../../examples/hpc-cluster-slurm-v5.yaml).
```
tpdownes, line 35: This doesn't yet exist in this PR and just want to note that as an issue to track if the filename changes. I'm not really sure how I feel about an example "snippet" w/o any settings at all but I'll give it thought.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -0,0 +1,141 @@
...
+[slurm\_controller\_instance] and [slurm\_instance\_template] modules.
+
+More information about Slurm On GCP can be found at the
+[project's GitHub page][SchedMD/slurm-gcp] and in the
+[Slurm on Google Cloud User Guide][slurm-ug].
+
+The [user guide][slurm-ug] provides detailed instructions on customizing and
+enhancing the Slurm on GCP cluster as well as recommendations on configuring the
+controller for optimal performance at different scales.
+
```
tpdownes, line 13: I believe there should be something here in **bold** text indicating the requirement to install Python and several dependencies in the working environment if enabling `var.x` or `var.y` (replace with links to the settings in the README). You should not replicate the instructions but link to them. If they do not exist (or are not accurate), write what we believe and document lack of instructions as an issue.

tpdownes (changes requested):
This is a first pass at the "optics". I plan on doing a more technical test later.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
tpdownes: I think you might consider output-ing the ID and self-link.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/versions.tf
tpdownes: @nick-stroud I appreciate your input on this question. I think the "right" thing to do is not to have any `provider` block whatsoever. In fact, I think that adding it would cause TFLint to fail. What I don't know one way/other is if one can add `provider_meta` in the absence of `provider`. I bet TFLint wouldn't care but Terraform itself might.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
heyealex: Updated both description, thanks for catching that.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
heyealex: It looks like I may not need the controller module to be output after all. It was needed before to specify the dependency, but it looks like that was fixed in recent updates.
I'll run some tests and if it works without adding the explicit `depends_on`, I'll convert this to self_link.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
heyealex: That's a good point, I removed the validation for this and other passthrough variables to avoid duplication of effort (both for us and for terraform).

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Removed, and added a note to add it back in when I submit the example PR. Thanks for catching this.

Also, I added a setting for machine_type. Nothing is required if the network and partition are provided via `use`, but it doesn't hurt to show how it can be customized otherwise.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Added a warning in the README. Take a look and let me know if it's sufficiently clear.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
heyealex: The module output is no longer needed, I removed it. In the process I also fixed a bug related to filestore, so I'll be running a larger set of tests to ensure that didn't break anything else.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/versions.tf
nick-stroud: provider_meta is only needed when a google (or google beta) provider is called out, and those are only needed when there is a resource or data block using the provider, so not here. 

I am not sure if this file is still needed/ if the `required_version` is needed if everything is a submodule. You might still want the `required_version` called out if there are features being used in the locals block that are not available before a certain point. 

Summary, `required_providers ` and `provider_meta` not needed. Beyond that if tflint doesn't complain, then I won't.

tpdownes: This module is so simple I think we should probably lower the version to 0.13 but leave `required_version` in. (`for_each` and `count` for modules was introduced in 0.13 even if we somehow end up using them)

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -0,0 +1,152 @@
...
+> **_WARNING:_** The variables [enable\_cleanup\_compute] and
+> [enable\_cleanup\_subscriptions], if set to true, require addition
```
tpdownes, line 15: "additional"

tpdownes: The links to variables don't work. I think you should use `[var.enable_cleaunp_compute](#input_enable_cleanup_compute)` or similar

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/versions.tf
heyealex: I removed it and all the tests still passed. Fixed in the last pushed commit.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Fixed

tpdownes (approved):

PR #380, opened 2022-06-28
Extend periodic cleanup to reset Filestore API

tpdownes:
Cloud Filestore has a limit on the number of internal networks that can be created to peer the Filestore managed service to the user's VPCs. One can avoid this problem by reusing names of Filestore instances (undesirable if one wants to run builds in parallel) or periodically resetting the Filestore API. This commit implements the reset.

https://cloud.google.com/filestore/docs/troubleshooting#system_limit_for_internal_resources_has_been_reached_error_when_creating_an_instance

I do not believe that either of the `elif` blocks should constitute failure of the script. It's OK for them to exit that way, but if we are regularly exiting via those blocks we should consider scheduling the cleanup at a different time of day.

### Submission Checklist

  install`)

### DIFF
--- a/tools/clean-filestore-limit.sh
+++ b/tools/clean-filestore-limit.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


cboneti (approved):

tpdownes:
@cboneti I see an issue that may cause this to fail in the real environment (vs my shell testing). Please remove approval for now.

tpdownes:
Sorry for the noise - I updated all `gcloud` commands to use the same PROJECT_ID inherited by same mechanism as `clean-resource-policies.sh`.

nick-stroud (approved):

tpdownes:
@nick-stroud the checks got all confused and refused to run. I tricked them by force-pushing the same branch with new commit history. No changes.

nick-stroud (approved):

PR #381, opened 2022-06-29
Modules/slurm gcp v5 partition

heyealex:
### Description
Adds a module for the slurm-gcp partition that can be used with the slurm-gcp controller ([PR#378](https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/378))

To test, a test_config has been added. This will create a cluster similar to examples/hpc-cluster-small.yaml, but without a login node and with the v5 modules.

### Submission Checklist

  install`)


community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
```diff
@@ -0,0 +1,298 @@
...
+variable "source_image_family" {
+  type        = string
+  description = "Source image family."
+  default     = ""
+}
+
+variable "source_image" {
+  type        = string
+  description = "Source disk image."
+  default     = ""
```
nick-stroud, line 107: I am confused by the image variable defaults. I would have expected to see this pointing to the slurm image by default.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
```diff
@@ -0,0 +1,298 @@
...
+  default     = 10
+}
+
+variable "source_image_project" {
+  type        = string
+  description = "Project where the source image comes from. If it is not provided, the provider project is used."
+  default     = ""
+}
+
+variable "source_image_family" {
```
nick-stroud, line 98: Are we able to use a more concise and consistent image definition such as [this](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/b0a5f6f1ef6298ccda812e9c332be7a195e1f117/community/modules/compute/SchedMD-slurm-on-gcp-partition/variables.tf#L46-L49).

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
```diff
@@ -0,0 +1,298 @@
...
+variable "min_cpu_platform" {
+  description = "The name of the minimum CPU platform that you want the instance to use."
+  type        = string
+  default     = null
+}
+
+variable "on_host_maintenance" {
+  type        = string
+  description = "Instance availability Policy"
+  default     = "MIGRATE"
```
nick-stroud, line 196: Placement is enabled by default which I believe conflicts with live migration. If this is ignored when placement is true then we should say so in the description.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
```diff
@@ -0,0 +1,298 @@
...
+variable "subnetwork_self_link" {
+  type        = string
+  description = "Subnet to deploy to. Only one of network or subnetwork should be specified."
+  default     = ""
+}
+
+variable "exclusive" {
+  description = "Exclusive job access to nodes"
+  type        = bool
+  default     = false
```
nick-stroud, line 277: I am not sure how exclusive and placement interact on the Slurm side, but as discussed previously, ideally exclusive would follow the setting for placement unless they are explicitly set to be different. My recommendation would be to have the default be null, and set it to follow placement in locals (if not explicitly set). At a minimum we should document when this is being ignored. 

The default doesn't really make sense at the moment (placement on, exclusive off)

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: It defaults to the schedmd-public-images project and the default family name for the hpc-centos-7 slurm image in the underlying module. I've updated the description to make this more clear in a coming commit.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: I can do that. I was following the format used by schedmd: https://github.com/SchedMD/slurm-gcp/blob/master/terraform/slurm_cluster/modules/slurm_instance_template/variables.tf#L210

I don't have a preference for either personally, so if the concise version is preferred by others I'm fine with that.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: Good point - I don't see this handled anywhere in the submodules. From what I can tell, both on_host_maintenance=MIGRATE and enable_placement=true can be set and neither override at the terraform level. 

I infer from [the docs](https://cloud.google.com/compute/docs/instances/define-instance-placement#support_and_restrictions ) that if on_host_maintenance is not TERMINATE, placement will not be set. Does that seem right?

I'll run some tests to check the behavior and edit the defaults/description accordingly.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: Exclusive is set based on placement: https://github.com/SchedMD/slurm-gcp/blob/master/terraform/slurm_cluster/modules/slurm_partition/main.tf#L29

So this works, but is not intuitive. I'll update the default to true.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: I was right in my interpretation - I updated the default to TERMINATE so that placement policy will not be overwritten. I also added an extra note in this description to make that behavior clear.

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
heyealex: After looking into this a bit further, this is also the default format for "instance_template" in terraform. I'd rather keep our version consistent with the other sources. I'm open to discussion on that, but for now I'm not going to take action to update this.
https://registry.terraform.io/modules/terraform-google-modules/vm/google/latest/submodules/instance_template

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
nick-stroud: nit: it might be worth 1 line in the description to say, "exclusive will always be enabled when placement is enabled".

modules/README.md
```diff
@@ -33,9 +33,12 @@ Modules that are still in development and less stable are labeled with the
 * **[vm-instance]** ![core-badge] : Creates one or more VM instances.
 * **[SchedMD-slurm-on-gcp-partition]** ![community-badge] : Creates a partition
   to be used by a [slurm-controller][schedmd-slurm-on-gcp-controller].
+* **[schedmd-slurm-gcp-v5-partition]** ![community-badge] ![experimental-badge] :
+  Creates a partition to be used by a [slurm-controller][schedmd-slurm-gcp-v5-controller].
 
 [vm-instance]: compute/vm-instance/README.md
 [schedmd-slurm-on-gcp-partition]: ../community/modules/compute/SchedMD-slurm-on-gcp-partition/README.md
+[schedmd-slurm-on-gcp-partition-v5]: ../community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
```
nick-stroud, line 41: How do we feel about the difference in capitalization between v4 and v5? I don't have any strong opinion, I just wanted to make sure the choice was conscious.

tools/validate_configs/test_configs/hpc-cluster-small-slurm-v5.yaml
```diff
@@ -0,0 +1,73 @@
...
+    id: debug_partition
+    use:
+    - network1
+    - homefs
+    settings:
+      partition_name: debug
+      node_count_dynamic_max: 4
+      enable_placement: false
+      machine_type: n2-standard-2
+      is_default: true
```
nick-stroud, line 54: What happens if this is not set on any partition? Does it resolve gracefully? This is a new requirement for v5 and I could imagine customers leaving it off.

heyealex: There would not be a default partition, which is entirely valid, but could be a bit annoying.

My preference would be to set the first listed partition as default, but ran into some errors when I tried to access and update an element of the list in terraform from the controller module, something similar to:
```
The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created.
```

I'll spend a bit more time on this and see if I can find another workaround.

modules/README.md
heyealex: This was discussed post standup a couple weeks back, though I think you might not have been there. This was the consensus, but there is room to update this if you have objections. 

I also noticed that this link is not correct, I'll update.

heyealex: Link has been updated.

nick-stroud (approved):

PR #382, opened 2022-06-30
Ignore threads_per_core for unsupported machine types in vm-instance

kkr16:
### Description

Adds a condition to the `advanced_machine_features` block in the `vm-instance` module that is used to bypass the `threads_per_core` parameter for machine types that do not support setting an SMT parameter (t2d instances, shared core instances, and instances with <2 vCPU).

The default `threads_per_core` has been set to `0`, which translates to an intelligent default behavior that turns off SMT on supported machine types, and removes the parameter where it is not supported.

To test, `test_configs/threads_per_core.yaml` has been expanded to include more machine types, and `threads_per_core=null`, and `threads_per_core=0`.

### Submission Checklist

  install`)


modules/compute/vm-instance/main.tf
```diff
@@ -45,6 +45,15 @@ locals {
     "ENABLE"  = "TRUE"
   }
   enable_oslogin = var.enable_oslogin == "INHERIT" ? {} : { enable-oslogin = lookup(local.oslogin_api_values, var.enable_oslogin, "") }
+
+  machine_vals   = split("-", var.machine_type)
+  machine_family = local.machine_vals[0]
+  machine_vcpus  = length(local.machine_vals) > 2 ? parseint(local.machine_vals[2], 10) : 1
```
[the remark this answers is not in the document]
tpdownes: Consider as "nit" unless @nick-stroud adds it to a list of requested changes.

modules/compute/vm-instance/variables.tf
```diff
@@ -195,11 +195,13 @@ variable "tags" {
...
-  will have 60 virtual cores with threads_per_core equal to 2. With
-  threads_per_core equal to 1 (SMT turned off), only the 30 physical cores will
-  be available on the VM.
+  to 2, Simultaneous Multithreading (SMT) is enabled extending the total number
+  of virtual cores. For example, a machine of type c2-standard-60 will have 60
+  virtual cores with threads_per_core equal to 2. With threads_per_core equal
+  to 1 (SMT turned off), only the 30 physical cores will be available on the VM.
+
+  The setting will be silently ignored for machine types that do not support 
+  SMT (t2d, shared-core instances, or instances with less than 2 vCPU).
```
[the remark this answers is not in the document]
tpdownes: I definitely prefer `null` as the default value and hadn't considered the simple

```hcl
locals {
...
  threads_per_core = local.smt_capable && var.threads_per_core == null ? 1 : var.threads_per_core
...
} 
```

With the obvious substitution in the `for_each` block. Otherwise code (I think) remains the same.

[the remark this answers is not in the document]
kkr16: >I prefer that we avoid silently ignoring values whenever possible.

Agreed. I added a precondition on the vm_instance resource to make the manifest fail when threads_per_core is set on an unsupported instance. Here is an example error:

```
│ Error: Resource precondition failed
│ 
│   on modules/vm-instance/main.tf line 157, in resource "google_compute_instance" "compute_vm":
│  157:       condition     = var.threads_per_core == null || local.smt_capable
│     ├────────────────
│     │ local.smt_capable is false
│     │ var.threads_per_core is 2
│ 
│ machine type n1-standard-1 does not support threads_per_core setting on instance n1-1-thread-0
```

> I would suggest that we set the default value to null in [variables.tf](http://variables.tf/)

I considered this, but noticed that _(while unrelated to 'ghpc')_ the [GKE documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/configure-smt) mentions 'null' as being the Compute Engine VM default. 
I decided against altering the behavior of `null` for consistency. Let me know if you think otherwise. 
If we are altering the default behavior of GCE, I think we should be explicit about it in our default vars.

> Then in the locals block we look for a value of null and replace with an intelligent default (1 where SMT off is supported, 2 everywhere else).

It would be "1 where SMT off is supported, and no config block everywhere else". We cannot have a `threads_per_core` parameter on instances where the SMT parameter is not supported, regardless of the value.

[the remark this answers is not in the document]
tpdownes: Whatever version the final PR requires must also be accurately reflected in `modules/compute/vm-instance/versions.tf`.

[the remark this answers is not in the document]
kkr16: > If the user explicitly sets the value (var.threads_per_core != null), then we don't do any validation and pass through the users value regardless of correctness.

I was trying to avoid pass-through of the users value regardless of correctness, because this would only fail at the `terraform apply` stage, and would not be caught at `terraform plan`, however, after giving this a bit more thought, I believe this is not really an issue.

The behavior I have in mind is very similar to what @nick-stroud has described, with minor changes:

- in variables.tf keep default `threads_per_core = 1`
- ignore the smt parameter block if `threads_per_core == null`
- pass through whatever value the user provided if `threads_per_core != null`
 and let the GCE resource fail if an unsupported parameter is provided

With this, we do not need to make any assertions as to whether an machine type supports SMT or not, since we rely on the GCE API to fail, which makes us compatible with future instance types, without having to modify our logic.

And, we still give the user possibility to overwrite our explicit default value of `threads_per_core = 1` to `null` which will remove the config block for unsupported machine types.

tpdownes:
@kkr16 does this merit review now?

Please edit the PR to target the `develop` branch. You may need to `rebase`. All PRs initially land on `develop` before release.

kkr16:
> @kkr16 does this merit review now?

No, @nick-stroud and I had a chat on the target behavior. I'll push a change accordingly.

> Please edit the PR to target the `develop` branch. You may need to `rebase`. All PRs initially land on `develop` before release.

Will do. Thanks!

tpdownes:
Please squash these commits as advised in #411

modules/compute/vm-instance/main.tf
```diff
@@ -45,6 +45,18 @@ locals {
...
+
+  machine_vals            = split("-", var.machine_type)
+  machine_family          = local.machine_vals[0]
+  machine_not_shared_core = length(local.machine_vals) > 2
+  machine_vcpus           = try(parseint(local.machine_vals[2], 10), 1)
+
+  smt_capable_family = !contains(["t2d"], local.machine_family)
+  smt_capable_vcpu   = local.machine_vcpus >= 2
+
+  smt_capable = local.smt_capable_family && local.smt_capable_vcpu && local.machine_not_shared_core
```
nick-stroud, line 57: nit: I don't think you need to define and use `machine_not_shared_core` since its value will always match that of `smt_capable_vcpu`. For me, it adds extra complexity, but I would also understand if you want to keep it since the behavior is kind of relying on the fallback value of `machine_vcpus`.

modules/compute/vm-instance/main.tf
```diff
@@ -45,6 +45,18 @@ locals {
...
+  smt_enable  = var.threads_per_core != null && (var.threads_per_core == 0 && local.smt_capable || try(var.threads_per_core >= 1, false))
```
nick-stroud, line 58: nit: I think `set_threads_per_core` or `should_set_threads_per_core` would be a better variable name here. `smt_enable` is confusing to me because when we are setting `threads_per_core=1` (aka smt disabled) this value will be true.

modules/compute/vm-instance/main.tf
```diff
@@ -45,6 +45,18 @@ locals {
...
+  smt_value   = var.threads_per_core == 2 ? 2 : 1
```
nick-stroud, line 59: nit: I prefer `threads_per_core` as the variable name here as it is more explicit. Basically, what is a _"Simultaneous Multithreading value"_. Terraform should be able to handle having a var and local of the same name. So you would have: 

Line 59
```
threads_per_core   = var.threads_per_core == 2 ? 2 : 1
```

Line 148
```
threads_per_core = local.threads_per_core
```

nick-stroud:
@kkr16, I just had a few "nits". For context, "nits" are suggestions that, in my opinion, would improve the code but are not required for my approval. It is your choice whether or not to accept these suggestions. 

If you want to decline the suggestions, then just respond "Ack." to the comment and resolve the thread. This just lets me know that you saw it and are not going to act on it. If you do make any of the suggested changes then I would just add it as one additional commit titled something like "Addressing feedback". Either way once you are done I will approve and merge.

modules/compute/vm-instance/main.tf
kkr16: ACK. 
I added it based on https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/382#discussion_r911420156 
I prefer keeping `machine_not_shared_core` explicit for code readability/maintainability.

modules/compute/vm-instance/main.tf
kkr16: Changed. Thanks!

modules/compute/vm-instance/main.tf
kkr16: Changed. Thanks!

nick-stroud (approved):

PR #391, opened 2022-07-08
Add strict positional arguments checking to the create and expand

danielahlin:
subcommands.

### Submission Checklist

* [ x] Have you installed and run this change against pre-commit? (`pre-commit
  install`)
* [ x] Are all tests passing? (`make tests`)
* [ x] Have you followed the guidelines in our Contributing document?

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -64,6 +64,7 @@ var (
 		Short: "Create a new deployment.",
 		Long:  "Create a new deployment based on a provided blueprint.",
 		Run:   runCreateCmd,
+		Args:  cobra.ExactArgs(1),
 	}
 )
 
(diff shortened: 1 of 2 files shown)


danielahlin:
Concrete example where an error message would be helpful:

While incorrect  - the following should produce an error message:
./ghpc create examples/hpc-cluster-small.yaml --vars project_id=xxxxxxxxxx deployment_name=some-test-small

user (me in this case) can not really tell that  deployment_name=some-test-small is just ignored due to the incorrect syntax. Failing silently is not helpful.

tpdownes:
I agree that this improves invalid commands by producing following:

```shell
$ ghpc create examples/hpc-cluster-small.yaml --vars project_id=my-secret-project deployment_name=my-secret-deployment
Error: accepts 1 arg(s), received 2
Usage:
  ghpc create BLUEPRINT_NAME [flags]

Flags:
...
```

PR #417, opened 2022-07-13
Install HTCondor autoscaler into filesystem and fix node deletion bug

tpdownes:
- Enhance htcondor-configure
  - anticipate that access point (not central manager) will run autoscaler in most installations
  - there are some harmless variable name changes to harmonize style
- Enhance htcondor-install
  - fix bug in node deletion from regional MIGs (status quo simply doesn't work)
  - create runner that installs a Python virtual environment with autoscaler dependencies
  - create runner that installs the actual autoscaler Python code

This PR does not actually cause the autoscaler to execute. That will be in a forthcoming PR.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/htcondor-configure/variables.tf
+++ b/community/modules/scheduler/htcondor-configure/variables.tf
@@ -28,6 +28,7 @@ variable "access_point_roles" {
   description = "Project-wide roles for HTCondor Access Point service account"
   type        = list(string)
   default = [
+    "roles/compute.instanceAdmin",
     "roles/monitoring.metricWriter",
     "roles/logging.logWriter",
     "roles/storage.objectViewer",
@@ -38,7 +39,6 @@ variable "central_manager_roles" {
   description = "Project-wide roles for HTCondor Central Manager service account"
   type        = list(string)
   default = [
-    "roles/compute.instanceAdmin",
     "roles/monitoring.metricWriter",
     "roles/logging.logWriter",
     "roles/storage.objectViewer",
--- a/community/modules/scripts/htcondor-install/files/autoscaler.py
+++ b/community/modules/scripts/htcondor-install/files/autoscaler.py
@@ -87,12 +87,9 @@ class AutoScaler:
             self.instanceGroupManagers = self.service.instanceGroupManagers()
 
     # Remove specified instance from MIG and decrease MIG size
-    def deleteFromMig(self, instance):
+    def deleteFromMig(self, instance, zone):
         instanceUrl = "https://www.googleapis.com/compute/v1/projects/" + self.project
-        if self.multizone:
-            instanceUrl += "/regions/" + self.region
-        else:
-            instanceUrl += "/zones/" + self.zone
+        instanceUrl += "/zones/" + zone
         instanceUrl += "/instances/" + instance
 
         instances_to_delete = {"instances": [instanceUrl]}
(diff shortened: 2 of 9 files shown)


community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -295,22 +296,22 @@ def scale(self):
 
                     if server not in node_busy:
                         if status == "Unclaimed":
-                            node_busy[server] = False
+                            node_busy[server] = zone
```
heyealex, line 299: The `node_busy` map is not well named anymore. You could consider renaming it to something like `server_zone` or setting it to a tuple or map of (busy:bool, zone:string).

community/modules/scripts/htcondor-install/files/install-htcondor-autoscaler-deps.yml
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+- name: Install but do not activate HTCondor autoscaler
+  become: true
+  hosts: localhost
+  tasks:
+  - name: Install Python 3 pip
+    ansible.builtin.yum:
```
heyealex, line 21: Will this always be run on centos/redhat images? If not, ansible.buildtin.package would be a more flexible choice here.

heyealex (changes requested):

community/modules/scheduler/htcondor-configure/variables.tf
```diff
@@ -28,6 +28,7 @@ variable "access_point_roles" {
   description = "Project-wide roles for HTCondor Access Point service account"
   type        = list(string)
   default = [
+    "roles/compute.instanceAdmin",
```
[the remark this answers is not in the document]
tpdownes: In Slurm-speak: the central manager is the controller while the access point is a login node from which one can submit jobs. It is definitely not ideal to put this on the access point but the autoscaler.py presently requires it. Changing that is relatively minor fix but something for post-release.

Recall that, right now, POSIX users other than `root` and`condor` are blocked from the instance metadata server anyhow.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -87,12 +87,12 @@ def __init__(self, multizone=False):
...
 
     # Remove specified instance from MIG and decrease MIG size
-    def deleteFromMig(self, instance):
+    def deleteFromMig(self, instance, zone):
         instanceUrl = "https://www.googleapis.com/compute/v1/projects/" + self.project
-        if self.multizone:
-            instanceUrl += "/regions/" + self.region
-        else:
-            instanceUrl += "/zones/" + self.zone
+        # if self.multizone:
```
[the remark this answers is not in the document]
tpdownes: Will do.

community/modules/scripts/htcondor-install/files/install-htcondor-autoscaler-deps.yml
tpdownes: I will check the package names on Debian derivatives. If so, I will make the change. There are all sorts of CentOS 7 assumptions here and throughout but we should not be adding more.

tpdownes: Resolved on HEAD of PR

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: Resolved on HEAD of PR. Called it `idle_node_zones`. This code is in serious need of re-implementation (shell call-outs are unnecessary as HTCondor has Python interface).

community/modules/scripts/htcondor-install/files/autoscaler.py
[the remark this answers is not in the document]
tpdownes: Resolved on HEAD of PR

PR #419, opened 2022-07-13
Add public example for HTCondor Pool

tpdownes:
This is blocked by PR #417 and #418. Once all 3 are merged I will follow up with

- PR to replace HTCondor integration test to use this example
- PR to remove functionality warnings from `htcondor-*` README

I will preemptively discuss a likely concern:

- OS Login is explicitly disabled because empirically HTCondor's internal notion of users fails to detect the username properly when using the extremely high UIDs generated by OS Login but works with UIDs ~1000 (or perhaps a problem with the `_` underscore characters); I interpret this an HTCondor bug which I will discuss with them

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/htcondor-pool.yaml
+++ b/community/examples/htcondor-pool.yaml
@@ -1,0 +1,136 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+blueprint_name: htcondor-pool
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: htcondor-001
+  region: us-central1
+  zone: us-central1-c
+
+deployment_groups:
+- group: htcondor
+  modules:
+  - source: modules/network/vpc
+    kind: terraform
+    id: network1
+    settings:
+      network_name: htcondor-pool
+      subnetwork_name: htcondor-pool-usc1
+    outputs:
+    - network_name
+
+  - source: community/modules/scripts/htcondor-install
+    kind: terraform
+    id: htcondor_install
+
+  - source: community/modules/project/service-enablement
+    kind: terraform
+    id: htcondor_services
+    use:
+    - htcondor_install
+
+  - source: community/modules/scheduler/htcondor-configure
+    kind: terraform
+    id: htcondor_configure
+
+  - source: modules/scripts/startup-script
+    kind: terraform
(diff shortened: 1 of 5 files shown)


community/examples/htcondor-pool.yaml
```diff
@@ -0,0 +1,136 @@
...
+- group: htcondor
+  modules:
+  - source: modules/network/vpc
+    kind: terraform
+    id: network1
+    settings:
+      network_name: htcondor-pool
+      subnetwork_name: htcondor-pool-usc1
+    outputs:
+    - network_name
```
nick-stroud, line 34: nit: inline this: `outputs: [network_name]`

tpdownes:
After rebasing when #418 was merged, I have force pushed. `make tests` succeeds locally. I unchecked the "make tests" box and re-checked it for good measure.

community/examples/htcondor-pool.yaml
```diff
@@ -0,0 +1,136 @@
...
+    settings:
+      runners:
+      - type: shell
+        source: modules/startup-script/examples/install_ansible.sh
+        destination: install_ansible.sh
+      - $(htcondor_install.install_htcondor_runner)
+      - $(htcondor_install.install_autoscaler_deps_runner)
+      - $(htcondor_install.install_autoscaler_runner)
+      - $(htcondor_configure.access_point_runner)
+      - $(htcondor_execute_point.configure_autoscaler_runner)
```
nick-stroud, line 73: nit: At one point we had talked about only being able to reference modules that had already been defined above (as a cheap way to avoid cycles). This would violate. Maybe move each startup script next to the module that is consuming them.

community/examples/htcondor-pool.yaml
```diff
@@ -0,0 +1,136 @@
...
+    settings:
+      metadata:
+        central-manager: ((module.htcondor_cm.internal_ip[0]))
+        enable-oslogin: "FALSE"
+      service_account:
+        email: $(htcondor_configure.execute_point_service_account)
+        scopes:
+        - cloud-platform
+
+  - source: modules/compute/vm-instance
```
nick-stroud, line 119: No action: Part of me wants to see a htcondor access and htcondor cm modules. It would definitely be possible to make variable passing easier, cut down on some verbosity, and maybe even take the startup scripts inside of the module. With that said I think what you have here gets it down with much less overhead, so no objection.

nick-stroud (approved):

community/examples/htcondor-pool.yaml
tpdownes: Upon original design we had not yet used GitHub URL based approach to use Toolkit modules. There is a need for enforcing the `central-manager` metadata existence and correctness. Maybe that could be addressed with `use`. I'll take a look / create a task.

community/examples/htcondor-pool.yaml
tpdownes: Performed re-org due to increased clarity of connection.

community/examples/htcondor-pool.yaml
tpdownes: Ack

community/examples/htcondor-pool.yaml
tpdownes: Addressing this nit requires a re-approval @nick-stroud

nick-stroud (approved):

PR #428, opened 2022-07-19
fix errors when missing deployment_name

kkr16:
### Changes:
- Used existing func `DeploymentName()` to validate `deployment_name` is a valid string, and fail early with a user-friendly `DeploymentNameError()` otherwise. If `deployment_name` is:
  - **not set**: `deployment_name must be a string and cannot be empty, cause: deployment_name variable not defined.`
  - **an empty string** `""`: `deployment_name must be a string and cannot be empty, cause: deployment_name was an empty string.`
  - **not a string** (i.e. an integer `100`): `deployment_name must be a string and cannot be empty, cause: deployment_name was not of type string.`
  - **empty**: `deployment_name must be a string and cannot be empty, cause: deployment_name was not of type string.`
-  removed `getDeploymentName()` that returned `undefined` when `deployment_name` was not set. Early error checking renders this obsolete.
- Added a default `deployment_name` to test configs returned from `func getBasicDeploymentConfigWithTestModule()` and `func getDeploymentConfigForTest()` used during unit tests, and updated relevant unit tests accordingly

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -523,8 +523,8 @@ func ConvertMapToCty(iMap map[string]interface{}) (map[string]cty.Value, error) 
 // global variables, then they are replaced by the cty.Value of the
 // corresponding entry in the origin. All other cty.Values are unmodified.
 // ERROR: if (somehow) the cty.String cannot be converted to a Go string
-// ERROR: rely on HCL TraverseAbs to bubble up "diagnostics" when the global variable
-//        being resolved does not exist in b.Vars
+// ERROR: rely on HCL TraverseAbs to bubble up "diagnostics" when the global
+//        variable being resolved does not exist in b.Vars
 func ResolveVariables(
 	ctyMap map[string]cty.Value,
 	origin map[string]cty.Value,
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -396,6 +396,10 @@ func checkUsedModuleNames(
 
 // validateConfig runs a set of simple early checks on the imported input YAML
 func (dc *DeploymentConfig) validateConfig() {
+	_, err := dc.Config.DeploymentName()
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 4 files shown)


ghpc.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    https://www.apache.org/licenses/LICENSE-2.0
```
heyealex, line 8: Was this change intentional? It appears this line is 4 spaces in other files, it's better to keep it consistent.

pkg/config/config.go
```diff
@@ -519,8 +523,8 @@ func ConvertMapToCty(iMap map[string]interface{}) (map[string]cty.Value, error)
 // global variables, then they are replaced by the cty.Value of the
 // corresponding entry in the origin. All other cty.Values are unmodified.
 // ERROR: if (somehow) the cty.String cannot be converted to a Go string
-// ERROR: rely on HCL TraverseAbs to bubble up "diagnostics" when the global variable
-//        being resolved does not exist in b.Vars
+// ERROR: rely on HCL TraverseAbs to bubble up "diagnostics" when the global
+// variable being resolved does not exist in b.Vars
```
heyealex, line 527: I found this more clear before, though I'm not opposed to the change. Was this made by the go linter?

heyealex (changes requested):
Thanks Karim, this is great! I just had a couple comments related to some formatting changes that might have been an issue with to the go linter.

ghpc.go
kkr16: Not intentional. Change was proposed by the linter, and I just went with it. 
I reverted it for now.

pkg/config/config.go
kkr16: A worse change change was proposed by the linter, and this is how I was able to appease the linter. 
I reverted it for now.

heyealex (dismissed):

heyealex (approved):

PR #431, opened 2022-07-20
Add Cloud Batch job submission to integration test

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/cloud-batch-login-node/variables.tf
+++ b/community/modules/scheduler/cloud-batch-login-node/variables.tf
@@ -70,3 +70,13 @@ variable "batch_job_directory" {
   type        = string
   default     = "/home/batch-jobs"
 }
+
+variable "enable_oslogin" {
+  description = "Enable or Disable OS Login with \"ENABLE\" or \"DISABLE\". Set to \"INHERIT\" to inherit project OS Login setting."
+  type        = string
+  default     = "ENABLE"
+  validation {
+    condition     = var.enable_oslogin == null ? false : contains(["ENABLE", "DISABLE", "INHERIT"], var.enable_oslogin)
+    error_message = "Allowed string values for var.enable_oslogin are \"ENABLE\", \"DISABLE\", or \"INHERIT\"."
+  }
+}
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
@@ -1,0 +1,35 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


community/modules/scheduler/cloud-batch-login-node/variables.tf
```diff
@@ -70,3 +70,13 @@ variable "batch_job_directory" {
   type        = string
   default     = "/home/batch-jobs"
 }
+
+variable "enable_oslogin" {
+  description = "Enable or Disable OS Login with \"ENABLE\" or \"DISABLE\". Set to \"INHERIT\" to inherit project OS Login setting."
```
heyealex, line 75: Is there a reason to not just set this to either "TRUE", "FALSE" or null (which implies "INHERIT")? ENABLE/DISABLE seems to add another layer that's inconsistent with the actual API. 

Alternatively, you could just have a bool that defaults to null, if not set it inherits.

tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: Wait for startup script to complete
+  become: true
+  ansible.builtin.wait_for:
+    path: /var/log/messages
+    search_regex: '.*{{ remote_node }}.*startup-script exit status ([0-9]+)'
+    timeout: 600
+  register: startup_status
```
heyealex, line 21: There should be an additional task that fails out if the exit status is not 0. 
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/tools/cloud-build/daily-tests/ansible_playbooks/spack.yml#L25
I also noticed it's not checked for the monitoring test. I'll add a task to update that.

tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: Batch Job Block
+  block:
+  - name: Submit batch job
+    ansible.builtin.command: gcloud alpha batch jobs submit {{ deployment_name }} --config=/home/batch-jobs/cloud-batch-{{ deployment_name }}.json --location=us-central1 --project={{ custom_vars.project }}
+  - name: Wait for job to run
+    ansible.builtin.command: "gcloud alpha batch jobs describe {{ deployment_name }} --location=us-central1 --project={{ custom_vars.project }}"
```
heyealex, line 27: nit: this one is quoted, the last one isn't. It makes no functional difference, but it would look cleaner if they were consistent.

heyealex (changes requested):

community/modules/scheduler/cloud-batch-login-node/variables.tf
nick-stroud: I am following the same pattern as [vm-instance](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/199ab9a0a6026b1fb1474bced569f02f7716851f/modules/compute/vm-instance/variables.tf#L211). I might try to provide more context offline.

tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
nick-stroud: Done.

tools/cloud-build/daily-tests/ansible_playbooks/test-batch-submission.yml
nick-stroud: Done.

heyealex (approved):

PR #436, opened 2022-07-21
Update Batch integration test to run in series

nick-stroud:
Without this failure of one test will kill the parallel test and leave its resources up.

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/integration-group-4.yaml
+++ b/tools/cloud-build/daily-tests/integration-group-4.yaml
@@ -17,7 +17,8 @@
 # and test health of the pool
 # ├── build_ghpc
 # └── fetch_builder
-#    ├── htcondor (group 4)
+#    └── htcondor (group 4)
+#       └── Cloud Batch
 
 
 timeout: 14400s  # 4hr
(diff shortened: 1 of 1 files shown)


tpdownes:
@nick-stroud does this logic also apply to `tools/cloud-build/daily-tests/integration-group-3.yaml`?

nick-stroud:
> @nick-stroud does this logic also apply to `tools/cloud-build/daily-tests/integration-group-3.yaml`?

I believe so. When one task fails it kills all other parallel tasks without allowing for a graceful cleanup. In group 3 you will see that omnia waits for monitoring and luster waits for omnia. It looks like packer does not follow that pattern. Not sure if that is on purpose or a mistake.

tpdownes:
I agree with this and think I just observed it while testing #435.  Can you make the appropriate change in both builds?

nick-stroud:
> I agree with this and think I just observed it while testing #435. Can you make the appropriate change in both builds?

Are you referring to group 3 and packer. I am less familiar with that one and was not sure if parallelism is intentional. Generally I would prefer to handle in separate PR since this has already been reviewed.

tpdownes (approved):
Please create a task to address group 3

tpdownes:
No need here to wait on #435

PR #437, opened 2022-07-21
Breakout startup wait to new file & update Batch test

nick-stroud:
Breakout the common Ansible task of waiting for startup script to finish so that it can be reused in various tests.

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
@@ -1,0 +1,30 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: Assert variables are defined
+  ansible.builtin.assert:
+    that:
+    - vm_name is defined
+
+- name: Wait for startup script to complete
+  become: true
+  ansible.builtin.wait_for:
+    path: /var/log/messages
+    search_regex: '.*{{ vm_name }}.*startup-script exit status ([0-9]+)'
+    timeout: 600
+  register: startup_status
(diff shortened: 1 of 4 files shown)


nick-stroud:
@tpdownes & @heyealex, I am looking for initial feedback on this. I waited to apply this in more places until we agree it is a good idea. If we like this pattern then I will go ahead and swap out this task anywhere we are waiting for startup script.

tools/cloud-build/daily-tests/ansible_playbooks/spack.yml
```diff
@@ -14,18 +14,11 @@
 
 ---
 
-- name: Wait for startup script to complete
-  become: true
-  wait_for:
-    path: /var/log/messages
-    search_regex: '.*{{ login_node }}.*startup-script exit status ([0-9]+)'
-    timeout: 7200
-    state: present
```
nick-stroud, line 23: According to [ansible docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/wait_for_module.html#parameter-state), when used for a file/search string (which is what we are doing), `present` has the same effect as the default `started`.

tools/cloud-build/daily-tests/ansible_playbooks/spack.yml
```diff
@@ -14,18 +14,11 @@
...
-  register: startup_status
-- name: Fail if startup script exited with a non-zero return code
-  fail:
-    msg: There was a failure in the startup script
-  when: startup_status['match_groups'][0] != "0"
+- name: Include wait for startup script
```
heyealex, line 17: I'm curious if you could do the same thing with a role, similar to this: https://github.com/dellhpc/omnia/blob/v1.2.2/omnia.yml#L20

I think you just need to create a "roles" directory with that name in it, then you can run multiple roles with one task. I wouldn't put much time into it, but if it's easy it could be worth it and decrease the boilerplate.

tools/cloud-build/daily-tests/ansible_playbooks/spack.yml
tpdownes: I agree with the logic. I'd raise at a "nit" level that I prefer being explicit for state for clarity and for possibility of default behavior changes. This is probably not one of those cases.

tpdownes (approved):
I like the idea here and believe it's a step in right direction.

@nick-stroud I prefer you wait until #435 is merged

tools/cloud-build/daily-tests/ansible_playbooks/spack.yml
nick-stroud: Ack.

nick-stroud:
This had to be rebased past ansible-lint upgrade and needs re-approval. Integration tests previously passed but I kicked them off again anyway.

tpdownes (approved):

PR #444, opened 2022-07-22
Add test-mount to Batch integration & move variables into custom_vars

nick-stroud:
This change:
- moves `mount` and `partition` vars into `custom_vars`
- breaks out the `test-mounts.yml` into its own file so it can be called by tests that want to check mounts but don't have partitions
- Adds `test-mounts` to Cloud Batch integration test

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
@@ -19,6 +19,10 @@
     - custom_vars.partitions is defined
     - custom_vars.mounts is defined
 
+- name: Include test mounts
+  ansible.builtin.include_tasks: "test-mounts.yml"
+  vars:
+    custom_vars: "{{ custom_vars }}"
 - name: Get partition info
   ansible.builtin.command: sinfo --format='%P' --noheader
   changed_when: False
@@ -38,16 +42,6 @@
     msg: Test Check Partitions failed
   when: item not in partition_output.stdout
   loop: "{{ custom_vars.partitions }}"
-- name: Get mount info
-  ansible.builtin.stat:
-    path: "{{ item }}"
-  register: stat_mounts
-  loop: "{{ custom_vars.mounts }}"
-- name: Check if mount exists
-  ansible.builtin.fail:
-    msg: "{{ item.item }} not mounted"
-  when: not item.stat.exists
-  loop: "{{ stat_mounts.results }}"
 - name: Test Mounts on partitions
   register: srun_mounts
   changed_when: srun_mounts.rc == 0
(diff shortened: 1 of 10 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
```diff
@@ -13,7 +13,16 @@
 # limitations under the License.
 
 ---
+- name: Assert variables are defined
+  ansible.builtin.assert:
+    that:
+    - custom_vars.partitions is defined
+    - custom_vars.mounts is defined
 
+- name: Include test mounts
```
heyealex, line 22: Sounds like `test_mounts` should be a separate test then. This could be converted to `test_partitions` and added as a test alongside `test_mounts`. I imagine it would still take `mounts` as a parameter to test whether the mounts exist on the partitions.

tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
```diff
@@ -32,27 +41,17 @@
   ansible.builtin.fail:
```
heyealex, line 41: Line above this is incorrect (probably a prior mistake of mine). While your changes these, could you change the name of the task to "Check partitions exist"? If you'd rather, I can create a quick bugfix PR for that instead.

heyealex (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
nick-stroud: Ya, I debated this. It is tricky because the original test is not only testing mounts on the login node but also testing them on each partition. We probably still want testing mounts on each partition as part the partition test but it should not be in the mounts only test. 

How it was structured:
- test-mounts-and-partitions.yml
  - tests mounts
  - tests partitions
  - tests mounts on partitions

How the pr is currently written:
- test-mounts.yml
  - tests mounts
- test-mounts-and-partitions.yml
  - tests mounts (via test-mounts.yml)
  - tests partitions
  - tests mounts on partitions

Alternative option:
- test-mounts.yml
  - tests mounts
- test-partitions.yml (note: is this naming reasonable)
  - tests partitions
  - tests mounts on partitions

Is your preference the alternative option? Obviously then we would add both as post deploy tests that previously had test-mounts-and-partitions.yml.

heyealex: > Is your preference the alternative option? Obviously then we would add both as post deploy tests that previously had test-mounts-and-partitions.yml.

Yes, that's how I'd organize it. I'm also open to creating follow-on task to do that since tests are already running.

nick-stroud: Done.

tools/cloud-build/daily-tests/ansible_playbooks/test-mounts-and-partitions.yml
nick-stroud: Done.

heyealex (approved):

heyealex:
Approved, but pending tests passing

PR #474, opened 2022-08-12
Change `source` to `.` when calling startup scripts in shell runner

heyealex:
Update the startup script shell runner invocation from source, which is specific to bash, to `.` which is compatible with more shells. 

A set of new test_configs have been added that test startup scripts against different OS images. This change has been deployed and tested against these new test_configs.

### Submission Checklist

  install`)

### DIFF
--- a/modules/scripts/startup-script/templates/startup-script-custom.tpl
+++ b/modules/scripts/startup-script/templates/startup-script-custom.tpl
@@ -29,7 +29,6 @@ stdlib::runner() {
 
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
-    # shellcheck source=/dev/null
     shell) . $destpath/$filename $args;;
   esac
   
--- a/modules/scripts/startup-script/templates/startup-script-custom.tpl
+++ b/modules/scripts/startup-script/templates/startup-script-custom.tpl
@@ -29,7 +29,7 @@ stdlib::runner() {
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
     # shellcheck source=/dev/null
-    shell)  sh -c "source '$destpath/$filename' $args";;
+    shell)  sh -c ". '$destpath/$filename' $args";;
   esac
 }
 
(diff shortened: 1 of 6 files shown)


modules/scripts/startup-script/templates/startup-script-custom.tpl
```diff
@@ -29,7 +29,7 @@ stdlib::runner() {
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
     # shellcheck source=/dev/null
-    shell)  sh -c "source '$destpath/$filename' $args";;
+    shell)  sh -c ". '$destpath/$filename' $args";;
```
cboneti, line 32: Here, on line 13, I would add a set -x so it fails when the runner fails. 
Also, I recommend replacing line 13 by: 
  . '$destpath/$filename' $args"
  exit_code=$?
  stdlib::debug "=== $object finished with exit_code=$exit_code ==="

tpdownes:
I think we also need to add documentation that is very explicit about what type of shell scripts we support. I never interpreted it to mean "POSIX" but that is the status quo.

modules/scripts/startup-script/templates/startup-script-custom.tpl
```diff
@@ -29,7 +29,7 @@ stdlib::runner() {
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
     # shellcheck source=/dev/null
```
nick-stroud, line 31: What is this line doing? Do we still need `source=/dev/null` now that you are using `.`?

heyealex:
A note as I'm running through some tests:
* Using `.` to source within `sh -c` in ubuntu (and probably debian) means arguments aren't passed to the script.
* Using `.` to source without `sh -c` can have some odd effects with commands like `exit 1`. The entire process will exit with a failure in this case. That may be what we want, but it's worth noting as I ran into it during testing.

modules/scripts/startup-script/templates/startup-script-custom.tpl
heyealex: Done

modules/scripts/startup-script/templates/startup-script-custom.tpl
heyealex: Removed

cboneti (approved):

heyealex:
> I think we also need to add documentation that is very explicit about what type of shell scripts we support. I never interpreted it to mean "POSIX" but that is the status quo.

The recent change removed the `sh -c`, so the shell used will be the default shell, and not strictly a POSIX shell.

heyealex:
The integration test failure is one that has been exposed by this change, Omnia is not installing properly anymore ands needs to be updated to a more recent version.

PR #476, opened 2022-08-12
Making slurm_cluster_name optional

cboneti:
It will default to the first 8 characters of the deployment name
(removing any invalid characters).

### Submission Checklist

  install`)


community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -60,13 +60,18 @@ locals {
       instance_template = null
     },
   ]
+
+  # Since deployment name may be used to create a cluster name, we remove any invalid character from the begining
+  # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
+  tmp_name           = lower(substr(replace(var.deployment_name, "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8))
```
heyealex, line 66: This would remove any upper case letters before lower gets a chance to do anything if I'm reading this right. I'd run lower first, then replace then substr.

heyealex: nit: `tmp_name` isn't very descriptive. I'd go with `filtered-deployment-name` or `filtered-deploy` or something like that.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -60,13 +60,18 @@ locals {
       instance_template = null
     },
   ]
+
+  # Since deployment name may be used to create a cluster name, we remove any invalid character from the begining
```
heyealex, line 64: spelling:  "begining" -> "beginning"

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
```diff
@@ -17,14 +17,15 @@
...
-  description = "Cluster name, used for resource naming and slurm accounting."
+}
 
-  validation {
-    condition     = can(regex("(^[a-z][a-z0-9]*$)", var.slurm_cluster_name))
-    error_message = "Variable 'slurm_cluster_name' must be composed of only alphanumeric values and begin with a leter. regex: '(^[a-z][a-z0-9]*$)'."
-  }
+variable "slurm_cluster_name" {
+  type        = string
+  description = "Cluster name, used for resource naming and slurm accounting.If not provided it will default to the first 8 characters of the deployment name (removing any invalid characters)."
```
heyealex, line 27: Space is needed after the period: accounting.If

heyealex (changes requested):

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
cboneti: characters (a-z) are not being touched by the regexp. It removes any non character happening in the beginning of the string, and anything that is not a character or a number happening anywhere. 
The reason I do it last is that since the substr already cuts to just the first 8 characters, lower has to run on a smaller string...

cboneti: changed tmp_name to tmp_cluster_name

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
cboneti: good catch, thanks

community/modules/compute/schedmd-slurm-gcp-v5-partition/variables.tf
cboneti: fixed, thanks

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
cboneti: sorry, I misunderstood your comment on the regexp... you are right.. I fixed it now.

heyealex (approved):

PR #479, opened 2022-08-15
Adding Slurm-on-GCP V5 Ubuntu example 

cboneti:
Also renaming the previous example to clearly show it was Centos7 based.

### Submission Checklist

  install`)


community/examples/slurm-gcp-v5-centos7.yaml
```diff
@@ -14,7 +14,7 @@
 
 ---
 
-blueprint_name: slurm-gcp-v5-cluster
+blueprint_name: slurm-gcp-v5-centos7
```
heyealex, line 17: Should this be "slurm-gcp-v5-hpc-centos7"? slurm-gcp also builds an image for the base centos7 image: https://github.com/SchedMD/slurm-gcp/blob/master/packer/example.pkrvars.hcl#L83

community/examples/slurm-gcp-v5-ubuntu2004.yaml
```diff
@@ -0,0 +1,85 @@
...
+---
+
+blueprint_name: slurm-gcp-v5-ubuntu2004
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: slurm-gcp-v5
+  region: us-central1
+  zone: us-central1-c
+  slurm_cluster_name: ghpc
```
heyealex, line 24: The PR that removes this as requirement has been merged, right? This could be removed from the example as well.

examples/README.md
```diff
@@ -203,9 +204,42 @@ For this example the following is needed in the selected region:
 * Compute Engine API: Resource policies: **one for each job in parallel** -
   _only needed for `compute` partition_
 
-[slurm-gcp-v5-cluster.yaml]: ../community/examples/slurm-gcp-v5-cluster.yaml
+[slurm-gcp-v5-centos7.yaml]: ../community/examples/slurm-gcp-v5-centos7.yaml
 [slurm-gcp]: https://github.com/SchedMD/slurm-gcp/tree/v5.0.2
 
+### [slurm-gcp-v5-ubuntu2004.yaml] ![community-badge]
+
+Similar to the previous example, but using Ubuntu 20.04 instead of CentOS 7.
```
heyealex, line 212: I'd add a note that other images are provided by slurm-gcp and that the source_image_family can be updated to use them in this example, specifying that it's not tested or officially supported of course. 
Docs: https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#published-image-family

```markdown
SchedMD supports
[4 base images](https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#published-image-family)
for its terraform modules. The deployment variable `source_image_family` can be updated in this
example to use any of these or a custom image built using the
[Slurm on GCP packer template](https://github.com/SchedMD/slurm-gcp/tree/v5.0.3/packer).
Note that only ubuntu and hpc-centos7 have been tested with the HPC Toolkit, other images may require additional configuration. 
```

community/examples/slurm-gcp-v5-ubuntu2004.yaml
```diff
@@ -0,0 +1,85 @@
...
+  source_image_family: "schedmd-v5-slurm-22-05-2-ubuntu-2004-lts"
+  source_image_project: "projects/schedmd-slurm-public/global/images/family"
```
heyealex, line 26: Is this just for clarity? It shouldn't be needed if I recall correctly. The source_image_project will default to this regardless of whether the image family has been set.

I actually like including it, but wanted to check if that was the purpose.

community/examples/slurm-gcp-v5-centos7.yaml
cboneti: good catch. I tried to reduce this as much as possible, as we have a limited number of characters, but I think the potential confusion justifies having the hpc in here.

community/examples/slurm-gcp-v5-ubuntu2004.yaml
cboneti: good catch!

examples/README.md
cboneti: good catch. Added a note.

community/examples/slurm-gcp-v5-ubuntu2004.yaml
cboneti: Correct. That is the default value for source_image_project, but I felt this would make it more clear that this works because the project has the right value.

heyealex (approved):

PR #481, opened 2022-08-16
Adding option to enable public ips on Slurm-GCP V5

cboneti:
Through disable_login_public_ips in login node and disable_controller_public_ips in the controller.
It a weird semmantic, since it is true by default, but it was chosen
this way to maintain compatibility with V4.

By default, both the controller and the login node will have disable
public ips set to true. If set to false, and access_config is not
provided, it will generated a random public ip for each one of these

The partition doesn't seem to have the option available, and since it is
also typically less desirable, I did not add it.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
@@ -27,12 +27,16 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  enable_public_ip_access_config = var.disable_controller_public_ips ? [] : [{ nat_ip = null, network_tier = null }]
+  access_config                  = length(var.access_config) == 0 ? local.enable_public_ip_access_config : var.access_config
+
 }
 
 module "slurm_controller_instance" {
   source = "github.com/SchedMD/slurm-gcp.git//terraform/slurm_cluster/modules/slurm_controller_instance?ref=v5.0.3"
 
-  access_config                = var.access_config
+  access_config                = local.access_config
   slurm_cluster_name           = local.slurm_cluster_name
   instance_template            = module.slurm_controller_template.self_link
   project_id                   = var.project_id
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
@@ -103,6 +103,12 @@ variable "deployment_name" {
   type        = string
 }
 
+variable "disable_controller_public_ips" {
+  description = "If set to false. The controller will have a random public IP assigned to it. Ignored if access_config is set."
+  type        = bool
+  default     = true
+}
+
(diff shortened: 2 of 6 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
```diff
@@ -103,6 +103,12 @@ variable "deployment_name" {
   type        = string
 }
 
+variable "disable_controller_public_ips" {
```
heyealex, line 106: I can't find the official docs on it anymore, but I remember tflint used to complain about negating boolean variables. In general, I agree that boolean variables should be enabling if true, disabling if false.

On the other hand, this keeps the naming the same as in v4, which is valuable. 

There's a good case for either, so I'll leave it up to you to keep it as is or invert it to "enable_controller_public_ips".

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
```diff
@@ -27,12 +27,16 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  enable_public_ip_access_config = var.disable_controller_public_ips == false ? [{ nat_ip = null, network_tier = null }] : []
```
heyealex, line 31: nit: You could simplify this a littler with:
`= var.disable_controller_public_ips ? [] : [{ nat_ip = null, network_tier = null }]`

community/modules/scheduler/schedmd-slurm-gcp-v5-login/main.tf
```diff
@@ -23,6 +23,9 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  enable_public_ip_access_config = var.disable_login_public_ips == false ? [{ nat_ip = null, network_tier = null }] : []
```
heyealex, line 27: Same as with the controller

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
cboneti: I really wanted to keep the same exact signature as v4, even though I hate the double negation.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
cboneti: I thought about this one, in terraform, I have been inclined towards explicit (long) tests, like == false.  But it maybe silly, do you feel strongly about this?

cboneti: I re-read it, and I think you are right, it's more direct and more clear.

community/modules/scheduler/schedmd-slurm-gcp-v5-login/main.tf
cboneti: fixed as well

heyealex (approved):

PR #482, opened 2022-08-16
Additional validation for blueprint_name

kkr16:
### Submission Checklist

  install`)

### Changes
Proposing a change to validate `blueprint_name` against label value naming constraints as defined in the Google Cloud [docs](https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -591,6 +591,13 @@ func (b Blueprint) ResolveGlobalVariables(ctyVars map[string]cty.Value) error {
 	return ResolveVariables(ctyVars, origin)
 }
 
+// isValidLabelValue checks if a string is a valid value for a GCP label.
+// For more information on valid label values, see the docs at:
+// https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements
+func isValidLabelValue(value string) bool {
+	return regexp.MustCompile(matchLabelExp).MatchString(value)
+}
+
 // DeploymentName returns the deployment_name from the config and does approperate checks.
 func (b *Blueprint) DeploymentName() (string, error) {
 	nameInterface, found := b.Vars["deployment_name"]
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -402,6 +402,10 @@ func (dc *DeploymentConfig) validateConfig() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	err = dc.Config.checkBlueprintName()
```
cboneti, line 405: I feel that everything that will eventually become a label (blueprint name, deployment name, etc) should run this test. Should you not generalize to check label requirements? and then run it for both blueprint name and deployment name?

pkg/config/config.go
```diff
@@ -596,3 +609,14 @@ func (b *Blueprint) DeploymentName() (string, error) {
 	}
 	return deploymentName, nil
 }
+
+// checkBlueprintName returns an error if blueprint_name does not comply with
+// requirements for correct label values as per https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements
+func (b *Blueprint) checkBlueprintName() error {
```
heyealex, line 615: Based on what Carlos suggested, I'd change this function to `checkLabelValue` with the same validation. With that, you could run the same test against deployment_name in the Config.DeploymentName() function as well.

This also helps generalize as we add features. There may be other labels that we track and having the function ready to go will be very valuable.

pkg/config/config_test.go
```diff
@@ -344,6 +345,30 @@ func (s *MySuite) TestDeploymentName(c *C) {
...
+	dc.Config.BlueprintName = ""
+	err = dc.Config.checkBlueprintName()
+	c.Check(errors.As(err, &e), Equals, true)
+
+	// Is blueprint_name longer than 63 characters?
+	dc.Config.BlueprintName = "blueprint-name-blueprint-name-blueprint-name-blueprint-name-0123"
+	err = dc.Config.checkBlueprintName()
+	c.Check(errors.As(err, &e), Equals, true)
+
+	// Does deployment_name contain special characters other than dashes or underscores?
```
heyealex, line 366: A couple more tests you could include:
* Test with an underscore (succeed)
* Test with a capital letter (fail)

heyealex (changes requested):

cboneti (approved):

PR #484, opened 2022-08-17
Log runner that is being executed

nick-stroud:
I have wanted this information several times in the past. It will help to make it clear which runner logs are coming from as well as confirming that all runners have executed. 

### Submission Checklist

  install`)

### DIFF
--- a/modules/scripts/startup-script/templates/startup-script-custom.tpl
+++ b/modules/scripts/startup-script/templates/startup-script-custom.tpl
@@ -27,6 +27,7 @@ stdlib::runner() {
 
   stdlib::get_from_bucket -u "gs://${bucket}/$object" -d "$destpath" -f "$filename"
 
+  stdlib::info "=== start executing runner: $object ==="
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
     shell) . $destpath/$filename $args;;


modules/scripts/startup-script/templates/startup-script-custom.tpl
```diff
@@ -27,10 +27,12 @@ stdlib::runner() {
...
 
+  echo "start executing runner: $filename"
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
     shell) . $destpath/$filename $args;;
   esac
+  echo "finished executing runner: $filename"
   
   exit_code=$?
   stdlib::info "=== $object finished with exit_code=$exit_code ==="
```
heyealex, line 38: This line already announces when the runner has completed. I'd stick with this format for the "start executing runner" line as well.

cboneti: What is the difference between using echo here, vs stdlib::info? Shouldn't the latter be preferred?

nick-stroud: I honestly had never noticed that line in the logs. 

I have:
- removed the finished line
- switched to using stdlib::info
- updated start line format to look similar to finished

We can either merge or close. I think it is marginally better to also print the begining to make it clear the runner has started.

heyealex: It was added recently, with the switch away from source. 
I like including the begin line as well, if a runner stalls you don't have to go back to the blueprint to figure out which one was active currently.

heyealex (approved):

PR #485, opened 2022-08-18
Update ansible install script

heyealex:
Overhaul of the install_ansible startup script runner. The goal of this update is to provide a more generalizable process for installing python, pip and ansible on a variety of VM OS images and environments.

This has been tested against the following VM images using the startups script test_configs:
* debian-10
* centos8
* hpc-centos7
* ubuntu 20.04
* rocky8

Other OS images have been tested against the install_ansible.sh script alone (i.e not testing against follow-on runners like install-nfs.sh)

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/nfs-server/README.md
+++ b/community/modules/file-system/nfs-server/README.md
@@ -6,6 +6,10 @@ files with other clients over a network. `nfs-server` can be used by
 [vm-instance](../../../../modules/compute/vm-instance/README.md) and SchedMD
 community modules that create compute VMs.
 
+> **_WARNING:_** This module has only been tested against the HPC centos7 OS
+> disk image (the default). Using other images may work, but have not been
+> verified.
+
 [disk]: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk
 
 ### Example
--- a/community/modules/scripts/spack-install/templates/install_spack.tpl
+++ b/community/modules/scripts/spack-install/templates/install_spack.tpl
@@ -9,6 +9,11 @@ if [[ $EUID -ne 0 ]]; then
   exit 1
 fi
 
+# Activate ghpc-venv virtual environment if it exists
+if [ -d /usr/local/ghpc-venv ]; then
+  source /usr/local/ghpc-venv/bin/activate
+fi
+
 # Only install and configure spack if ${INSTALL_DIR} doesn't exist
 if [ ! -d ${INSTALL_DIR} ]; then
 
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -53,6 +53,34 @@ Each runner receives the following attributes:
   Therefore`args` should not include any arguments that alter this behavior,
   such as `--connection`, `--inventory`, or `--limit`.
 
+### Runner dependencies
+
+The `ansible-local` runner requires ansible to be installed in the VM before
+running. To support other playbook runners in the HPC Toolkit, we require
+version 2.11 of ansible-core or higher. Note that this is distinct from the
+package version used to install ansible with pip. The minimum pip package
+of ansible is 4.10.0.
+
+To install ansible, a runner supplied by this module can be added as a prior
+runner. An example of this can be found in the [Example](#example) section below
+as the first runner in the list of runners. This script will do the following in
+your VM instance:
+
+- Install python3 if not already installed using system package managers (yum,
+  apt-get, etc)
+- Install pip3 if not already installed and upgrade pip3 if the version is not
+  at least 18.0.
+- Install and create a virtualenv located at `/usr/local/ghpc-venv`.
+- Install ansible into this virtualenv if the current version of ansible is not
+  version 2.11 or higher.
+
+To use the virtualenv created by this script, you can activate it by running the
+following commmand on the VM:
+
+```shell
+source /usr/local/ghpc-venv/bin/activate
+```
+
 ### Staging the runners
 
 Runners will be uploaded to a
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -66,13 +66,13 @@ runner. An example of this can be found in the [Example](#example) section below
 as the first runner in the list of runners. This script will do the following in
 your VM instance:
 
-- Install python3 if not already installed using system package managers (yum,
-  apt-get, etc)
-- Install pip3 if not already installed and upgrade pip3 if the version is not
-  at least 18.0.
-- Install and create a virtualenv located at `/usr/local/ghpc-venv`.
-- Install ansible into this virtualenv if the current version of ansible is not
-  version 2.11 or higher.
+- Install system-wide python3 if not already installed using system package
+  managers (yum, apt-get, etc)
+- Install system-wide pip3 if not already installed and upgrade pip3 if the
+  version is not at least 18.0.
+- Install and create a virtual environment located at `/usr/local/ghpc-venv`.
+- Install ansible into this virtual environment if the current version of
+  ansible is not version 2.11 or higher.
 
 To use the virtualenv created by this script, you can activate it by running the
 following commmand on the VM:
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -68,18 +68,33 @@ your VM instance:
 
 - Install system-wide python3 if not already installed using system package
   managers (yum, apt-get, etc)
+- Install `python3-distutils` system-wide in debian and ubuntu based
+  environments. This can be a missing dependency on system installations of
+  python3 for installing and upgrading pip.
 - Install system-wide pip3 if not already installed and upgrade pip3 if the
   version is not at least 18.0.
 - Install and create a virtual environment located at `/usr/local/ghpc-venv`.
 - Install ansible into this virtual environment if the current version of
   ansible is not version 2.11 or higher.
 
-To use the virtualenv created by this script, you can activate it by running the
-following commmand on the VM:
+To use the virtual environment created by this script, you can activate it by
+running the following commmand on the VM:
 
 ```shell
 source /usr/local/ghpc-venv/bin/activate
 ```
+
+You may also need to provide the correct python interpreter as the python3
+binary in the virtual environment. This can be done by adding the following flag
+when calling `ansible-playbook`:
+
+```shell
+-e ansible_python_interpreter=/usr/local/ghpc-venv/bin/activate
+```
+
+> **_NOTE:_** ansible-playbook and other ansible command line tools will only be
+> accessible from the command line (and in your PATH variable) after activating
+> this environment.
 
 ### Staging the runners
 
--- a/modules/scripts/startup-script/examples/install_ansible.sh
+++ b/modules/scripts/startup-script/examples/install_ansible.sh
@@ -34,7 +34,7 @@ apt_wait() {
 install_python_deps() {
 	if [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release 2>/dev/null ||
 		grep -qi ubuntu /etc/os-release 2>/dev/null; then
-		apt install -y python3-distutils
+		apt-get install -y python3-distutils
 	fi
 }
 
@@ -68,15 +68,15 @@ get_python_minor_version() {
 install_python3_yum() {
 	## TODO restrict repos to search through in centos to decrease overhead
 	yum install -y python3
-	python_path=$(rpm -ql python3 | grep 'python3$')
+	python_path=$(rpm -ql python3 | grep 'bin/python3$')
 }
 
 # Install python3 with the apt package manager. Updates python_path to the
 # newly installed packaged.
 install_python3_apt() {
 	apt_wait
-	apt install -y python3 python3-distutils
-	python_path=$(dpkg -L python3 | grep 'python3$')
+	apt-get install -y python3 python3-distutils
+	python_path=$(which python3)
 }
 
 install_python3() {
--- a/modules/scripts/startup-script/examples/install_ansible.sh
+++ b/modules/scripts/startup-script/examples/install_ansible.sh
@@ -15,21 +15,21 @@
 
 REQ_ANSIBLE_VERSION=2.11
 REQ_ANSIBLE_PIP_VERSION=4.10.0
-REQ_PIP_VERSION=18
+REQ_PIP_MINOR_VERSION=18
 REQ_PYTHON3_VERSION=6
 
 apt_wait() {
-	while fuser /var/lib/dpkg/lock 2>/dev/null 2>&1; do
+	while fuser /var/lib/dpkg/lock >/dev/null 2>&1; do
 		echo "Sleeping for dpkg lock"
 		sleep 3
 	done
-	while fuser /var/lib/apt/lists/lock 2>/dev/null 2>&1; do
+	while fuser /var/lib/apt/lists/lock >/dev/null 2>&1; do
 		echo "Sleeping for apt lists lock"
 		sleep 3
 	done
 	if [ -f /var/log/unattended-upgrades/unattended-upgrades.log ]; then
 		echo "Sleeping until unattended-upgrades finishes"
-		while fuser /var/log/unattended-upgrades/unattended-upgrades.log 2>/dev/null 2>&1; do
+		while fuser /var/log/unattended-upgrades/unattended-upgrades.log >/dev/null 2>&1; do
 			sleep 3
 		done
 	fi
@@ -165,8 +165,7 @@ main() {
 	fi
 	pip_version=$(${python_path} -m pip --version | sed -nr 's/^pip ([0-9]+\.[0-9]+).*$/\1/p')
 	pip_major_version=$(echo "${pip_version}" | cut -d '.' -f 1)
-	echo "pip version: ${pip_version}"
-	if [ "${pip_major_version}" -lt "${REQ_PIP_VERSION}" ]; then
+	if [ "${pip_major_version}" -lt "${REQ_PIP_MINOR_VERSION}" ]; then
 		${python_path} -m pip install --upgrade pip
 	fi
 
--- a/modules/scripts/startup-script/examples/install_ansible.sh
+++ b/modules/scripts/startup-script/examples/install_ansible.sh
@@ -14,40 +14,117 @@
 # limitations under the License.
 
 apt_wait() {
-	while fuser /var/lib/dpkg/lock >/dev/null 2>&1; do
+	while fuser /var/lib/dpkg/lock 2>/dev/null 2>&1; do
 		echo "Sleeping for dpkg lock"
 		sleep 3
 	done
-	while fuser /var/lib/apt/lists/lock >/dev/null 2>&1; do
+	while fuser /var/lib/apt/lists/lock 2>/dev/null 2>&1; do
 		echo "Sleeping for apt lists lock"
 		sleep 3
 	done
 	if [ -f /var/log/unattended-upgrades/unattended-upgrades.log ]; then
 		echo "Sleeping until unattended-upgrades finishes"
-		while fuser /var/log/unattended-upgrades/unattended-upgrades.log >/dev/null 2>&1; do
+		while fuser /var/log/unattended-upgrades/unattended-upgrades.log 2>/dev/null 2>&1; do
 			sleep 3
 		done
 	fi
 }
 
-if [ ! -h /usr/bin/ansible-playbook ] || [ ! -f /usr/bin/ansible-playbook ]; then
-	if [ -f /etc/centos-release ] || [ -f /etc/redhat-release ] || [ -f /etc/oracle-release ] || [ -f /etc/system-release ]; then
-		if [ ! -f /bin/pip ]; then
-			curl -Os https://bootstrap.pypa.io/pip/2.7/get-pip.py
-			/usr/bin/python get-pip.py
-		fi
-		/usr/bin/python -m pip install virtualenv
-		/usr/bin/python -m virtualenv /usr/local/toolkit
-		/usr/local/toolkit/bin/python -m pip install wheel
-		/usr/local/toolkit/bin/python -m pip install ansible==2.9.27
-		ln -s /usr/local/toolkit/bin/ansible-playbook /usr/bin/ansible-playbook
-	elif [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release || grep -qi ubuntu /etc/os-release; then
-		echo 'WARNING: unsupported installation of ansible in debian / ubuntu'
-		apt_wait
-		apt-get update
-		DEBIAN_FRONTEND=noninteractive apt-get install -y ansible
-	else
-		echo 'Unsupported distribution'
-		exit 1
+# Installs any dependencies needed for python based on the OS
+install_python_deps() {
+	if [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release 2>/dev/null ||
+		grep -qi ubuntu /etc/os-release 2>/dev/null; then
+		apt install -y python3-distutils
 	fi
-fi
+}
+
+# Gets the name of the python executable for python starting with python3, then
+# checking python. Sets the variable to an empty string if neither are found.
+get_python_path() {
+	python_path=""
+	if [ -f /bin/python3 ]; then
+		python_path="/bin/python3"
+	elif [ -f /bin/python ]; then
+		python_path="/bin/python"
+	fi
+}
+
+# Returns the python major version. If provided, it will use the first argument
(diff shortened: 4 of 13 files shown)


modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+	curl -Os ${get_pip_url}
+	${python_path} get-pip.py
+
+	# Create pip virtual environment for HPC Toolkit
+	${python_path} -m pip install virtualenv
+	${python_path} -m virtualenv /usr/local/ghpc-venv
+	python_path=/usr/local/ghpc-venv/bin/python3
+
+	# Install ansible
+	${python_path} -m pip install ansible==4.10.0
```
nick-stroud, line 127: Previously we were doing `pip install wheel` also. Not fully sure the purpose. Was there a reason for dropping it?

modules/scripts/startup-script/templates/startup-script-custom.tpl
```diff
@@ -1,12 +1,18 @@
 
 
 stdlib::run_playbook() {
+  if [ -d /usr/local/ghpc-venv ]; then
+    . /usr/local/ghpc-venv/bin/activate
```
nick-stroud, line 5: We are somewhat hiding the ansible install from the user. Say a user runs a shell script, that then calls ansible. They will probably get that `ansible-playbook is not a command`. Is this by design? What would be our recommendation to that user. 

The previous install did a `ln -s /usr/local/toolkit/bin/ansible-playbook /usr/bin/ansible-playbook` to make it accessible to others.

tools/validate_configs/test_configs/rocky-linux.yaml
```diff
@@ -53,8 +53,6 @@ deployment_groups:
       spack_url: https://github.com/spack/spack
       spack_ref: v0.17.0
       spack_cache_url:
-      - mirror_name: 'gcs_cache'
-        mirror_url: gs://example-buildcache/linux-centos7
```
nick-stroud, line 57: Should this have been replaced by something? I see you are removing them in several places. I am missing context.

community/modules/file-system/nfs-server/README.md
```diff
@@ -6,6 +6,10 @@ files with other clients over a network. `nfs-server` can be used by
 [vm-instance](../../../../modules/compute/vm-instance/README.md) and SchedMD
```
nick-stroud, line 6: I see the following behavior changes coming from this PR:
- We now may be installing python in addition to Ansible. Python takes much longer to install. People may be surprised to see a jump in startup time. I guess work around for now would be to tell people to use an older version of the script, or pick an image that already has python3.
- We are limiting access to the installed ansible behind the virtual environment. If I was installing ansible and expecting it to be available on my vm, I might be surprised to find it is not there.

Consider some documentation update.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+}
+
+# Gets the name of the python executable for python starting with python3, then
+# checking python. Sets the variable to an empty string if neither are found.
+get_python_path() {
+	python_path=""
+	if [ -f /bin/python3 ]; then
+		python_path="/bin/python3"
+	elif [ -f /bin/python ]; then
+		python_path="/bin/python"
```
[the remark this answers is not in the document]
heyealex: In the images I tested with, /bin/python3 was typically the source of /usr/bin/python3 - both existed. By looking for /bin/python{3,}, we are only looking for the system level python. Any other installation may be in user space and not something we want to interfere with.

Alternatively, this could be a variable, letting the user specify the correct python to use in the startup script module, defaulting to /bin/python3.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+	python_path=${1:-python}
+	python_minor_version=$(${python_path} -c "import sys; print(sys.version_info.minor)")
+}
+
+# Install python3 with the yum package manager. Updates python_path to the
+# newly installed packaged.
+install_python3_yum() {
+	## TODO restrict repos to search through in centos to decrease overhead
+	yum install -y python3
+	python_path=$(rpm -ql python3 | grep 'bin/python3$')
```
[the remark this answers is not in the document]
heyealex: Python3 is not installed on centos7, so this path was tested.

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: I believe that was related to our use of python2. I'll ask @tpdownes to verify.

modules/scripts/startup-script/templates/startup-script-custom.tpl
heyealex: I like keeping our environment isolated (open to discussion of course).
This should be documented however. We should tell the user how to user and/or copy the environment either in their scripts or post-install. I'll include that in the next push.

tools/validate_configs/test_configs/rocky-linux.yaml
heyealex: I removed them for testing. I can add them back in, but I'm not sure how much value these lines add to the validate configs test; other tests cover a broad range of spack-install variables.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+# as the python executable, otherwise it will default to simply "python".
+get_python_minor_version() {
+	python_path=${1:-python}
+	python_minor_version=$(${python_path} -c "import sys; print(sys.version_info.minor)")
+}
+
+# Install python3 with the yum package manager. Updates python_path to the
+# newly installed packaged.
+install_python3_yum() {
+	## TODO restrict repos to search through in centos to decrease overhead
```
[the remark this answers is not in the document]
tpdownes: "base" and "updates" are, I believe, the correct answer. From RHEL 7.7 onward, the OS repos themselves contain Python 3 and EPEL is, by its own policies, prohibited from duplicating packages in the OS repos.

I may be getting 7.7 wrong. Could have been 7.6. Point being, the status quo is as I describe.

modules/scripts/startup-script/examples/install_ansible.sh
tpdownes: Strictly speaking, I believe it was duplicative in the old script because the get-pip bootstrap script also installed it.

That said, Ansible prefers to build a wheel when you run `pip install ansible` rather than default to the legacy `setup.py` approach. If you see words to that effect, it means `wheel` is not installed. It probably should be.

modules/scripts/startup-script/templates/startup-script-custom.tpl
tpdownes: Among the issues with not having it in the system path is the fact that we would have to rewrite all our playbooks to use a specific path for the ansible binary and the python binary. _This may be a good thing ensuring that our playbooks use a specific release even if other 3rd party playbooks do as they see fit_. But, it's a thing tp be discussed.

This is at least 1 part of what we'd have to look at:

https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html

tpdownes: Whether we link to `ansible-playbook` from a default PATH location is at least somewhat distinct from the claim that we are or are not isolating the environment. It's not like we'd be copying everything in to a system directory.

community/modules/file-system/nfs-server/README.md
tpdownes: There is a problem here in that CentOS 7 is a massive outlier in that it does not have Python 3 baked in while everything else will. Of course, it's also very popular. The advice I gave @heyealex on this was not to use `yum install python3` but to use something like `yum install https://path/to/python3.rpm` where the URL is guaranteed (or nearly so) to live "forever".  We identified a few together. The advantage is that (I believe) it will bypass all repo analysis in favor of just trying to install the RPM (and dependencies). So it might need to be a short sequence of yum  (or `rpm -U`) install commands to install the dependencies in order. Likewise, we could consider a short-lived GCS/HTTPS mirror of these RPMs.

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: Quick test on centos7 to confirm what I remembered seeing:
Installing pip with get-pip.py also installs wheel by default:
```shell
[heyealex@centos7-test-1 ~]$ curl -Os https://bootstrap.pypa.io/pip/3.6/get-pip.py
[heyealex@centos7-test-1 ~]$ python3 get-pip.py 
Defaulting to user installation because normal site-packages is not writeable
Collecting pip<22.0
  Downloading pip-21.3.1-py3-none-any.whl (1.7 MB)
     |████████████████████████████████| 1.7 MB 8.0 MB/s            
Collecting wheel
  Downloading wheel-0.37.1-py2.py3-none-any.whl (35 kB)
Installing collected packages: wheel, pip
Successfully installed pip-21.3.1 wheel-0.37.1
```

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+	yum install -y python3
+	python_path=$(rpm -ql python3 | grep 'bin/python3$')
+}
+
+# Install python3 with the apt package manager. Updates python_path to the
+# newly installed packaged.
+install_python3_apt() {
+	apt_wait
+	apt-get install -y python3 python3-distutils
```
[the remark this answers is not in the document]
cboneti: I would actually leave it here and only call install_python_deps IF we do not install python3. That would effectively avoid calling apt-get twice.

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
heyealex: I did a bit more testing, and one OS did in fact fail to match that pattern, ubuntu. For the sake of not adding significantly more complexity to handle that one case, I'm adding a 3rd fallback option for `which python3`. 

The changes has been added and pushed.

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
heyealex: I added the same enabled repos as we did with the nfs utils install, which is "base,epel" for centos7 and "baseos" for 8. I'm leaving it as unsupported for anything beyond that for now.

The fix has been added and pushed.

modules/scripts/startup-script/templates/startup-script-custom.tpl
heyealex: I added a section to the README that tells the user about this environment and how to use it.

tools/validate_configs/test_configs/rocky-linux.yaml
heyealex: As we discussed offline, I left them out, and removed them consistently for all of these *-ss.yaml tests.

community/modules/file-system/nfs-server/README.md
heyealex: I looked into using rpms for installing python, and the result was not very clean. It requires us to track down the dependencies ourselves, many of which also have dependencies. This doesn't seem like a good approach unfortunately.

For the yum install, I did add in the login to only use specified repos, just like what we did for the nfs-utils install. This should help cut down on installation time at least.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
-		apt_wait
-		apt-get update
-		DEBIAN_FRONTEND=noninteractive apt-get install -y ansible
+	fi
+	install_python_deps
+
+	# Install and/or upgrade pip
+	get_python_minor_version "${python_path}"
+	if [ "${python_minor_version}" -lt 7 ]; then
+		get_pip_url="https://bootstrap.pypa.io/pip/${python_major_version}.${python_minor_version}/get-pip.py"
```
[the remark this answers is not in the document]
heyealex: Fixed

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
heyealex: Fixed

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -14,40 +14,117 @@
...
+	else
+		echo "Error: Unsupported Distribution"
+		return 1
+	fi
+}
+
+main() {
+	# Get the python3 executable, or install it if not found
+	get_python_path
+	if [ "${python_path}" = "" ]; then
```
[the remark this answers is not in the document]
heyealex: Fixed.

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
heyealex: I set the variable in the script as a global for now. Having to convert this script to a template will be a bit more complicated and disruptive than I anticipated. That can come in a follow-up PR, I think it would make more sense alongside the PR where we automate adding the install script when ansible-local runners are detected.

heyealex:
I believe everything has been addressed. Please take a look again when you get a chance.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -13,41 +13,183 @@
...
+# Gets the name of the python executable for python starting with python3, then
+# checking python. Sets the variable to an empty string if neither are found.
+get_python_path() {
+	python_path=""
+	if [ -f /bin/python3 ]; then
+		python_path="/bin/python3"
+	elif [ -f /bin/python ]; then
+		python_path="/bin/python"
+	else
+		python_path=$(which python3 2>/dev/null)
```
cboneti, line 55: nit: what if only python2 is in the path?

modules/scripts/startup-script/templates/startup-script-custom.tpl
nick-stroud: @tpdownes, When you mention having to rewrite our playbooks to invoke a specific path, are you thinking of ansible running as runners? I believe that any ansible running as a runner will have the virtualenv activated before running so it will be added to the path (and then deactivated afterward). This will only affect ansible that is being run as a non-runner, such as after startup.

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: Thanks for catching this, changed this to an elif. The variable should only be set if python3 is found.

cboneti: and what if python is in the path? should you not register it?

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -13,41 +13,181 @@
...
+		return 1
+	fi
+}
+
+main() {
+	# Get the python3 executable, or install it if not found
+	get_python_path
+	get_python_major_version "${python_path}"
+	get_python_minor_version "${python_path}"
+	if [ "${python_path}" = "" ] || [ "${python_major_version}" = "2" ] || [ "${python_minor_version}" -lt "${REQ_PYTHON3_VERSION}" ]; then
```
heyealex, line 150: Change the variable name to REQ_PYTHON3_MINOR_VERSION

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -13,41 +13,181 @@
...
+	if which ansible-playbook 2>/dev/null; then
+		ansible_version=$(ansible-playbook --version 2>/dev/null | sed -nr 's/^ansible-playbook.*([0-9]+\.[0-9]+\.[0-9]+).*/\1/p')
+		ansible_major_vers=$(echo "${ansible_version}" | cut -d '.' -f 1)
+		ansible_minor_vers=$(echo "${ansible_version}" | cut -d '.' -f 2)
+		ansible_req_major_vers=$(echo "${REQ_ANSIBLE_VERSION}" | cut -d '.' -f 1)
+		ansible_req_minor_vers=$(echo "${REQ_ANSIBLE_VERSION}" | cut -d '.' -f 2)
+	fi
+	if [ -z "${ansible_version}" ] || [ "${ansible_major_vers}" -ne "${ansible_req_major_vers}" ] ||
+		[ "${ansible_minor_vers}" -lt "${ansible_req_minor_vers}" ]; then
+		${python_path} -m pip install ansible==${REQ_ANSIBLE_PIP_VERSION}
```
tpdownes, line 189: This is making the choice to install 4.10.0 on all platforms (aligning all platforms to the last version compatible with CentOS 7 and Rocky 8). That is a reasonable choice. It might also be reasonable to install whatever `pip` thinks is the best version. Anyhow, I think this is a choice we should all understand.

In particular, I cannot find documentation for 4.10.0 at https://docs.ansible.com/

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -13,41 +13,181 @@
...
-if [ ! -h /usr/bin/ansible-playbook ] || [ ! -f /usr/bin/ansible-playbook ]; then
-	if [ -f /etc/centos-release ] || [ -f /etc/redhat-release ] || [ -f /etc/oracle-release ] || [ -f /etc/system-release ]; then
-		if [ ! -f /bin/pip ]; then
-			curl -Os https://bootstrap.pypa.io/pip/2.7/get-pip.py
-			/usr/bin/python get-pip.py
+# Installs any dependencies needed for python based on the OS
+install_python_deps() {
+	if [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release 2>/dev/null ||
+		grep -qi ubuntu /etc/os-release 2>/dev/null; then
+		apt-get install -y python3-distutils
```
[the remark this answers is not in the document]
heyealex: I ran into a dependency when upgrading pip in ubuntu and debian that required distutils to be installed first.

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: The pip package 4.10.0 will install version 2.11 of ansible-core, which I am able to find listed in the documentation: https://docs.ansible.com/ansible-core/2.11/index.html

It is however the oldest version listed currently. 

I'm inclined to keep the hard version in place for now and investigate installing as `python -m pip install ansible>=4.10.0` when we can run further tests against it (I'll a task for it in the backlog). How do you all feel about that?

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: Updated to only use which, moving away from hard coded paths.

modules/scripts/startup-script/README.md
```diff
@@ -53,6 +53,34 @@ Each runner receives the following attributes:
...
+version 2.11 of ansible-core or higher. Note that this is distinct from the
+package version used to install ansible with pip. The minimum pip package
+of ansible is 4.10.0.
+
+To install ansible, a runner supplied by this module can be added as a prior
+runner. An example of this can be found in the [Example](#example) section below
+as the first runner in the list of runners. This script will do the following in
+your VM instance:
+
+- Install python3 if not already installed using system package managers (yum,
```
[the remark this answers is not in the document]
heyealex: Updated

modules/scripts/startup-script/README.md
```diff
@@ -53,6 +53,34 @@ Each runner receives the following attributes:
...
+  apt-get, etc)
+- Install pip3 if not already installed and upgrade pip3 if the version is not
+  at least 18.0.
+- Install and create a virtualenv located at `/usr/local/ghpc-venv`.
```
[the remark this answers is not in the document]
heyealex: I've heard people refer to a virtual environment created by virtualenv as a virtualenv. Doesn't mean it's correct, but it slipped into my vocabulary since.

Updated to the correct verbiage.

modules/scripts/startup-script/examples/install_ansible.sh
```diff
@@ -13,41 +13,181 @@
...
 # limitations under the License.
 
+REQ_ANSIBLE_VERSION=2.11
+REQ_ANSIBLE_PIP_VERSION=4.10.0
+REQ_PIP_VERSION=18
+REQ_PYTHON3_VERSION=6
+
 apt_wait() {
-	while fuser /var/lib/dpkg/lock >/dev/null 2>&1; do
+	while fuser /var/lib/dpkg/lock 2>/dev/null 2>&1; do
```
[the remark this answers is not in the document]
heyealex: Fixed

modules/scripts/startup-script/examples/install_ansible.sh
heyealex: Fixed

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
tpdownes: I'd put a comment somewhere in the script that captures this. It's not obvious from the package description:

https://packages.debian.org/bullseye/python3-distutils

modules/scripts/startup-script/examples/install_ansible.sh
tpdownes: I like that idea as the best way to make progress.

modules/scripts/startup-script/README.md
```diff
@@ -53,6 +53,34 @@ Each runner receives the following attributes:
...
+
+- Install system-wide python3 if not already installed using system package
+  managers (yum, apt-get, etc)
+- Install system-wide pip3 if not already installed and upgrade pip3 if the
+  version is not at least 18.0.
+- Install and create a virtual environment located at `/usr/local/ghpc-venv`.
+- Install ansible into this virtual environment if the current version of
+  ansible is not version 2.11 or higher.
+
+To use the virtualenv created by this script, you can activate it by running the
```
tpdownes, line 77: please use "virtual environment"

modules/scripts/startup-script/templates/startup-script-custom.tpl
tpdownes: I think the main issue is running Ansible by some other means.

(1) I think the README should make more clear that Ansible will only be in the PATH if you activate the virtual environment
(2) If one wants to invoke Ansible via SSH, do we need to set ansible_python_interpreter?

I am happy to address (2) as a new task for after the PR. I do think (1) should be more clear in the PR.

heyealex: (1) has been documented. I also added a sentence on (2) and a suggestion of setting the `ansible_python_interpreter` variable when using ansible in the environment. A task has been added to investigate better ways of handling that.

modules/scripts/startup-script/examples/install_ansible.sh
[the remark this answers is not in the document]
heyealex: Done

modules/scripts/startup-script/README.md
heyealex: Done

nick-stroud (approved):

community/modules/scripts/spack-install/templates/install_spack.tpl
```diff
@@ -9,6 +9,11 @@ if [[ $EUID -ne 0 ]]; then
   exit 1
 fi
 
+# Activate ghpc-venv virtual environment if it exists
+if [ -d /usr/local/ghpc-venv ]; then
+  source /usr/local/ghpc-venv/bin/activate
```
cboneti, line 14: This is a sad side repercussion of having virtual env. Should we rather fix all scripts that expect ansible, or just install in the system?

heyealex: Yeah, I don't like this. 
Right now, I'm tempted to pull the plug on the virtual environment for now and put it back in later when I/we have a better understanding of the effects.

nick-stroud: You could do what @tpdownes had done before which is to still install into a virtual env but then symbolic link to the that virtual env, exposing it to the system: [example](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8525bf230cbf67ae30d58295243c777ee82f7f24/modules/scripts/startup-script/examples/install_ansible.sh#L43).

I am also fine to do that as a follow on as we have people waiting for a newer version of ansible as of this moment.

nick-stroud (approved):

community/modules/scripts/spack-install/templates/install_spack.tpl
tpdownes: It is also possible to invoke any pip-installed binary with a full path. It will do the "right thing" in terms of Python libraries e.g. swap `/usr/bin/ansible-playbook` for `/usr/local/ghpc-venv/bin/ansible-playbook`.

PR #489, opened 2022-08-22
Fixing deployment_name conflict

thiagosgobe:
Fixing conflict when working with multiple blueprints by given it a unique deployment_name.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/slurm-gcp-v5-hpc-centos7.yaml
+++ b/community/examples/slurm-gcp-v5-hpc-centos7.yaml
@@ -18,7 +18,7 @@ blueprint_name: slurm-gcp-v5-hpc-centos7
 
 vars:
   project_id:  ## Set GCP Project ID Here ##
-  deployment_name: slurm-gcp-v5
+  deployment_name: slurm-gcp-v5-centos7
   region: us-central1
   zone: us-central1-c
 
--- a/community/examples/slurm-gcp-v5-hpc-centos7.yaml
+++ b/community/examples/slurm-gcp-v5-hpc-centos7.yaml
@@ -18,7 +18,7 @@ blueprint_name: slurm-gcp-v5-hpc-centos7
 
 vars:
   project_id:  ## Set GCP Project ID Here ##
-  deployment_name: slurm-gcp-v5-centos7
+  deployment_name: slurm-centos-v5
   region: us-central1
   zone: us-central1-c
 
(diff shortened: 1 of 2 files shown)


cboneti:
Does that really work? I believe deployment name gets shortened to the first 8 characters for the slurm resources and this change only differs at the end of the deployment name. Did you deploy both simultaneously?

cboneti (changes requested):
Did you test if this really work deploying the two at the same time? I believe deployment name gets shortened to the first 8 characters for the slurm resources and this change only differs at the end of the deployment name. Did you deploy both simultaneously?

thiagosgobe:
> Did you test if this really work deploying the two at the same time? I believe deployment name gets shortened to the first 8 characters for the slurm resources and this change only differs at the end of the deployment name. Did you deploy both simultaneously?

You are right. I changed the name again in order to differ when the name is shorted and this time I tested deploying both simultaneously. Thank you!

community/examples/slurm-gcp-v5-hpc-centos7.yaml
```diff
@@ -18,7 +18,7 @@ blueprint_name: slurm-gcp-v5-hpc-centos7
 
 vars:
   project_id:  ## Set GCP Project ID Here ##
-  deployment_name: slurm-gcp-v5
```
cboneti, line 21: V5 pertains to slurm-gcp version, so I think your changes make it a little confusing. What about moving the distro to the beginning?  eg: 
centos-slurm-gcp-v5
ubuntu-slurm-gcp-v5

cboneti (changes requested):

community/examples/slurm-gcp-v5-hpc-centos7.yaml
thiagosgobe: > V5 pertains to slurm-gcp version, so I think your changes make it a little confusing. What about moving the distro to the beginning? eg: centos-slurm-gcp-v5 ubuntu-slurm-gcp-v5

Make sense. I would like to keep at least the "slurm" after name get shorted but I don't like the idea of abbreviate distro. I can't think other possibilities given the scenario. If you have any, will be welcome otherwise let go with them.

cboneti: Ok, so I talked with the team internally and we believe there is no good solution here. We will talk to SchedMD and try to remove this limitation of 10 characters in the future. 
Let's close this PR now and accept that these two deployments will temporarily clash until we fix this in a better way. 
What do you think?

cboneti:
I talked with the team internally and we believe there is no good solution here. We will talk to SchedMD and try to remove this limitation of 10 characters in the future.
Let's close this PR now and accept that these two deployments will temporarily clash until we fix this in a better way.

PR #493, opened 2022-08-23
Synchronize develop with release branch

tpdownes:
This PR will have the effect of merging #492 into develop so that main and develop remain synchronized.

### Submission Checklist

  install`)


nick-stroud (changes requested):
There is a bug in the wait for on lustre.

Before we had `lustre` waiting for `omnia` waiting for `monitoring`

The desired state after disabling would be `luster` waiting for `monitoring`

The disable pr removed wait for from the luster test but did not establish the wait for the monitoring, so now we will have tests running in parallel

nick-stroud (approved):

PR #506, opened 2022-08-29
Document use of -parallelism when using compact placement with vm-ins…

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/modules/compute/vm-instance/README.md
+++ b/modules/compute/vm-instance/README.md
@@ -60,6 +60,13 @@ Use the following settings for compact placement:
       availability_domain_count: null
 ```
 
+> **Warning** When creating a compact placement with more than 10 VMs, you must
+> add `-parallelism=<n>` argument on apply. For example if you have 15 VMs in a
+> placement group: `terraform apply -parallelism=15`. This is because terraform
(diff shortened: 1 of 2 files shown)


modules/compute/vm-instance/README.md
```diff
@@ -60,6 +60,13 @@ Use the following settings for compact placement:
       availability_domain_count: null
 ```
 
+> **Warning** When creating a compact placement with more than 10 VMs, you must
+> add `-parallelism=<n>` argument on apply. For example if you have 15 VMs in a
+> placement group: `terraform apply -parallelism=15`. This is because terraform
+> self limits to 10 parallel requests by default but the create instance
+> requests will not succeed until all VMs in the placement group have been
+> requested, forming a deadlock.
+
```
cboneti, line 69: My understanding is that this will only happen when deploying more than 10 VMs.  Could we warn the user during the ghpc run?

cboneti (approved):

modules/compute/vm-instance/README.md
nick-stroud: I have created a task along these lines: We could add the argument into the suggested terraform commands. Some offline discussion on that topic would be good.

PR #508, opened 2022-08-30
Add validator test_apis_enabled

tpdownes:
- add infrastructure for new validator
- implement stub function only for validation
- implement basic unit testing of stub function
- TODO: extend stub function with API calls implementing test

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/validate.go
+++ b/pkg/config/validate.go
@@ -282,6 +282,7 @@ func (dc *DeploymentConfig) getValidators() map[string]func(validatorConfig) err
 		testRegionExistsName.String():  dc.testRegionExists,
 		testZoneExistsName.String():    dc.testZoneExists,
 		testZoneInRegionName.String():  dc.testZoneInRegion,
+		testApisEnabledName.String():   dc.testApisEnabled,
 	}
 	return allValidators
 }
@@ -471,4 +472,20 @@ func (dc *DeploymentConfig) getStringValue(inputReference interface{}) (string, 
 		}
 	}
 	return "", fmt.Errorf("the value %s is not a deployment variable or was not defined", inputReference)
+}
+
+func (dc *DeploymentConfig) testApisEnabled(validator validatorConfig) error {
+	requiredInputs := []string{}
+	funcName := testApisEnabledName.String()
+
+	if validator.Validator != funcName {
+		return fmt.Errorf("passed wrong validator to %s implementation", funcName)
+	}
+
+	err := testInputList(validator.Validator, validator.Inputs, requiredInputs)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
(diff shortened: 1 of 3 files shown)


tpdownes:
I think there is some potential for reduced replication of this pattern in the validators:

```go
        requiredInputs := []string{"project_id", "region", "zone"}
        funcName := testZoneInRegionName.String()
        funcErrorMsg := fmt.Sprintf("validator %s failed", funcName)

        if validator.Validator != funcName {
                return fmt.Errorf("passed wrong validator to %s implementation", funcName)
        }

        err := testInputList(validator.Validator, validator.Inputs, requiredInputs)
        if err != nil {
                return err
        }
...
```

I prefer to capture this as a task.

pkg/config/validate.go
```diff
@@ -472,3 +473,19 @@ func (dc *DeploymentConfig) getStringValue(inputReference interface{}) (string,
 	}
 	return "", fmt.Errorf("the value %s is not a deployment variable or was not defined", inputReference)
 }
+
+func (dc *DeploymentConfig) testApisEnabled(validator validatorConfig) error {
```
nick-stroud, line 477: move function above `getStringValue` to be adjacent with other validators. 

If we expect this validator to be run first then it might even be best to put it first in the file to keep logical ordering.

pkg/config/validate.go
```diff
@@ -282,6 +282,7 @@ func (dc *DeploymentConfig) getValidators() map[string]func(validatorConfig) err
 		testRegionExistsName.String():  dc.testRegionExists,
 		testZoneExistsName.String():    dc.testZoneExists,
 		testZoneInRegionName.String():  dc.testZoneInRegion,
+		testApisEnabledName.String():   dc.testApisEnabled,
```
nick-stroud, line 285: I don't think the ordering here has any effect but since we think we are going to run this validator first, to me it makes most sense to declare it first in the list.

pkg/config/validate.go
```diff
@@ -472,3 +473,19 @@ func (dc *DeploymentConfig) getStringValue(inputReference interface{}) (string,
...
+	if validator.Validator != funcName {
+		return fmt.Errorf("passed wrong validator to %s implementation", funcName)
+	}
+
+	err := testInputList(validator.Validator, validator.Inputs, requiredInputs)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
nick-stroud, line 490: Add a TODO here to actually run the validator. 

My personal preference would have been to write the function first, but I like that you are splitting it up.

pkg/config/validate.go
tpdownes: Done on `HEAD` of PR

pkg/config/validate.go
tpdownes: Done on `HEAD` of PR

pkg/config/validate.go
tpdownes: Done on `HEAD` of PR

nick-stroud (approved):

PR #514, opened 2022-09-01
Adds option to attach local SSD to vm-instance module

nick-stroud:
Tested:
- Bad input for interface
- Bad input for count (based on vm type)

In all cases error message was informative and showed upon terraform apply.

### Submission Checklist

  install`)

### DIFF
--- a/modules/compute/vm-instance/variables.tf
+++ b/modules/compute/vm-instance/variables.tf
@@ -49,6 +49,18 @@ variable "disk_type" {
   default     = "pd-standard"
 }
 
+variable "local_ssd_count" {
+  description = "The number of local SSD to attach to each VM. See https://cloud.google.com/compute/docs/disks/local-ssd."
+  type        = number
+  default     = 0
+}
+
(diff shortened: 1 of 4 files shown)


modules/compute/vm-instance/variables.tf
```diff
@@ -49,6 +49,18 @@ variable "disk_type" {
   default     = "pd-standard"
 }
 
+variable "local_ssd_count" {
+  description = "The number of local SSD to attach to each VM. See https://cloud.google.com/compute/docs/disks/local-ssd."
```
cboneti, line 53: nit: local SSDs? or local SSD?

cboneti (approved):
just one nit question, feel free to pick what you think is best.

cboneti (changes requested):
Could you add at least some validation tests in https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/tools/validate_configs/test_configs ?

cboneti (approved):

PR #515, opened 2022-09-01
Add install path variable to hybrid

heyealex:
A workaround to simplify hybrid deployment off controller. Rather than
have the full absolute path of the output_dir as the resume/suspend
program paths in cloud.conf, update the file with a sed command
replacing the output_dir with a provided install_dir, which acts
essentially as the prefix variable for the installation.

Alternatives attempted: Using a local_file block to read the data from
the generated cloud.conf file, use replace to update the content and
place the context back into the same file. This does not work as the
file does not exist at plan time, therefore terraform will fail.

Longer term solution: Build this into the setup_hybrid.py script used by
slurm-gcp in the underlying hybrid module.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/README.md
@@ -163,10 +163,13 @@ limitations under the License.
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.14.0 |
+| <a name="requirement_null"></a> [null](#requirement\_null) | ~> 3.0 |
 
 ## Providers
 
-No providers.
+| Name | Version |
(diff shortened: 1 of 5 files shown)


tpdownes:
@heyealex in my limited testing with the test config, this appears to be idempotent (i.e. the script doesn't continually have to "fix" the Terraform output). Will this always be the case?

heyealex:
Thanks for pointing that out, I did a bit more testing and made some changes to ensure that the install prefix is updated even on a re-apply if either variable output_dir or install_dir are updated. It will not run otherwise (though if it did, it should have no effect).

Verified:
* Works when install_dir is set
* Works when install_dir is not set
* Works when install_dir changes on second apply
* Works when output_dir changes on second apply
* Works when neither install_dir or output_dir change on a second apply (doesn't do anything)

tpdownes (approved):

PR #518, opened 2022-09-02
Making 'kind' default to terraform when omitted

cboneti:
In this PR, I am not changing google-sided documentation to indicate this is now optional, I am also not changing all of the examples. I believe these should be new tasks. 

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -216,6 +216,7 @@ type DeploymentConfig struct {
 
 // ExpandConfig expands the yaml config in place
 func (dc *DeploymentConfig) ExpandConfig() {
+	dc.addKindToModules()
 	dc.setModulesInfo()
 	dc.validateConfig()
 	dc.expand()
@@ -327,6 +328,18 @@ func createModuleInfo(
 		}
 	}
 	return modInfo
+}
+
+// addKindToModules sets the kind to 'terraform' when empty.
+func (dc *DeploymentConfig) addKindToModules() {
+	for iGrp, grp := range dc.Config.DeploymentGroups {
+		for iMod, mod := range grp.Modules {
+			if mod.Kind == "" {
+				dc.Config.DeploymentGroups[iGrp].Modules[iMod].Kind =
+					"terraform"
+			}
+		}
+	}
 }
 
 // setModulesInfo populates needed information from modules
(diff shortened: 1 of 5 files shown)


pkg/modulereader/resreader.go
```diff
@@ -52,6 +52,8 @@ type ModReader interface {
 }
 
 var kinds = map[string]ModReader{
+	// The default kind, when empty, is terraform
+	"":          TFReader{allModInfo: make(map[string]ModuleInfo)},
```
nick-stroud, line 56: My personal preference would be to replace empty string with `terraform` in the expand step so that the expanded blueprint still says `terraform`. WDYT? 

I would approve as is because it is a huge improvement and we could always add back the behavior I describe later, but wondering if you thought about the expanded blueprint.

cboneti: That was, and continues to be my preference as well. I considered 2-3 alternatives and we can talk about them offline. The TLDR is that originally, I thought this could be a one-line change in the reader and so I went for it. I also wanted to add another dependency (defaults). Let me take another look at this.

cboneti: Please, take a look at this alternative now. :)

pkg/config/config.go
```diff
@@ -329,6 +330,18 @@ func createModuleInfo(
 	return modInfo
 }
 
+// addKindToModules sets the kind to 'terraform' when empty.
+func (dc *DeploymentConfig) addKindToModules() {
```
nick-stroud, line 334: Can you add a test in `pkg/config/config_test.go` for this function. I think just something simple that passes in a `dc` that has empty string and verify that it comes out with terraform for kind. 

If you wanted to get fancier you could also test that `terraform` and `packer` do not change when defined explicitly, but I would settle for just the case above.

cboneti: good point. PTAL

nick-stroud (approved):

pkg/config/config_test.go
```diff
@@ -239,13 +239,81 @@ func getBasicDeploymentConfigWithTestModule() DeploymentConfig {
 	}
 }
 
+func getDeploymentConfigWithTestModuleEmtpyKind() DeploymentConfig {
```
tpdownes, line 242: Empty

PR #519, opened 2022-09-02
Add required API data

tpdownes:
- add fields to structs
- populate blueprint config from module info
- add static map of required APIs for embedded modules

TODO: populate module required APIs from metadata within modules
TODO: document full solution (will be done in future PR that connects this work to #508)

### Submission Checklist

  install`)
* [] Have you updated all applicable documentation?

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -45,7 +45,7 @@ const (
 func (dc *DeploymentConfig) expand() {
 	dc.addSettingsToModules()
 	if err := dc.addMetadataToModules(); err != nil {
-		log.Printf("could not add module metadata: %v", err)
+		log.Printf("could not determine required APIs: %v", err)
 	}
 
 	if err := dc.expandBackends(); err != nil {
@@ -92,7 +92,6 @@ func (dc *DeploymentConfig) addMetadataToModules() error {
 			if mod.RequiredApis == nil {
 				_, projectIDExists := dc.Config.Vars["project_id"].(string)
 				if !projectIDExists {
-					// it is OK to error
 					return fmt.Errorf("global variable project_id must be defined")
 				}
 
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -44,6 +44,10 @@ const (
 // ExpandConfig for the create and expand commands.
 func (dc *DeploymentConfig) expand() {
 	dc.addSettingsToModules()
+	if err := dc.addMetadataToModules(); err != nil {
+		log.Printf("could not add module metadata: %v", err)
+	}
+
 	if err := dc.expandBackends(); err != nil {
 		log.Fatalf("failed to apply default backend to deployment groups: %v", err)
 	}
@@ -80,6 +84,32 @@ func (dc *DeploymentConfig) addSettingsToModules() {
 			}
 		}
 	}
+}
+
+func (dc *DeploymentConfig) addMetadataToModules() error {
+	for iGrp, grp := range dc.Config.DeploymentGroups {
+		for iMod, mod := range grp.Modules {
+			if mod.RequiredApis == nil {
+				_, projectIDExists := dc.Config.Vars["project_id"].(string)
+				if !projectIDExists {
+					// it is OK to error
+					return fmt.Errorf("global variable project_id must be defined")
+				}
+
+				// handle possibility that ModulesInfo does not have this module in it
+				// this occurs in unit testing because they do not run dc.ExpandConfig()
+				// and dc.setModulesInfo()
+				requiredAPIs := dc.ModulesInfo[grp.Name][mod.Source].RequiredApis
+				if requiredAPIs == nil {
+					requiredAPIs = []string{}
+				}
+				dc.Config.DeploymentGroups[iGrp].Modules[iMod].RequiredApis = map[string][]string{
+					"(( var.project_id ))": requiredAPIs,
+				}
+			}
+		}
+	}
+	return nil
 }
 
 func (dc *DeploymentConfig) expandBackends() error {
--- a/pkg/sourcereader/embedded.go
+++ b/pkg/sourcereader/embedded.go
@@ -103,7 +103,123 @@ func (r EmbeddedSourceReader) GetModuleInfo(modPath string, kind string) (module
 	}
 
 	reader := modulereader.Factory(kind)
-	return reader.GetInfo(modDir)
+	mi, err := reader.GetInfo(modDir)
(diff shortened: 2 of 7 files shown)


pkg/sourcereader/embedded.go
```diff
@@ -103,7 +103,123 @@ func (r EmbeddedSourceReader) GetModuleInfo(modPath string, kind string) (module
...
+			"compute.googleapis.com",
+			"iam.googleapis.com",
+			"secretmanager.googleapis.com",
+		},
+		"community/modules/file-system/nfs-server": {
+			"compute.googleapis.com",
+		},
+		"community/modules/project/new-project": {
+			"admin.googleapis.com",
+			"appengine.googleapis.com",      // optional
```
heyealex, line 145: Should we be validating against these two if they are optional?

heyealex (changes requested):

pkg/sourcereader/embedded.go
```diff
@@ -103,7 +103,123 @@ func (r EmbeddedSourceReader) GetModuleInfo(modPath string, kind string) (module
...
+			"bigqueryconnection.googleapis.com",
+			"sqladmin.googleapis.com",
+		},
+		"community/modules/file-system/DDN-EXAScaler": {
+			"compute.googleapis.com",
+			"deploymentmanager.googleapis.com",
+			"iam.googleapis.com",
+			"runtimeconfig.googleapis.com",
+		},
+		"community/modules/file-system/Intel-DAOS": {
```
cboneti, line 135: This is an interesting example. The source for DAOS will never be the local module. There is no wrapper. It will be something like: 

source: github.com/daos-stack/google-cloud-daos.git//terraform/modules/daos_server?ref=v0.2.1

pkg/sourcereader/embedded.go
```diff
@@ -103,7 +103,123 @@ func (r EmbeddedSourceReader) GetModuleInfo(modPath string, kind string) (module
...
+	mi, err := reader.GetInfo(modDir)
+	mi.RequiredApis = defaultAPIList(modPath)
+	return mi, err
+}
+
+func defaultAPIList(source string) []string {
+	// API lists at
+	// https://console.cloud.google.com/apis/dashboard and
+	// https://console.cloud.google.com/apis/library
+	staticAPIMap := map[string][]string{
```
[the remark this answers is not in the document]
tpdownes: I prefer to keep the data confined to a region of code when we swap out this lookup for a solution that reads the values from modules (embedded or not). I also believe that it is not possible to create a `const map` in Go., if you literally mean a `const`.

Put another way: this is intended to be a short-lived stop-gap.

pkg/sourcereader/embedded.go
tpdownes: Similar issue to `pubsub` for SchedMD v5 Slurm solutions.

pkg/config/expand.go
```diff
@@ -82,6 +86,32 @@ func (dc *DeploymentConfig) addSettingsToModules() {
 	}
 }
 
+func (dc *DeploymentConfig) addMetadataToModules() error {
+	for iGrp, grp := range dc.Config.DeploymentGroups {
+		for iMod, mod := range grp.Modules {
+			if mod.RequiredApis == nil {
+				_, projectIDExists := dc.Config.Vars["project_id"].(string)
+				if !projectIDExists {
```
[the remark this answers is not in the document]
tpdownes: I definitely prefer to live in a world where I know that

- project ID exists and is at least potentially a valid Google Cloud project ID (length, characters, etc.)
- deployment name exists and is at least potentially a valid Google Cloud resource ID prefix (length, characters, etc.)

pkg/sourcereader/embedded.go
tpdownes: I understand that this particular element of the map will never be accessed, but I recorded this for completeness sake (and a starting point when we automate solution). Do you suggest removing it?

tpdownes:
I have responded (with questions) to a few of the issues posed. I don't think there are required changes just yet. Could you make sure to take a careful look at the unit test changes? I believe I've made the right set of changes, but I have made a "superset" of those changes that still happens to work.

pkg/sourcereader/embedded.go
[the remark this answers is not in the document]
heyealex: Sounds good, I figured something like this might have been the case.

pkg/sourcereader/embedded.go
heyealex: I don't want to be in a situation where we are halting execution because of a recommended, but not necessarily required API. The best case scenario in my mind is:
* Error when a required API is not enabled
* Warn when a recommended API is not enable
which of course requires that we track that piece of information alongside the API lists. 

For this case, my preference is to only include hard requirements for now.

pkg/sourcereader/embedded.go
heyealex: It's a good reminder that the DAOS modules will need to updated as well to incorporate the API requirements. I support keeping it in.

pkg/config/expand.go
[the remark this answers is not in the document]
heyealex: I see, the behavior here isn't any different since the error isn't fatal. How we handle the project_id requirement should be looked at further, and probably kept in line with how we handle deployment name now. 

I'll resolve this and add a task for it.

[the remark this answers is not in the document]
heyealex: Actual ran a quick test, and realized that we are now warning the user twice if a project doesn't exist:
```
./ghpc create examples/hpc-cluster-small.yaml 
2022/09/07 23:05:37 could not add module metadata: global variable project_id must be defined
deployment variable project_id was not set
```

If this function is called each time `ghpc create` is called, I'd suggest removing the other warning, otherwise I'd suggest dropping the Println for this error and let the other validation catch it.

pkg/sourcereader/embedded.go
tpdownes: Removed on `HEAD`.

pkg/sourcereader/embedded.go
tpdownes: Done (no change)

pkg/config/expand.go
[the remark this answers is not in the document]
heyealex: I actually ran into this when deploying networks in two separate projects, I had to fill project_id with a dummy value that was never used. I agree that it probably shouldn't be a requirement. 

Thanks for making the change!

heyealex (approved):

PR #529, opened 2022-09-08
Add quantum circuit simulator integration test

tpdownes:
### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/integration-group-3.yaml
+++ b/tools/cloud-build/daily-tests/integration-group-3.yaml
@@ -127,3 +127,25 @@ steps:
     ansible-playbook tools/cloud-build/daily-tests/ansible_playbooks/packer-integration-test.yml \
       --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT}" \
       --extra-vars="@tools/cloud-build/daily-tests/tests/packer.yml"
+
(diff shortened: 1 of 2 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/simple-deploy-test.yml
```diff
@@ -0,0 +1,67 @@
+# Copyright 2022 Google LLC
```
nick-stroud, line 1: My initial instinct would have been to use `base-integration-test.yml` without any post deploy tests. I realise that has some extra steps in it that you do not need but I think the code duplication in this file is adding tech debt to our tests. Doing a diff this file is almost a perfect subset of `base-integration-test.yml` with exception of the terraform validate, which is a great example of how having the duplication can cause drift.

The fast solution would be to reuse `base-integration-test.yml`. I would be even happier if you were able to break out the common tasks into their own file and reuse from both `base-integration-test.yml` and `simple-deploy-test.yml` as needed. 

[here](https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/tools/cloud-build/daily-tests/ansible_playbooks/tasks) is an example of where this has been done prior.

tpdownes:
A group 3 build is running even if it is not shown on GitHub. Please review once it's complete (and hopefully successful).

nick-stroud (approved):
I am counting on you (@tpdownes) to confirm tests are passing.

PR #530, opened 2022-09-08
Add image to cloud batch example B/C default image gcloud version is lagging

nick-stroud:
This will resolve the failing Batch integration test.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/cloud-batch.yaml
+++ b/community/examples/cloud-batch.yaml
@@ -60,6 +60,9 @@ deployment_groups:
     settings:
       runnable: "cat /sw/hello.txt"
       machine_type: n2-standard-4
+      image:
+        family: rocky-linux-8-optimized-gcp
+        project: rocky-linux-cloud
 
   - id: batch-login
     source: community/modules/scheduler/cloud-batch-login-node


community/examples/cloud-batch.yaml
```diff
@@ -60,6 +60,9 @@ deployment_groups:
     settings:
       runnable: "cat /sw/hello.txt"
       machine_type: n2-standard-4
+      image:
+        family: centos-stream-8
+        project: centos-cloud
```
tpdownes, line 65: My instinct would be to use

```yaml
image:
  family: rocky-linux-8-optimized-gcp
  project: rocky-linux-cloud
```

as something more closely approximating the HPC VM Image and what HPC users might use. 

Obviously the actual testing being done here is pretty simple and the image is not very relevant. Was there a reason to use this image?

tpdownes (commented):
Requesting info, but not a change (yet)

community/examples/cloud-batch.yaml
nick-stroud: I originally had `centos-7` but that gets worse startup time because it does not have python3 so I just pushed to centos-8. I will test with your suggestion and go with that if it all looks good.

nick-stroud: Done.

tpdownes (approved):

PR #532, opened 2022-09-08
Add customized version output for debugging improvement

thiagosgobe:
go/hpc-toolkit-version-info

### Submission Checklist

  install`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,16 @@ MIN_GOLANG_VERSION=1.16 # for building ghpc
 ENG = ./cmd/... ./pkg/...
 TERRAFORM_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.tf" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
 PACKER_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.pkr.hcl" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
+
+ifneq (, $(shell which git))
+## GIT IS PRESENT
+ifneq (,$(wildcard .git))
+## GIT DIRECTORY EXISTS
 GIT_TAG_VERSION=$(shell git tag --points-at HEAD)
 GIT_BRANCH=$(shell git branch --show-current)
 GIT_COMMIT_INFO=$(shell git describe --tags --dirty --long)
+endif
+endif
 
 # RULES MEANT TO BE USED DIRECTLY
 
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,15 @@ MIN_GOLANG_VERSION=1.16 # for building ghpc
 ENG = ./cmd/... ./pkg/...
 TERRAFORM_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.tf" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
 PACKER_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.pkr.hcl" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
+GIT_TAG_VERSION=$(shell git tag --points-at HEAD)
+GIT_BRANCH=$(shell git branch --show-current)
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -34,12 +42,25 @@ HPC deployments on the Google Cloud Platform.`,
...
 		},
-		Version: "v1.4.1",
+		Version:     "v1.4.1",
+		Annotations: annotation,
 	}
 )
 
 // Execute the root command
 func Execute() error {
+	if len(GitBranch) > 0 {
```
nick-stroud, line 52: I would use `GitCommitInfo` here instead. It is possible that they are using git but not on any branch (detached HEAD mode). This is possible if you do `git checkout <sha>`.

cmd/root.go
```diff
@@ -34,12 +42,25 @@ HPC deployments on the Google Cloud Platform.`,
...
+		if len(GitTagVersion) == 0 {
```
nick-stroud, line 53: I would do something similar for git branch. If git branch is empty then you can say something like 'detached HEAD`.

Makefile
```diff
@@ -13,12 +13,15 @@ MIN_GOLANG_VERSION=1.16 # for building ghpc
...
+GIT_TAG_VERSION=$(shell git tag --points-at HEAD)
+GIT_BRANCH=$(shell git branch --show-current)
+GIT_COMMIT_INFO=$(shell git describe --tags --dirty --long)
 
 # RULES MEANT TO BE USED DIRECTLY
 
 ghpc: warn-go-version warn-terraform-version warn-packer-version $(shell find ./cmd ./pkg ghpc.go -type f)
 	$(info **************** building ghpc ************************)
-	go build ghpc.go
+	@go build -ldflags="-X 'main.gitTagVersion=$(GIT_TAG_VERSION)' -X 'main.gitBranch=$(GIT_BRANCH)' -X 'main.gitCommitInfo=$(GIT_COMMIT_INFO)'" ghpc.go
```
nick-stroud, line 24: I tried making when not in a git repo and the output from make looks like:
```
~/dev/test-zip$ make
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
```

In this case the binary still built and it correctly fell back to the in text version, but if we are considering this a valid use case we probably shouldn't be printing fatal when everything is actually fine. Can we suppress this specific warning, or perhaps tell the user that they are not building from a git repository. 

I also want to make sure we don't suppress all output since there could be something else useful. Perhaps we could first check if git is available and then have a different build command based on yes or now.

Steps to repro:
- checkout your branch
- git archive HEAD --format=zip -o toolkit-zip.zip
- cd ..
- mkdir test-zip
- mv hpc-toolkit/toolkit-zip.zip test-zip/
- cd test-zip/
- unzip toolkit-zip.zip
- make

nick-stroud: correction, this is the case that I have git installed but I am not in a git repo.

Makefile
```diff
@@ -13,9 +13,16 @@ MIN_GOLANG_VERSION=1.16 # for building ghpc
 ENG = ./cmd/... ./pkg/...
 TERRAFORM_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.tf" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
 PACKER_FOLDERS=$(shell find ./modules ./community/modules ./tools -type f -name "*.pkr.hcl" -not -path '*/\.*' -exec dirname "{}" \; | sort -u)
+
+ifneq (, $(shell which git))
+## GIT IS PRESENT
+ifneq (,$(wildcard .git))
+## GIT DIRECTORY EXISTS
```
nick-stroud, line 20: This LGTM but I am not expert with shell/make syntax. @cboneti , could you give the syntax a once over.

nick-stroud: Second thought, I see this syntax used elsewhere in the make file. I am good to approve.

tpdownes:
@thiagosgobe please rebase your branch off the current state of develop. There are fixes on develop that are necessary for PR-validation checks to pass.

nick-stroud (approved):

nick-stroud (approved):

PR #533, opened 2022-09-09
Fix spack setup scripts in tutorials.

nick-stroud:
The `spack env loads -r` line was causing `Error: No module available for package hdf5@1.12.2%gcc@8.2.0~cxx+fortran+hl~ipo~java+mpi+shared~szip~threadsafe+tools`. The error halts the startup script so the wget for the input deck never takes place and submit_gromacs.sh never gets loaded onto the machine. The result is that the sbatch command fails.

From reading the [spack docs](https://spack.readthedocs.io/en/latest/environments.html#loading) it seems like this load command is unnecessary for our tutorial. 

I have conferred with @douglasjacobsen, and they agree it is fine to remove it.

I have manually tested this and the tutorial succeeds.

### Submission Checklist

  install`)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -34,7 +34,7 @@ HPC deployments on the Google Cloud Platform.`,
 				log.Fatalf("cmd.Help function failed: %s", err)
 			}
 		},
-		Version: "v1.4.0",
+		Version: "v1.4.1",
 	}
 )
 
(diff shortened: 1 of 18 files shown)


cmd/root.go
```diff
@@ -34,7 +34,7 @@ HPC deployments on the Google Cloud Platform.`,
 				log.Fatalf("cmd.Help function failed: %s", err)
 			}
 		},
-		Version: "v1.4.0",
+		Version: "v1.4.1",
```
nick-stroud, line 37: Should we roll the patch version for this? I have it in here now but can remove if we decide otherwise.

cboneti (approved):

cmd/root.go
nick-stroud: We decided it should be a patch version update.

PR #542, opened 2022-09-12
Address permadiff in vm-instance module

tpdownes:
When used to provision A2 machine types, using the empty list as the default value of guest_accelerator causes a permadiff because the Compute Engine API reports the A2's A100 GPU card and we have chosen an explicit value. If, instead, we use a null value for guest_accelerator, Terraform treats the difference as not requiring action.

### Submission Checklist

  install`)


heyealex (approved):
Ran a quick test to be sure, and everything looks good.

PR #548, opened 2022-09-13
Add auto-delete boot disk as an option on vm-instance

nick-stroud:
If even only for testing, I often want to turn auto-delete off. This exposes the option without changing the default behavior.

### Submission Checklist

  install`)

### DIFF
--- a/modules/compute/vm-instance/variables.tf
+++ b/modules/compute/vm-instance/variables.tf
@@ -49,6 +49,12 @@ variable "disk_type" {
   default     = "pd-standard"
 }
 
+variable "auto_delete_boot_disk" {
+  description = "Controls if boot disk should be auto-deleted when instance is deleted."
+  type        = bool
+  default     = true
+}
+
 variable "local_ssd_count" {
   description = "The number of local SSDs to attach to each VM. See https://cloud.google.com/compute/docs/disks/local-ssd."
   type        = number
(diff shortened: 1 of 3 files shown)


modules/compute/vm-instance/variables.tf
```diff
@@ -49,6 +49,12 @@ variable "disk_type" {
   default     = "pd-standard"
 }
 
+variable "auto_delete_boot_disk" {
+  description = "Controlls if boot disk should be auto-deleted when instance is deleted."
```
tpdownes, line 53: Typo on "Controlls".

nick-stroud:
> it's not clear to me what workflow you are trying to enable.

In the current state, If I have waited a long time for startup scripts to install something and then I want to tweak a machine parameter that forces replacement, I will lose the disk and have to wait for startup to run again.

tpdownes:
I have some concerns about effectively managing the destruction of the disk twice (explicitly and implicitly) that this change does not address. I also suspect you could address some of your concerns by shutting the machine down, running Terraform, and then turning it back on.

That said, it's a knob that arguably merits existence.

In medium term I think we should probably manage the creation and destruction of the boot disk entirely implicitly (with auto deletion either on or off) and allow the user to supply externally-created secondary disks.

tpdownes (approved):

PR #551, opened 2022-09-15
Upgrade DDN-EXAScaler to v6.1.0

nick-stroud:
Upgraded needed to expose `ddn_exascaler.client_config` which will be added in subsequent PR.

DDN now only supports specifying image family (not name)

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/DDN-EXAScaler/variables.tf
+++ b/community/modules/file-system/DDN-EXAScaler/variables.tf
@@ -189,25 +189,27 @@ variable "boot" {
   type = object({
     disk_type   = string
     auto_delete = bool
+    script_url  = string
   })
   default = {
     disk_type   = "pd-standard"
     auto_delete = true
+    script_url  = null
   }
 }
 
 # Source image properties
 # project: project name
-# name: image name
+# family: image family name
 variable "image" {
   description = "Source image properties"
   type = object({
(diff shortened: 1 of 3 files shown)


community/modules/file-system/DDN-EXAScaler/variables.tf
```diff
@@ -189,25 +189,27 @@ variable "boot" {
...
     auto_delete = true
+    script_url  = null
   }
 }
 
 # Source image properties
 # project: project name
-# name: image name
+# family: image family name
 variable "image" {
```
heyealex, line 204: A suggestion (feel free to disregard):
Take all of project, name and family as input but validate with an error if name is used, indicating how the interface has changed. That way, the error received if someone is still using the old format clearly indicated how to fix it, rather than a confusing "variable not found" error.

heyealex (approved):

nick-stroud:
Test failed for unrelated `Heavy Weight read quota` error after luster install was successful. I have also run seperate manual test.

community/modules/file-system/DDN-EXAScaler/variables.tf
nick-stroud: I like this and will implement in a follow on PR.

PR #552, opened 2022-09-15
Add Epilog/Prolog scripts to install path in hybrid

heyealex:
Updated the install path sed command to also catch EpilogSlurmctld and PrologSlurmctld when exclusive partitions are being used.

Updated the hybrid test config to include an exclusive partition and to reformat module definitions in line with recent improvements in the codebase.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/main.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/main.tf
@@ -21,7 +21,7 @@ locals {
   }]
   install_dir         = var.install_dir != null ? var.install_dir : abspath(var.output_dir)
   install_dir_pattern = replace(local.install_dir, ".", "\\.")
-  install_path_cmd    = "sed -i -E 's|Program=/.*/(resume\\|suspend).py|Program=${local.install_dir_pattern}/\\1\\.py|g' cloud.conf"
+  install_path_cmd    = "sed -i -E 's|(Program\\|logSlurmctld)=/.*/(resume\\|suspend).py|\\1=${local.install_dir_pattern}/\\2\\.py|g' cloud.conf"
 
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/main.tf
```diff
@@ -21,7 +21,7 @@ locals {
   }]
   install_dir         = var.install_dir != null ? var.install_dir : abspath(var.output_dir)
   install_dir_pattern = replace(local.install_dir, ".", "\\.")
-  install_path_cmd    = "sed -i -E 's|Program=/.*/(resume\\|suspend).py|Program=${local.install_dir_pattern}/\\1\\.py|g' cloud.conf"
+  install_path_cmd    = "sed -i -E 's|(Program\\|logSlurmctld)=/.*/(resume\\|suspend).py|\\1=${local.install_dir_pattern}/\\2\\.py|g' cloud.conf"
```
nick-stroud, line 24: nit: Could we add a comment somewhere in this file. Either at the top of these locals or on the null resource.

Something that hints at what we are doing here. It is alluded to in the `install_dir` descriptions , but beyond that this is all pretty crypict. 

Maybe something like: "this command and the null resource it is consumed by will insert the install dir into the slurm configs for cases that the install dir is not the same as the output dir".

nick-stroud (approved):

community/modules/scheduler/schedmd-slurm-gcp-v5-hybrid/main.tf
heyealex: Agreed, I added a (lengthy) comment before the variables and a smaller comment before the null resource. I also took the opportunity  to break up the sed string a bit further for clarity.

nick-stroud (approved):

PR #553, opened 2022-09-15
Integrate DDN Lustre install script with startup-script

nick-stroud:
This change adds runners that are dynamically created by the DDN Lustre module to be able to install the luster client (and mount) on vm-instance.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/DDN-EXAScaler/outputs.tf
+++ b/community/modules/file-system/DDN-EXAScaler/outputs.tf
@@ -24,13 +24,13 @@ output "ssh_console" {
   value       = module.ddn_exascaler.ssh_console
 }
 
-output "client_config" {
+output "client_config_script" {
   description = "Script that will install DDN EXAScaler lustre client. The machine running this script must be on the same network & subnet as the EXAScaler."
   value       = module.ddn_exascaler.client_config
 }
 
 output "install_ddn_lustre_client_runner" {
-  description = "Runner that encapsulates the `client_config` output on this module."
+  description = "Runner that encapsulates the `client_config_script` output on this module."
   value = {
     "type"        = "shell"
     "content"     = module.ddn_exascaler.client_config
@@ -46,7 +46,7 @@ locals {
 }
 
 output "mount_command" {
-  description = "Command to mount the file system. `client_config` script must be run first."
+  description = "Command to mount the file system. `client_config_script` must be run first."
   value       = local.mount_cmd_w_mkdir
 }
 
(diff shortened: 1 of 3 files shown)


community/modules/file-system/DDN-EXAScaler/outputs.tf
```diff
@@ -24,16 +24,47 @@ output "ssh_console" {
   value       = module.ddn_exascaler.ssh_console
 }
 
+output "client_config" {
```
heyealex, line 27: nit: The name makes me think it's a config file, not a script. I'd append _script to the name.

Disregard if you disagree.

community/modules/file-system/DDN-EXAScaler/outputs.tf
```diff
@@ -24,16 +24,47 @@ output "ssh_console" {
...
+    "destination" = "mount-ddn-lustre.sh"
+  }
 }
 
 output "http_console" {
   description = "HTTP address to access the system web console."
   value       = module.ddn_exascaler.http_console
 }
 
+
```
heyealex, line 67: nit: Unneeded extra space

community/modules/file-system/DDN-EXAScaler/outputs.tf
```diff
@@ -24,16 +24,47 @@ output "ssh_console" {
   value       = module.ddn_exascaler.ssh_console
 }
 
+output "client_config" {
+  description = "Script that will install DDN EXAScaler lustre client. The machine running this script must be on the same network & subnet as the EXAScaler."
```
heyealex, line 28: For my own knowledge, why is it that the machine must be on the same network/subnet?

community/modules/file-system/DDN-EXAScaler/outputs.tf
nick-stroud: I agree, I was matching the ddn name but I think appending script satisfies both inclinations. Will update.

community/modules/file-system/DDN-EXAScaler/outputs.tf
nick-stroud: The script downloads a binary from the ddn server. Then invokes the binary. I know the binary does something with firewall rules, in addition to installing the client. So something to do with all that although I don't know in detail. The requirement is documented by DDN [here](https://github.com/DDNStorage/exascaler-cloud-terraform/tree/master/gcp#install-new-exascaler-cloud-clients).

community/modules/file-system/DDN-EXAScaler/outputs.tf
nick-stroud: Done.

PR #558, opened 2022-09-16
Replace Spack installation in AMD example with a builder VM

tpdownes:
This PR implements a builder VM on which Spack will be built and made available to a Slurm cluster via NFS.

It does not solve every problem (e.g. communicating to user clearly that Spack setup has complete) but does enable progress on provision clusters using `v5` of the Slurm-on-GCP solution.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
+++ b/community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
@@ -65,8 +65,25 @@ deployment_groups:
       - type: shell
         source: modules/startup-script/examples/install_ansible.sh
         destination: install_ansible.sh
+      - $(swfs.install_nfs_client_runner)
+      - $(swfs.mount_runner)
       - $(spack.install_spack_deps_runner)
       - $(spack.install_spack_runner)
+      - type: shell
+        content: "shutdown -h +1"
+        destination: shutdown.sh
+
+  - id: slurm_startup
+    source: modules/scripts/startup-script
+    settings:
+      runners:
+      - type: data
+        destination: /etc/profile.d/spack.sh
+        content: |
+          #!/bin/sh
+          if [ -f /sw/spack/share/spack/setup-env.sh ]; then
+                  . /sw/spack/share/spack/setup-env.sh
+          fi
       # the following installation of AOCC may be automated in the future
       # with a clear direction to the user to read the EULA at
       # https://developer.amd.com/aocc-compiler-eula/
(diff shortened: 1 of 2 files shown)


community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
```diff
@@ -65,20 +65,46 @@ deployment_groups:
       - type: shell
         source: modules/startup-script/examples/install_ansible.sh
         destination: install_ansible.sh
+      - $(swfs.install_nfs_client_runner)
+      - $(swfs.mount_runner)
       - $(spack.install_spack_deps_runner)
       - $(spack.install_spack_runner)
       # the following installation of AOCC may be automated in the future
```
heyealex, line 72: These comments should be in the next block, correct?

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
```diff
@@ -65,20 +65,46 @@ deployment_groups:
...
         content: |
           #!/bin/bash
+          source /sw/spack/share/spack/setup-env.sh
           spack install aocc@3.2.0 +license-agreed
           spack load aocc@3.2.0
           spack compiler find --scope site
           spack -d install -v openmpi@4.1.3 %aocc@3.2.0 +legacylaunchers +pmi schedulers=slurm
 
+  # must restart vm to re-initiate subsequent installs
+  - id: spack-builder
```
heyealex, line 101: nit: The IDs of other modules are using underscores rather than dashes.

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
```diff
@@ -65,20 +65,46 @@ deployment_groups:
...
         destination: install_ansible.sh
+      - $(swfs.install_nfs_client_runner)
+      - $(swfs.mount_runner)
       - $(spack.install_spack_deps_runner)
       - $(spack.install_spack_runner)
       # the following installation of AOCC may be automated in the future
       # with a clear direction to the user to read the EULA at
       # https://developer.amd.com/aocc-compiler-eula/
+      - type: shell
+        content: "shutdown -h +1"
```
heyealex, line 76: Is there a reason for the +1m? (I imagine it doesn't hurt, just curious)

tpdownes: I had this discussion with @nick-stroud. Empirically, this forks into the background and allows the startup script to complete. I like the idea of allowing the SystemD unit running the startup script to complete properly rather than potentially being shutdown prematurely when SystemD is triggered to shutdown.

I wish we could do `+15s` or similar, but 1 minute seems fine.

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
tpdownes: I will change that. Personal opinion: no dashes in IDs since they will be picked up by TFLint as a no-no.

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
tpdownes: Yes, I moved it. Thank you.

tpdownes: Should be done.

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
tpdownes: Should be done.

community/examples/AMD/hpc-cluster-amd-slurmv5.yaml
heyealex: Thanks for the context, that makes sense

heyealex (approved):

PR #560, opened 2022-09-16
Enable DDN Lustre client install with pre-existing-network-storage

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/modules/file-system/pre-existing-network-storage/outputs.tf
+++ b/modules/file-system/pre-existing-network-storage/outputs.tf
@@ -61,18 +61,18 @@ locals {
 
 output "client_install_runner" {
   description = "Runner that performs client installation needed to use file system."
-  value = lookup(local.install_scripts, var.fs_type, null) == null ? null : {
+  value = {
     "type"        = "shell"
-    "content"     = lookup(local.install_scripts, var.fs_type, "")
+    "content"     = lookup(local.install_scripts, var.fs_type, "echo 'skipping: client_install_runner not yet supported for ${var.fs_type}'")
     "destination" = "install_filesystem_client${replace(var.local_mount, "/", "_")}.sh"
   }
 }
 
 output "mount_runner" {
   description = "Runner that mounts the file system."
-  value = lookup(local.mount_commands, var.fs_type, null) == null ? null : {
+  value = {
     "type"        = "shell"
-    "content"     = local.mount_script
+    "content"     = (lookup(local.mount_commands, var.fs_type, null) == null ? "echo 'skipping: mount_runner not yet supported for ${var.fs_type}'" : local.mount_script)
     "destination" = "mount_filesystem${replace(var.local_mount, "/", "_")}.sh"
   }
 }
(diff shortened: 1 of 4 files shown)


modules/file-system/pre-existing-network-storage/outputs.tf
```diff
@@ -24,3 +24,55 @@ output "network_storage" {
     mount_options = var.mount_options
   }
 }
+
+locals {
```
heyealex, line 28: I did not know you could put locals blocks in outputs.tf... Or at least never thought to. 
I like it, it's a lot cleaner keeping anything output related only with the output.

modules/file-system/pre-existing-network-storage/outputs.tf
```diff
@@ -24,3 +24,55 @@ output "network_storage" {
...
+  ddn_lustre_client_install_script = templatefile(
+    "${path.module}/templates/ddn_exascaler_luster_client_install.tftpl",
+    {
+      server_ip    = split("@", var.server_ip)[0]
+      remote_mount = var.remote_mount
+      local_mount  = var.local_mount
+    }
+  )
+
+  install_scripts = {
```
heyealex, line 39: Just to clarify my understanding, the idea is that eventually all mount/install scripts and templates will be copied here (just nfs for now really) and automatically generated if the fs_type is the same?

modules/file-system/pre-existing-network-storage/outputs.tf
```diff
@@ -24,3 +24,55 @@ output "network_storage" {
...
+  value = lookup(local.install_scripts, var.fs_type, null) == null ? null : {
+    "type"        = "shell"
+    "content"     = lookup(local.install_scripts, var.fs_type, "")
+    "destination" = "install_filesystem_client${replace(var.local_mount, "/", "_")}.sh"
+  }
+}
+
+output "mount_runner" {
+  description = "Runner that mounts the file system."
+  value = lookup(local.mount_commands, var.fs_type, null) == null ? null : {
```
heyealex, line 73: What happens if a null runner is accidentally added to a startup-script runner list?

heyealex (commented):
Overall looks good. I just had a couple questions (and a comment).

modules/file-system/pre-existing-network-storage/outputs.tf
nick-stroud: It fails on startup script validation, but...

I discovered a larger issue which is that destination must be deterministic at apply time since it is used as a key in a map later (terraform being picky). To satisfy that I cannot do the dynamic substitution for null. I have pushed a version working version which still delivers an "empty runner" which just prints that it is skipping. Happy to have some discussion on this topic.

modules/file-system/pre-existing-network-storage/outputs.tf
nick-stroud: Yes.

modules/file-system/pre-existing-network-storage/outputs.tf
nick-stroud: I am pretty sure that with terraform, anything can go anywhere. I personally agree that it is clearest to keep this locals block in output but also am not too opinionated.

heyealex (approved):

PR #564, opened 2022-09-19
Allow generic git sources

heyealex:
This PR allows generic git repos to be sourced as modules using the `git::` prefix. The underlying implementation is identical to the github source, so the same sourcereader is being used. 

This has been tested against a git repo hosted outside of github or gitlab and the hpc-cluster-small.yaml example (all modules were converted to pull from that git repo instead of from the binary or locally).

A bug also had to be fixed, the group directory was not always created when all modules are sourced from git, so now a directory is created explicitly for all groups, rather than depending on the writer.

### Submission Checklist

  install`)

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -259,6 +259,14 @@ Toolkit vpc module, use:
 [tfsubdir]: https://www.terraform.io/language/modules/sources#modules-in-package-sub-directories
 [daos-cluster.yaml]: ../community/examples/intel/daos-cluster.yaml
 
+#### Generic Git Modules
+To use a Terraform module available in a non-GitHub git repository such as
+gitlab, set the source to a path starting `git::`. Two Standard git protocols
+are supported, `git::https://` for HTTPS or `git::git@github.com` for SSH.
+
+Additional formatting and features after `git::` are identical to that of the
+[GitHub Modules](#github-modules) described above.
+
 ### Kind (May be Required)
 
 `kind` refers to the way in which a module is deployed. Currently, `kind` can be
--- a/pkg/modulewriter/modulewriter.go
+++ b/pkg/modulewriter/modulewriter.go
@@ -115,7 +115,7 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
 func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) {
 	for iGrp, grp := range *deploymentGroups {
 		for iMod, module := range grp.Modules {
-			if sourcereader.IsGitHubPath(module.Source) {
+			if sourcereader.IsGitPath(module.Source) {
 				continue
 			}
 
--- a/pkg/modulewriter/modulewriter.go
+++ b/pkg/modulewriter/modulewriter.go
@@ -84,6 +84,10 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
 		return err
 	}
 
+	if err := createGroupDirs(deploymentDir, &blueprint.DeploymentGroups); err != nil {
+		return err
+	}
+
 	for _, grp := range blueprint.DeploymentGroups {
 
 		deploymentName, err := blueprint.DeploymentName()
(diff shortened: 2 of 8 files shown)


modules/README.md
```diff
@@ -259,6 +259,14 @@ Toolkit vpc module, use:
 [tfsubdir]: https://www.terraform.io/language/modules/sources#modules-in-package-sub-directories
 [daos-cluster.yaml]: ../community/examples/intel/daos-cluster.yaml
 
+#### Generic Git Modules
```
nick-stroud, line 262: nit: Does this also support GitHub? Maybe we should only document this method. We can still leave the other functionality intact but it would give us less surface area to maintain.

pkg/modulewriter/modulewriter.go
```diff
@@ -112,17 +115,25 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
...
-func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) {
+func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) error {
+
 	for iGrp, grp := range *deploymentGroups {
+		basePath := filepath.Join(deploymentPath, grp.Name)
+		// Create the deployment group directory if not already created.
+		if _, err := os.Stat(basePath); errors.Is(err, os.ErrNotExist) {
+			deploymentio := deploymentio.GetDeploymentioLocal()
+			if err := deploymentio.CreateDirectory(basePath); err != nil {
+				return fmt.Errorf("failed to create directory at %s for deployment group %s: err=%w", basePath, grp.Name, err)
```
nick-stroud, line 126: Could you write a unit test for this? If it is too closely tied to copying source then maybe having it as its own function `createGroupDirs()` would make it easier to test.

tpdownes:
General comment that I do not think should hold up a reasonable, small change that moves things forward.

There is existing functionality in go-getter, esp `getter.Detect` that may be relevant for a proper redesign.

https://github.com/hashicorp/terraform/blob/51b0aee36cc2145f45f5b04051a01eb6eb7be8bf/internal/getmodules/package.go#L7
https://github.com/hashicorp/go-getter/blob/main/detect.go

modules/README.md
heyealex: It covers both, but I kept them documented separately since the simplified option is available with github (doesn't require the git:: prefix). Terraform and go-getter (the package we and terraform use) both also document them separately:
* https://www.terraform.io/language/modules/sources#generic-git-repository
* https://www.terraform.io/language/modules/sources#github

I could see making github a subsection of git potentially, but I personally prefer this layout for consistency.

pkg/modulewriter/modulewriter.go
heyealex: Done

modules/README.md
nick-stroud: Ack.

pkg/modulewriter/modulewriter.go
```diff
@@ -115,17 +119,25 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
...
 }
 
+func createGroupDirs(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) error {
+	deploymentio := deploymentio.GetDeploymentioLocal()
+	for _, grp := range *deploymentGroups {
+		groupPath := filepath.Join(deploymentPath, grp.Name)
+		// Create the deployment group directory if not already created.
+		if _, err := os.Stat(groupPath); errors.Is(err, os.ErrNotExist) {
+			if err := deploymentio.CreateDirectory(groupPath); err != nil {
+				return fmt.Errorf("failed to create directory at %s for deployment group %s: err=%w",
```
nick-stroud, line 129: After taking a closer look at all of this, I feel like there is a smell here. We are:
- checking the path does not exist (here) and skipping if it does
- checking the path does not exist (deploymentio.CreateDirectory) and erroring if it does
- os.MkdirAll already handles case that dir already exists [ref](https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/os/path.go;l=16-17;drc=2291cae2af659876e93a3e1f95c708abb1475d02).

Personally I don't really understand the need for deploymentio to manage CreateDirectory. Without it, this would become the following:

```

L125	groupPath := filepath.Join(deploymentPath, grp.Name)
	if err := os.MkdirAll(groupPath, 0755); err != nil {
		return fmt.Errorf("failed to create directory at %s for deployment group %s: err=%w", 
			groupPath, grp.Name, err)}
```

heyealex: I agree, partially. blueprintio wasn't needed and obfuscated some of the logic needlessly. However, I would rather use Mkdir over MkdirAll. The expectation is that the deployment directory already exists, so if for some reason that fails to work in the future, I'd rather it fail than be covered up by using an overly permissive MkdirAll command here.

nick-stroud (approved):

PR #568, opened 2022-09-22
Perform regular cleanup of Filestore VPC network peerings

tpdownes:
### Submission Checklist

  install`)


nick-stroud (approved):
LGTM, @tpdownes This shell script is one level beyond my own knowledge. This is a case where I agree with the concept but shouldn't really be counted on for detail review. Happy if you want to move forward based on your own assurance or also if you would like to get another to review.

nick-stroud:
nit: Also I would recommend manually testing this before we set it into the wild.

tpdownes:
I have manually tested the diff as a standalone script on `hpc-toolkit-dev` and on my own personal project. I feel confident about it.

PR #569, opened 2022-09-26
Fix filestore peering network cleanup script

tpdownes:
The filestore network peering script failed over the weekend due to an apparent parsing error. The solution worked for me locally before submitting, but I am providing what is likely a more portable fix by explicitly reading in a bash array using `read -ra`. The commenting is also improved to guide future developers understanding the solution.

### Submission Checklist

  install`)


heyealex (approved):
I tried out on my end, and everything seems to work as expected (at least locally). LGTM

PR #571, opened 2022-09-26
Avoid spurious errors in Spack log

tpdownes:
- do not attempt to re-create a Spack environment that already exists

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scripts/spack-install/templates/install_spack.tpl
+++ b/community/modules/scripts/spack-install/templates/install_spack.tpl
@@ -99,6 +99,7 @@ echo "$PREFIX Installing root spack specs..."
 echo "$PREFIX Configuring spack environments"
 %{if ENVIRONMENTS != null ~}
 %{for e in ENVIRONMENTS ~}
+if [ ! -d ${INSTALL_DIR}/var/spack/environments/${e.name} ]; then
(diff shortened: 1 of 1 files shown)


community/modules/scripts/spack-install/templates/install_spack.tpl
```diff
@@ -99,6 +99,7 @@ echo "$PREFIX Installing root spack specs..."
 echo "$PREFIX Configuring spack environments"
 %{if ENVIRONMENTS != null ~}
 %{for e in ENVIRONMENTS ~}
+if [ ! -d ${INSTALL_DIR}/var/spack/environments/${e.name} ]; then
```
douglasjacobsen, line 102: The only issue I could see with this is if spack ever lets you change the path to where environments live.

It might be better to do something like:
`if [ ! -z "$(spack env list | grep ${e.name}) ]; then`

So you're using spacks internals rather than checking a hardcoded path.

FWIW, I don't think it's currently possible to change the path that named environments are created in though.

douglasjacobsen (approved):
Outside of my one comment (which honestly, doesn't need to be addressed, I just wanted to make sure it was known), this LGTM.

nick-stroud (approved):

PR #574, opened 2022-09-27
Add ability to pass non-text variables through the cli `--vars` flag

kkr16:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -450,7 +450,11 @@ func (dc *DeploymentConfig) SetCLIVariables(cliVariables []string) error {
 			return fmt.Errorf("invalid format: '%s' should follow the 'name=value' format", cliVar)
 		}
 
-		key, value := arr[0], arr[1]
+		// Convert the variable's string litteral to its equivalent default type.
+		var out interface{}
+		yaml.Unmarshal([]byte(arr[1]), &out)
+
+		key, value := arr[0], out
 		dc.Config.Vars[key] = value
 	}
 
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


cboneti:
gcbrun/

pkg/config/config.go
```diff
@@ -450,7 +450,11 @@ func (dc *DeploymentConfig) SetCLIVariables(cliVariables []string) error {
 			return fmt.Errorf("invalid format: '%s' should follow the 'name=value' format", cliVar)
 		}
 
-		key, value := arr[0], arr[1]
+		// Convert the variable's string litteral to its equivalent default type.
+		var out interface{}
+		yaml.Unmarshal([]byte(arr[1]), &out)
```
cboneti, line 455: Maybe a silly question, what happens if this fails? should you not check for an error?

pkg/config/config_test.go
```diff
@@ -542,12 +543,18 @@ func (s *MySuite) TestSetCLIVariables(c *C) {
 	cliRegion := "cli_region"
 	cliZone := "cli_zone"
 	cliKeyVal := "key=val"
+	cliKeyBool := "true"
```
heyealex, line 546: I'd like to see the full set of requirements from the design being tested here, at least in a simple form:

> Parse simple variables provided under a --vars flag
> - Boolean: true
> - Integers: 3
> - Floats: 3.14
> - Strings: “foo”
>
> Parse complex variables provided under a --vars flag
> - Array: [1, 2, 3]
> - Map: {a: 1, b: 2}
> - Array with maps: [foo, {a: 1, b: 2}]
> - Map with arrays: {foo: [1, 2, 3], bar: [a, b, c]}

heyealex (changes requested):
Thanks for the PR and the great addition to the Toolkit!
I added a comment for covering a couple more test cases, once those have been added this should be good to go.

pkg/config/config_test.go
kkr16: Done.

kkr16:
Thanks for the review @heyealex !

I updated the tests cases based on your recommendations. 
I've also updated [cmd/README.md](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/410cf5d9d60e08f20fcf4f5926a2e536579b0313/cmd/README.md#flags---create) with an exhaustive list of tested examples, showing escape characters for bash.

cboneti:
gcbrun/

heyealex (approved):
Thanks for adding the new tests and updating the documentation! Approved

PR #575, opened 2022-09-28
Allow for DDN Lustre to be ready to install client

nick-stroud:
Somehow in the tree of dependencies network_storage is released as soon as the msg ddn server is up while client_config waits until lustre is fully deployed. The pre-existing-network-storage install is trying execute before lustre is up, which I believe is causing issues. 

Also, the curl to download the client-setup-tool can fail when performed right after DDN EXAScaler finishes deploying or when multiple clients are trying to download simultaneously. I am working with DDN to try to add retry logic to this operation but for now the sleep in integration tests should improve congestion.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/DDN-EXAScaler/outputs.tf
+++ b/community/modules/file-system/DDN-EXAScaler/outputs.tf
@@ -73,4 +73,8 @@ output "network_storage" {
     fs_type       = "lustre"
     mount_options = ""
   }
+  depends_on = [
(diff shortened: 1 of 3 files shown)


nick-stroud:
I manually executed test on cloud build and verified pass. Running full group 3 test. I will open task to follow up on added TODO.

cboneti (approved):

PR #581, opened 2022-09-29
Fix Ansible module for upgrading setuptools in HTCondor autoscaler

tpdownes:
The Ansible module for installing the HTCondor autoscaler has a bug in which the version number is being supplied to the `state` parameter rather than the `version` parameter. I suspect this is a behavioral change in the pip module above ansible 2.9. The consequence of the bug is that the virtual environment for the HTCondor autoscaler fails to install and the  the HTCondor pool community example stops mid-way through its configuration. This has the user-facing consequence that the access point ("login node") fails to join the pool and users cannot submit jobs.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scripts/htcondor-install/files/install-htcondor-autoscaler-deps.yml
+++ b/community/modules/scripts/htcondor-install/files/install-htcondor-autoscaler-deps.yml
@@ -24,13 +24,13 @@
   - name: Create virtual environment for HTCondor autoscaler
     ansible.builtin.pip:
       name: pip
-      version: 21.3.1  # last Python 2.7-compatible release
+      version: 21.3.1  # last Python 3.6-compatible release
       virtualenv: /usr/local/htcondor
       virtualenv_command: /usr/bin/python3 -m venv
   - name: Install latest setuptools
     ansible.builtin.pip:
       name: setuptools
-      state: 44.1.1  # last Python 2.7-compatible release
+      version: 59.6.0  # last Python 3.6-compatible release
       virtualenv: /usr/local/htcondor
       virtualenv_command: /usr/bin/python3 -m venv
   - name: Install HTCondor autoscaler dependencies


community/modules/scripts/htcondor-install/files/install-htcondor-autoscaler-deps.yml
```diff
@@ -24,13 +24,13 @@
...
       name: pip
-      version: 21.3.1  # last Python 2.7-compatible release
+      version: 21.3.1  # last Python 3.6-compatible release
       virtualenv: /usr/local/htcondor
       virtualenv_command: /usr/bin/python3 -m venv
   - name: Install latest setuptools
     ansible.builtin.pip:
       name: setuptools
-      state: 44.1.1  # last Python 2.7-compatible release
+      version: 59.6.0  # last Python 3.6-compatible release
```
nick-stroud, line 33: Since you are proposing a hotfix on main, could you explain the outcome of this bug. Basically, justify the hotfix.

nick-stroud, line 33: Since you are merging to main, I think you should add a patch version roll to this PR. You could do it in a follow on PR but since all integration tests must be run for merge to main it would be easier to do all at once.

nick-stroud, line 33: Reminder to run integration tests for merge to main.

tpdownes: I thought these automatically ran. I will trigger them once I address the other issues.

tpdownes: I have updated the PR description. Should have been clearer. Thanks.

nick-stroud: no need since not a hot fix to main.

nick-stroud (approved):

PR #593, opened 2022-10-03
Fix: passing non-existent cache was causing failures in some projects

nick-stroud:
We (@douglasjacobsen and myself) have observed that some projects fail on `gs://optionally_set_spack_cache_bucket` while others do not. Root cause is not yet known, but until it is found I am changing the structure of the blueprint to eliminate the failure. 

### Submission Checklist

  install`)


tpdownes (approved):
LGTM

PR #596, opened 2022-10-04
Support local module required APIs

tpdownes:
If a local resource could plausibly be a local development copy of an embedded module, then ensure that its required API list defaults to that of the embedded module. This also provides a pathway for development to change and test the default value while modifying the embedded module.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/sourcereader/local.go
+++ b/pkg/sourcereader/local.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"hpc-toolkit/pkg/modulereader"
 	"os"
+	"strings"
 )
 
 // LocalSourceReader reads modules from a local directory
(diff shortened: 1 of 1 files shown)


pkg/sourcereader/local.go
```diff
@@ -30,7 +31,13 @@ func (r LocalSourceReader) GetModuleInfo(modPath string, kind string) (modulerea
 	}
 
 	reader := modulereader.Factory(kind)
-	return reader.GetInfo(modPath)
+	mi, err := reader.GetInfo(modPath)
+	if idx := strings.Index(modPath, "/community/modules/"); idx != -1 {
+		mi.RequiredApis = defaultAPIList(modPath[idx+1:])
+	} else if idx := strings.Index(modPath, "/modules/"); idx != -1 {
+		mi.RequiredApis = defaultAPIList(modPath[idx+1:])
```
cboneti, line 38: Nit: since defaultAPIList is now used by embedded and local, should it be moved to sourcereader.go instead of embedded.go? (I'm ok with not moving it if you think it is better this way).

cboneti (approved):
Thank you for submitting this. I think it makes a positive difference in usability. 
One nit that you may choose to ignore.

pkg/sourcereader/local.go
tpdownes: I had same thought:

- rename defaultRequiredAPIs
- move to sourcereader.go

but wanted a simple commit for purpose of later / clearer reversion when there is more automation here.

I will merge this while noting that we should consider the above if/when we touch this more (e.g. a `GitSourceReader` implementation).

PR #598, opened 2022-10-04
Clarify and compact validator failure message

tpdownes:
A user who attempts to create a blueprint with a validator failure in it will now receive a more compact warning with shorter `goo.gle` URLs to our GitHub documentation. The links below point to a mix of main and develop due to the current state of the branches; it is my intent for them all to point to main when all the files land on main.

```
baremetalsolution.googleapis.com: service is disabled in project fake-project-1701
baremetalsolution.googleapis.com: can be enabled at https://console.cloud.google.com/apis/library/baremetalsolution.googleapis.com?project=fake-project-1701
one or more required APIs are disabled in project fake-project-1701, please enable them as instructed above
warning: validator test_apis_enabled failed

validators are shown below.

Validation failures were treated as a warning, continuing to create blueprint.

Terraform group was successfully created in directory ansible/primary
To deploy, run the following commands:
  terraform -chdir=test/primary init
  terraform -chdir=test/primary validate
  terraform -chdir=test/primary apply
```

### Submission Checklist

  install`)


tpdownes:
Force push reflects a minor change in text you suggested in internal review.

pkg/config/validate.go
```diff
@@ -109,20 +110,22 @@ func (dc DeploymentConfig) executeValidators() error {
 	}
 
 	if warned || errored {
-		log.Println("validator failures can indicate a credentials problem.")
-		log.Println("cloud credentials associated with your Google Cloud account can be generated by the following command:")
-		log.Println("  $ gcloud auth application-default login")
```
nick-stroud, line 114: One side effect of this change is that the recommendation to call `gcloud auth application-default login` is getting buried/dropped. 

1. Do we think this is still a valid reason for failure (even if it is one of several)?
2. Do we print anything about `gcloud auth` in the messages referenced by "See messages above for suggested"?
3. Should we add what a gcloud auth failure looks like in troubleshooting doc?

tpdownes: Sorry, I meant to point this out explicitly in the description.

 (1) = yes, (2) = yes, if you count a link to those instructions. We handle it better now.

- Recent changes address [the case when there are no loadable application default credentials](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/6526126fe1fa6bc4972c7c754e8aee52ee799de6/pkg/validators/validators.go#L41-L48)
  - Every `test_x_y` validator trips over this code 

This handles the "user has no application default credentials" case which that message addresses.

(3)

Otherwise, it is hard and maybe impossible to disambiguate "I have the wrong credentials"  from "I have the right credentials but my boss won't give me the right IAM role from "I'm logged in under my personal gmail account but forgot". I believe all cases return a simple 404 Not Found because you don't want unauthorized users to detect the presence of projects by just poking around. (I am willing to believe that I may have something wrong here, though)

My instinct is that these cases can't be addressed in code but in improved documentation guiding the user around the difference between `gcloud auth login` and "application default credentials" and how to know which you're using. I'm a bit hesitant to start writing things without a bit of real world experience of the failures users have. We should rely on cloud.google.com as much as possible.

nick-stroud: I am happy with (2) in place.

pkg/config/validate.go
```diff
@@ -109,20 +110,22 @@ func (dc DeploymentConfig) executeValidators() error {
...
-		log.Println("- can be disabled entirely by using the create command")
-		log.Println("  with the flag \"--validation-level IGNORE\"")
-		log.Println("- a custom set of validators can be configured following")
-		log.Println("  instructions at:")
-		log.Println("https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/blueprint-validation.md")
+		log.Println("One or more blueprint validators has failed. See messages above for suggested")
+		log.Println("actions. General troubleshooting guidance and instructions for configuring")
+		log.Println("validators are shown below.")
+		log.Println("")
+		log.Println("- https://goo.gle/hpc-toolkit-troubleshooting")
```
nick-stroud, line 117: I will just note here that I recognize you are using the short links. On one hand this looks cleaner, on the other there is an added dependency on the maintenance of the short links. I could go either way but am fine to use them.

nick-stroud (approved):

pkg/config/validate.go
tpdownes: I suggest trying them and reverting if it becomes a problem.

PR #601, opened 2022-10-04
Add slurm-gcp-v5-ubuntu2004 example to integration tests

heyealex:
Dependent on https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/587 being merged. This will be rebased before submitting for review.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/slurm-gcp-v5-ubuntu2004.yaml
+++ b/community/examples/slurm-gcp-v5-ubuntu2004.yaml
@@ -68,6 +68,8 @@ deployment_groups:
     - debug_partition
     - compute_partition
     - homefs
+    settings:
+      disable_controller_public_ips: false
 
   - id: slurm_login
     source: community/modules/scheduler/schedmd-slurm-gcp-v5-login
@@ -76,3 +78,4 @@ deployment_groups:
     - slurm_controller
     settings:
       machine_type: n2-standard-4
+      disable_login_public_ips: false
--- a/community/examples/slurm-gcp-v5-ubuntu2004.yaml
+++ b/community/examples/slurm-gcp-v5-ubuntu2004.yaml
@@ -21,7 +21,7 @@ vars:
   deployment_name: slurm-gcp-v5
   region: us-central1
   zone: us-central1-c
-  source_image_family: "schedmd-v5-slurm-22-05-2-ubuntu-2004-lts"
+  source_image_family: "schedmd-v5-slurm-22-05-3-ubuntu-2004-lts"
   source_image_project: "projects/schedmd-slurm-public/global/images/family"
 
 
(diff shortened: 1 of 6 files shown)


cboneti (approved):
lgtm
q: why adding it to the same group as the other V5?

heyealex:
Partially just because I could, and there was space to do so in that group, but also I like having a test that deploys 2 v5 clusters in the same group in parallel as a bit of a stress test. It seems to handle it pretty well actually!

tpdownes (changes requested):
Unless there is a good reason not to, I believe that we should the [current published image family](https://github.com/SchedMD/slurm-gcp/blob/0abeab3f907778674aab7ffcfed5e48b9463ee3a/docs/images.md#published-image-family): `schedmd-v5-slurm-22-05-3-ubuntu-2004-lts`

tpdownes:
> Partially just because I could, and there was space to do so in that group, but also I like having a test that deploys 2 v5 clusters in the same group in parallel as a bit of a stress test. It seems to handle it pretty well actually!

I think the problem with parallel builds is that a failure in one will cause the other to pause rather than cleanup. Group 3 and 4 are examples of doing it linearly for this reason.

tpdownes:
It won't let me make a comment in an unchanged section, but line 20 of the integration group 3 YAML file should be updated to include the new test in whatever dependency order we decide upon.

heyealex:
> Unless there is a good reason not to, I believe that we should the [current published image family](https://github.com/SchedMD/slurm-gcp/blob/0abeab3f907778674aab7ffcfed5e48b9463ee3a/docs/images.md#published-image-family): `schedmd-v5-slurm-22-05-3-ubuntu-2004-lts`

Fixed. I also ran another test, and it also fixed the issue with SMT core counts, so I reverted to using the default for `disable_smt` as well. Thanks for catching that!

heyealex:
> > Partially just because I could, and there was space to do so in that group, but also I like having a test that deploys 2 v5 clusters in the same group in parallel as a bit of a stress test. It seems to handle it pretty well actually!
> 
> I think the problem with parallel builds is that a failure in one will cause the other to pause rather than cleanup. Group 3 and 4 are examples of doing it linearly for this reason.

Good point. I changed it to sequential, if it takes too much time I'll also move one of the tests to another group.

tpdownes (dismissed):
Assuming the check passes, LGTM

heyealex:
Switching to sequential execution did take longer, so I moved the ubuntu image to group 4 instead (both ran at around 30min before, now are ~50min). I will need a follow-up review when you have time.

cboneti (approved):

tpdownes (approved):

PR #604, opened 2022-10-05
Escape variables capability

thiagosgobe:
### Submission Checklist

  install`)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -798,7 +798,7 @@ defined before creating a deployment, making debugging quicker and easier.
 
 ### Escape Variables
 
-Under circumstances where is needed escape either a literal variable or blueprint variables format, a non-quoted backslash `\` is the escape character. It preserves the literal value of the next character that follows:
+Under circumstances where the variable notation conflicts with the content of a setting or string, for instance when defining a startup-script runner that uses a subshell like in the example below, a non-quoted backslash (`\`) can be used as an escape character. It preserves the literal value of the next character that follows:
 
 * `\$(not.bp_var)` evaluates to `$(not.bp_var)`.
 * `\((not.literal_var))` evaluates to `((not.literal_var))`.
--- a/examples/README.md
+++ b/examples/README.md
@@ -795,3 +795,28 @@ everything inside will be provided as is to the module.
 Whenever possible, blueprint variables are preferred over literal variables.
 `ghpc` will perform basic validation making sure all blueprint variables are
 defined before creating a deployment, making debugging quicker and easier.
+
+### Escape Variables
+
+Under circumstances where is needed escape either a literal variable or blueprint variables format, a non-quoted backslash `\` is the escape character. It preserves the literal value of the next character that follows:
+
+* `\$(not.bp_var)` evaluates to `$(not.bp_var)`.
+* `\((not.literal_var))` evaluates to `((not.literal_var))`.
+
+```yaml
+deployment_groups:
+  - group: primary
+     modules:
+       - id: resource1
+         source: path/to/module/1
+         settings:
+            key1: \((not.literal_var))   ## Evaluates to "((not.literal_var))".
+         ...
+       - id: resource2
+         source: path/to/module/2
+         ...
+         settings:
+            key1: |
+              #!/bin/bash
+              echo \$(cat /tmp/file1)    ## Evaluates to "echo $(cat /tmp/file1)"
+```
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -34,13 +34,11 @@ const (
 	roleLabel             string = "ghpc_role"
 	simpleVariableExp     string = `^\$\((.*)\)$`
 	deploymentVariableExp string = `^\$\(vars\.(.*)\)$`
-	anyVariableExp        string = `\$\((.*)\)`
-	literalExp            string = `^\(\((.*)\)\)$`
-	escapeVariableExp     string = `\\\$\((.*?)\)`
-	// Checks if a non-escaped variable exists only as a substring, ex:
+	// Checks if a variable exists only as a substring, ex:
 	// Matches: "a$(vars.example)", "word $(vars.example)", "word$(vars.example)", "$(vars.example)"
 	// Doesn't match: "\$(vars.example)", "no variable in this string"
-	nonEscapeVariableExp string = `(?:\s|^|[[:word:]]|[^\\])\$\((.*)\)`
+	anyVariableExp string = `(^|[^\\])\$\((.*)\)`
+	literalExp     string = `^\(\((.*)\)\)$`
 	// the greediness and non-greediness of expression below is important
 	// consume all whitespace at beginning and end
 	// consume only up to first period to get variable source
@@ -403,13 +401,6 @@ type varContext struct {
 	blueprint  Blueprint
 }
 
-// Escape blueprint variable
-// Convert \$(var.variable) to $(var.variable)
-func expandEscapeVariable(str string) string {
-	re := regexp.MustCompile(escapeVariableExp)
-	return re.ReplaceAllString(str, (`$(${1})`))
-}
-
 // Needs DeploymentGroups, variable string, current group,
 func expandSimpleVariable(
 	context varContext,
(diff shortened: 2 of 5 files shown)


pkg/config/expand.go
```diff
@@ -503,6 +512,28 @@ func isSimpleVariable(str string) bool {
 	return matched
 }
 
+// isEscapeVariable checks if the entire string is an escaped variable \$(not.var)
+func isEscapeVariable(str string) bool {
+	// if string has multiple variables all of them must have escape character
+	if hasNoEscapeVariable(str) {
```
heyealex, line 518: I'm not sure I understand the reason for this. Is it not possible to work around existing variables? I want to make sure it's eventually possible to support the following case when expandVariable is implemented:
`\$(ls -l $(vars.list_path))`
In this case, the outer "variable" is escaped to let this run as a subshell command, but access the vars.list_path variable inside of that command.

examples/README.md
```diff
@@ -795,3 +795,28 @@ everything inside will be provided as is to the module.
 Whenever possible, blueprint variables are preferred over literal variables.
 `ghpc` will perform basic validation making sure all blueprint variables are
 defined before creating a deployment, making debugging quicker and easier.
+
+### Escape Variables
+
+Under circumstances where is needed escape either a literal variable or blueprint variables format, a non-quoted backslash `\` is the escape character. It preserves the literal value of the next character that follows:
```
heyealex, line 801: nit: I reworded this a bit for clarity.
```
Under circumstances where the variable notation conflicts with the content of a setting or string,
for instance when defining a startup-script runner that uses a subshell like in the example below,
a non-quoted backslash (`\`) can be used as an escape character. It preserves the literal value of
the next character that follows:
```

pkg/modulewriter/tfwriter.go
```diff
@@ -65,8 +65,12 @@ func createBaseFile(path string) error {
 }
 
 func handleLiteralVariables(hclBytes []byte) []byte {
+	// Convert ((var.variable)) to var.variable
 	re := regexp.MustCompile(`"\(\((.*?)\)\)"`)
-	return re.ReplaceAll(hclBytes, []byte(`${1}`))
+	hclBytes = re.ReplaceAll(hclBytes, []byte(`${1}`))
+	// Convert \((not.variable)) to ((not.variable))
+	re = regexp.MustCompile(`\\\\\(\((.*?)\)\)`)
```
heyealex, line 72: Is this matching `\\((*))` or `\((*))`?

heyealex (commented):
This is looking good, I added a couple comments on how it could be improved. I also have some ideas around improving the flow of `handleVariable`, but we can talk through that offline.

pkg/config/expand.go
```diff
@@ -503,6 +512,28 @@ func isSimpleVariable(str string) bool {
 	return matched
 }
 
+// isEscapeVariable checks if the entire string is an escaped variable \$(not.var)
```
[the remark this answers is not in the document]
thiagosgobe: Yes. That's right. I rephrased that to "isEscapeVariable checks if variables within string are escaped variables"

pkg/config/expand.go
```diff
@@ -36,6 +36,8 @@ const (
 	deploymentVariableExp string = `^\$\(vars\.(.*)\)$`
 	anyVariableExp        string = `\$\((.*)\)`
 	literalExp            string = `^\(\((.*)\)\)$`
+	escapeVariableExp     string = `\\\$\((.*?)\)`
+	nonEscapeVariableExp  string = `(?:\s|^|[[:word:]]|[^\\])\$\((.*)\)`
```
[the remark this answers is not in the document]
thiagosgobe: Indeed it's complex that I can't remember exactly the meaning but I tested dropping `[[:word:]]` and `\s` didn't work properly but dropping only `[[:word:]]` it looks work fine. Just need to test a bit more to be sure.

pkg/modulewriter/tfwriter.go
thiagosgobe: It is matching `\\((*))`. When converted to byte it double the backslash for some reason.

pkg/config/expand.go
thiagosgobe: I added this `hasNoEscapeVariable` function to avoid this situation meanwhile we do not have expandVariable implemented but I think I got your point. Should be better call expandVariable function explicitly in such situations, is that what you thought?

pkg/config/expand.go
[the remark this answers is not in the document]
thiagosgobe: You were right. I changed that to ``(^|[^\\])\$\((.*)\)``

thiagosgobe:
@heyealex  ready for reviewing. I resolved the comments that doesn't make sense anymore.

pkg/modulewriter/hcl_utils.go
```diff
@@ -17,11 +17,24 @@ package modulewriter
...
 	"path/filepath"
+	"regexp"
 
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/zclconf/go-cty/cty"
 )
 
+func escapeBlueprintVariables(hclBytes []byte) []byte {
+	// Convert \$(not.variable) to $(not.variable)
+	re := regexp.MustCompile(`\\\\\$\((.*?)\)`)
```
heyealex, line 28: A couple suggestions here:
1. You could drop the final `\)` and the captured text, since it wouldn't matter in the final expression. Example below of what I'm thinking.
1. You should also check that the backslash itself isn't being escaped. If it is, both should be kept.

Example 1:
```go
re := regexp.MustCompile(`\\\\\$\(`)
return re.ReplaceAll(hclBytes, []byte(`$(`)
```

Example 2:
```go
re := regexp.MustCompile(`[^\\]\\\\\$\((.*?)\)`)
return re.ReplaceAll(hclBytes, []byte(`$(${1})`))
```

Combined:
```go
re := regexp.MustCompile(`[^\\]\\\\\$\(`)
return re.ReplaceAll(hclBytes, []byte(`$(`)
```

pkg/modulewriter/hcl_utils.go
```diff
@@ -17,11 +17,24 @@ package modulewriter
...
+	return re.ReplaceAll(hclBytes, []byte(`$(${1})`))
+}
+
+func escapeLiteralVariables(hclBytes []byte) []byte {
+	// Convert \((not.variable)) to ((not.variable))
+	re := regexp.MustCompile(`\\\\\(\((.*?)\)\)`)
```
heyealex, line 34: Same suggestion as above for this function.

```go
re := regexp.MustCompile(`[^\\]\\\\\(\(`)
return re.ReplaceAll(hclBytes, []byte(`((`))
```

pkg/modulewriter/tfwriter.go
```diff
@@ -64,9 +64,19 @@ func createBaseFile(path string) error {
 	return err
 }
 
+func handleBlueprintVariables(hclBytes []byte) []byte {
```
heyealex, line 67: Since they are in the same class, you could simply use `escapeBlueprintVariables` from hcl_utils.go in tfwriter. That would replace the calls to `handleBlueprintVariables()` with `escapeBlueprintVariables()` in the write* functions.

Eventually, the goal is to move much more from tfwriter and packerwriter into shared functions, this is a good opportunity to do that.

pkg/modulewriter/tfwriter.go
```diff
@@ -64,9 +64,19 @@ func createBaseFile(path string) error {
...
+	return re.ReplaceAll(hclBytes, []byte(`$(${1})`))
+}
+
 func handleLiteralVariables(hclBytes []byte) []byte {
+	// Convert ((var.variable)) to var.variable
 	re := regexp.MustCompile(`"\(\((.*?)\)\)"`)
-	return re.ReplaceAll(hclBytes, []byte(`${1}`))
+	hclBytes = re.ReplaceAll(hclBytes, []byte(`${1}`))
+	// Convert \((not.variable)) to ((not.variable))
+	re = regexp.MustCompile(`\\\\\(\((.*?)\)\)`)
```
heyealex, line 78: Similar to above, you could call `escapeLiteralVariables` here since it performs the same task:
```go
...
hclBytes = re.ReplaceAll(hclBytes, []byte(`${1}`))
return escapeLiteralVariables(hclBytes)
```

heyealex (commented):
Thank you so much for adding the tests and making these updates! It's looking really good. 

I have just a couple minor comments I'd like to see addressed. Let me know if you have any questions on any of them.

pkg/modulewriter/hcl_utils.go
thiagosgobe: 1. Yes. Let's make it as simple as possible.
2. If I understood correctly, It actually already happens

Let's say we have a blueprint file:

```yaml
dummyAttributeName = abc \\$(escape.backslash) abc
```

golang add 2 escape backslash `\` and when converted to bytes it is doubled:

```golang
dummyAttributeName = \"abc \\\\\\\\$(escape.backslash) abc\"\n\n"
```

Calling func `escapeBlueprintVariables()` removes 4 backslashs:

```golang
dummyAttributeName = \"abc \\\\$(escape.backslash) abc\"\n\n"
```

Then when go writes to file:

```hcl
dummyAttributeName = abc \\$(escape.backslash) abc
```

pkg/modulewriter/tfwriter.go
thiagosgobe: Absolutely! I should have thought that before.

heyealex:
This all looks good! I kicked off one full integration test that includes a packer test and a couple terraform tests. Assuming that succeeds, this will be ready to merge.

heyealex (approved):

PR #605, opened 2022-10-05
Remove deployment_name var from vpc example as it is passed implicitly

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/modules/network/vpc/README.md
+++ b/modules/network/vpc/README.md
@@ -109,16 +109,10 @@ compact set of subnetworks possible.
 - id: network1
   source: modules/network/vpc
   settings:
-  - deployment_name: $(vars.deployment_name)
+    network_name: cluster-net
 ```
 
-This creates a new VPC network named based on the `deployment_name` variable
-with `_net` appended. `network_name` can be set manually as well as part of the
-settings.
-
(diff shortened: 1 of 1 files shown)


modules/network/vpc/README.md
```diff
@@ -108,8 +108,6 @@ compact set of subnetworks possible.
 ```yaml
 - id: network1
   source: modules/network/vpc
-  settings:
-  - deployment_name: $(vars.deployment_name)
 ```
 
 This creates a new VPC network named based on the `deployment_name` variable
```
heyealex, line 113: There's a **_NOTE:_** that should be updated or removed as well.
I agree that it shouldn't be in the default example anymore, but I think it was initially included to demonstrate how to set some settings. 
It might be worth including `network_name: custom-vpc`(or something similar) as a secondary example; suggestion only, feel free to disregard.

heyealex (changes requested):

modules/network/vpc/README.md
nick-stroud: Removed the note and added alt setting to keep user aware of settings in general as per our offline discussion.

heyealex (approved):

PR #610, opened 2022-10-07
Filestore cleanup fails on existing instances

heyealex:
### Description
This PR ensures a failure will occur in the daily-project-cleanup.yaml cloud config that notifies the team when the test project has pre-existing filestore instances that have not been cleaned up. The trigger will not fail if there are active cloud builds, assuming that any active builds should be prioritized, are intentional and do not imply any invalid state.

A README was added to the cloud-build directory as well to document the purpose of the cloud-build tools in the repo.

### Submission Checklist

  install`)


tpdownes (approved):
Based on the log entries of previous runs, it seems evident that this would address our concern.

PR #611, opened 2022-10-07
Add image support documentation

heyealex:
### Description
Documents our Ubuntu support. In writing this, I validated against all of our modules and ran tests against modules and startup scripts that take image as an input where it hadn't been verified already or explicitly only supports specific images.

### Submission Checklist

  install`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,11 @@ Be aware that Cloud Shell has [several limitations][cloud-shell-limitations],
 in particular an inactivity timeout that will close running shells after 20
 minutes. Please consider it only for blueprints that are quickly deployed.
 
+## VM Image Support
+
+The HPC Toolkit supports the HPC CentOS 7 VM image and the Ubuntu 20.04 LTS
+image family. For more information, see [docs/vm-images.md](docs/vm-images.md).
+
 ## Blueprint Validation
 
 The Toolkit contains "validator" functions that perform basic tests of the
--- a/README.md
+++ b/README.md
@@ -103,12 +103,13 @@ minutes. Please consider it only for blueprints that are quickly deployed.
 
 ## VM Image Support
 
-The HPC Toolkit supports the following VM images:
+The HPC Toolkit officially supports the following VM images:
 
 * HPC CentOS 7
 * Ubuntu 20.04 LTS
 
-For more information, see [docs/vm-images.md](docs/vm-images.md).
+For more information on these and other images, see
+[docs/vm-images.md](docs/vm-images.md).
 
 ## Blueprint Validation
 
--- a/README.md
+++ b/README.md
@@ -103,8 +103,12 @@ minutes. Please consider it only for blueprints that are quickly deployed.
 
 ## VM Image Support
 
-The HPC Toolkit supports the HPC CentOS 7 VM image and the Ubuntu 20.04 LTS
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -101,6 +101,11 @@ Be aware that Cloud Shell has [several limitations][cloud-shell-limitations],
 in particular an inactivity timeout that will close running shells after 20
 minutes. Please consider it only for blueprints that are quickly deployed.
 
+## VM Image Support
+
+The HPC Toolkit supports the HPC CentOS 7 VM image and the Ubuntu 20.04 LTS
```
nick-stroud, line 106: nit: I would add the word `officially` in here. I have used other images with the Toolkit with some success. Optionally, make some statement unsupported images, that they might work but are not actively tested.

nick-stroud, line 106: nit: Does it matter to make the distinction between centos 7 and the hpc centos 7? I might consider listing both.

nick-stroud, line 106: nit: I would turn this into a list as I expect it to grow over time and it will make it more readable.

docs/vm-images.md
```diff
@@ -0,0 +1,54 @@
...
+* [htcondor-configure]: Only provides support for the HPC CentOS 7 image.
+* [nfs-server]: Only provides support for CentOS 7 images for the server itself.
+* [DDN-EXAScaler]: The underlying terraform module has a limitted set of
+  supported images [documented][exascalerimages] in the exascaler-cloud-terraform
+  github repo. **This requirement only applies to the servers, not the clients.**
+* [omnia-install]: Only provides support for Rocky 8.
+
+HPC Toolkit modules and startup scripts
+have been evaulated against the `ubuntu-2004-lts` image family. For more
+information about the Ubuntu Google Cloud images, see the Canonical
```
nick-stroud, line 32: nit: should Canonical be capitalized?

nick-stroud (approved):

README.md
heyealex: Added both.

heyealex: There are changes in the HPC images that may affect how well we support other modules. I'd rather keep this strict for "official" support for now. I am open to further discussion on that though.

heyealex: Done. Also added a table of contents in the vm-images doc that will help when the list gets bigger.

docs/vm-images.md
heyealex: It's the company name that publishes ubuntu: https://canonical.com/

nick-stroud: Ha. I see.

README.md
nick-stroud: Still nit: My lingering concern is that someone will read this section and be like, "Well, if they don't support Rocky then I guess I am not interested".

I think the section you added in the images doc is great. I am wondering if we can tie it in somehow here to keep from losing these people. 

Maybe saying:

>The HPC Toolkit **officially** supports the following VM images:
>...
>
>For more information **on these and other images**, see [docs/vm-images.md](docs/vm-images.md).

(bolding used to show additions)

nick-stroud (approved):

README.md
heyealex: I see, I was looking at the wrong file when I updated. It should be updated now.

nick-stroud (approved):

PR #614, opened 2022-10-10
Add additional network interfaces to vm-instance

heyealex:
This commit adds the option to specify more than one network interface. Additional network interfaces more closely match the terraform format, and allow a lot more direct configuration. This way, the simple route is provided via defaults and the first network interface and additional network interfaces are targeted more towards knowledgeable users with specific requirements.

access_config has been simplified to `disable_public_ips` similar to the default network interface. This could be expanded later on to provide full customization.

alias_ip_range and ipv6_access_config have not been included, but could be if they are needed at a future time.

I'm very open to discussing any of these decisions further. 

### Submission Checklist

  install`)


modules/compute/vm-instance/main.tf
```diff
@@ -133,6 +133,24 @@ resource "google_compute_instance" "compute_vm" {
     nic_type   = local.enable_gvnic ? "GVNIC" : null
   }
 
+  dynamic "network_interface" {
+    for_each = var.additional_network_interfaces
```
cboneti, line 137: I think it is a good practice to keep parameter names whenever possible. In this case, you are effectively renaming network interface with "additional_network_interfaces". I would suggest you keep the same name. In fact, why not keep it the same variable?

tools/validate_configs/test_configs/2-network-interfaces.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - subnet_name: primary-subnet-three
+        subnet_region: $(vars.region)
+        new_bits: 8
+
+  - id: workstation
+    source: ./modules/compute/vm-instance
+    use:
+    - network1
+    settings:
+      additional_network_interfaces:
```
cboneti, line 58: I don't know exactly how I would do it, but I would discourage departing too much from the terraform usability. Therefore, if network interface is defined, I would void the use of network1 and essentially require that the user defines any of the interfaces. That would ensure that this module and the terraform underlying vm remain sufficiently aligned.

modules/compute/vm-instance/main.tf
```diff
@@ -133,6 +133,24 @@ resource "google_compute_instance" "compute_vm" {
...
+
+    content {
+      network            = lookup(network_interface.value, "network", null)
+      subnetwork         = lookup(network_interface.value, "subnetwork", null)
+      subnetwork_project = lookup(network_interface.value, "subnetwork_project", null)
+      network_ip         = lookup(network_interface.value, "network_ip", null)
+      nic_type           = lookup(network_interface.value, "nic_type", null)
+      stack_type         = lookup(network_interface.value, "stack_type", null)
+      queue_count        = lookup(network_interface.value, "queue_count", null)
+      dynamic "access_config" {
```
cboneti, line 147: I would rather have an external variable, disable_public_ips that determine default values here if the interface is not defined or if the value vas omitted, but I would not rename the structure member. Furthermore, if a user passes access config, it would be ignored here. 
Access config contains other interesting options such as an IP address, a networking tier, or the domain name.

cboneti (changes requested):
A few recommendations, PTAL

tools/validate_configs/test_configs/2-network-interfaces.yaml
```diff
@@ -0,0 +1,65 @@
...
+      network_address_range: 10.0.0.0/16
+      subnetworks:
+      - subnet_name: primary-subnet-three
+        subnet_region: $(vars.region)
+        new_bits: 8
+
+  - id: workstation
+    source: ./modules/compute/vm-instance
+    use:
+    - network1
```
nick-stroud, line 56: I will start by saying that I have thought about this zero. What would it take to make multiple nics work with use? aka: `use: [network1, network2]`

cboneti: good point!

tpdownes (changes requested):
The pattern you are emulating here is the pattern we've chosen to deprecate in the `vpc` module. I think we should take some lessons from that deprecation and emulate the pattern we ultimately settled upon.

In that module, ignoring the deprecated input variables, we have `var.subnetworks` which is just a list of explicitly defined subnetworks. By default it is the empty list. If the user supplies no value, then it is replaced by a default subnetwork constructed from other inputs.

I think it would be ambitious to scope in supporting `use: [network0, ..., networkN]`. In my view the right way to do that is to build a module (if that feels heavyweight, call it a "module-let", "meta-pattern", "support for locals blocks") that does nothing but generate network interfaces in the right format for this module.

I also think this worsens the situation with respect to the external and internal IP address output values. Right now, we output the first external and internal IP as a list for all VMs. In the common case of creating only 1 VM, this is a list of length 1 and we have to use `(( module.id.internal_ip[0] ))` to extract the value elsewhere in a blueprint.

Right now, the outputs aren't capturing anything except the first IP address. I'd want to understand how to present this info to the user as what might become a MxN matrix, where M is the number of instances and N is the number of interfaces per instance. In this case, the output values (singular "ip") are _really_ misnamed.

heyealex:
Update from offline discussion:
As a short-term solution that can be expanded in the future, support both the original variables and additional network interfaces by doing the following:
* Remove the `additional_` from the variable name so it can be used more generically in the future
* If `network_interfaces` is supplied, ignore the default variables (`network_self_link`, `subnetwork_self_link`, etc)
* Convert from a map to an object
* Update the test_config to define all subfields

modules/compute/vm-instance/main.tf
heyealex: Variable name was updated. 
Terraform doesn't appear to support setting blocks as variables: https://discuss.hashicorp.com/t/use-block-variables/5527/3

tools/validate_configs/test_configs/2-network-interfaces.yaml
heyealex: Agreed, the variable has been updated to align with terraforms network_interface.

modules/compute/vm-instance/main.tf
heyealex: access_config has been added, and in fact it now supports multiple if desired. 

I chose to only apply `var.disable_public_ips` (top level variable) to the network interface as defined by `subnetwork_self_link` and `network_self_link`. If a value is passed into `network_interfaces`, the assumption is that the user knows what they are doing  and wouldn't want their configuration overwritten. What do you think about that?

tools/validate_configs/test_configs/2-network-interfaces.yaml
heyealex: Agreed; that can now be accomplished via a follow-up PR by creating a `network_interfaces` output from vpc and pre-existing-vpc. I'm considering that out of scope based on our discussion this morning.

modules/compute/vm-instance/variables.tf
```diff
@@ -136,6 +136,33 @@ variable "subnetwork_self_link" {
...
+    network            (string, required if subnetwork is not supplied)
+    subnetwork         (string, required if network is not supplied)
+    subnetwork_project (string, optional)
+    network_ip         (string, optional)
+    disable_public_ips (bool, optional, whether or not to create a public IP for the instance in this network, defaults to true)
+    nic_type           (string, optional, choose from ["GVNIC", "VIRTIO_NET"])
+    stack_type         (string, optional, choose from ["IPV4_ONLY", "IPV4_IPV6"])
+    queue_count        (number, optional)
+    EOT
+  type        = list(map(any))
```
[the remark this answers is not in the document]
heyealex: network_interface is now defined as an object, as are all sub-fields that not primitives.

modules/compute/vm-instance/main.tf
```diff
@@ -61,6 +61,22 @@ locals {
...
+    subnetwork_project = var.project_id
+    network_ip         = null
+    nic_type           = local.enable_gvnic ? "GVNIC" : null
+    stack_type         = null
+    queue_count        = null
+    access_config      = var.disable_public_ips ? [] : [{}]
+    ipv6_access_config = []
+    alias_ip_range     = []
+  }
+  network_interfaces = length(var.network_interfaces) == 0 ? [local.base_network_interface] : var.network_interfaces
```
tpdownes, line 79: Nit: you might consider [coalescelist](https://www.terraform.io/language/functions/coalescelist). This function dates back to the 0.x days.

tpdownes: Above a nit, below a show-stopper: I prefer "default" to "base" which suggests something that might always be there.

modules/compute/vm-instance/variables.tf
```diff
@@ -136,6 +136,73 @@ variable "subnetwork_self_link" {
...
+  default = []
+  validation {
+    condition = alltrue([
+      for ni in var.network_interfaces : (ni.network == null) != (ni.subnetwork == null)
+    ])
+    error_message = "All additional network interfaces must define exactly one of \"network\" or \"subnetwork\"."
+  }
+  validation {
+    condition = alltrue([
+      for ni in var.network_interfaces : ni.nic_type == "GVNIC" || ni.nic_type == "VIRTIO_NET" || ni.nic_type == null
```
tpdownes, line 194: nit: above 3 values, I'd start considering [contains](https://www.terraform.io/language/functions/contains)

tpdownes (changes requested):
Running the test YAML (overriding only project_id) gives me the following error. The line #s are from vm-instance but I can't obviously see what part of the blueprint it traces back to.

```
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 155, in resource "google_compute_instance" "compute_vm":
│  155:           nat_ip                 = access_config.value.nat_ip
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "nat_ip".
╵
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 155, in resource "google_compute_instance" "compute_vm":
│  155:           nat_ip                 = access_config.value.nat_ip
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "nat_ip".
╵
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 156, in resource "google_compute_instance" "compute_vm":
│  156:           public_ptr_domain_name = access_config.value.public_ptr_domain_name
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "public_ptr_domain_name".
╵
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 156, in resource "google_compute_instance" "compute_vm":
│  156:           public_ptr_domain_name = access_config.value.public_ptr_domain_name
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "public_ptr_domain_name".
╵
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 157, in resource "google_compute_instance" "compute_vm":
│  157:           network_tier           = access_config.value.network_tier
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "network_tier".
╵
╷
│ Error: Missing map element
│
│   on modules/vm-instance/main.tf line 157, in resource "google_compute_instance" "compute_vm":
│  157:           network_tier           = access_config.value.network_tier
│     ├────────────────
│     │ access_config.value is empty map of string
│
│ This map does not have an element with the key "network_tier".
```

tools/validate_configs/test_configs/2-network-interfaces.yaml
```diff
@@ -0,0 +1,159 @@
...
+  region: us-central1
+  zone: us-central1-c
+
+deployment_groups:
+- group: primary
+  modules:
+  # Source is an embedded module, denoted by "modules/*" without ./, ../, /
+  # as a prefix. To refer to a local or community module, prefix with ./, ../ or /
+  # Example - ./modules/network/vpc
+  - id: default
```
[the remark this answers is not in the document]
heyealex: Fixed

modules/compute/vm-instance/main.tf
heyealex: coalescelist is much cleaner, thanks for the suggestion. Fixed both.

modules/compute/vm-instance/variables.tf
heyealex: I like the idea, but `contains` doesn't seem to play well with values that can be null:
```
│ 
│   on modules/vm-instance/variables.tf line 200, in variable "network_interfaces":
│  200:       for ni in var.network_interfaces : ni.stack_type == null || contains(["IPV4_ONLY", "IPV4_IPV6"], ni.stack_type)
│     ├────────────────
│     │ ni.stack_type is null
│ 
```

So the options seem to be:
1. keep it as it is
1. require these values to be specified every time

heyealex:
I've addressed your comments @tpdownes. 
The error was due to how the default network interface was being created, and it seems like the final test I did before pushing didn't include that change somehow. The default if `disable_public_ips` is false (default) is to provide an empty map as the access_config, but now that it's been updated to an object, the fields are required.

tpdownes (approved):
My only concern is that we do not appear to be exercising anything to do with IPV6 or alias_ip_range. I do not think that should hold up this feature.

modules/compute/vm-instance/main.tf
cboneti: agree, as long as it is well documented in the variables description.

PR #616, opened 2022-10-11
Improve Batch output instructions and add README to login node

nick-stroud:
This change:
- Adds more explicit instructions about created Batch job template and instance template
- Adds a README.md file to the folder generated on the login node
- Renames Google Cloud Batch to Batch in documentation

Example output:
```
instructions_batch-job = <<EOT

A Batch job template file has been created locally at:
  /***/hpc-toolkit/simple-batch/primary/cloud-batch-simple-batch.json

An instance template was created for your Batch job & referenced in the above file:
  https://www.googleapis.com/compute/v1/projects/***/global/instanceTemplates/simple-batch-instance-template-20221011005718936600000001


Use the following commands to:
Submit your job:
  gcloud alpha batch jobs submit simple-batch --config=/***/hpc-toolkit/simple-batch/primary/cloud-batch-simple-batch.json --location=us-central1 --project=***
  
Check status:
  gcloud alpha batch jobs describe simple-batch --location=us-central1 --project=*** | grep state:
  
Delete job:
  gcloud alpha batch jobs delete simple-batch --location=us-central1 --project=***

List all jobs:
  gcloud alpha batch jobs list --project=***

EOT
```

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/cloud-batch-job/outputs.tf
+++ b/community/modules/scheduler/cloud-batch-job/outputs.tf
@@ -15,8 +15,8 @@
  */
 
 locals {
-  provided_instance_tpl_msg  = "An instance template was provided to be used by your Batch job"
-  generated_instance_tpl_msg = "An instance template was created for your Batch job"
+  provided_instance_tpl_msg  = "The Batch job template uses the existing VM instance template:"
+  generated_instance_tpl_msg = "The Batch job template uses a new VM instance template created matching the provided settings:"
 }
 
 output "instructions" {
@@ -26,9 +26,8 @@ output "instructions" {
   A Batch job template file has been created locally at:
     ${abspath(local.job_template_output_path)}
 
-  ${var.instance_template == null ? local.generated_instance_tpl_msg : local.provided_instance_tpl_msg} & referenced in the above file:
+  ${var.instance_template == null ? local.generated_instance_tpl_msg : local.provided_instance_tpl_msg}
     ${local.instance_template}
-
 
   Use the following commands to:
   Submit your job:
--- a/community/modules/scheduler/cloud-batch-job/outputs.tf
+++ b/community/modules/scheduler/cloud-batch-job/outputs.tf
@@ -14,11 +14,23 @@
  * limitations under the License.
  */
 
+locals {
+  provided_instance_tpl_msg  = "An instance template was provided to be used by your Batch job"
+  generated_instance_tpl_msg = "An instance template was created for your Batch job"
+}
+
 output "instructions" {
-  description = "Instructions for submitting Google Cloud Batch job."
+  description = "Instructions for submitting the Batch job."
   value       = <<-EOT
+
+  A Batch job template file has been created locally at:
+    ${abspath(local.job_template_output_path)}
+
+  ${var.instance_template == null ? local.generated_instance_tpl_msg : local.provided_instance_tpl_msg} & referenced in the above file:
+    ${local.instance_template}
+
+
   Use the following commands to:
-  
   Submit your job:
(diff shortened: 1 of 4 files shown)


community/modules/scheduler/cloud-batch-job/outputs.tf
```diff
@@ -14,11 +14,23 @@
...
+
 output "instructions" {
-  description = "Instructions for submitting Google Cloud Batch job."
+  description = "Instructions for submitting the Batch job."
   value       = <<-EOT
+
+  A Batch job template file has been created locally at:
+    ${abspath(local.job_template_output_path)}
+
+  ${var.instance_template == null ? local.generated_instance_tpl_msg : local.provided_instance_tpl_msg} & referenced in the above file:
```
tpdownes, line 29: The `&` seems a bit out of place in the code and in the output. I'd just use "and".

tpdownes: In fact, overall I would simplify to:

```
The Batch job template uses the existing instance template:
```

or

```
The Batch job template uses a new instance template created matching the provided settings:
```

community/modules/scheduler/cloud-batch-login-node/outputs.tf
```diff
@@ -22,21 +22,15 @@ output "login_node_name" {
 output "instructions" {
   description = "Instructions for accessing the login node and submitting Google Cloud Batch jobs"
   value       = <<-EOT
-  Use the following commands to:
 
+  A Batch job template file will be placed on the Batch login node at:
+    ${local.job_template_destination}
+
+
```
tpdownes, line 29: Do we need 2 blank lines here?

tpdownes (changes requested):
Without testing each instructed command in detail, these changes seem simple and align our documentation with the branding.

My primary concern is with the use of Terraform outputs as a means by which to guide the user for next actions. (A) we are probably relying alphanumeric order of module ID and output name (B) we are relying on the user to explicitly add the `outputs` in the blueprint.

Re: (A) it is a better solution than the primary alternative (nothing at all)

Re: (B) I assume our cloud-batch example only adds instructions for the cloud-batch-login-node because they would be duplicative of the cloud-batch-job instructions?

community/modules/scheduler/cloud-batch-job/outputs.tf
nick-stroud: Done.

community/modules/scheduler/cloud-batch-login-node/outputs.tf
nick-stroud: It was an aesthetic choice that I do not feel strongly about. I have removed the 2nd blank line.

nick-stroud:
>(A) we are probably relying alphanumeric order of module ID and output name (B) we are relying on the user to explicitly add the `outputs` in the blueprint.
> 
> Re: (A) it is a better solution than the primary alternative (nothing at all)
> 
> Re: (B) I assume our cloud-batch example only adds instructions for the cloud-batch-login-node because they would be duplicative of the cloud-batch-job instructions?

I agree there are some deficiencies with this pattern but I do not know a better way to go about this. We have used this pattern elsewhere, for example in the [dashboard](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/develop/modules/monitoring/dashboard/README.md#output_instructions) module. I especially do not like that the instructions do not print when there is a failure, for example if you add a wait for startup and it fails then you never see the instructions.

For (B) you are correct, the information is basically duplicate but environment specific so they only need to print one.

I don't think there is any action here, LMK if you have ideas.

tpdownes (approved):

tpdownes:
No action needed, I just think we need to be sure we understand the limits of `output` instructions and understand the particular choices in our Batch module.

PR #618, opened 2022-10-11
Warn on git hash mismatch

kkr16:
This change makes the CLI throw a warning to `strerr` in case of git version mismatch between git hash used to build the binary, and the current working directory git hash. Warning looks like:

`WARNING: ghpc binary was built from a different commit (warn-cli-branch/54270c1) than the working directory (main/54270c1). You can rebuild the binary by running 'make'.`

In case of error (e.g. working directory is not a git repo, or user doesn't have access to working directory), the check will fail silently and gracefully, to avoid confusing the user. 

### Submission Checklist

  install`)


cboneti:
gcbrun/

cmd/root.go
```diff
@@ -50,8 +54,22 @@ HPC deployments on the Google Cloud Platform.`,
 // Execute the root command
 func Execute() error {
 	if len(GitCommitInfo) > 0 {
+		dir, err := os.Getwd()
```
cboneti, line 57: nit: this is assuming your are working from the same location as ghpc. But what if ghpc is in path and the user is somewhere else? Could we also check where the binary is located? (also difficult to figure out what is best).

cboneti: also, what if you are not in a hpc-toolkit repo, but in another git repo?

cboneti (commented):
Thank you for adding this. I will let Nick do the final review. In the meantime, I have two small comments.

cmd/root.go
```diff
@@ -50,8 +54,22 @@ HPC deployments on the Google Cloud Platform.`,
...
+		if err == nil {
+			repo, err := git.PlainOpen(dir)
+			if err == nil {
+				head, err := repo.Head()
+				if err == nil && GitCommitHash != head.Hash().String() {
+					b := head.Name().Short()
+					h := head.Hash().String()
+					fmt.Fprintf(os.Stderr, "WARNING: ghpc binary was built from a different commit (%s/%s) than the working directory (%s/%s). You can rebuild the binary by running 'make'\n",
+						GitBranch, GitCommitHash[0:7], b, h[0:7])
```
nick-stroud, line 66: I would like to see some unit tests for this. It is a little hard for a few reasons:
- result is a printed line
- interacts with git/ file system
- requires built binary

I think it could be testable if you put in a function and pass in binary flags. AKA:

`func warnGitHashMismatch(GitCommitHash) -> string`

by passing in the constant you can essentially mock the state of the binary. By returning the string you can test the output. The the layer above in the stack would be responsible for passing in the git hash (from build flag) and printing a string if returned. You could run tests like:
- In test get the current commit hash, pass it in and see that no warning is returned
- In test, get previous commit hash, pass it in and see warning is returned
- Pass in random hash (not in repo), and see no warning is returned (case of different repo)

go.mod
```diff
@@ -29,10 +30,16 @@ require (
 require (
 	cloud.google.com/go v0.102.1 // indirect
 	cloud.google.com/go/iam v0.3.0 // indirect
+	github.com/Microsoft/go-winio v0.4.16 // indirect
```
nick-stroud, line 33: I never quite know what to do with all these go.mod changes. I am leaving this open as a reminder for me to follow up.

cmd/root.go
kkr16: **To the first comment:**, my thinking for using workdir instead of binary site was that if the user had an official release installed under `/usr/bin` or `~/bin` or in another directory added to their `PATH` AND was developing in a working directory with ghpc source code, it would still warn the user in case they call `$ ghpc` instead of `$ ./ghpc`.

**To the second comment: **, we could also  compare the initial commits of the working directory vs binary using something similar to `git rev-list --max-parents=0 HEAD` which would prevent false negatives in situations where the user has a working directory with a git branch that is a few commits behind their binary.

kkr16: I updated the logic as follows. Code will traverse the working directory tree until it hits the root of the file system, looking for a hpc-toolkit repo. If repo not found, we will then look at the hash of the call site.

`hpc-toolkit` repo is positively identified by searching for the initial commit (injected at build time) in the current repository.

cmd/root.go
```diff
@@ -50,8 +54,22 @@ HPC deployments on the Google Cloud Platform.`,
...
+				}
+			}
+		}
```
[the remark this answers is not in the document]
kkr16: ACK and fixed. Thanks for the feedback.

cmd/root.go
kkr16: I added a new file `root_test.go` and reached 44.6% coverage - if needed, we could improve coverage further by adding test cases to the `rootCmd.SetVersionTemplate` code, though those parts may require a slight rewrite to make them more testable.   
Happy to pick it up separately after the PR is merged.

This is the first "production grade" go test file I write from scratch, please be thorough in your review if possible, and nit away! :)

cmd/root.go
```diff
@@ -68,3 +82,81 @@ Commit info: {{index .Annotations "commitInfo"}}
...
+		return b, h, dir
+	}
+	return "", "", ""
+}
+
+// hpcToolkitRepo will find the path of the directory containing the hpc-toolkit
+// starting with the working directory and evaluating the parent directories until
+// the toolkit repository is found. If the HPC Toolkit repository is not found by
+// traversing the path, then the executable directory is checked.
+func hpcToolkitRepo() (*git.Repository, string, error) {
```
nick-stroud, line 115: What case is all of this protecting against?   
1. If I call `git rev-list --max-parents=0 HEAD` in a subdirectory of a git repo it still works just fine. For example I get the same  response if I call it from `hpc-toolkit/` or `hpc-toolkit/examples/`. I would assume `git.PlainOpen(dir)` would work even if you are in a subdir (I could be wrong). The only thing I can think that we are protecting against by crawling up the dir tree is if there is a git repo inside of the hpc-toolkit repo. This is possible but I don't think it is a requirement and I don't know if the extra complexity is worth it.
2. I don't think we should ever be falling back to the executable directory. I tried to bring this up in [this comment](https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/618#discussion_r993661237). Let's discuss.

go.mod
```diff
@@ -16,23 +17,29 @@ require (
 	github.com/zclconf/go-cty v1.11.1
 	golang.org/x/exp v0.0.0-20220907003533-145caa8ea1d0
 	google.golang.org/genproto v0.0.0-20221014213838-99cd37c6964a
-	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
```
nick-stroud, line 20: Any idea why this rolled? Mostly just curious. It is updated to the latest version, which is still from 2020 so I see no issue here.

cmd/root_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+
+	"github.com/go-git/go-billy/v5"
+	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/storage/memory"
+	. "gopkg.in/check.v1"
+)
+
+type MySuite struct{}
```
nick-stroud, line 32: Tests generally look good except I would expect them to change a bit when other comments are closed.

cmd/root.go
nick-stroud: I am asking the group what to do here. Currently `make tests` fails because coverage is not over 80%, but it is an obvious improvement from the previous 0% so I would like to take as is. More on this...

cmd/root.go
```diff
@@ -68,3 +82,81 @@ Commit info: {{index .Annotations "commitInfo"}}
 }
 
 func init() {}
+
+// checkGitHashMismatch will compare the hash of the git repository vs the git
+// hash the ghpc binary was compiled against
+func checkGitHashMismatch() (string, string, string) {
```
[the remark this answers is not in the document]
kkr16: Thank you!

cmd/root.go
kkr16: `git.PlainOpen(dir)` expects `dir` to be the root of the repo and to contain the `.git` directory, so I initially put this to walk the parent directories' path until the root of the git repo is identified. 

Following your comment (thanks!!), I looked again and found I can instead do something like the below, and rely on the library to find the `.git` directory.

```
o := git.PlainOpenOptions{DetectDotGit: true}
repo, _ := git.PlainOpenWithOptions(repoDir, &o)
```

With that change, the only benefit of this block would be if ./ghpc is called from within a subfolder that contains a different git repo.

I'm leaving this open and unchanged for now, pending our discussion.

cmd/root.go
kkr16: As discussed, with @nick-stroud I've disabled the new tests for the cmd package, and I'll take on a task to write tests to get coverage > 80%

cmd/root_test.go
```diff
@@ -29,100 +29,132 @@ import (
...
-	// built in a test dir that is not in a git repository
-	os.Chdir(filepath.Dir(expected))
+	// find repo when workdir is root of repo
+	os.Chdir(repoDir)
+	_, got, err = hpcToolkitRepo()
+	c.Assert(got, Equals, repoDir)
+	c.Assert(err, IsNil)
+
+	// find repo when workDir is another git repo under ./hpc-toolkit
+	dir := ".." + "/hpc-toolkit/modules/network/vpc/.terraform/modules/cloud_router/.git"
```
nick-stroud, line 68: Update to not depend on the existence of this git repo. (discussed offline)

cmd/root_test.go
```diff
@@ -29,100 +29,132 @@ import (
...
+	err = os.Chdir(dir)
+	_, got, err = hpcToolkitRepo()
+	c.Assert(got, Equals, repoDir)
+	c.Assert(err, IsNil)
+
+	// Try to find repo when workdir is outside root of repo. Normal execution
+	// returns binary site in. Empty value expected during tests since test
+	// binary is built in a test dir that is not in a git repository
+	err = os.Chdir(filepath.Dir(repoDir))
 	_, got, _ = hpcToolkitRepo()
```
nick-stroud, line 78: nit: What is our defined api for when we can't find the repo. In the code you check for if err != nil, here you check that path is empty string. I would check for err in test instead of path.

cmd/root_test.go
```diff
@@ -29,100 +29,132 @@ import (
...
+	GitInitialHash = "6c50deedff0520887a50b8ad2ca84724ff1adae8"
+	GitCommitHash = "f4ed7c1ca864c77ea8cc21bd9286b0c48d28db95"
+	mismatch, b, h, dir = checkGitHashMismatch()
+	c.Assert(mismatch, Equals, false)
+	c.Assert(b, Equals, "")
+	c.Assert(h, Equals, "")
 
 	// verify current working directory git hash against a wrong random hash
+	GitInitialHash = "8fc4768edbef9b3f115a41eaf2a5740d41758cff"
 	GitCommitHash = "6c50deedff0520887a50b8ad2ca84724ff1adae8"
```
nick-stroud, line 127: nit: It took me a minute to wrap my head around these cases. I think readability could be improved with some intermediate variable naming like:

```
randomGitHash := f4ed7c1ca864c77ea8cc21bd9286b0c48d28db95
initalGitHash := 8fc4768edbef9b3f115a41eaf2a5740d41758cff
oldGitHashFromMain := f4ed7c1ca864c77ea8cc21bd9286b0c48d28db95

// and then the setup would read

GitInitialHash = initalGitHash
GitCommitHash = randomGitHash
```

nick-stroud (approved):

PR #625, opened 2022-10-13
Refactor mount shell script to take args

nick-stroud:
This change:
- is a no-op
- refactors the pre-existing-network-storage mount command to be a shell script that takes args
- fixes complaints from shell linter

Rational: 
- The shell script is more portable than a templated string
- The shell script is now checked by pre-commit

### Submission Checklist

  install`)


nick-stroud:
Test failed on omnia (I believe unrelated) before the lustre test could run. I am rerunning.

tpdownes (approved):
This seems like an improvement for the reasons you note and better to scope it to refactor without other considerations.

PR #628, opened 2022-10-14
Mount script also makes entry in /etc/fstab

nick-stroud:
This change:
- Adds functionality for mount options to mount.sh script
- Adds entry for mount in /etc/fstab

Tested:
- manually tested mounting and fstab entry with and without options

Note: ATM this script is only used for mounting lustre with pre-existing-network-storage but the intent is to make it more broadly used in follow on PR.

### Submission Checklist

  install`)

### DIFF
--- a/modules/file-system/pre-existing-network-storage/scripts/mount.sh
+++ b/modules/file-system/pre-existing-network-storage/scripts/mount.sh
@@ -30,3 +30,13 @@ if ! findmnt --source "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" --target "${LO
 else
 	echo "Skipping mounting source: ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}, already mounted to target:${LOCAL_MOUNT}"
 fi
+
+MOUNT_IDENTIFIER="${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}[[:space:]]${LOCAL_MOUNT}"
+if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
+	echo "Adding ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} -> ${LOCAL_MOUNT} to /etc/fstab"
+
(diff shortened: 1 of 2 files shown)


modules/file-system/pre-existing-network-storage/scripts/mount.sh
```diff
@@ -17,11 +17,26 @@ SERVER_IP=$1
...
+
+MOUNT_IDENTIFIER="${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}[[:space:]]${LOCAL_MOUNT}"
+if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
+	echo "Adding ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} -> ${LOCAL_MOUNT} to /etc/fstab"
+
+	[[ -z "${MOUNT_OPTIONS}" ]] && POPULATED_MOUNT_OPTIONS="defaults" || POPULATED_MOUNT_OPTIONS="${MOUNT_OPTIONS}"
+	echo "${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} ${LOCAL_MOUNT} ${FS_TYPE} ${POPULATED_MOUNT_OPTIONS} 0 0" >>/etc/fstab
+else
+	echo "Skipping editing /etc/fstab as entry already exists"
+fi
```
cboneti, line 42: This may be a nit, but why do you mount first and then add to the fstab? 
Should you not, instead: 
1. check if not in fstab and add it if needed.
2. mount <target> (so it will pull from fstab)

This could help with idempotency.

That should guarantee consistency between what is mounted and what is in the fstab.... (in case what was previously added there is somewhat different than what is specified here. 

Also, there are a number of potential problems here that we are choosing to ignore and that maybe we should at the very least document and think about it: 
1. What if the mount identified is already provided, but with different options or local mount? 
2. What if another mount identifier is mounting to the same destination? (i.e. /mnt, more probable than the first).
3. What if you are rebooting, and that the mount was already done during the boot ? (should be ok here).

cboneti (approved):
Ok to submit as is, or to pursue the improvements described in my comments.

modules/file-system/pre-existing-network-storage/scripts/mount.sh
nick-stroud: To unblock other development I am going to merge this as is. I will document follow on actions in the task and follow up with incremental improvements in a separate PR.

PR #629, opened 2022-10-14
Install Ansible automatically when other runners depend upon it

tpdownes:
### Submission Checklist

  install`)


modules/scripts/startup-script/variables.tf
```diff
@@ -79,3 +79,9 @@ EOT
   }
   default = []
 }
+
+variable "prepend_ansible_installer" {
+  description = "Prepend Ansible installation script if any of the specified runners are of type ansible-local"
+  type        = bool
+  default     = true
+}
```
cboneti, line 87: Nit; missing an empty line at the end of the file

modules/scripts/startup-script/main.tf
```diff
@@ -45,13 +45,23 @@ locals {
...
+
+  ansible_installer = {
+    type        = "shell"
+    source      = "${path.module}/examples/install_ansible.sh"
+    destination = "install_ansible.sh"
+  }
+
+  ansible_local_runners     = [for r in var.runners : r if r.type == "ansible-local"]
+  prepend_ansible_installer = length(local.ansible_local_runners) > 0 && var.prepend_ansible_installer
+  runners                   = local.prepend_ansible_installer ? concat([local.ansible_installer], var.runners) : var.runners
```
cboneti, line 64: super-nit: I typically find it more elegant to define variables before they are used (call it a reminder of my mid 90's C programming). So I would have defined these lines at the top... ;)

cboneti (commented):
I realize this may not be ready for review yet, but I had a few comments that I wanted to add. Feel free to add me as actual reviewer if you are looking for one.

modules/scripts/startup-script/variables.tf
tpdownes: Addressed nit while I was addressing a build failure.

modules/scripts/startup-script/main.tf
tpdownes: Addressed nit while I was addressing a build failure.

tpdownes:
To address a build failure, I have:

- removed all explicit `install_ansible.sh` from our blueprints (both examples and tests)
- named the destination for the automatic one to avoid colliding with `install_ansible.sh`

There are several places where this runner is output:

- modules/file-system/filestore/outputs.tf
- community/modules/scripts/spack-install/outputs.tf
- community/modules/scripts/omnia-install/outputs.tf
- community/modules/file-system/nfs-server/outputs.tf

I think we should address them but I feel it should be done more carefully than the mass deletion I am submitting here.

tpdownes:
Had to force push after noticing an unrelated tweak snuck into the branch. I am going to leave the 4 builds going although the link to GitHub is broken due to the new commit. I'll report success/failure.

cboneti:
* https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/6a7c8744a2ead97b9c5c2f77b685b06046dea595/modules/file-system/filestore/outputs.tf doesn't actually output the running. But now, it erroneously guides the user to add the install ansible. That should probably be fixed as part of this PR. 
I think you should address these comments there as they do not really install anything, but as is, it is broken documentation.

cboneti (approved):

PR #636, opened 2022-10-17
Improve Ansible linting

tpdownes:
Most of the action in this PR is in ebb1daf3ffb5cdeee6fa106d59ffc79841970fb1. The theory here is that the ansible-lint pre-commit hook does not behave like typical hooks. This behavior is reasonable for a repo aimed at Ansible playbooks, but not our own.

The status quo:

- always trigger
- run ansible-lint on the entire directory structure and allow it to make smart decisions (e.g. differentiate between blueprint YAML and ansible YAML)

The change:

- trigger hook only when a commit includes YAML files
- BUT do not trigger on YAML files in blueprint directories
- configure ansible-lint so that it does not operate on files in blueprint directories (because it will run on the whole repo regardless of the trigger source)
- exclude jinja failures because we do not have default values / ansible directory structure for many of our templating variables

These changes combined with #435 enable better-targeted real world linting and also allow us to upgrade ansible-lint pre-commit hook itself. The linter now catches errors in our Ansible code and I have resolved approximately 40 failures along this PR branch.

You can test this by editing an Ansible code to use `command` instead of `ansible.builtin.command` under `tools` and it will properly report the FQCN violation. Meanwhile, editing a blueprint under `{community/}examples` doesn't trigger the hook at all.

### Submission Checklist

  install`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,9 +57,14 @@ repos:
   - id: go-critic
     args: [-disable, "#experimental,sloppyTypeAssert"]
 - repo: https://github.com/ansible/ansible-lint.git
-  rev: v5.4.0
+  rev: v6.8.2
   hooks:
   - id: ansible-lint
+    always_run: false
+    exclude: '^(docs/|examples/|community/examples/)'
+    types: [yaml]
+    additional_dependencies:
+    - ansible==6.*
 - repo: https://github.com/adrienverge/yamllint.git
   rev: v1.28.0
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
     name: terraform-readme
     entry: tools/autodoc/terraform_docs.sh
     language: script
-    types: ['terraform']
+    types: [terraform]
     exclude: \.terraform\/.*$
     pass_filenames: true
     require_serial: true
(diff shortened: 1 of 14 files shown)


.pre-commit-config.yaml
```diff
@@ -57,9 +57,14 @@ repos:
   - id: go-critic
     args: [-disable, "#experimental,sloppyTypeAssert"]
 - repo: https://github.com/ansible/ansible-lint.git
-  rev: v5.4.0
+  rev: v6.8.2
   hooks:
   - id: ansible-lint
+    always_run: false
+    exclude: '^(docs/|examples/|community/examples/)'
+    types: [yaml]
```
nick-stroud, line 65: I am assuming (as proven by the files in which errors were caught) that this will also pick up files with `.yml` suffix.

nick-stroud (approved):

.pre-commit-config.yaml
tpdownes: if you follow the chain starting at [the pre-commit docs](https://pre-commit.com/#filtering-files-with-types) you find this [section of code](https://github.com/pre-commit/identify/blob/56aa6afbe66c9da96dcdb66d9a61632f45c6b44b/identify/extensions.py#L253-L256) that demonstrates both `yml` and `yaml` match the `yaml` type.

PR #643, opened 2022-10-18
Improve and document mount.sh

nick-stroud:
This change:
- makes edit in fstab first then mounts from fstab
- documents undefined behavior when local mount already exists
- undoes mounting and fstab entry automatically added by DDN client auto installation (more discussion below)

The DDN lustre client installer does a host of things, including making an entry in fstab and mounting the lustre. The issue here is that:
- the installer is slow, even when called a second time
- there is no way to set mount options for these steps

For these reasons we really want install and mount to be separate. The two lines added to undo the fstab entry and mounting done by the auto installer while leaving the client in place.

### Submission Checklist

  install`)

### DIFF
--- a/modules/file-system/pre-existing-network-storage/README.md
+++ b/modules/file-system/pre-existing-network-storage/README.md
@@ -24,6 +24,22 @@ This creates a pre-existing-network-storage module in terraform at the
 provided IP in `server_ip` of type nfs that will be mounted at `/home`. Note
 that the `server_ip` must be known before deployment.
 
+### Mounting
+
+For some `fs_type`, this module will provide `client_install_runner` and
+`mount_runner` outputs. These can be used to create a startup script to mount
+the network storage system.
+
+Supported `fs_type`:
+
+- lustre (DDN)
+
+[scripts/mount.sh](./scripts/mount.sh) is used as the contents of
+`mount_runner`. This script will update `/etc/fstab` and mount the
+network storage. Behavior is undefined if:
+
+- System already has an entry in `/etc/fstab` for `local_mount`.
+
 ## License
 
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/modules/file-system/pre-existing-network-storage/scripts/mount.sh
+++ b/modules/file-system/pre-existing-network-storage/scripts/mount.sh
@@ -19,18 +19,6 @@ LOCAL_MOUNT=$3
 FS_TYPE=$4
 MOUNT_OPTIONS=$5
 
-if ! findmnt --source "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" --target "${LOCAL_MOUNT}" &>/dev/null; then
-	echo "Mounting --source ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} --target ${LOCAL_MOUNT}"
-	mkdir -p "${LOCAL_MOUNT}"
-	if [ -z "${MOUNT_OPTIONS}" ]; then
-		mount -t "${FS_TYPE}" "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" "${LOCAL_MOUNT}"
-	else
-		mount -t "${FS_TYPE}" -o "${MOUNT_OPTIONS}" "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" "${LOCAL_MOUNT}"
-	fi
-else
-	echo "Skipping mounting source: ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}, already mounted to target:${LOCAL_MOUNT}"
-fi
-
 MOUNT_IDENTIFIER="${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}[[:space:]]${LOCAL_MOUNT}"
 if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
 	echo "Adding ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} -> ${LOCAL_MOUNT} to /etc/fstab"
@@ -39,4 +27,12 @@ if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
 	echo "${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} ${LOCAL_MOUNT} ${FS_TYPE} ${POPULATED_MOUNT_OPTIONS} 0 0" >>/etc/fstab
 else
 	echo "Skipping editing /etc/fstab as entry already exists"
+fi
+
+if ! findmnt --source "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" --target "${LOCAL_MOUNT}" &>/dev/null; then
+	echo "Mounting --target ${LOCAL_MOUNT} from fstab"
+	mkdir -p "${LOCAL_MOUNT}"
+	mount --target "${LOCAL_MOUNT}"
+else
+	echo "Skipping mounting source: ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}, already mounted to target:${LOCAL_MOUNT}"
 fi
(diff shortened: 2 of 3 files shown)


modules/file-system/pre-existing-network-storage/README.md
```diff
@@ -24,6 +24,22 @@ This creates a pre-existing-network-storage module in terraform at the
 provided IP in `server_ip` of type nfs that will be mounted at `/home`. Note
 that the `server_ip` must be known before deployment.
 
+### Mounting
+
+For some `fs_type`, this module will provide `client_install_runner` and
```
cboneti, line 29: nit: s/For some `fs_type`/For the `fs_type` listed below,/

cboneti: also, what will happen for the unsuported ones? (meaning, it will still populate the data structure, but will not provide the mounters, which is fine, as long as these are part of the image...)

modules/file-system/pre-existing-network-storage/README.md
```diff
@@ -24,6 +24,22 @@ This creates a pre-existing-network-storage module in terraform at the
...
+
+Supported `fs_type`:
+
+- lustre (DDN)
+
+[scripts/mount.sh](./scripts/mount.sh) is used as the contents of
+`mount_runner`. This script will update `/etc/fstab` and mount the
+network storage. Behavior is undefined if:
+
+- System already has an entry in `/etc/fstab` for `local_mount`.
```
cboneti, line 41: this is not an fs type, I think it is confusing to have it this way.

modules/file-system/pre-existing-network-storage/scripts/mount.sh
```diff
@@ -19,18 +19,6 @@ LOCAL_MOUNT=$3
...
-		mount -t "${FS_TYPE}" "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" "${LOCAL_MOUNT}"
-	else
-		mount -t "${FS_TYPE}" -o "${MOUNT_OPTIONS}" "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" "${LOCAL_MOUNT}"
-	fi
-else
-	echo "Skipping mounting source: ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}, already mounted to target:${LOCAL_MOUNT}"
-fi
-
 MOUNT_IDENTIFIER="${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH}[[:space:]]${LOCAL_MOUNT}"
 if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
```
cboneti, line 23: I believe here you should look for only ${SERVER_IP}:${REMOTE_MOUNT_WITH_SLASH} and not the local mount. The reason is that you may want to mount the same file system in multiple places.. right? 
However, looking for only the SERVER_IP will not be sufficient either, because multiple remote mounts may exist under the same ip.

modules/file-system/pre-existing-network-storage/templates/ddn_exascaler_luster_client_install.tftpl
```diff
@@ -46,3 +46,5 @@ do
 done
 chmod +x /usr/sbin/esc-client
```
cboneti, line 47: do you think the error here will be clear if the curl never succeeded? (maybe it will)

modules/file-system/pre-existing-network-storage/templates/ddn_exascaler_luster_client_install.tftpl
```diff
@@ -46,3 +46,5 @@ do
...
 esc-client auto setup --config /etc/esc-client.conf
+esc-client unmount -m "${local_mount}"
+mv /etc/fstab.last /etc/fstab
```
cboneti, line 50: what are these lines here? why do you need to remove it from the fstab? (won't the mount be a no-op in that case?)

cboneti (commented):
a few comments

modules/file-system/pre-existing-network-storage/README.md
nick-stroud: In the case it is unsupported there will still be a runner  added that just prints something like "Skipping, not supported". See the first and last lines of [this section](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/54270c169a6edbea50adbb50a1e919b9fc5d8cad/modules/file-system/pre-existing-network-storage/outputs.tf#L66-L75).

modules/file-system/pre-existing-network-storage/templates/ddn_exascaler_luster_client_install.tftpl
nick-stroud: Each curl failure is printed in the startup script, so it will look like 20 curl failures and then a file does not exist. To me that is pretty clear what happened but maybe it could be made better. Do you have a suggestion?

modules/file-system/pre-existing-network-storage/README.md
nick-stroud: Case 1: remote mount is different and local mount is the same from existing fstab entry:
- entry will be added to fstab
- the device will never be mounted as the local mount will already be busy
- even if the local mount point doesn't have anything mounted, when you call mount target it will select the first matching item in the list, so your device will never be mounted

Case 2: remote mount is same and local mount is different from existing fstab entry:
- entry will be added to fstab
- device will mount normally (you will have the same device mounted in 2 locations)

So how ever we say it the real problem arises if local mount is the same, which is what I am trying to call out here. Maybe I should change the `MOUNT_IDENTIFIER` to only look for common local mount. Basically it doesn't matter if remote mount is different, if there is already an entry for that local mount then we error. WDYT?

Then I could change the wording to say, "mounting will fail if there is an entry in fstab that already uses the same local mount". This is a much clearer and definitive statement.

modules/file-system/pre-existing-network-storage/README.md
nick-stroud: The wording from your original substitution suggestion has been incorporated. Leaving comment open in case there is more discussion here.

cboneti: yes, but it will still populate the data structure, is that structure no longer being added to instance metadata?

modules/file-system/pre-existing-network-storage/scripts/mount.sh
```diff
@@ -40,3 +28,11 @@ if ! grep -q "${MOUNT_IDENTIFIER}" /etc/fstab; then
 else
 	echo "Skipping editing /etc/fstab as entry already exists"
 fi
+
+if ! findmnt --source "${SERVER_IP}":"${REMOTE_MOUNT_WITH_SLASH}" --target "${LOCAL_MOUNT}" &>/dev/null; then
+	echo "Mounting --target ${LOCAL_MOUNT} from fstab"
+	mkdir -p "${LOCAL_MOUNT}"
+	mount --target "${LOCAL_MOUNT}"
```
cboneti, line 35: Maybe just check for a return value > 0 here. 
See mount return codes here: https://linux.die.net/man/8/mount

cboneti (commented):
back to you

modules/file-system/pre-existing-network-storage/README.md
nick-stroud: It is still being added to network storage. As it stand now it will always emit a runner, they will just be more or less a no-op if it is unsupported. 

Another valid option would be to set runners to `null` in the unsupported case (although it is not set up that way now). Would this be your preference? In one case we silently emit null, in the other case we emit a runner even when unsupported, but the runner logs that it is unsupported. 

`Null` maybe seems more appropriate but I could go either way and don't want to change unless that is also what you want.

nick-stroud: As discussed offline, we will leave as is: populate a runner with a warning.

modules/file-system/pre-existing-network-storage/templates/ddn_exascaler_luster_client_install.tftpl
nick-stroud: `set -e` has been added to this script and tested (as discussed offline).

modules/file-system/pre-existing-network-storage/README.md
nick-stroud: Done. The wording has been updated.

modules/file-system/pre-existing-network-storage/scripts/mount.sh
nick-stroud: This has been resolved by using `set -e` on this script.

cboneti (approved):

PR #650, opened 2022-10-18
Default to the compute engine service account if null is provided

heyealex:
### Description
Get and set the service account to the default compute engine service account if it was not provided. 

The enable_reconfigure setting of the controller currently requires an explicit service_account to be defined, as the default email set in the underlying modules is still just "default", which is not accepted when running the reconfig scripts. This is a temporary workaround until the underlying modules handle this situation in a similar way.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
@@ -28,25 +28,28 @@ locals {
       node_conf              = var.node_conf
 
       # Template By Definition
-      additional_disks         = var.additional_disks
-      bandwidth_tier           = var.bandwidth_tier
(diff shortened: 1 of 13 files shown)


community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -68,6 +71,9 @@ locals {
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
 }
 
+data "google_compute_default_service_account" "default" {
```
nick-stroud, line 74: nit: It is not clear to me why this change was needed now. Maybe just add a note in the description with the "why".

Feel free to merge without sending back to me.

PR #653, opened 2022-10-19
Add documentation for Slurm reconfigure option

heyealex:
More content may be added as testing continues

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
@@ -50,6 +50,32 @@ This creates a controller node with the following attributes:
 For a complete example using this module, see
 [slurm-gcp-v5-cluster.yaml](../../../examples/slurm-gcp-v5-cluster.yaml).
 
+### Reconfigure Option
+The schedmd-slurm-gcp-v5-controller module supports the reconfiguration of
+partitions and slurm configuration in a running, active cluster. This option is
+activated through the `enable_reconfigure` setting:
+
+```yaml
+- id: slurm_controller
+  source: community/modules/scheduler/schedmd-slurm-gcp-v5-controller
+  settings:
+    enable_reconfigure: true
+```
+
+Reconfigure has some additional requirements:
+
+* The Pub/Sub API must be activated in the target project:
+  `gcloud services enable file.googleapis.com --project "<<PROJECT_ID>>"`
+* Python package dependencies need to be installed with pip in the environment
+  deployment the cluster. See the [Slurm on GCP Documentation][optdeps] for more
+  info.
+* The project in your gcloud config must match the project the cluster is being
+  deployed onto due to a known issue with the reconfigure scripts. To set your
+  default config project, run the following command:
+  `gcloud config set core/<<PROJECT ID>>`
+
+[optdeps]: https://github.com/SchedMD/slurm-gcp/tree/v5.1.0/terraform/slurm_cluster#optional
+
 ## Support
 The HPC Toolkit team maintains the wrapper around the [slurm-on-gcp] terraform
 modules. For support with the underlying modules, see the instructions in the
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -50,6 +50,32 @@ This creates a controller node with the following attributes:
...
+activated through the `enable_reconfigure` setting:
+
+```yaml
+- id: slurm_controller
+  source: community/modules/scheduler/schedmd-slurm-gcp-v5-controller
+  settings:
+    enable_reconfigure: true
+```
+
+Reconfigure has some additional requirements:
```
cboneti, line 65: This option has ...

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -50,6 +50,32 @@ This creates a controller node with the following attributes:
 For a complete example using this module, see
 [slurm-gcp-v5-cluster.yaml](../../../examples/slurm-gcp-v5-cluster.yaml).
 
+### Reconfigure Option
```
cboneti, line 53: I this the feature official name? was that already defined somewhere?  Is it clear? 
Otherwise, we could maybe call it Live Cluster Reconfiguration (`enable_reconfigure`)

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -50,6 +50,32 @@ This creates a controller node with the following attributes:
...
+  source: community/modules/scheduler/schedmd-slurm-gcp-v5-controller
+  settings:
+    enable_reconfigure: true
+```
+
+Reconfigure has some additional requirements:
+
+* The Pub/Sub API must be activated in the target project:
+  `gcloud services enable file.googleapis.com --project "<<PROJECT_ID>>"`
+* Python package dependencies need to be installed with pip in the environment
```
cboneti, line 69: Python must be installed ([Python](https://github.com/SchedMD/slurm-gcp/blob/3979e81fc5e4f021b5533a23baa474490f4f3614/docs/glossary.md#python) is installed.
Required Version: >= 3.6.0, < 4.0.0) and python packages dependencies ...

It is pretty hard impossible to follow these instructions as they are there, since it will say: 
pip3 install -r ../../../scripts/requirements.txt --user

But that file is not in our repo, and so they have to find its contents in the slurm on gcp repo, and now that file contains 63 lines of dependencies, which we suspect are not all needed for reconfigure.. I started a thread with SchedMD to see if we could narrow down the dependencies and, at least, point them to the right file here.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -50,6 +50,32 @@ This creates a controller node with the following attributes:
...
+  deployment the cluster. See the [Slurm on GCP Documentation][optdeps] for more
+  info.
+* The project in your gcloud config must match the project the cluster is being
+  deployed onto due to a known issue with the reconfigure scripts. To set your
+  default config project, run the following command:
+  `gcloud config set core/<<PROJECT ID>>`
+
```
cboneti, line 76: There is more: 
Pub/Sub Admin (roles/pubsub.admin)
Required when enable_reconfigure=true.

cboneti (commented):
I left a few comments.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Updated

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Updated

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: I clarified the python dependency in the warning of this module. I also added `enable_reconfigure` as one of the triggering options for the requirement. Rather than rewrite them here, I added a link to the description and inform the user to learn more there.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
heyealex: Added

cboneti (approved):

PR #661, opened 2022-10-24
Add documentation on zone policies and egress risks.

heyealex:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
@@ -48,9 +48,9 @@ The Slurm on GCP partition modules provide the option to set policies regarding
 which zone the compute VM instances will be created in through the
 `zone_policy_allow` and `zone_policy_deny` variables.
 
-As an example, the following module will show preference for zones
+As an example, see the the following module: will show preference for zones
 `us-central1-a` and `us-central1-b` while explicitly not allowing any VM
-instances to be created in `us-central1-f`:
+instances to be created in `us-central1-f`
 
 ```yaml
 - id: partition-with-zone-policy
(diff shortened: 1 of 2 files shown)


community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
```diff
@@ -32,6 +32,42 @@ The following code snippet creates a partition module with:
...
+- id: partition-with-zone-policy
+  source: community/modules/compute/schedmd-slurm-gcp-v5-partition
+  settings:
+    zone_policy_allow:
+    - us-central1-a
+    - us-central1-b
+    zone_policy_deny: [us-central1-f]
+```
+
+> **_NOTE:_** The default policy for  a zone is to allow them, therefore
```
tpdownes, line 65: There is some extra whitespace here. In addition to that, I'd probably use a more "active" voice and clarify whether `us-central-c` would be allowed given the above configuration.

 Something like

>`zone_policy_allow` does not guarantee the use of specified zones because zones are allowed by default. Configure `zone_policy_deny` to ensure that zones outside the allowed list are not used.

It's not leaping out at me what the value of allow is.

tpdownes (changes requested):

community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
heyealex: Good suggestion, I've added it and made the example explanation more clear.

community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
```diff
@@ -32,6 +32,51 @@ The following code snippet creates a partition module with:
...
+>
+> To avoid egress charges, ensure your compute nodes are created in the same
+> zone as the other resources that share data with them by setting
+> `zone_policy_deny` to all other zones in the region.
+
+The Slurm on GCP partition modules provide the option to set policies regarding
+which zone the compute VM instances will be created in through the
+`zone_policy_allow` and `zone_policy_deny` variables.
+
+As an example, see the the following module: will show preference for zones
```
tpdownes, line 51: I would not use "preference" here but a more neutral word like allow.

community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
```diff
@@ -32,6 +32,51 @@ The following code snippet creates a partition module with:
...
+    zone_policy_deny: [us-central1-f]
+```
+
+In this module, the following is defined:
+
+* `us-central1-a` and `us-central1-b` zones have been explicitly allowed.
+* `us-central1-f` has been explicitly denied, therefore no nodes in this
+  partition will be created in that zone.
+* Since `us-central1-c` was not included in the zone policy, it will default to
+  "Allow", which means the partition has the same liklihood of created a node in
```
tpdownes, line 71: - misspelling of "liklihood"
- you use "region" when you mean "zone"
- you use "created" when you mean "creating"
- I would probably not say "2 listed under" and say instead "as zones explicitly listed..."

tpdownes (changes requested):
I think we're close but since this involves billing, we should get this right.

community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
tpdownes: Unless this actually causes those zone to be more highly ranked than 1-c.

heyealex: It was a mistake to include that part, everything after the : should have been removed in favor of the bulleted list below. It's been fixed.

community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
heyealex: Fixed

tpdownes (approved):

PR #662, opened 2022-10-25
Add zone option to slurm-gcp v5 partition

heyealex:
Dependent on #661 

Update the partition to take a "zone" variable and explicitly deny all other zones in that region to make it happen.

I decided to prioritize zone policies over zone because `zone` is a common deployment variable in our examples, and by prioritizing `zone`, it would be very easy to accidentally override intended zone policies. It's easy enough to go the other way, I welcome discussion on the topic.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
@@ -73,6 +73,7 @@ locals {
   uses_zone_policies = length(var.zone_policy_allow) + length(var.zone_policy_deny) > 0
   excluded_zones     = var.zone == null ? [] : [for z in data.google_compute_zones.available.names : z if z != var.zone]
   zone_policy_deny   = local.uses_zone_policies ? var.zone_policy_deny : local.excluded_zones
+  zone_policy_allow  = local.uses_zone_policies || var.zone == null ? var.zone_policy_allow : [var.zone]
 }
 
 data "google_compute_zones" "available" {
@@ -95,7 +96,7 @@ module "slurm_partition" {
   partition_name          = var.partition_name
   project_id              = var.project_id
   region                  = var.region
-  zone_policy_allow       = var.zone_policy_allow
+  zone_policy_allow       = local.zone_policy_allow
   zone_policy_deny        = local.zone_policy_deny
   subnetwork              = var.subnetwork_self_link == null ? "" : var.subnetwork_self_link
   partition_conf          = local.partition_conf
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
@@ -69,6 +69,15 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  uses_zone_policies = length(var.zone_policy_allow) + length(var.zone_policy_deny) > 0
+  excluded_zones     = var.zone == null ? [] : [for z in data.google_compute_zones.available.names : z if z != var.zone]
+  zone_policy_deny   = local.uses_zone_policies ? var.zone_policy_deny : local.excluded_zones
+}
+
+data "google_compute_zones" "available" {
+  project = var.project_id
+  region  = var.region
 }
 
 data "google_compute_default_service_account" "default" {
@@ -87,7 +96,7 @@ module "slurm_partition" {
   project_id              = var.project_id
   region                  = var.region
   zone_policy_allow       = var.zone_policy_allow
-  zone_policy_deny        = var.zone_policy_deny
+  zone_policy_deny        = local.zone_policy_deny
   subnetwork              = var.subnetwork_self_link == null ? "" : var.subnetwork_self_link
   partition_conf          = local.partition_conf
 }
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/README.md
@@ -73,6 +73,25 @@ In this module, the following is defined:
 > because zones are allowed by default. Configure `zone_policy_deny` to ensure
 > that zones outside the allowed list are not used.
 
+#### Setting a Single Zone
+
+The `zone` variable is another option for setting the zone policy. If `zone` is
+set and neither `zone_policy_deny` nor `zone_policy_allow` are set, the
+policy will be configured as follows:
+
+* All _currently active_ zones in the region **at deploy time** will be set in the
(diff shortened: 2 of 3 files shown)


community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -69,6 +69,15 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  uses_zone_policies = length(var.zone_policy_allow) + length(var.zone_policy_deny) > 0
+  excluded_zones     = var.zone == null ? [] : [for z in data.google_compute_zones.available.names : z if z != var.zone]
+  zone_policy_deny   = local.uses_zone_policies ? var.zone_policy_deny : local.excluded_zones
+}
+
+data "google_compute_zones" "available" {
```
cboneti, line 78: What happens if a new zone become available while this cluster is running? will bulk api start scheduling work there?

heyealex: That's a good question. Since we are creating a single list based on what's known at the time, if a new zone is created in the region, my assumption is that it would be set to allow by default (as all zones are) and could lead to a risk of egress fees.

I don't know of a good way to ensure all created zones are denied. @tpdownes, do you have any thoughts on this?

This would only happen if the zone was entirely unknown at deploy time, as all zones, up or down, are captured by this terraform data structure. I'll add some more documentation to clarify this risk. I still believe having this option is worth it, since this risk is rare and it greatly reduces the overhead for a user to specify a single zone.

cboneti: The question really is, what does zone_policy_allow do? can you use that instead of zone_policy_deny?

heyealex: From my research and experience, zone_policy_allow doesn't do much more than make the default explicit. 

In this case, it wouldn't change anything functionally, but it would make sense to explicitly define it. I'll follow up with that after the dependent PR is merged. 

The only clear way I've found of guaranteeing a zone is to use [instances.bulkInsert] rather than [regionInstances.bulkInsert] when `zone` is specified. This would require a change to the logic of the resume script.

[instances.bulkInsert]: https://cloud.google.com/compute/docs/reference/rest/v1/instances/bulkInsert
[regionInstances.bulkInsert]: https://cloud.google.com/compute/docs/reference/rest/v1/regionInstances/bulkInsert

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -69,6 +69,16 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  uses_zone_policies = length(var.zone_policy_allow) + length(var.zone_policy_deny) > 0
+  excluded_zones     = var.zone == null ? [] : [for z in data.google_compute_zones.available.names : z if z != var.zone]
```
tpdownes, line 74: I think I might name this something more like `other_zones_in_region`.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -69,6 +69,16 @@ locals {
...
+  zone_policy_deny   = local.uses_zone_policies ? var.zone_policy_deny : local.excluded_zones
+  zone_policy_allow  = local.uses_zone_policies || var.zone == null ? var.zone_policy_allow : [var.zone]
+}
+
+data "google_compute_zones" "available" {
+  project = var.project_id
+  region  = var.region
```
tpdownes, line 81: Strictly speaking you are making the assumption that `var.zone` is in `var.region` which we do not know to be the case. I think you are making the most reasonable choice here but we should be aware that this solution is fragile because I believe it would have the impact of rejecting all zones if both `var.region` and `var.zone` are set without setting explicit allow/deny lists.

This would be a good use case for Terraform 1.2 pre/post conditions which could allow a test of whether `var.zone` was found in the zones data source.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -69,6 +69,16 @@ locals {
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
   tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
   slurm_cluster_name = var.slurm_cluster_name != null ? var.slurm_cluster_name : local.tmp_cluster_name
+
+  uses_zone_policies = length(var.zone_policy_allow) + length(var.zone_policy_deny) > 0
+  excluded_zones     = var.zone == null ? [] : [for z in data.google_compute_zones.available.names : z if z != var.zone]
+  zone_policy_deny   = local.uses_zone_policies ? var.zone_policy_deny : local.excluded_zones
+  zone_policy_allow  = local.uses_zone_policies || var.zone == null ? var.zone_policy_allow : [var.zone]
```
tpdownes, line 76: I'll admit to not loving that the pattern for these 2 lines differs since you're really sneaking in the default value for var.zone_policy_allow when var.zone is null. Can you just add a comment at line 72 which explains the logic? e.g.

```
# use var.zone_policy_allow and var.zone_policy_deny if either or both are set
# if neither are set and var.zone is unset: choose empty list for deny and allow
# if neither are set and var.zone is unset: choose empty list for deny and [var.zone] for allow
# use var.zone_policy_allow in all cases except when var.zone is set
```


FYI: this is a good case where you would want `nullable = false` for these 2 variables because then you don't have to worry about them being `null`. Right now, you could pass an explicit null and it would break. I do not think you should develop against that scenario, though.

tpdownes (changes requested):
Minor changes just above nit

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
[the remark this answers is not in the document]
tpdownes: These 3 comments are all right around nit level. I am going to collect them in a task and approve PR.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
tpdownes: These 3 comments are all right around nit level. I am going to collect them in a task and approve PR.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
tpdownes: These 3 comments are all right around nit level. I am going to collect them in a task and approve PR.

tpdownes (approved):
LGTM

PR #663, opened 2022-10-26
Automatic VM instance mounting of network storage

cboneti:
### Submission Checklist

  install`)


modules/compute/vm-instance/startup_from_network_storage.tf
```diff
@@ -0,0 +1,64 @@
...
+    type        = "shell"
+  }]
+
+  full_runner_list = concat(
+    local.storage_client_install_runners,
+    local.mount_runners,
+    local.startup_script_runner
+  )
+
+  startup_from_network_storage = length(local.full_runner_list) > 0 ? module.netstorage_startup_script.startup_script : null
```
nick-stroud, line 53: nit: currently length will never be <1 since `startup_scirpt_runner` is static.

modules/compute/vm-instance/variables.tf
```diff
@@ -203,6 +205,11 @@ variable "network_interfaces" {
   }
 }
 
+variable "region" {
```
nick-stroud, line 208: nit: We could compute this from the zone by hacking off the end. It might be more robust to pass it so I will let you make the call.

nick-stroud (dismissed):

nick-stroud:
Resolve failing tests.

modules/compute/vm-instance/variables.tf
cboneti: I considered it originally and, while this is a breaking change, I think most blueprints already receive both region and zone. I would prefer not to create a precedent having modules compute zone automatically.

modules/compute/vm-instance/startup_from_network_storage.tf
cboneti: good point. Fixed.

nick-stroud (approved):

PR #664, opened 2022-10-26
Update nfs-server network_storage to use shell mount script

nick-stroud:
This change swaps out the ansible runner for shell runners in the `nfs-server` `network_storage` variable. This removes the requirement for ansible to be installed when mounting with the `use` field. It also makes it possible to mount both a `nfs-server` alongside a DDN lustre network file system. 

Note: `nfs-server` differs from other file systems in that it takes multiple local mount points. A mount script is generated for each mount point. Since the `mount_runner` output only provides a single runner we must still use a runner that can mount to all locations. The simple fix here is to continue to use the ansible runner for this output. We may want to consider rewriting the mount script to take multiple local mount points.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/nfs-server/outputs.tf
+++ b/community/modules/file-system/nfs-server/outputs.tf
@@ -16,14 +16,14 @@
 # render the content for each folder
 output "network_storage" {
   description = "export of all desired folder directories"
-  value = [for mount in var.local_mounts : {
+  value = [for mount, mount_runner in local.mount_runners : {
     remote_mount          = "/exports${mount}"
     local_mount           = mount
-    fs_type               = "nfs"
-    mount_options         = "defaults,hard,intr"
-    server_ip             = google_compute_instance.compute_instance.network_interface[0].network_ip
+    fs_type               = local.fs_type
+    mount_options         = local.mount_options
+    server_ip             = local.server_ip
     client_install_runner = local.install_nfs_client_runner
-    mount_runner          = local.mount_runner
+    mount_runner          = mount_runner
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


community/modules/file-system/nfs-server/outputs.tf
```diff
@@ -16,14 +16,14 @@
...
 output "network_storage" {
   description = "export of all desired folder directories"
-  value = [for mount in var.local_mounts : {
+  value = [for mount, mount_runner in local.mount_runners : {
     remote_mount          = "/exports${mount}"
     local_mount           = mount
-    fs_type               = "nfs"
-    mount_options         = "defaults,hard,intr"
-    server_ip             = google_compute_instance.compute_instance.network_interface[0].network_ip
+    fs_type               = local.fs_type
```
cboneti, line 22: nit: why the indirection? Do you expect it to change?

cboneti (approved):
A nit / dummy question, but ok to merge if the right tests passed.

community/modules/file-system/nfs-server/outputs.tf
nick-stroud: It is also used [here](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/5ac4aab7417ed3cbadaf63d9ce0df82773996605/community/modules/file-system/nfs-server/main.tf#L35). So instead of using the same string literal twice I created a variable (kind of like a constant). Just my DRY coding habits kicking in. 

You could argue that the value from having the variable defined once is negated by diminished readability. But since you opened it as nit I will Ack and press forward.

PR #680, opened 2022-11-01
Improve startup-script validation

tpdownes:
Handle case that user has supplied an explicit null value to either the startup script content (string) or source (filepath). Modules will accept null values by default, while resources will treat these as values that have not been supplied by the user.

### Submission Checklist

  install`)


heyealex (approved):
LGTM

FYI - I tried a few combinations manually (supplied by the user), and having `content=...` and `source=null` failed, which I think is expected.

I'm assuming that in the use-case you were describing this morning, the difference is that the value would supplied by a resource directly, and therefore is seen as not being supplied at all.

heyealex (approved):

PR #681, opened 2022-11-02
Add PBS Pro initial modules

tpdownes:
### Submission Checklist

  install`)


community/modules/scripts/pbspro-install/variables.tf
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "rpm_url" {
+  description = "Path to PBS Pro RPM file for select PBS host type (server, client, execution)"
+  type        = string
+}
+
+variable "pbs_exec" {
```
heyealex, line 22: nit: exec sounds like its pointing to the executable itself. Can this be updated to "pbs_exec_dir", "pbs_install_prefix" or similar?

community/modules/scripts/pbspro-install/variables.tf
```diff
@@ -0,0 +1,65 @@
...
+variable "pbs_home" {
+  description = "PBS working directory"
+  type        = string
+  default     = "/var/spool/pbs"
+}
+
+variable "pbs_license_server" {
+  description = "IP address or DNS name of PBS license server"
+  type        = string
+  default     = "CHANGE_THIS_TO_PBS_PRO_LICENSE_SERVER_HOSTNAME"
```
heyealex, line 43: Why not just make this a mandatory variable? I'm assuming this default doesn't provide any value.

community/modules/scripts/pbspro-install/variables.tf
```diff
@@ -0,0 +1,65 @@
...
+variable "pbs_license_server_port" {
+  description = "Networking port of PBS license server"
+  type        = number
+  default     = 6200
+}
+
+variable "pbs_server" {
+  description = "IP address or DNS name of PBS server host"
+  type        = string
+  default     = "CHANGE_THIS_TO_PBS_PRO_SERVER_HOSTNAME"
```
heyealex, line 55: Same here.

community/modules/scripts/pbspro-preinstall/README.md
```diff
@@ -0,0 +1,141 @@
+## Description
+
+This module uploads PBS Pro RPM packages and, optionally, a license file to
+Google Cloud Storage. This enables machines in a PBS cluster to rapidly
+download and install PBS at boot or during the building of an image.
+
+### Example
+
+The following code snippet uses this module to
```
heyealex, line 9: Missing the end of the sentence.

community/modules/scripts/pbspro-preinstall/main.tf
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ */
+
+locals {
+  location = coalesce(var.location, var.region)
+
+  bucket_name = "pbspro-packages"
+
+  retention_policy = {
+    (local.bucket_name) = var.retention_policy
```
heyealex, line 23: I've not used this syntax in TF before, what is the purpose of the () around local.bucket_name?

community/modules/scripts/pbspro-install/variables.tf
tpdownes: These values come directly from the PBS parameter names (a mix of environment variables, settings in `/etc/pbs.conf` and settable via qmgr). See section 3.5.2.2:

[bigbook]: https://2021.help.altair.com/2021.1/PBSProfessional/PBS2021.1.pdf

community/modules/scripts/pbspro-install/variables.tf
tpdownes: There is a bit of a challenge here in that this input is required only for the use case of installing the server. So, it can either be required for use cases where it is not actually required or it can be never strictly required. Let me know what you think.

community/modules/scripts/pbspro-install/variables.tf
tpdownes: This has the same challenge as `pbs_license_info` (not required for servers, but required for clients and execution hosts). It is also literally the default value PBS gives this parameter. See [section 3.5.2.2][bigbook].

[bigbook]: https://2021.help.altair.com/2021.1/PBSProfessional/PBS2021.1.pdf

community/modules/scripts/pbspro-preinstall/main.tf
tpdownes: Because quotation marks are optional (and generally not preferred) for fields/keys in objects/maps, it is the only way to guide Terraform into realizing that you want the resolved value for that variable.

tpdownes: https://developer.hashicorp.com/terraform/language/expressions/types#maps-objects

community/modules/scripts/pbspro-preinstall/README.md
tpdownes: Addressed on HEAD of PR. I think the text went missing during some branch re-arrangement.

community/modules/scripts/pbspro-install/variables.tf
heyealex: That makes sense. I'd suggest adding that to the description, because it's not clear on first glance that this is only sometimes mandatory.

"IP address or DNS nam of PBS license server, mandatory when installing the pbs server."

community/modules/scripts/pbspro-install/variables.tf
heyealex: Same as above then, I'd suggest adding that to the description: 
"IP address or DNS name of PBS server host, required for client and execution hosts."

heyealex (approved):
Approved, but with a couple suggestions. Up to you if you think they make sense before merging.

community/modules/scripts/pbspro-install/variables.tf
tpdownes: I will address this in a shortly forthcoming PR.

community/modules/scripts/pbspro-install/variables.tf
tpdownes: I will address this in a shortly forthcoming PR.

PR #691, opened 2022-11-03
Update variables in node_group for consistency

heyealex:
### Description
Update to some node group variables to provide consistency between this module and others, as well as Slurm on GCP.

### Submission Checklist

  install`)


community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -73,37 +73,35 @@ variable "metadata" {
...
-    EOD
-  default     = null
+  validation {
+    condition = length(var.instance_image) == 0 || (
+    can(var.instance_image["family"]) || can(var.instance_image["name"])) == can(var.instance_image["project"])
+    error_message = "The project is required if family or name are provided in var.instance_image."
+  }
+  validation {
+    condition     = length(var.instance_image) == 0 || can(var.instance_image["family"]) != can(var.instance_image["name"])
+    error_message = "Only one of family and name can be provided in var.instance_image."
```
nick-stroud, line 103: The problem here is that if I pass `{ dog: good }` into here, validation passes. 

Maybe just single validation:
```
length(var.instance_image) == 0 ||
(can(family) && can(project)) ||
(can(name)&& can(project))
```

heyealex: It was caught for me:
```
$ terraform -chdir=im-test/primary apply
╷
│ Error: Invalid value for variable
│ 
│   on main.tf line 96, in module "node_group4":
│   96:   instance_image = {
│   97:     dog = "good"
│   98:   }
│ 
│ Only one of family and name can be provided in var.instance_image.
│ 
│ This was checked by the validation rule at
│ modules/schedmd-slurm-gcp-v5-node-group/variables.tf:101,3-13.
```

The error message could stand to be updated though. If both name and family are not set the second validation will fail.

heyealex: Updated the error messages in the latest commit.

nick-stroud: You are correct. I did not actually run the code. Looks good.

nick-stroud (approved):

PR #693, opened 2022-11-03
Add module for PBS Pro execution hosts

tpdownes:
### Submission Checklist

  install`)


community/modules/compute/pbspro-execution/main.tf
```diff
@@ -0,0 +1,103 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  resource_prefix = var.name_prefix != null ? var.name_prefix : "${var.deployment_name}-exec"
+  # gather local (inside VM) mountpoints with trailing slashes
+  mountpoints = [for ns in var.network_storage : endswith(ns.local_mount, "/") ? ns.local_mount : "${ns.local_mount}/"]
```
nick-stroud, line 20: nit: It is not clear where the trailing slash requirement is coming from. Since you are already adding a comment it would be great to leave a breadcrumb.

community/modules/compute/pbspro-execution/scripts/pbs_usecp.yml
```diff
@@ -0,0 +1,40 @@
...
+    mountpoints: ""
+  tasks:
+  - name: Allow PBS MoM to transfer files on shared filesystems using cp
+    ansible.builtin.lineinfile:
+      path: "{{ pbs_home }}/mom_priv/config"
+      line: $usecp *:{{ item }} {{ item }}
+    loop: "{{ mountpoints.split(',') }}"
+    notify:
+    - Restart PBS
+  - name: Ensure PBS is running
```
nick-stroud, line 32: nit: order of operations are not clear to me, but probably just because I am ansible noob. My understanding from https://docs.ansible.com/ansible/latest/user_guide/playbooks_handlers.html#controlling-when-handlers-run is that `Restart PBS` will not run until after `Ensure PBS is running`. Is that the desired behavior?

community/modules/compute/pbspro-execution/variables.tf
```diff
@@ -0,0 +1,314 @@
...
+variable "network_interfaces" {
+  description = <<-EOT
+    A list of network interfaces. The options match that of the terraform
+    network_interface block of google_compute_instance. For descriptions of the
+    subfields or more information see the documentation:
+    https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#nested_network_interface
+
+    **_NOTE:_** If `network_interfaces` are set, `network_self_link` and
+    `subnetwork_self_link` will be ignored, even if they are provided through
+    the `use` field. `bandwidth_tier` and `disable_public_ips` also do not apply
```
nick-stroud, line 184: nit: `disable_public_ips` should be `enable`

modules/README.md
```diff
@@ -40,12 +40,15 @@ Modules that are still in development and less stable are labeled with the
 * **[htcondor-execute-point]** ![community-badge] ![experimental-badge] :
   Manages a group of execute points for use in an [HTCondor
   pool][htcondor-configure].
+* **[pbspro-execution]** ![community-badge] ![experimental-badge] :
```
nick-stroud, line 43: nit: Probably a little late in the game to bring this up but almost every other module (maybe all) are a noun. `pbspro-execution` sounds like a verb to me. Would `pbspro-executor` be a better fit, or `pbspro-execution-host(s)`.

nick-stroud (approved):

community/modules/compute/pbspro-execution/scripts/pbs_usecp.yml
tpdownes: By default, Ansible handlers run at the end of a play if-and-only-if they were notified. In this case, the notification will be generated any time the list of mount points changes (increase, decrease, rename).

I am not quite the Ansible expert and I'm going by memory, but I think you might need to pair "ensure it's running" with "restart it" because SystemD may complain about restarting a service that is not actually started. The typical SystemD implementation of restart is "stop then start" and stop will fail if not active.

tpdownes: The apache example in your link demonstrates this pattern.

modules/README.md
tpdownes: The PBS nomenclature is:

- Server hosts
- Client / Submit hosts (sometimes I see Submission)
- Execution hosts

I prefer the brevity to adding "host" to everything but we can consider changes while experimental.

community/modules/compute/pbspro-execution/scripts/pbs_usecp.yml
tpdownes: "ACK"

community/modules/compute/pbspro-execution/main.tf
tpdownes: Resolved on HEAD of PR. Your comment also alerted me to another instance of `endswith` which was introduced in Terraform 1.3 (I'm surprised it's so new). I've replaced that with a Terraform 1.0-friendly solution.

community/modules/compute/pbspro-execution/variables.tf
tpdownes: Resolved on HEAD of PR

nick-stroud (approved):

PR #695, opened 2022-11-03
Enable tests on ./cmd and enforce lower (40%) coverage

kkr16:
I've started writing unit tests for `./cmd` and I'm lowering the threshold to 40% prevent `enforce_coverage.pl` from failing due to insufficient coverage. I'll increase it as coverage of `./cmd` improves.

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/hpc-toolkit-pr-validation.yaml
+++ b/tools/cloud-build/hpc-toolkit-pr-validation.yaml
@@ -40,6 +40,7 @@ steps:
 - id: make-tests
   waitFor:
   - pre-commits-setup
+  - git-fetch-unshallow
   name: >-
     us-central1-docker.pkg.dev/$PROJECT_ID/hpc-toolkit-repo/hpc-toolkit-builder
   entrypoint: /bin/bash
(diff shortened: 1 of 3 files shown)


tools/cloud-build/hpc-toolkit-pr-validation.yaml
```diff
@@ -15,6 +15,9 @@
 ---
 
 steps:
+- id: git-fetch-unshallow
```
nick-stroud, line 18: I think we need to add this step to the `waitFor` block of `make-tests` on line 41. Right now it is not failing because `pre-commit-setup` takes longer than `git-fetch-unshallow` but it would be correct to have the explicit dependency. In this case you would have two waitFor entries ([example](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/f4ed7c1ca864c77ea8cc21bd9286b0c48d28db95/tools/cloud-build/daily-tests/integration-group-2.yaml#L49-L51)).

kkr16: Good catch, thanks!

nick-stroud (approved):

PR #701, opened 2022-11-07
make overwrite error exit with rc=1

kkr16:
Fixing a bug where an overwrite error causes `ghpc` to exit with an error but with with rc=0 instead of rc=1

```
$ ./ghpc create examples/hpc-cluster-small.yaml --vars project_id=project-id 

Failed to overwrite existing deployment.

Use the -w command line argument to enable overwrite.
If overwrite is already enabled then this may be because you are attempting to remove a deployment group, which is not supported.
original error: The directory already exists: hpc-small
$ echo $?
0
```

After fix:
```
$ ./ghpc create examples/hpc-cluster-small.yaml --vars project_id=project-id 

Failed to overwrite existing deployment.

Use the -w command line argument to enable overwrite.
If overwrite is already enabled then this may be because you are attempting to remove a deployment group, which is not supported.
original error: The directory already exists: hpc-small
$ echo $?
1
```

### Submission Checklist

  install`)

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -23,6 +23,7 @@ import (
 	"hpc-toolkit/pkg/config"
 	"hpc-toolkit/pkg/modulewriter"
 	"log"
+	"os"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


heyealex (approved):
Thanks Karim, and we appreciate your patience during SC!

PR #707, opened 2022-11-08
Disable OS Login in HTCondor example

tpdownes:
Although a UID mapping problem was resolved in htcondor/htcondor#780, some issues are still observed when running Docker jobs. This change disables OS Login in the HTCondor example until it is further diagnosed.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/htcondor-pool.yaml
+++ b/community/examples/htcondor-pool.yaml
@@ -20,6 +20,7 @@ vars:
   deployment_name: htcondor-001
   region: us-central1
   zone: us-central1-c
+  enable_oslogin: "DISABLE"
 
 # Documentation for each of the modules used below can be found at
 # https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md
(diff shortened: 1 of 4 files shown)


community/examples/htcondor-pool.yaml
```diff
@@ -38,11 +39,6 @@ deployment_groups:
   - id: htcondor_install
     source: community/modules/scripts/htcondor-install
 
-  - id: htcondor_services
```
heyealex, line 41: Why was this module removed?

tpdownes: It was added prior to the development of `test_enabled_apis` validation and it requires the user running Terraform to have the ability to enable services. There is a known issue that Terraform cannot guarantee that APIs are globally enabled after the API call to enable them is complete. So, the functionality isn't perfect and we have new functionality which has better guarantees. Should have noted that in the PR description but I think we are better without it.

heyealex: SG - That's what I figured, but wanted to verify it was intentional.

heyealex (approved):

PR #722, opened 2022-11-10
Update pbspro-preinstall

tpdownes:
Require the user to supply "devel" RPM; this matters for compiling OpenMPI using Spack as it needs to `-I` include within it.

This only provides the functionality of uploading the RPM to Cloud Storage. Future PRs will implement its use within a PBS Pro cluster.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scripts/pbspro-preinstall/variables.tf
+++ b/community/modules/scripts/pbspro-preinstall/variables.tf
@@ -29,6 +29,15 @@ variable "region" {
   type        = string
 }
 
+variable "devel_rpm" {
+  description = "Absolute path to PBS Pro Development RPM file"
+  type        = string
+  validation {
+    condition     = can(regex("^/", var.devel_rpm))
+    error_message = "Path to RPM must be an absolute path beginning with \"/\"."
+  }
+}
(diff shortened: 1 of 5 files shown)


community/modules/scripts/pbspro-preinstall/variables.tf
```diff
@@ -29,6 +29,15 @@ variable "region" {
   type        = string
 }
 
+variable "devel_rpm" {
```
heyealex, line 32: It looks like this is going to be a requirement going forward as written. Would it make sense to make this optional, in case the user is not using spack?

I'll defer to your knowledge on this; I'm familiar enough with PBS to know if that is a reasonable requirement.

heyealex (approved):

community/modules/scripts/pbspro-preinstall/variables.tf
tpdownes: Offline conversation: this RPM is included with standard tarball from Altair. It is reasonable to require uploading it to bucket. Separate question whether and where to require installing it.

PR #725, opened 2022-11-11
Update default spack version to v0.19.0

douglasjacobsen:
This merge updates the default spack version used from v0.18.0 to v0.19.0.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/spack-gromacs.yaml
+++ b/community/examples/spack-gromacs.yaml
@@ -49,8 +49,6 @@ deployment_groups:
     source: community/modules/scripts/spack-install
     settings:
       install_dir: /sw/spack
-      spack_url: https://github.com/spack/spack
-      spack_ref: v0.18.0
       log_file: /var/log/spack.log
       configs:
       - type: single-config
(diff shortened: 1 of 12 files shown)


community/examples/spack-gromacs.yaml
```diff
@@ -50,7 +50,7 @@ deployment_groups:
     settings:
       install_dir: /sw/spack
       spack_url: https://github.com/spack/spack
-      spack_ref: v0.18.0
+      spack_ref: v0.19.0
```
tpdownes, line 53: On the balance, I think we should probably have our examples not specify `spack_url` ever and specify `spack_ref` only when we know there is a good reason to.

The environment/package specifications ought to be enough to be Spack-version-forward-compatible right?

tpdownes (changes requested):
Let me know what you think.

community/examples/spack-gromacs.yaml
douglasjacobsen: Spack sometimes changes the formats of these files.

They recently (I think in v0.18.0) changed the format of the packages file. They give fairly useful errors however, and it's likely important that we catch these changes.

However, removing it from the examples would just let the default come in. So, that's probably fine, I can edit this to just delete them instead of changing them all.

douglasjacobsen:
@tpdownes Back to you

heyealex (approved):
LGTM - Thanks Doug!

PR #730, opened 2022-11-15
Add Slurm on GCP v5 High IO example

heyealex:
Also: Improved error handling of partition name, clearer validation error message.

A follow-up PR (#731) adds this example as a test

### Submission Checklist

* [x] Have you installed and run this change against pre-commit? (`pre-commit
  install`)
* [x] Are all tests passing? (`make tests`)
* [x] Have you written unit tests to cover this change?
* [x] Is unit test coverage still above 80%?
* [x] Have you updated all applicable documentation?
* [x] Have you followed the guidelines in our Contributing document?

PR #734, opened 2022-11-16
Default enable reconfig high io

heyealex:
### Description
With the update to the build in #733, enable_reconfigure will be able to
run without failing at a python dependency, so we are setting active
cluster configuration by default in the high io blueprint.

### Dependent PRs

* #730 
* #733 

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/slurm-gcp-v5-high-io.yaml
+++ b/community/examples/slurm-gcp-v5-high-io.yaml
@@ -26,7 +26,7 @@ vars:
   disable_public_ips: false
   # Set to true for active cluster reconfiguration. Note that setting this
   # option requires additional dependencies to be installed locally.
-  enable_reconfigure: false
+  enable_reconfigure: true
 
 # Documentation for each of the modules used below can be found at
 # https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md


community/examples/slurm-gcp-v5-high-io.yaml
```diff
@@ -26,7 +26,7 @@ vars:
   disable_public_ips: false
   # Set to true for active cluster reconfiguration. Note that setting this
   # option requires additional dependencies to be installed locally.
-  enable_reconfigure: false
+  enable_reconfigure: true
```
nick-stroud, line 29: It is not obvious to me why we want this on in this example. We do not specify it in other v5 examples and the default is false. Is there something special about this example? Do we just want test coverage of this option?

I see good documentation for the required dependencies for this option but am not sure how the errors actually manafest and so am wondering if this is something we want in a public example.

heyealex: The high IO example in general has been more "full featured", so I wanted to include this since it's one of the most impactful new features in v5. 

Any errors will be clear in the terraform output - they are needed for a "null resource" script that's run locally, and the error specifies clearly which package is needed.

There is also the benefit of having it included in our tests, though I'd rather find a better way of adding additional command line variables in tests then set it by default for everyone if that's the only reason.

nick-stroud (approved):

PR #735, opened 2022-11-17
Added newlines and group names after ghpc create

omartin2010:
### Submission Checklist

  install`)
* [ ] Have you written unit tests to cover this change? no
* [X] Is unit test coverage still above 80%? yes (except cmd with 44.5% coverage)
* [ ] Have you updated all applicable documentation? not applicable
* [ ] Have you followed the guidelines in our Contributing document? yes

### DIFF
--- a/pkg/modulewriter/modulewriter.go
+++ b/pkg/modulewriter/modulewriter.go
@@ -170,8 +170,8 @@ func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGrou
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


heyealex (changes requested):
Thanks for your contribution! This makes the output much cleaner and easier to read.

I added one comment/suggestion that I'd like addressed before merging; it's fairly minor and I could push a commit with the update from my end as well if you would prefer.

heyealex (approved):

PR #736, opened 2022-11-17
Set unique script names for all mount points in nfs-server

heyealex:
Solves an error when deploying with multiple mount points that share the same destination filenames (including the default value for local_mounts.

Tested against `tools/validate_configs/test_configs/hpc-centos-ss.yaml`

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/nfs-server/variables.tf
+++ b/community/modules/file-system/nfs-server/variables.tf
@@ -103,11 +103,16 @@ variable "scopes" {
 variable "local_mounts" {
   description = "Mountpoint for this NFS compute instance"
   type        = list(string)
+  default     = ["/tools", "/data"]
+
   validation {
     condition = alltrue([
       for m in var.local_mounts : substr(m, 0, 1) == "/"
     ])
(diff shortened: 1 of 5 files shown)


community/modules/file-system/nfs-server/variables.tf
```diff
@@ -103,11 +103,16 @@ variable "scopes" {
 variable "local_mounts" {
   description = "Mountpoint for this NFS compute instance"
   type        = list(string)
+  default     = ["/tools", "/data"]
```
nick-stroud, line 106: nit: maybe a separate PR (or not at all). But would it make more sense to have a single mount as the default. Maybe just '/data'. Seems like having multiple mount points is an interesting feature but maybe not the desired default use case. 🤷 .

nick-stroud (dismissed):

community/modules/file-system/nfs-server/variables.tf
heyealex: I agree. I do also want to make sure we have a test that covers the multi-local-mount case, so I updated one of the test configs as well.

PR #737, opened 2022-11-17
Update file system documentation: mount and link to primary network storage doc

heyealex:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/nfs-server/README.md
+++ b/community/modules/file-system/nfs-server/README.md
@@ -37,6 +37,37 @@ volume as an external file system.
 > you will need to manually delete the disk after destroying a deployment group
 > with `nfs-server`.
 
+## Mounting
+
+To mount the NFS Server you must first ensure that the NFS client has been
+installed the and then call the proper `mount` command.
+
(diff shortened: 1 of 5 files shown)


community/modules/file-system/nfs-server/README.md
```diff
@@ -34,6 +37,37 @@ volume as an external file system.
 > you will need to manually delete the disk after destroying a deployment group
 > with `nfs-server`.
 
+## Mounting
+
+To mount the NFS Server you must first ensure that the NFS client has been
+installed the and then call the proper `mount` command.
+
+When mounting to a Slurm resource, a `vm_instance`, or a `batch-job-template`,
```
nick-stroud, line 45: I would point to the compatibility matrix instead of listing out modules here. Now we also have HTCondor and PBS Pro., and it will just be easier to keep a single source of truth up to date.

nick-stroud: Same thing for filestore and pre-existing-network-storage.

heyealex: Done

nick-stroud (approved):

PR #739, opened 2022-11-18
Fix bug where controller ID can't be obtained when destroying after a partial build

heyealex:
Fixes index not found error when destroying after a partial build. With this fix, the controller ID will be an empty list rather than not found at index 0 if the controller had not been created yet.

Tested by deploying the high-io-slurm-gcp-bv5 example in a project without enough quota for filestore, killing the deployment while the high scale filestore was "creating" and then destroying. The deployment destroyed as expected with this fix.

### Submission Checklist

  install`)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/outputs.tf
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-output "controller_instance_id" {
+output "controller_instance_ids" {
   description = "The server-assigned unique identifier of the controller compute instance."
-  value       = module.slurm_controller_instance.slurm_controller_instance.instances_details[0].id
+  value       = module.slurm_controller_instance.slurm_controller_instance.instances_details[*].id
```
nick-stroud, line 19: I would suggest using `one` here instead of changing to a list. I think it would avoid the failure and pass null instead of breaking on a lookup. 

https://developer.hashicorp.com/terraform/language/functions/one

heyealex: Good suggestion, it's a lot cleaner this way. 
Fixed

cboneti (approved):

PR #748, opened 2022-11-29
Sourcereader wrapfs workaround

thiagosgobe:
### Submission Checklist

  install`)


pkg/sourcereader/embedded.go
```diff
@@ -15,95 +15,35 @@
...
 	if !IsEmbeddedPath(modPath) {
 		return modulereader.ModuleInfo{}, fmt.Errorf("Source is not valid: %s", modPath)
 	}
 
-	modDir, err := copyFSToTempDir(ModuleFS, modPath)
-	defer os.RemoveAll(modDir)
-	if err != nil {
-		err = fmt.Errorf("failed to copy embedded module at %s to tmp dir %s: %v",
-			modPath, modDir, err)
+	mod, diag := tfconfig.LoadModuleFromFilesystem(tfconfig.WrapFS(ModuleFS), modPath)
```
heyealex, line 39: This needs to be reorganized a bit. `tfconfig` shouldn't be used directly in the `sourcereader` package, but instead in the `modulereader` package.

`sourcereader` essentially handles a lot of the IO and access to the raw contents, `modulereader` is specific to a module kind (terraform, packer, etc) and is able to provide more detailed information like Inputs/Outputs/etc of the module being referenced.

This function should pass the `modPath` into [`reader.GetInfo()`](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/pkg/modulereader/tfreader.go#L32), which calls [`hcl_utils.getHCLInfo()`](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/pkg/modulereader/hcl_utils.go#L27). The `getHCLInfo()` function can now check if it's embedded or not. If it isn't, it will do what it does today. If it is, it will call `LoadModuleFromFilesystem` rather than `LoadModule`.

The flow will look like this:
GetModuleInfo() -> reader.GetInfo() -> getHCLInfo() -> LoadModule() for local or LoadModuleFromFilesystem() for embedded

pkg/sourcereader/embedded.go
```diff
@@ -246,13 +186,11 @@ func (r EmbeddedSourceReader) GetModule(modPath string, copyPath string) error {
 		return fmt.Errorf("Source is not valid: %s", modPath)
```
heyealex, line 186: For this function, you want to copy a directory from the embedded FS to disk. The path itself appears relative, so it would only be able to copy what is relative to this directory in this implementation. Instead, you'd want to do something similar to `copyFSToTempDir`, but instead of copying to a temp directory, copy to the `copyPath` directly.

In fact, I'd recommend keeping it as is and updating it in a follow-up PR. There is some additional safety builtin when copying to a temp directory first and using the same logic as the other sourcereader's (see the git.go for an example).

heyealex (changes requested):
A couple changes are needed. Let me know if you need any clarification on my suggestions.

heyealex:
The PR-main-Integration-Group-5 check failure will go away after a rebase. That test group was added recently.

pkg/sourcereader/embedded.go
```diff
@@ -94,16 +94,8 @@ func (r EmbeddedSourceReader) GetModuleInfo(modPath string, kind string) (module
 		return modulereader.ModuleInfo{}, fmt.Errorf("Source is not valid: %s", modPath)
 	}
 
-	modDir, err := copyFSToTempDir(ModuleFS, modPath)
```
heyealex, line 97: I believe ModuleFS and BaseFS (at the top) are no longer used and can be removed from this file. Correct me if I'm mistaken, I haven't checked everywhere.

pkg/sourcereader/sourcereader_test.go
```diff
@@ -41,6 +42,8 @@ const (
 // Setup GoCheck
 type MySuite struct{}
 
+//go:embed modules
+var moduleFS embed.FS
```
heyealex, line 46: For clarity, can you rename this to something like `testModuleFS`?

pkg/sourcereader/embedded_test.go
```diff
@@ -119,19 +120,20 @@ func (s *MySuite) TestCopyFSToTempDir(c *C) {
...
 
 func (s *MySuite) TestGetModuleInfo_Embedded(c *C) {
-	ModuleFS = getTestFS()
+	ModuleFS = moduleFS
+	modulereader.ModuleFS = moduleFS
 	reader := EmbeddedSourceReader{}
 
 	// Success
-	moduleInfo, err := reader.GetModuleInfo("modules/network/vpc", tfKindString)
+	moduleInfo, err := reader.GetModuleInfo("modules/test_data/network/vpc", tfKindString)
```
heyealex, line 128: Suggestion to make this path less deep, but still clarify that it's a test module:
`"modules/test_role/test_module"`
This removed one level of depth and wouldn't lead to any confusion about the content of the module itself.

pkg/sourcereader/embedded_test.go
```diff
@@ -119,19 +120,20 @@ func (s *MySuite) TestCopyFSToTempDir(c *C) {
...
 	c.Assert(err, IsNil)
 	c.Assert(moduleInfo.Inputs[0].Name, Equals, "test_variable")
 	c.Assert(moduleInfo.Outputs[0].Name, Equals, "test_output")
 
 	// Invalid: No embedded modules
 	badEmbeddedMod := "modules/does/not/exist"
 	moduleInfo, err = reader.GetModuleInfo(badEmbeddedMod, tfKindString)
-	expectedErr := "failed to copy embedded module at .*"
+	expectedErr := "failed to get info using tfconfig for terraform module at modules/does/not/exist: Source to module does not exist: modules/does/not/exist"
```
heyealex, line 136: To avoid hard coding the fake path, try using .* instead. The following should be sufficient:
`expectedErr := "failed to get info using tfconfig for terraform module at .*"`

It's enough to ensure the error string is the same, but decreases maintenance since we don't have to update the expected string whenever the test path changes.

pkg/sourcereader/modules/test_data/network/vpc/README.md
```diff
@@ -0,0 +1,50 @@
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
heyealex, line 1: You could exclude this test module from terraform docs at [this line](https://github.com/thiagosgobe/hpc-toolkit/blob/sourcereader-wrapfs-workaround/.pre-commit-config.yaml#L32) of the precommit config.

Something like the following should work (not fully tested!):
`exclude: \[.terraform|pkg]\/.*$`

This would exclude any path in pkg, therefore anything in the go code.

When you do that, you can also remove this file.

heyealex (changes requested):
Thanks for the update! I've requested a few minor updates, once those are in this should be good to go.

Also, to bypass the failing check, you'll need to rebase on upstream/develop:
```shell
git fetch upstream
git rebase -i upstream/develop
```

pkg/sourcereader/modules/test_data/network/vpc/README.md
thiagosgobe: It works like that:

`exclude: \.terraform|pkg\/.*$`

thiagosgobe:
Thank you for reviewing it and also for all guidance :)
Updates are done let me know if there is anything else.

heyealex (approved):
Thanks for making those changes! This is looks good.

Issue #746, opened 2022-11-23
Filesystems - Invalid options in Performance tier drop down menu when creating new Filestore instance


saltysoup:
When creating new filestore instance, options for STANDARD and PREMIUM exists in the drop down menu, which should be removed as they are no longer valid params.

**Terraform log output**

`
Error: Invalid value for variable

  on main.tf line 20, in module "htk-fe-filestore":
  20:   filestore_tier       = "STANDARD"
    ├────────────────
    │ var.filestore_tier is "STANDARD"

The preferred name for STANDARD tier is now BASIC_HDD
https://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier.

This was checked by the validation rule at
modules/filestore/variables.tf:74,3-13.

Error: Invalid value for variable

  on main.tf line 20, in module "htk-fe-filestore":
  20:   filestore_tier       = "STANDARD"
    ├────────────────
    │ var.filestore_tier is "STANDARD"

Allowed values for filestore_tier are
'BASIC_HDD','BASIC_SSD','HIGH_SCALE_SSD','ENTERPRISE'.
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier
https://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier.

This was checked by the validation rule at
modules/filestore/variables.tf:82,3-13.
`

cboneti:
Hi Injae, 

Just to confirm, are you using the front end, or deploying using a specific blueprint? 

Thanks!

saltysoup:
This is for the front end when deploying a new filesystem

cboneti:
Thanks, I will investigate.

cboneti:
[mattstreet-nag](https://github.com/mattstreet-nag) already had a PR for this. He is submitting a fix soon.

heyealex:
PR #749 has been merged into develop, which removes the invalid filestore options. 
We will keep this issue open until the change makes it into a release on the main branch, which should be in a couple of weeks. Thank you for reporting this!

heyealex:
Since the code in question was not part of a release yet and its fixed on develop, I'm closing this issue as completed.

PR #749, opened 2022-11-29
Fix for issue #746 - removed invalid options for filesystem (STANDARD…

mattstreet-nag:
… and PREMIUM are now deprecated) and added ENTERPRISE.

Added files to .gitignore to ensure safer commits.

### Submission Checklist

  install`)

### DIFF
--- a/community/front-end/ofe/website/ghpcfe/models.py
+++ b/community/front-end/ofe/website/ghpcfe/models.py
@@ -1230,11 +1230,10 @@ class GCPFilestoreFilesystem(Filesystem):
 
     FILESTORE_TIER = (
         ("u", "TIER_UNSPECIFIED"),
-        ("s", "STANDARD"),
-        ("p", "PREMIUM"),
         ("bh", "BASIC_HDD"),
         ("bs", "BASIC_SSD"),
         ("hs", "HIGH_SCALE_SSD"),
+        ("en", "ENTERPRISE"),
     )
     capacity = models.PositiveIntegerField(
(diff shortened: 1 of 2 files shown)


community/front-end/ofe/website/ghpcfe/models.py
```diff
@@ -1230,11 +1230,10 @@ class GCPFilestoreFilesystem(Filesystem):
 
     FILESTORE_TIER = (
         ("u", "TIER_UNSPECIFIED"),
```
heyealex, line 1232: What is the behavior when the tier is unspecified?

mattstreet-nag: @heyealex 

Terraform creation fails, as expected with the same validator in #746, and filesystem is not created.
<pre>

  on main.tf line 27, in module "fs1":
  27:   filestore_tier       = "TIER_UNSPECIFIED"
    ├────────────────
    │ var.filestore_tier is "TIER_UNSPECIFIED"
</pre>

Would agree this is a bit pointless in the Front End - was probably added for completeness.

heyealex: Could it be removed as well without breaking anything? I can see the need for a null value, but a user shouldn't be able to select it.

heyealex: I just removed it and ran a quick test, everything worked as expected.

heyealex (approved):

PR #760, opened 2022-12-06
Add a validator for unused modules in the "use" list

heyealex:
### Description
Adds a struct for tracking information about connections between modules. This is only used for modules in the "use" list for now, but will eventually be useful for tracking all module connections and for storing the types where used and any other useful data.

Using that data structure, a validator has been added to check if any modules in the "use" list are not actually being used.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -216,11 +216,39 @@ type Blueprint struct {
 	TerraformBackendDefaults TerraformBackend  `yaml:"terraform_backend_defaults"`
 }
 
+// ConnectionKind defines the kind of module connection, defined by the source
+// of the connection. Currently, only Use is supported.
+type ConnectionKind int
+
+const (
+	undefinedConnection ConnectionKind = iota
+	useConnection
+	// explicitConnection
+	// globalConnection
+)
+
 // ModConnection defines details about connections between modules. Currently,
 // only modules connected with "use" are tracked.
 type ModConnection struct {
 	toID   string
-	unused bool
+	fromID string
+	// Currently only supports useConnection
+	kind ConnectionKind
+	// List of variables shared from module `fromID` to module `toID`
+	sharedVariables []string
+}
+
+// Returns true if a connection does not functionally link the outputs and
+// inputs of the modules. This can happen when a module is connected with "use"
+// but none of the outputs of fromID match the inputs of toID.
+func (mc *ModConnection) isEmpty() (isEmpty bool) {
+	isEmpty = false
+	if mc.kind == useConnection {
+		if len(mc.sharedVariables) == 0 {
+			isEmpty = true
+		}
+	}
+	return
 }
 
 // DeploymentConfig is a container for the imported YAML data and supporting data for
@@ -232,7 +260,7 @@ type DeploymentConfig struct {
 	// Maps module ID to group index
 	ModuleToGroup     map[string]int
 	expanded          bool
-	moduleConnections map[string][]ModConnection
+	moduleConnections []ModConnection
 }
 
 // ExpandConfig expands the yaml config in place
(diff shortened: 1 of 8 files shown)


pkg/validators/validators.go
```diff
@@ -47,6 +49,25 @@ func handleClientError(e error) error {
...
+	foundUnused := false
+	for mod, unusedMods := range unusedModules {
+		foundUnused = true
+		for _, unusedMod := range unusedMods {
+			log.Printf(unusedModuleMsg, mod, unusedMod)
+		}
+	}
+
+	if foundUnused {
+		return fmt.Errorf("one or more used variables could not have their settings and outputs linked")
```
tpdownes, line 65: "used variables" should probably read "used modules"

heyealex: Fixed

pkg/config/config.go
```diff
@@ -213,15 +216,23 @@ type Blueprint struct {
...
 type DeploymentConfig struct {
 	Config Blueprint
 	// Indexed by Resource Group name and Module Source
 	ModulesInfo map[string]map[string]modulereader.ModuleInfo
 	// Maps module ID to group index
-	ModuleToGroup map[string]int
-	expanded      bool
+	ModuleToGroup     map[string]int
+	expanded          bool
+	moduleConnections map[string][]ModConnection
```
nick-stroud, line 235: Normally storing graph like data is either done as:
1. nodes as objects, edges as pointers
2. list of edges
3. matrix (ignored in this case)

What it looks like we are doing here is kind of #1 but pointers being a data structure that contains the name of the next node. I am wondering if #2 might be a better representation here. In this case you could just have a list of connections which contain both the up and down stream id (as suggested elsewhere). 

I think just storing as a list would simplify the operations done in this CL. I am not sure if it would make other future operations harder.

Thoughts? This might just be personal preference bleeding through.

pkg/config/expand.go
```diff
@@ -167,7 +167,8 @@ func useModule(
 	modInputs map[string]string,
 	useOutputs []modulereader.VarInfo,
 	changedSettings map[string]bool,
-) {
+) bool {
```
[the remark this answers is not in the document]
heyealex: I wasn't aware of that feature, it's included (though the return type changed)

pkg/config/config.go
```diff
@@ -213,15 +216,23 @@ type Blueprint struct {
 	TerraformBackendDefaults TerraformBackend  `yaml:"terraform_backend_defaults"`
 }
 
+// ModConnection defines details about connections between modules. Currently,
+// only modules connected with "use" are tracked.
+type ModConnection struct {
+	toID   string
```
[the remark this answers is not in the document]
heyealex: Done

pkg/config/config.go
```diff
@@ -213,15 +216,23 @@ type Blueprint struct {
...
+	unused bool
```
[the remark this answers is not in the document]
heyealex: I refactored a bit to include "kind" and "sharedVariables" with a function to detect if the connection is empty or not based on those.

pkg/config/expand.go
```diff
@@ -193,23 +194,27 @@ func useModule(
...
 }
 
 // applyUseModules applies variables from modules listed in the "use" field
 // when/if applicable
 func (dc *DeploymentConfig) applyUseModules() error {
 	for iGrp := range dc.Config.DeploymentGroups {
 		group := &dc.Config.DeploymentGroups[iGrp]
+		grpModsInfo := dc.ModulesInfo[group.Name]
 		for iMod := range group.Modules {
 			mod := &group.Modules[iMod]
```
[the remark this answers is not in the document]
heyealex: Done

pkg/config/expand.go
```diff
@@ -219,7 +224,18 @@ func (dc *DeploymentConfig) applyUseModules() error {
 					return fmt.Errorf("could not find module %s used by %s in group %s",
 						useModID, mod.ID, group.Name)
 				}
-				useModule(mod, useMod, modInputs, useInfo.Outputs, changedSettings)
+				isUsed := useModule(mod, useMod, modInputs, useInfo.Outputs, changedSettings)
+				connection := ModConnection{
+					unused: !isUsed,
+					toID:   useModID,
+				}
+				if _, exists := dc.moduleConnections[mod.ID]; exists {
```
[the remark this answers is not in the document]
heyealex: Switching to a connection list simplified these lines away

pkg/config/config.go
heyealex: Looking at it again with fresh eyes, you're absolutely right, I made the change and it simplified a lot of the code.

It might be helpful to access connections for all fromID/toID modules quickly at some point, but that can be added later.

pkg/config/config.go
```diff
@@ -238,6 +277,22 @@ func (dc *DeploymentConfig) ExpandConfig() error {
 	return nil
 }
 
+// listUnusedModules provides a mapping of modules to modules that are in the
+// "use" field, but not actually used.
+func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
```
nick-stroud, line 282: nit: personally I would prefer if this returned type `[]ModConnection` and then any processing needed happens downstream. By creating this map of lists we are introducing a new data structure that is kind of specific to the validation function.

nick-stroud (approved):

pkg/config/config.go
heyealex: Not a bad idea, but it creates an import cycle if ModConnection needs to be referenced in validators (config imports validators, validators imports config).

I'm going to keep it as is for now, and consider that a follow-up improvement.

PR #765, opened 2022-12-07
Add no_comma_params option in "cloud_parameters"

heyealex:
### Submission Checklist

  install`)


tpdownes:
For the record:

- this exposes an existing option in slurm-gcp module that defaults to false (in slurm-gcp)
- the default value for the new object sets this to false (no behavioral change)
- users who set `var.cloud_parameters` must add this new field to their objects. it will be discovered at time of `terraform plan/apply`

SGTM, but noting "half" breaking change.

tpdownes (approved):

PR #785, opened 2022-12-14
Add partition level startup script variables

heyealex:
Fixes a bug with the slurm_cluster_name in the partition module that prevented the partition setup from recognizing custom startup scripts in the metadata.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
@@ -15,13 +15,17 @@
  */
 
 locals {
+  ghpc_startup_script = var.startup_script == "" ? [] : [{
+    filename = "ghpc_startup.sh"
+    content  = var.startup_script
+  }]
 
   # Default to value in partition_conf if both set "Default"
   partition_conf = merge(var.is_default == true ? { "Default" : "YES" } : {}, var.partition_conf)
 
   # Since deployment name may be used to create a cluster name, we remove any invalid character from the beginning
   # Also, slurm imposed a lot of restrictions to this name, so we format it to an acceptable string
-  tmp_cluster_name   = substr(replace(lower(var.deployment_name), "/^[^a-z]*|[^a-z0-9]/", ""), 0, 8)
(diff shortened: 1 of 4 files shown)


tools/validate_configs/test_configs/hpc-cluster-small-slurm-v5.yaml
```diff
@@ -68,6 +68,10 @@ deployment_groups:
     - compute_node_group
     settings:
       partition_name: compute
+      startup_script: |
```
nick-stroud, line 71: Could we do something similar in the controler? We previously had a regression there.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
```diff
@@ -15,13 +15,17 @@
  */
 
 locals {
+  ghpc_startup_script = var.startup_script == "" ? [] : [{
```
nick-stroud, line 18: This type of logic previously caused a regression in the controller and login node (see #777).

When using startup script module, the script is not known at apply time so this variable is also not known at apply time. Parts of slurm depend on the filename being known at apply time.

tools/validate_configs/test_configs/hpc-cluster-small-slurm-v5.yaml
nick-stroud: Also it would be better to test with the actual startup script module as the startup script module output is not known at apply time which can reveal other issues.

nick-stroud: As you pointed out offline, we might want to include this in one of our actual integration tests to make sure terraform plan is being called.

heyealex: I updated the test_config and ran the test manually. 
As we discussed offline, it will be best to add the test_config to the integration tests in a separate PR. The v5 tests are all based on examples currently, and adding dummy startup-script modules to those tests would be fragile. Adding a new test based on this test_config will be a cleaner way of incorporating startup-scripts.

community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf
heyealex: Updated

PR #790, opened 2022-12-15
Add slurm gcp v5 integration test with startup scripts

heyealex:
### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
+++ b/tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
@@ -21,7 +21,7 @@ deployment_name: "ss-v5-{{ build }}"
 slurm_cluster_name: "ssv5{{ build[0:6] }}"
 zone: us-central1-c
 workspace: /workspace
-blueprint_yaml: "{{ workspace }}/tools/validate_configs/test_configs/hpc-cluster-small-slurm-v5.yaml"
+blueprint_yaml: "{{ workspace }}/tools/validate_configs/test_configs/slurm-gcp-v5-startup-scripts.yaml"
 network: "{{ deployment_name }}-net"
 max_nodes: 5
 # Note: Pattern matching in gcloud only supports 1 wildcard, centv5*-login-* won't work.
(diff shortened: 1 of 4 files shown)


tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
```diff
@@ -0,0 +1,37 @@
...
+login_node: "{{ slurm_cluster_name }}-login-*"
+controller_node: "{{ slurm_cluster_name }}-controller"
+post_deploy_tests:
+- test-partitions.yml
+custom_vars:
+  partitions:
+  - compute
+  - debug
+  mounts:
+  - /home
```
nick-stroud, line 37: Now that this is going to be an integration test, can we remove the filestore from the config to speed up deployment. I don't see that it is really adding any benefit.

tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
```diff
@@ -0,0 +1,37 @@
...
+max_nodes: 5
+# Note: Pattern matching in gcloud only supports 1 wildcard, centv5*-login-* won't work.
+login_node: "{{ slurm_cluster_name }}-login-*"
+controller_node: "{{ slurm_cluster_name }}-controller"
+post_deploy_tests:
+- test-partitions.yml
+custom_vars:
+  partitions:
+  - compute
+  - debug
```
nick-stroud, line 35: We run the startup script on both partitions. Is this adding value. I am wondering if we would get better coverage by running the startup script on 1 partition and having no startup script on the other partition. This way we test both cases. I realize we probably test the no startup script scenario in other tests. Any thoughts, feel free to push back.

tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
heyealex: I converted to a nfs-server. It's faster, and it provides a bit more coverage since none of the other slurm-gcp v5 tests include one.

tools/cloud-build/daily-tests/tests/slurm-v5-startup-scripts.yml
heyealex: That makes sense. I also renamed the test to make it more startup-script centric, so even if it's covered elsewhere, testing it here will reasonably be in scope.

Fixed, but rerunning test group 2 to verify everything still works.

nick-stroud (approved):

PR #791, opened 2022-12-15
Add disable_public_ips option to node group module

heyealex:
### Description
This option was missing before, but is present in the v4 modules as "disable_compute_public_ips" on this controller. This option will set a public IP via the access_config if the access_config is not set, otherwise it will do nothing.

### Submission Checklist

  install`)


community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -360,3 +360,9 @@ variable "access_config" {
   }))
   default = []
 }
+
+variable "disable_public_ips" {
```
nick-stroud, line 364: I am leaning towards making this, `enable_public_ip`. We have a task to deprecate all "disable" style variables, and in this case there is not even an underlying variable that we are trying to stay consistent with. I recognize that this creates an odd temporary situation where the variables are different on different components, so I could be convinced otherwise.

heyealex: I'd rather have consistency between modules rather than fixing the variable name in one module. 

I see two options that I'd be okay with:
1. Push this PR through now as is and make the breaking change for all modules later
1. Make the breaking change for all of them now, perhaps with backward compatibility built in.

I'd prefer (1) so that we can discuss and get agreement from everyone after the new year. Let me know your thoughts.

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
nick-stroud: OK 1 is fine.

PR #793, opened 2022-12-19
Run dependabot at repeatable weekly interval

tpdownes:
I believe this will not take effect until it hits the `main` branch. I am happy for that to go through the normal process of release, though.

See [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleday) for more details.

### Submission Checklist

  install`)


.github/dependabot.yml
```diff
@@ -19,5 +19,8 @@ updates:
 - package-ecosystem: gomod
   directory: /
   schedule:
-    interval: daily
+    interval: weekly
+    day: monday
```
nick-stroud, line 23: I had thought we had decided on Sunday so they would be ready first thing Monday morning. If we want to avoid weekend activity then how about 5AM Central on Monday.

tpdownes: We kinda left it open in discussion but I'm happy to move it to the suggested time which is as early as anyone rightly ought to be working in US ET.

nick-stroud (approved):

PR #810, opened 2023-01-05
Pass family through instead of looking up image to allow compute nodes to pick up new version within family

nick-stroud:
By passing the family instead of the specific image we allow for the compute nodes to grab the latest version of the image without any updates to the cluster. 

Tested: 
- I have manually verified that compute nodes will grab the latest image in a family without applying any updates to the cluster. The testing was done by:
  - Deploy cluster using custom image family in my project
  - Run job to confirm version of image used
  - While cluster is up, create a new image in the same family
  - Run subsequent job and confirm that new image is used

Follow on work might include:
- Using this same methodology for the controller and login images
- Verifying behavior on Slurm V5

I have opted to submit this PR as stand alone the compute nodes are most critical and additional testing should be completed for other updates. 

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/SchedMD-slurm-on-gcp-partition/outputs.tf
+++ b/community/modules/compute/SchedMD-slurm-on-gcp-partition/outputs.tf
@@ -13,11 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-data "google_compute_image" "compute_image" {
-  family  = var.instance_image.family
-  project = var.instance_image.project
-}
-
 output "partition" {
   description = "The partition structure containing all the set variables"
   value = {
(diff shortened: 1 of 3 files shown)


community/modules/compute/SchedMD-slurm-on-gcp-partition/outputs.tf
```diff
@@ -26,7 +21,7 @@ output "partition" {
     static_node_count : var.static_node_count
     max_node_count : var.max_node_count
     zone : var.zone
-    image : data.google_compute_image.compute_image.self_link
+    image : "projects/${var.instance_image.project}/global/images/family/${var.instance_image.family}"
```
cboneti, line 24: what if the image does not have a family? did you try that? 
Are we unable to specify an image directly?

cboneti (commented):
Just one question as before I recall being able to use images and not families. I thought we were still able to do that.

community/modules/compute/SchedMD-slurm-on-gcp-partition/outputs.tf
cboneti: I think the fix as is already brings value. I will approve it now and follow up on another task.

cboneti (approved):

community/modules/compute/SchedMD-slurm-on-gcp-partition/outputs.tf
nick-stroud: SG I will experiment to bring back non-family image usage in a separate pr.

PR #812, opened 2023-01-05
Add slash to network storage output for pre-existing file systems

heyealex:
The slash was added for the mount scripts directly, but not in the network storage output directly.

### Submission Checklist

  install`)

### DIFF
--- a/modules/file-system/pre-existing-network-storage/outputs.tf
+++ b/modules/file-system/pre-existing-network-storage/outputs.tf
@@ -18,7 +18,7 @@ output "network_storage" {
   description = "Describes a remote network storage to be mounted by fs-tab."
   value = {
     server_ip             = var.server_ip
-    remote_mount          = var.remote_mount
+    remote_mount          = contains(local.mount_vanilla_supported_fstype, var.fs_type) ? local.remote_mount_with_slash : var.remote_mount
     local_mount           = var.local_mount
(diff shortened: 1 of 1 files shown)


modules/file-system/pre-existing-network-storage/outputs.tf
```diff
@@ -18,7 +18,7 @@ output "network_storage" {
   description = "Describes a remote network storage to be mounted by fs-tab."
   value = {
     server_ip             = var.server_ip
-    remote_mount          = var.remote_mount
+    remote_mount          = contains(local.mount_vanilla_supported_fstype, var.fs_type) ? local.remote_mount_with_slash : var.remote_mount
```
nick-stroud, line 21: nit: from a readability standpoint it is getting a bet hard to follow when we need remote_mount_with_slash and when we just use remote mount. I feel like there is maybe an opportunity to say early on, "is it a gcs bucket, if not add the slash, and then just use remote mount for the rest of the module". If you think it could be easily implemented go for it otherwise feel free to ignore.

nick-stroud (approved):

modules/file-system/pre-existing-network-storage/outputs.tf
heyealex: I agree, and made the change. I'm running a couple tests before actually merging just to be safe.

PR #814, opened 2023-01-06
Enable Slurm v4 image to be specified by name

nick-stroud:
Tested:
- manually validated that specifying controller, login, and compute nodes by name and family work.

### Submission Checklist

  install`)


cboneti (approved):
There is a failed test, if that is an unrelated problem, then this is good to go.

nick-stroud:
> There is a failed test, if that is an unrelated problem, then this is good to go.

It is a V5 test that failed. I will merge this and follow up separately with the v5 failure.

PR #820, opened 2023-01-06
Adding support for existing GCS bucket in startup script module.

soumyapani:
### Submission Checklist

  install`)

### DIFF
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -162,6 +162,24 @@ sudo journalctl -u google-startup-scripts.service
   use: [homefs, startup]
 ```
 
+```yaml
+- id: startup
+  source: ./modules/scripts/startup-script
+  settings:
+    gcs_bucket_full_path: gs://user-test-bucket/folder1/folder2
+    runners:
+      - type: shell
+        source: modules/startup-script/examples/install_ansible.sh
+        destination: install_ansible.sh
+      - type: shell
+        source: modules/startup-script/examples/install_cloud_ops_agent.sh
+        destination: install_cloud_ops_agent.sh
+
+- id: compute-cluster
+  source: ./modules/compute/vm-instance
+  use: [startup]
+```
+
 ## License
 
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
@@ -219,6 +237,7 @@ No modules.
 | <a name="input_prepend_ansible_installer"></a> [prepend\_ansible\_installer](#input\_prepend\_ansible\_installer) | Prepend Ansible installation script if any of the specified runners are of type ansible-local | `bool` | `true` | no |
 | <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Project in which the HPC deployment will be created | `string` | n/a | yes |
 | <a name="input_region"></a> [region](#input\_region) | The region to deploy to | `string` | n/a | yes |
+| <a name="input_gcs_bucket_full_path"></a> [gcs_bucket_full_path](#input\_gcs_bucket_full_path) | The GCS full path for storage bucket and the object. | `string` | n/a | yes |
 | <a name="input_runners"></a> [runners](#input\_runners) | List of runners to run on remote VM.<br>    Runners can be of type ansible-local, shell or data.<br>    A runner must specify one of 'source' or 'content'.<br>    All runners must specify 'destination'. If 'destination' does not include a<br>    path, it will be copied in a temporary folder and deleted after running.<br>    Runners may also pass 'args', which will be passed as argument to shell runners only. | `list(map(string))` | `[]` | no |
 
 ## Outputs
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -162,8 +162,8 @@ sudo journalctl -u google-startup-scripts.service
   use: [homefs, startup]
 ```
 
-In the above example, a new GCS bucket is created to upload the startup-scripts. 
-But in the case where the user wants to reuse existing GCS bucket or folder, 
+In the above example, a new GCS bucket is created to upload the startup-scripts.
+But in the case where the user wants to reuse existing GCS bucket or folder,
 they are able to do so by using the `gcs_bucket_path` as shown in the below example
 
 ```yaml
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -162,11 +162,15 @@ sudo journalctl -u google-startup-scripts.service
   use: [homefs, startup]
 ```
 
+In the above example, a new GCS bucket is created to upload the startup-scripts. 
+But in the case where the user wants to reuse existing GCS bucket or folder, 
+they are able to do so by using the `gcs_bucket_path` as shown in the below example
+
 ```yaml
 - id: startup
   source: ./modules/scripts/startup-script
   settings:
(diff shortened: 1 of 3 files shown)


modules/scripts/startup-script/main.tf
```diff
@@ -25,10 +25,25 @@ locals {
...
+    var.gcs_bucket_full_path == null
+    ? try(module.config_storage_bucket[0], null)
+    : try(data.google_storage_bucket.existing_bucket[0], null)
+  )
+
+  storage_bucket_name = (
+    var.gcs_bucket_full_path == null
+    ? "${var.deployment_name}-startup-scripts-${random_id.resource_name_suffix.hex}"
+    : regex("^gs://([^/]*)/*(.*)", var.gcs_bucket_full_path)[0]
+  )
```
nick-stroud, line 38: I think you can get rid of this block and just put each name with the corresponding bucket source (lookup or create). If we are creating a bucket then we are using the one based off the deployment name as before, if we are looking up then we are using the gcs_bucket_path.

modules/scripts/startup-script/main.tf
```diff
@@ -25,10 +25,25 @@ locals {
   prepend_ansible_installer = length(local.ansible_local_runners) > 0 && var.prepend_ansible_installer
   runners                   = local.prepend_ansible_installer ? concat([local.ansible_installer], var.runners) : var.runners
 
+  storage_bucket = (
+    var.gcs_bucket_full_path == null
+    ? try(module.config_storage_bucket[0], null)
+    : try(data.google_storage_bucket.existing_bucket[0], null)
```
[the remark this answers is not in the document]
nick-stroud: Since you are already making the count of each source dependent on the name I think it would be simpler to just write:
```
  storage_bucket = coalesce(one(module.config_storage_bucket, 
     one(data.google_storage_bucket.existing_bucket)
```

modules/scripts/startup-script/README.md
```diff
@@ -162,6 +162,24 @@ sudo journalctl -u google-startup-scripts.service
   use: [homefs, startup]
 ```
 
+```yaml
```
nick-stroud, line 165: We should add a line of context as to what this example is showing.

modules/scripts/startup-script/variables.tf
```diff
@@ -30,6 +30,12 @@ variable "region" {
 }
 
 
+variable "gcs_bucket_full_path" {
```
nick-stroud, line 33: I assume pre-commits are failing because docs need to be updated with this variable. This can be done automatically by pre-commits. I can help you to get set up or do it myself on Monday.

modules/scripts/startup-script/README.md
```diff
@@ -162,6 +162,24 @@ sudo journalctl -u google-startup-scripts.service
...
+- id: startup
+  source: ./modules/scripts/startup-script
+  settings:
+    gcs_bucket_full_path: gs://user-test-bucket/folder1/folder2
```
nick-stroud, line 169: I think we will want to add or update a test to use this. I can help with this next week.

modules/scripts/startup-script/main.tf
```diff
@@ -68,22 +83,30 @@ resource "random_id" "resource_name_suffix" {
   byte_length = 4
 }
 
-resource "google_storage_bucket" "configs_bucket" {
-  project                     = var.project_id
-  name                        = "${var.deployment_name}-startup-scripts-${random_id.resource_name_suffix.hex}"
-  uniform_bucket_level_access = true
+module "config_storage_bucket" {
```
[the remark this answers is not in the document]
soumyapani: using module here so that I can conditionally use it in line #28. I just picked this pattern from existing hpc-toolkit code.

modules/scripts/startup-script/README.md
soumyapani: done.

modules/scripts/startup-script/main.tf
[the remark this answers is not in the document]
soumyapani: done

modules/scripts/startup-script/main.tf
soumyapani: done

modules/scripts/startup-script/main.tf
```diff
@@ -25,10 +25,25 @@ locals {
...
+  )
+
+  storage_bucket_name = (
+    var.gcs_bucket_full_path == null
+    ? "${var.deployment_name}-startup-scripts-${random_id.resource_name_suffix.hex}"
+    : regex("^gs://([^/]*)/*(.*)", var.gcs_bucket_full_path)[0]
+  )
+
+  storage_folder_path = var.gcs_bucket_full_path == null ? null : regex("^gs://([^/]*)/*(.*)", var.gcs_bucket_full_path)[1]
+  storage_folder_path_prefix = local.storage_folder_path == null ? "" : "${local.storage_folder_path}/"
```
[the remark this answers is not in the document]
soumyapani: Trimming trailing `/` before parsing. So it is handled now.

modules/scripts/startup-script/variables.tf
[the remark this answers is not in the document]
soumyapani: done.

modules/scripts/startup-script/main.tf
```diff
@@ -84,20 +75,20 @@ resource "random_id" "resource_name_suffix" {
...
   location                    = var.region
   storage_class               = "REGIONAL"
   labels                      = var.labels
 }
 
 data "google_storage_bucket" "existing_bucket" {
-  count                       = var.gcs_bucket_full_path != null ? 1 : 0
-  name                        = local.storage_bucket_name
+  count                       = var.gcs_bucket_path != null ? 1 : 0
+  name                        = regex("^gs://([^/]*)/*(.*)", local.gcs_bucket_path_trimmed)[0]
```
nick-stroud, line 91: nit: You could store the regex string as a local constant since you use it twice. This way if it ever needs to be updated we won't forget to update it in both places (DRY).
```
locals {
  bucket_regex = "^gs://([^/]*)/*(.*)"
}
...
name = regex(local.bucket_regex, local.gcs_bucket_path_trimmed)[0]
```

modules/scripts/startup-script/main.tf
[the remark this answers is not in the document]
nick-stroud: A few things I noticed here. 

1. If we use the module (as you currently have), to preserve the existing behavior, we would have to set `force_destroy = true`. Otherwise terraform destroy will fail. 

2. The same `count` syntax works with the `resource`. I personally don't have a reason to prefer the resource to the module except that it would limit the change to a smaller scope. For this reason I might suggest just keeping this as the resource (and adding in the count pattern you have here).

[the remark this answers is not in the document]
soumyapani: Converted to resource in latest iteration and validated it works.

modules/scripts/startup-script/main.tf
soumyapani: done.

modules/scripts/startup-script/README.md
nick-stroud: I am creating a task for this to be done in a follow on PR. It is something that will be picked up by the HPC Toolkit team unless you (@soumyapani) explicitly want to take it on.

modules/scripts/startup-script/variables.tf
nick-stroud: resolved

nick-stroud (approved):

PR #825, opened 2023-01-10
Adding a ramble module

douglasjacobsen:
This module automates installation of ramble, and runs arbitrary ramble specific commands.

Additionally, this adds the path to the spack installation as an output from the spack module.

### Submission Checklist

  install`)


douglasjacobsen:
I just pushed an update to this based on the issues fixed in #826

community/modules/scripts/ramble/templates/ramble_install.tpl
```diff
@@ -0,0 +1,62 @@
...
+  - name: Install git
+    ansible.builtin.package:
+      name:
+      - git
+      state: latest
+  - name: Install python
+    ansible.builtin.package:
+      name:
+      - python
+      state: latest
```
nick-stroud, line 29: Any thoughts on `latest` vs `present`. We kind of know that we just installed a specific version of python to get ansible to run. I am worried that we are then turning around and installing python again. Maybe we don't care.

community/modules/scripts/ramble/variables.tf
```diff
@@ -0,0 +1,69 @@
...
+}
+
+variable "chgrp_group" {
+  description = "Group to chgrp the Ramble clone to"
+  default     = "root"
+  type        = string
+}
+
+variable "chmod_mode" {
+  description = "Mode to chmod the Ramble clone to."
```
nick-stroud, line 54: Documentation should explain what the special value of empty string means: that chmod is not run.

community/modules/scripts/ramble/templates/ramble_install.tpl
nick-stroud: From discussion: if there is an easy way to specify a lower bound then we could try that. Otherwise stick with latest.

nick-stroud:
Closing to keep inbox clean. We will reopen or harvest from this PR when redesign implementation is underway.

PR #827, opened 2023-01-11
Run config validation test in a different directory than the repo, fix bugs found when making the change

heyealex:
### Description
This goal of this PR was to run the `validate_configs.sh` validation outside of the HPC Toolkit repo. Doing so exposed a bug with the packer modulereader package where the packerreader attempts to copy files by default, without checking if the source is local vs embedded.

The fix to this is included as well to ensure tests are passing again. Some refactoring took place as well with the sourcereader and modulereader. It became difficult to avoid cyclical dependencies, so I instead moved some functionality out of sourcereader and into modulereader where it belongs, and now use sourcereader more actively when copying a module from various sources to the file system.

There is still work to do to improve these modules (e.g. function names need to improve), but I tried to keep it limited to what was needed to reasonably get tests passing again without adding additional tech debt.

### Changes made

Testing
* `make tests` now also builds and installs the binary
* Moved golang tests where applicable, I didn't actually rewrite any beyond making them work in a different module

Refactor
* Reversed the package dependency between modulereader and sourcereader
  * Now the dependency is more linear: modulereader -> sourcereader -> blueprintio
  * Functionality of modulereader started slipping into sourcereader, this attempts to fix that, since it was getting difficult to avoid cyclical dependencies.
* packerreader no longer copies the module to a temp directory itself, it relies on existing functionality in sourcereader
* Moved API list  and functionality to modulereader
* Moved GetModuleInfo from sourcereader implementations to the top level of modulereader (it didn't belong in sourcereader, and this decreases duplicated code)

### Submission Checklist

  install`)


pkg/modulereader/resreader_test.go
```diff
@@ -98,6 +102,64 @@ func (s *MySuite) TestFactory(c *C) {
...
+	badSource := "gcs::https://www.googleapis.com/storage/v1/GoogleCloudPlatform/hpc-toolkit/modules"
+	moduleInfo, err = GetModuleInfo(badSource, tfKindString)
+	expectedErr = "Source is not valid: .*"
+	c.Assert(err, ErrorMatches, expectedErr)
+}
+
+func (s *MySuite) TestGetModuleInfo_Git(c *C) {
+
+	// Invalid git repository - path does not exists
+	badGitRepo := "github.com:not/exist.git"
```
tpdownes, line 130: Do these tests for GitHub and https URIs introduce an implied expectation of network connectivity to run unit tests? (i.e. testing for failure implies we expect a good value to succeed)

I believe (without full review) that we do not otherwise actually require GCP credentials or network connections.

tpdownes (changes requested):
This is a very significant improvement in the code and its technical debt. Please address the minor regex change and concern about the unit tests.

Additionally, I suggest renaming `output.tf` to `outputs.tf` to comply with [standard module structure](https://developer.hashicorp.com/terraform/language/modules/develop/structure).

Once you make these changes, do you believe we should run all optional PR checks or ought they be unaffected?

tools/validate_configs/validate_configs.sh
```diff
@@ -21,16 +21,26 @@ run_test() {
 	exampleFile=$(basename "$example")
 	DEPLOYMENT=$(echo "${exampleFile%.yaml}-$(basename "${tmpdir##*.}")" | sed -e 's/\(.*\)/\L\1/')
 	PROJECT="invalid-project"
+	LOCAL_SOURCE_PATTERN='source: ./'
```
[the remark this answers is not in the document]
cboneti: per https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/modules#local-modules, this can be ./, ../ and /. right?

Makefile
```diff
@@ -47,7 +47,7 @@ install: install-user
 
 endif
 
-tests: warn-terraform-version warn-packer-version test-engine validate_configs packer-check
+tests: ghpc install warn-terraform-version warn-packer-version test-engine validate_configs packer-check
```
cboneti, line 50: why make install as part of tests? 

Does this mean you now need to be root to be able to make tests? what if the user doesn't wan to make install?  Is this really necessary?

cboneti: also, validate configs already depends on ghpc, so there is no need to add the ghpc dependency here.

tools/validate_configs/validate_configs.sh
```diff
@@ -21,16 +21,26 @@ run_test() {
...
-	./ghpc create -l IGNORE --vars "project_id=${PROJECT},deployment_name=${DEPLOYMENT}" "${tmpdir}"/"${exampleFile}" >/dev/null ||
+
+	# Only run from the repo directory if there are local modules, otherwise
+	# run the test from the test directory using the installed ghpc binary.
+	if grep -q "${LOCAL_SOURCE_PATTERN}" "${cwd}/${example}"; then
+		cd "${cwd}"
+	else
+		cd "${tmpdir}"
+	fi
+	ghpc create -l IGNORE --vars "project_id=${PROJECT},deployment_name=${DEPLOYMENT}" "${tmpdir}"/"${exampleFile}" >/dev/null ||
```
cboneti, line 36: Consider this can also run on the user development machine (their laptops or desktops). I frequently call make tests from my workstation and this is a use case we want to continue supporting. 

This line is testing the system's ghpc (from the path), and not the local binary that you are building. Tests should focus at what you are specifically building and testing (Tests should not have system wide effects).

You should, instead, call ${cwd}/ghpc which is the one you just built.

cboneti (changes requested):
Please see some of my comments here. I have not reviewed the go code, but I trust Tom went through it in detail.

tools/validate_configs/validate_configs.sh
tpdownes: I agree with this concern and didn't properly consider it since make install copies into my PATH even as a non-superuser. +1

pkg/modulereader/resreader.go
```diff
@@ -50,6 +55,47 @@ func (i ModuleInfo) GetOutputsAsMap() map[string]VarInfo {
...
+	default:
+		return ModuleInfo{}, fmt.Errorf("Source is not valid: %s", source)
+	}
+
+	reader := Factory(kind)
+	mi, err := reader.GetInfo(modPath)
+
+	// Add API list if source is known
+	if sourcereader.IsEmbeddedPath(source) {
+		mi.RequiredApis = defaultAPIList(modPath)
```
tpdownes, line 88: I'd modify the comment to read "add APIs required by the module, if known" because it reduces ambiguity.

tpdownes:
> Please see some of my comments here. I have not reviewed the go code, but I trust Tom went through it in detail.

The bulk of the changes to Go code are cosmetic due to the improved dependency structure. That said, there is some real action in `resreader.go` which I support. The mental path through the code is easier to traverse.

tools/validate_configs/validate_configs.sh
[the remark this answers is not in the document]
heyealex: Updated the pattern for all three valid sources and to escape . correctly

tools/validate_configs/validate_configs.sh
```diff
@@ -27,15 +27,15 @@ run_test() {
 	exampleFile=$(basename "$example")
 	DEPLOYMENT=$(echo "${exampleFile%.yaml}-$(basename "${tmpdir##*.}")" | sed -e 's/\(.*\)/\L\1/')
 	PROJECT="invalid-project"
-	local LOCAL_MOD_COUNT=$(grep 'source: ./' "${example}" | wc -l)
-	echo "LOCAL_MOD_COUNT=${LOCAL_MOD_COUNT}"
+	local LOCAL_MOD_COUNT
+	LOCAL_MOD_COUNT=$(grep -c 'source: ./' "${example}")
```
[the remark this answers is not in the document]
heyealex: done

Makefile
heyealex: I wanted to make it visible from anywhere, but it is better to just point to the binary in the repo directly, so that's updated.

You wouldn't need root regardless, `make install` will install locally in that case (~/bin/). Either way, I agree it's better to not mess with the installed version.

tools/validate_configs/validate_configs.sh
heyealex: Done

pkg/modulereader/resreader.go
heyealex: Done

pkg/modulereader/resreader_test.go
heyealex: It looks like there is a similar failure in the sourcereader tests when network is cut off, but this test passes (test logs below).

These tests as are were copied over as is from sourcereader. Given that we have other tests that require network connection, I'd prefer to add a follow-up task to handle it. It would also make sense to run PR-Validation tests without the network once that's fixed. 

Let me know if it's enough of a concern to update this now, otherwise I'll consider this comment resolved once the task is created.

```
$ unshare -r -n go test hpc-toolkit/pkg/...
ok      hpc-toolkit/pkg/config  (cached)
ok      hpc-toolkit/pkg/deploymentio    (cached)
ok      hpc-toolkit/pkg/modulereader    (cached)
ok      hpc-toolkit/pkg/modulewriter    (cached)
?       hpc-toolkit/pkg/validators      [no test files]

----------------------------------------------------------------------
FAIL: git_test.go:25: MySuite.TestCopyGitModules

git_test.go:35:
    c.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"error downloading 'https://github.com/terraform-google-modules/terraform-google-project-factory.git': /usr/bin/git exited with 128: Cloning into '/tmp/getter1748151968/temp'...\nssh: Could not resolve hostname github.com.corp.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.prod.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.prodz.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.net.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.in.goog.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com: Temporary failure in name resolution\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n"} ("error downloading 'https://github.com/terraform-google-modules/terraform-google-project-factory.git': /usr/bin/git exited with 128: Cloning into '/tmp/getter1748151968/temp'...\nssh: Could not resolve hostname github.com.corp.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.prod.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.prodz.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.net.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.google.com.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com.in.goog.: Temporary failure in name resolution\r\nssh: Could not resolve hostname github.com: Temporary failure in name resolution\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n")

OOPS: 10 passed, 1 FAILED
--- FAIL: Test (0.10s)
FAIL
FAIL    hpc-toolkit/pkg/sourcereader    0.219s
FAIL
```

heyealex:
I have integration tests running for another PR (#832), I will kick them off for this PR when they finish

tools/validate_configs/validate_configs.sh
```diff
@@ -21,16 +21,28 @@ run_test() {
 	exampleFile=$(basename "$example")
 	DEPLOYMENT=$(echo "${exampleFile%.yaml}-$(basename "${tmpdir##*.}")" | sed -e 's/\(.*\)/\L\1/')
 	PROJECT="invalid-project"
+	GHPC_PATH="${cwd}/ghpc"
+	# Cover the three possible starting sequences for local sources: ./ ../ /
+	LOCAL_SOURCE_PATTERN='source:\s\+\(\./\|\.\./\|/\)'
```
cboneti, line 26: nit, do not resubmit because of this, but I think you could have made it simpler with something like:
'source:\s\+\.\{0,2\}/'
I am never too sure of the different regexp syntaxes.. I *think* this one should hold.

cboneti (approved):

PR #832, opened 2023-01-12
Inject build id into Omnia test, drop to last build

heyealex:
There have been known issues with the naming of the resources created by the omnia example in automated testing. In order to allow parallel builds, this PR injects the build ID into the name prefix before running the test.

More thorough fixes are possible, but are not a priority for now. In addition, the test has been dropped to last in the group to ensure that all other tests can run if Omnia fails due to changes outside of our teams control.

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -81,10 +81,6 @@
         gcloud compute instances describe --zone={{ zone }} {{ remote_node }}
         --format='get(networkInterfaces[0].accessConfigs[0].natIP)'
       when: '"*" not in remote_node'
-    - name: Set remote_ip variable - Exact name provided
-      ansible.builtin.set_fact:
-        remote_ip: "{{ get_remote_ip.stdout }}"
-      when: '"*" not in remote_node'
     - name: Get IP of the remote node - Name pattern provided
       changed_when: false
       register: get_remote_ip
@@ -93,13 +89,9 @@
           --format='get(networkInterfaces[0].accessConfigs[0].natIP)' --limit=1 \
           --filter=NAME:{{ remote_node }}
       when: '"*" in remote_node'
-    - name: Set remote_ip variable - Name pattern provided
-      ansible.builtin.set_fact:
-        remote_ip: "{{ get_remote_ip.stdout }}"
-      when: '"*" in remote_node'
     - name: Print remote node's public IP
       ansible.builtin.debug:
-        var: remote_node
+        var: get_remote_ip.stdout
 
     ## Setup firewall for cloud build
     - name: Get Builder IP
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
```diff
@@ -74,12 +74,32 @@
...
+      when: '"*" not in remote_node'
+    - name: Get IP of the remote node - Name pattern provided
+      changed_when: false
+      register: get_remote_ip
+      ansible.builtin.command: >-
+        gcloud compute instances list \
+          --format='get(networkInterfaces[0].accessConfigs[0].natIP)' --limit=1 \
+          --filter=NAME:{{ remote_node }}
+      when: '"*" in remote_node'
+    - name: Set remote_ip variable - Name pattern provided
```
nick-stroud, line 96: Do you need this one twice (here and L84). Seems like you could do the two different gets and just set it once. And do we even need it at all, or could you just have each of the Gets set `remote_ip` directly?

tools/cloud-build/daily-tests/integration-group-3.yaml
```diff
@@ -123,7 +123,11 @@ steps:
     set -x -e
     BUILD_ID_FULL=$BUILD_ID
     BUILD_ID_SHORT=$${BUILD_ID_FULL:0:6}
+    OMNIA_EXAMPLE=community/examples/omnia-cluster.yaml
 
-    ansible-playbook tools/cloud-build/daily-tests/ansible_playbooks/packer-integration-test.yml \
+    # Inject the build ID into the name prefix of the vm-instance modules to avoid naming collisions
+    sed -i "s/name_prefix: \(.*\)/name_prefix: \1-$${BUILD_ID_SHORT}/" $${OMNIA_EXAMPLE}
```
nick-stroud, line 129: nit for future: This would have been easier to review if the rearrange had been done in a separate commit.

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
heyealex: It looks like it. I can't recall why it was needed for the slurm tests, but from the looks of it, it might not have been. Updated, tests pending.

tools/cloud-build/daily-tests/integration-group-3.yaml
heyealex: I was hoping the git diff would be cleaner, but it didn't turn out that way. I'll keep that in mind in the future.

heyealex:
I'm running tests against all groups that use the `base-integration-test.yml` playbook to be safe. It is ready to be reviewed again, but I will not merge until they all pass.

nick-stroud (approved):

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
heyealex: I see now why it had to be defined twice. There is a variable and a "ansible fact". The variable is updated for any task that registers it, even if it's skipped, so I get this back when it's not a pattern (or the most recent task to register the variable):

```
TASK [Print remote node's public IP] *******************************************
ok: [localhost] => {
    "get_remote_ip.stdout": "VARIABLE IS NOT DEFINED!"
}
```

I missed it when testing Omnia because the most recent task to set the variable actually set it, but it failed the other base integration tests. I'm going to roll back to change and re-test.

PR #838, opened 2023-01-17
Improve expandVariables error message

thiagosgobe:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -36,7 +36,7 @@ const (
 	// Checks if a variable exists only as a substring, ex:
 	// Matches: "a$(vars.example)", "word $(vars.example)", "word$(vars.example)", "$(vars.example)"
 	// Doesn't match: "\$(vars.example)", "no variable in this string"
-	anyVariableExp string = `(^|[^\\])\$\((.*)\)`
+	anyVariableExp string = `(^|[^\\])\$\((.*?)\)`
 	literalExp     string = `^\(\((.*)\)\)$`
 	// the greediness and non-greediness of expression below is important
 	// consume all whitespace at beginning and end
@@ -593,7 +593,21 @@ func expandSimpleVariable(
 func expandVariable(
 	context varContext,
 	modToGrp map[string]int) (string, error) {
-	return "", fmt.Errorf("%s: expandVariable", errorMessages["notImplemented"])
+	re := regexp.MustCompile(anyVariableExp)
+	matchall := re.FindAllString(context.varString, -1)
+	errHint := ""
+	for _, element := range matchall {
+		// the regex match will include the first matching character
+		// this might be (1) "^" or (2) any character EXCEPT "\"
+		// if (2), we have to remove the first character from the match
+		firstChars := element[0:2]
+		if firstChars != "$(" {
+			element = strings.Replace(element, element[0:1], "", 1)
+		}
+		errHint += "\\" + element + " will be rendered as " + element + "\n"
+	}
+	return "", fmt.Errorf("%s \n%s",
+		errorMessages["varWithinStrings"], errHint)
 }
 
 // isDeploymentVariable checks if the entire string is just a single deployment variable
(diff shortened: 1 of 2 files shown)


thiagosgobe:
Error message example:

```bash
2023/01/17 16:53:17 expandVariables: not yet implemented: String operations with blueprint variables "$()" are not yet implemented: 
#!/bin/bash
sudo yum -y install kernel-devel-$(uname -r) kernel-headers-$(uname -r)
sudo yum -y groupinstall "KDE desktop" "X Window System" "Fonts" "Development Tools" "Server with GUI"
curl -O https://storage.googleapis.com/nvidia-drivers-us-public/GRID/vGPU14.2/NVIDIA-Linux-x86_64-510.85.02-grid.run
sudo bash NVIDIA-Linux-x86_64-510.85.02-grid.run -silent
 
Use a non-quoted backslash as an escape character if the content conflicts with blueprint variable notation, e.g. \$(not.bp_var).
```

tpdownes (changes requested):
I definitely like guiding the user to where the problem is occurring (which we don't do at all right now). I have few concerns:

- printing the entire string, rather than the specific substrings where the problem occurs
- printing out guidance at both top and bottom
- "not yet implemented" appears twice
- there are 2 colons in the first message

This is a hard problem! The user may want the string as shown or they may be trying to do string interpolation. I suggest changes that result in the following type of message:

```
expandVariables: variables "$(...)" within strings are not yet implemented. remove them or add a backslash to render literally
\$(uname -r) will be rendered as $(uname-r)
\$(uname -r) will be rendered as $(uname-r)
```

pkg/config/expand.go
```diff
@@ -593,7 +593,14 @@ func expandSimpleVariable(
 func expandVariable(
 	context varContext,
 	modToGrp map[string]int) (string, error) {
-	return "", fmt.Errorf("%s: expandVariable", errorMessages["notImplemented"])
+	re := regexp.MustCompile(`\$\((.*?)\)`)
```
tpdownes, line 596: Shouldn't we be using `anyVariableExp` here?

tpdownes (changes requested):

pkg/config/expand.go
thiagosgobe: `anyVariableExp` in this context would return the precedent character:

`kernel-devel-$(uname -r)` would return `-$(uname -r)`

That's because anyVariable needs to know what comes first in order to escape when the precedent character is the backslash.

tpdownes: I'll state my concerns more explicitly:

1.`anyVariableExp` is the existing regular expression to search for variables.
2. If you provide the string "$(vars.example1) is like \$(vars.example2)", your message will advise the user to escape both `$(vars.example1)` AND `\$(vars.example2)`.

Please use `anyVariableExp` such that we don't encounter (2) and don't provide multiple definitions for the same thing (1). I'm not convinced it's actually true that the matching substring will include the first character, but it's easy to slice out if it does.

pkg/config/expand.go
```diff
@@ -593,7 +593,19 @@ func expandSimpleVariable(
 func expandVariable(
 	context varContext,
 	modToGrp map[string]int) (string, error) {
-	return "", fmt.Errorf("%s: expandVariable", errorMessages["notImplemented"])
+	re := regexp.MustCompile(anyVariableExp)
+	matchall := re.FindAllString(context.varString, -1)
+	errHint := ""
+	for _, element := range matchall {
+		// get first 2 chars
```
tpdownes, line 600: Summary of offline discussion... functionality is right. Let's improve commenting so future archaeologists can understand the regular expression and its behavior. Change:

```
// get first 2 chars
```

to something like

```
the regex match will include the first matching character
this might be (1) "^" or (2) any character EXCEPT "\"
if (2), we have to remove the first character from the match
```

tpdownes (changes requested):
Almost done!

tpdownes (approved):
LGTM

PR #839, opened 2023-01-17
Add documentation for accelerator support in the HPC Toolkit

heyealex:
### Submission Checklist

  install`)

### DIFF
--- a/docs/gpu-support.md
+++ b/docs/gpu-support.md
@@ -10,11 +10,13 @@
   * `schedmd-slurm-gcp-v5-*`
   * `SchedMD-slurm-on-gcp-*`
 * PBS Pro modules (`pbspro-*`)
+* Cloud Batch modules through custom instance templates
 
 ## Accelerator definition automation
 
 The schedmd-slurm-gcp-v5 modules ([node-group], [controller] and [login]),
-the [vm-instance] module and any module relying on [vm-instance] support
+the [vm-instance] module and any module relying on [vm-instance] (HTCondor,
+Omnia, PBS Pro) support
 automation for defining the `guest_accelerator` config. If the user supplies any
 value for this setting, the automation will be bypassed.
 
@@ -53,7 +55,9 @@ cannot be determined automatically like with `a2`.
 * To list accelerator types and availability by region, run
   `gcloud compute accelerator-types list`. The information is also available in
   the Google Cloud documentation [here](https://cloud.google.com/compute/docs/gpus/gpu-regions-zones).
-* Deployment time of VMs with many guest accelerators can take longer.
+* Deployment time of VMs with many guest accelerators can take longer. See the
+  [Timeouts when deploying a compute VM](#timeouts-when-deploying-a-compute-vm)
+  section below if you experience timeouts because of this.
 
 ### Slurm on GCP
 
@@ -80,8 +84,8 @@ provides consistency with the underlying terraform modules from the
 
 As mentioned above, VMs with many guest accelerators can take longer to deploy.
 Slurm sets timeouts for creating VMs, and it's possible for high GPU
-configurations to push past the default timeout. The timeout in the Slurm on GCP
-v4 HPC Toolkit modules (`SchedMD-slurm-on-gcp-*`) due to a timeout that cannot
+configurations to push past the default timeout. The timeout in the
+Slurm on GCP v4 HPC Toolkit modules (`SchedMD-slurm-on-gcp-*`) cannot
 be increased, therefore we recommend using the Slurm on GCP v5 modules.
 
 The v5 Toolkit modules (`schedmd-slurm-gcp-v5-*`) allow Slurm configuration
@@ -103,14 +107,19 @@ See the example below which increases the `resume_timeout` from the default of
     ...
 ```
 
-#### Using GPUs with Slurm
+#### Launching Slurm jobs with GPUs
 
 In order to utilize the GPUs in the compute VMs deployed by Slurm, the GPU
-count must be specified when starting a job with `srun` or `sbatch`:
+count must be specified when submitting a job with `srun` or `sbatch`. For
+instance, the following `srun` command launches a job that runs nvidia-smi in a
+partition called `gpu_partition` (`-p gpu_partition`) on a full node (`-N 1`)
+with 8 GPUs (`--gpus 8`):
 
 ```shell
 srun -N 1 -p gpu_partition --gpus 8 nvidia-smi
 ```
+
+An equivalent `sbatch` script:
 
 ```bash
 #!/bin/bash
--- a/docs/gpu-support.md
+++ b/docs/gpu-support.md
@@ -2,7 +2,7 @@
 
 ## Supported modules
 
-* [vm-instance] and therefore any module that relies on `vm-instance` including
+* [vm-instance] and therefore any module that relies on `vm-instance` including:
   * HTCondor modules including [htcondor-install], [htcondor-configure] and
     [htcondor-execute-point].
   * [omnia-install]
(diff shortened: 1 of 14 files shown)


docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
+# Deploying with Accelerators in the Cloud HPC Toolkit
+
+## Supported modules
```
cboneti, line 3: This makes it sound like only the following modules support GPUs. 
Doesn't Batch, HTCondor or Omnia support GPUs?

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+* Slurm on GCP modules where applicable, both version 4 and version 5
+  * `schedmd-slurm-gcp-v5-*`
+  * `SchedMD-slurm-on-gcp-*`
+* PBS Pro modules (`pbspro-*`)
+
+## Accelerator definition automation
+
+The [vm-instance] and schedmd-slurm-gcp-v5 [node-group], [controller] and
+[login] modules support automation for defining the `guest_accelerator` config.
+If the user supplies any value for this setting, the automation will be bypassed.
```
cboneti, line 15: What about modules that use vm-instance? should we not mention them here?

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+
+The automation is handled primary in the `gpu_definition.tf` files in these
+modules. This file assumes the existence of two input variables in the module:
+
+* `guest_accelerator`: A list of terraform objects with the attributes type and
+  count.
+* `machine_type`: Defines the machine type of the VM being created.
+
+`gpu_definition.tf` works by checking the `machine_type` to associate it with
```
cboneti, line 24: see suggestions at the doc

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+[node-group]: https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/community/modules/compute/schedmd-slurm-gcp-v5-node-group
+[controller]: https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/community/modules/scheduler/schedmd-slurm-gcp-v5-controller
+[login]: https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/community/modules/scheduler/schedmd-slurm-gcp-v5-login
+
+## Troubleshooting and tips
+
+* To list accelerator types and availability by region, run
+  `gcloud compute accelerator-types list`. The information is also available in
+  the Google Cloud documentation [here](https://cloud.google.com/compute/docs/gpus/gpu-regions-zones).
+* Deployment time of VMs with guest accelerators can be longer than a simple VM.
```
cboneti, line 43: nit: I would rephrase as: " Deployment time of VMs with many guest accelerators can take longer."

We should at least suggest how one will increase the timeout for Slurm V5.

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Slurm on GCP
+
+* When deploying a Slurm cluster with GPUs, we highly recommend using the
+  modules based on Slurm on GCP version 5 (`schedmd-slurm-gcp-v5-*`). The
+  interface is more consistent with HPC Toolkit standards and more functionality
+  is available to support, debug and workaround any issues related to GPU
+  resources.
+* `SchedMD-slurm-on-gcp-v5-*` modules have a different interface for defining
```
cboneti, line 52: you mean v4?

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+  attached accelerators, `gpu_type` and `gpu_count`. These must be set even if
+  the machine type implies GPUs.
+* Some GPUs will fail in Slurm on GCP v4 HPC Toolkit modules
+  (`SchedMD-slurm-on-gcp-*`) due to a timeout that cannot be customized.
+  * Slurm on GCP v5 HPC Toolkit modules (`schedmd-slurm-gcp-v5-*`) have a timeout
```
cboneti, line 57: I would make this a direct recommendation, and move this note to the troubleshooting and tips section.

cboneti: also, what is the name of the option? you should just recommend the setting=value pair (like 10 minutes).

docs/gpu-support.md
```diff
@@ -0,0 +1,73 @@
...
+    variables that can be adjusted to work around this. If you run into this
+    issue, consider migrating to the v5 modules.
+* The Slurm on GCP v5 HPC Toolkit modules (`schedmd-slurm-gcp-v5-*`) have two
+  variables that can be used to define attached GPUs, `guest_accelerators` for
+  HPC Toolkit consistency and `gpus` for consistency with the underlying
+  terraform modules from the [Slurm on GCP repo][slurm-gcp].
```
cboneti, line 63: We should be opinionated, we should recommend using the toolkit format and document that the other one will be used if specified...

cboneti (commented):
This is great, thanks! I added a few suggestions.

docs/gpu-support.md
heyealex: Done

docs/gpu-support.md
heyealex: Added. With the addition of code blocks, I also converted this away from bullets to text.

docs/gpu-support.md
heyealex: Yes, thanks for catching that.

docs/gpu-support.md
heyealex: This is still under "Troubleshooting and tips", but in a subsection focused on Slurm on GCP specific issues.

I updated the text to be more clear and added an example updating `cloud_parameters`

docs/gpu-support.md
heyealex: Done

docs/gpu-support.md
cboneti: I think it would still be important to mention that Batch supports it...

docs/gpu-support.md
cboneti: End users will probably not know what modules use it. I like it how you. mentioned them by name up in the first section and I would repeat that here, even if just like:
```
(...) relying on [vm-instance] (htcondor-*, omnia-install, pbs-pro-*) (...)
```

docs/gpu-support.md
cboneti: I like how you rewrote this, but I would at least point to the timeouts section so users who just glance at this will get it. Something like:
"See the [Timeouts when deploying a compute VM](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/20937d0bd5f8eabc892739bb6c2705858131e6bb/docs/gpu-support.md#timeouts-when-deploying-a-compute-vm) section below.

docs/gpu-support.md
```diff
@@ -0,0 +1,140 @@
...
+provides consistency with the underlying terraform modules from the
+[Slurm on GCP repo][slurm-gcp].
+
+#### Timeouts when deploying a compute VM
+
+As mentioned above, VMs with many guest accelerators can take longer to deploy.
+Slurm sets timeouts for creating VMs, and it's possible for high GPU
+configurations to push past the default timeout. The timeout in the Slurm on GCP
+v4 HPC Toolkit modules (`SchedMD-slurm-on-gcp-*`) due to a timeout that cannot
+be increased, therefore we recommend using the Slurm on GCP v5 modules.
```
cboneti, line 85: I think this phrase is fragmented. You probably mean:
```
The timeout in the Slurm on GCP
v4 HPC Toolkit modules (`SchedMD-slurm-on-gcp-*`) cannot
be increased and, therefore, (...)
```

docs/gpu-support.md
```diff
@@ -0,0 +1,140 @@
...
+```yaml
+- id: slurm_controller
+  source: community/modules/scheduler/schedmd-slurm-gcp-v5-controller
+  use: [...]
+  settings:
+    cloud_parameters:
+      resume_rate: 0
+      resume_timeout: 600  # Update this value, default is 300
+      suspend_rate: 0
+      suspend_timeout: 300
```
cboneti, line 101: I wonder if the suspend_timeout should also be increased... Since I am not sure, we can leave it as 300, but be vigilant around it, and maybe test it.

docs/gpu-support.md
```diff
@@ -0,0 +1,140 @@
...
+      no_comma_params: false
+    ...
+```
+
+#### Using GPUs with Slurm
```
cboneti, line 106: What about: 

### Launching Slurm Jobs with GPUS 
?? is it better ??

docs/gpu-support.md
```diff
@@ -0,0 +1,140 @@
...
+
+In order to utilize the GPUs in the compute VMs deployed by Slurm, the GPU
+count must be specified when starting a job with `srun` or `sbatch`:
```
cboneti, line 109: s/starting/submitting/

cboneti: also s/:/. For instance, the following srun command launches a job that runs nvidia-smi in a partition called gpu_partition (-p gpu_partition) and a full node (-N 1) with 8 GPUs (-gpus 8)./

cboneti (approved):
This is great. How are users going to find this file? should we add links to it from Slurm, vm instance and other places? Is there a GPU example that should also point to this?

docs/gpu-support.md
heyealex: Added

docs/gpu-support.md
heyealex: Done

docs/gpu-support.md
heyealex: Done

docs/gpu-support.md
heyealex: Done

docs/gpu-support.md
heyealex: Good point. I added a task to investigate this.

docs/gpu-support.md
heyealex: Yup, I like that better. It's been updated.

docs/gpu-support.md
heyealex: Done

PR #840, opened 2023-01-18
Add validation logic for references to outputs from prior modules

tpdownes:
This PR adds the validation logic for references to outputs from modules in earlier deployment groups. It does not yet add full support for these references; this will require further PRs to implement. However, it puts into place the validation logic such that these "intergroup references" are rejected for cases we never intend to support. For example, intergroup references are rejected when:

- the reference does not include an explicit group ID
- they refer to a later group

There is an implicit change here that allows the user to use references with explicit groups even in the "intragroup reference" case. The behavior of these references is identical to the status quo (`module_id.output_name`). I believe it would be more complicated to enforce and communicate "implicitness" requirements for intragroup references, so it is better to allow them.

With this PR, unit test coverage for the package improves from 84.9% to 85.3%.

### Submission Checklist

  install`)


pkg/config/expand.go
```diff
@@ -535,10 +533,52 @@ func expandSimpleVariable(
...
+
+	callingGroupID := context.blueprint.DeploymentGroups[context.groupIndex].Name
+	refStr := contents[1]
+
+	ref, err := identifySimpleVariable(refStr, callingGroupID)
+	if err != nil {
+		return "", err
+	}
+
+	isInterGroupReference, err := validateReference(ref, context, modToGrp)
```
heyealex, line 565: My preference for flow here (feel free to disregard if you disagree):

```go
// Checks based only on the string pattern. Could return a value or set a field of the struct
isInterGroupReference, err := ref.isInterGroupReference()

// This or include a field in ref for tracking intergroup or not
err = validateReference(ref, context, modToGrp, isInterGroupReference)
```

pkg/config/expand_test.go
```diff
@@ -447,64 +448,88 @@ func (s *MySuite) TestIdentifySimpleVariable(c *C) {
...
-	testVarContext.blueprint.Vars["globalExists"] = "existsValue"
-	testVarContext.varString = "$(vars.globalExists)"
-	got, err := expandSimpleVariable(testVarContext, testModToGrp)
+	testVarContext1.blueprint.Vars["globalExists"] = "existsValue"
+	testVarContext1.varString = "$(vars.globalExists)"
+	got, err := expandSimpleVariable(testVarContext1, testModToGrp)
 	c.Assert(err, IsNil)
-	expected := "((var.globalExists))"
-	c.Assert(got, Equals, expected)
+	expectedErr = "((var.globalExists))"
```
heyealex, line 516: Why `expectedErr` if this isn't an error?

heyealex (changes requested):

pkg/config/expand.go
tpdownes: My theory for `varReference` -- which does not have to be the only theory -- is that it's just converting the YAML string to a Go struct with well-defined fields. I am trying to avoid doing anything that is a "guess" (such as whether the target of the reference actually exists anywhere in the blueprint).

That said, because the source GroupID is a parameter to the function, we could add a field (or func interface) which would report whether the reference _superficially_ appears to be intergroup. I will look at the impact this would have on the code. It would probably not need the first line or the last parameter in the 2nd. I'd still want to be validating the intergroup-ness of a reference.

pkg/config/expand.go
```diff
@@ -473,52 +473,50 @@ func identifySimpleVariable(yamlReference string, callingGroupID string) (varRef
...
-	}
-
-	if ref.GroupID == "deployment" && ref.ID == "vars" { // Global variable
-		// Verify global variable exists
-		if _, ok := context.blueprint.Vars[ref.Name]; !ok {
-			return "", fmt.Errorf("%s: %s is not a deployment variable",
-				errorMessages["varNotFound"], context.varString)
+// if reference is valid, error will be nil; boolean value indicates
+// if is intergroup (yes: true, no: false)
+func validateReference(ref varReference, context varContext, modToGrp map[string]int) (bool, error) {
```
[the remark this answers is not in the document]
tpdownes: Based on discussion [below], I will explore whether adding a superficial IsIntergroup field to the varReference struct helps us here. The problem may still exist -- there is superficial intergroup-ness in the string and then there is the question of "do you actually see both groups in the blueprint" and "in the right order"?

I don't really want to end up adding a bunch of functions with similar parameter lists if I don't have to. But I will look at this after addressing.

[below]: https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/840/files#r1082740268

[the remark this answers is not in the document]
tpdownes: I've chosen not to split up this function, although I've simplified the code such that:

- varReference tracks explicit references only when intergroup
- validateReference returns only an `error` and errors appropriately

After this, `varReference` can be trusted to accurately express intergroup-ness.

pkg/config/expand.go
```diff
@@ -473,52 +473,50 @@ func identifySimpleVariable(yamlReference string, callingGroupID string) (varRef
...
+			}
+			return false, nil
 		}
-		return fmt.Sprintf("((var.%s))", ref.Name), nil
+		return false, fmt.Errorf("%s is not a valid ID for a deployment-wide reference", ref.ID)
 	}
 
-	// Module variable
-	// Verify module exists
+	// at this point, we know that
```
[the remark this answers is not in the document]
tpdownes: Yes, that must have gotten snipped at some point along the way. Will be resolved within the updated PR.

pkg/config/expand.go
tpdownes: I have modified varReference.Explicit to be varReference.ExplicitInterGroup. This value is validated by `validateReference` and can then be trusted by remaining code.

pkg/config/expand_test.go
tpdownes: Resolved on HEAD

pkg/config/expand.go
```diff
@@ -473,52 +473,50 @@ func identifySimpleVariable(yamlReference string, callingGroupID string) (varRef
...
+	if ref.GroupID == "deployment" {
+		if ref.ID == "vars" {
+			if _, ok := context.blueprint.Vars[ref.Name]; !ok {
+				return false, fmt.Errorf("%s: %s is not a deployment variable",
+					errorMessages["varNotFound"], ref.ID)
+			}
+			return false, nil
 		}
-		return fmt.Sprintf("((var.%s))", ref.Name), nil
+		return false, fmt.Errorf("%s is not a valid ID for a deployment-wide reference", ref.ID)
```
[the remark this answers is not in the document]
tpdownes: Done. This code should essentially never be traversed (there is a task to ensure that no deployment group is named "deployment").

heyealex (approved):
Thanks for the updates! LGTM

PR #858, opened 2023-01-25
Update varReference implementation and usage

tpdownes:
Match behavior of modReference arising from review of #852

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -467,20 +467,18 @@ func identifyModuleByReference(yamlReference string, dg DeploymentGroup) (modRef
 /*
 A variable reference has the following fields
   - ID: a module ID or "vars" if referring to a deployment variable
-  - GroupID: if ID is a module ID, GroupID must be the deployment group in
-    which the module is *expected* to be found. If ID is "vars", then it
-    should be set to "deployment" to indicate that the reference belongs
-    to the entire blueprint, rather than a deployment group.
   - Name: the name of the module output or deployment variable
-  - ExplicitInterGroup: a boolean value indicating whether the user made a
-    reference that superficially appears to be intergroup (i.e. they used
-    an explicit GroupID that differs from the context's GroupID)
+  - ToGroupID: the deployment group in which the module is *expected* to be found
+  - FromGroupID: the deployment group from which the reference is made
+  - Explicit: a boolean value indicating whether the user made a reference that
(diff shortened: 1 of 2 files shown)


pkg/config/expand.go
```diff
@@ -573,11 +573,11 @@ func (ref *modReference) validate(depGroups []DeploymentGroup, modToGrp map[stri
 // ref.ExplicitInterGroup: intergroup references must explicitly identify the
 // target group ID and intragroup references cannot have an incorrect explicit
 // group ID
-func validateReference(ref varReference, context varContext, modToGrp map[string]int) error {
+func (ref *varReference) validate(depGroups []DeploymentGroup, vars map[string]interface{}, modToGrp map[string]int) error {
```
nick-stroud, line 576: nit: This looks similar enough to me that I feel like there is an opportunity for code reuse between this and modReference. Something like an interface that does `validateModuleReference` and then the varReference would do additional validation. Probably not congruent with my other comment about `ID` name.

I think it would look something like [this](https://stackoverflow.com/q/31997753) in practice.

I realize I already brought this up before and we moved past it. So don't feel like you need to do it.

nick-stroud (approved):

pkg/config/expand.go
```diff
@@ -467,20 +467,18 @@ func identifyModuleByReference(yamlReference string, dg DeploymentGroup) (modRef
...
+  - FromGroupID: the deployment group from which the reference is made
+  - Explicit: a boolean value indicating whether the user made a reference that
+    explicitly identified ToGroupID rather than inferring it using FromGroupID
 */
 type varReference struct {
-	ID                 string
-	GroupID            string
-	Name               string
-	ExplicitInterGroup bool
+	ID          string
```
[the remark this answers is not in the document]
tpdownes: The problem with this is that, in the model, `ref.ID` can also be "vars" which is not a module. It might also complicate your suggestion at https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/858#discussion_r1087170425. I'd rather not address this issue at the time. The meaning is documented which is an improvement.

pkg/config/expand.go
tpdownes: I do want to consider a `type reference interface {...}` as a means of implementing `validate`. At the moment, that is complicated by the fact that the parameter list is different for each. I prefer to tackle that as its own task, probably as part of a general attempt to simplify blueprint parsing and validation.

PR #871, opened 2023-01-31
Adds HCLS Blueprint and deploy instructions

nick-stroud:
This change is a starting point for the HCLS blueprint to get a basis from which those involved can collaborate and iterate.

Items to follow:
- sbatch script and config files that uses the 1FJS protein
- Updates to spack install and slurm partition to run gromacs with GPUs 

### Submission Checklist

  install`)

### DIFF
--- a/docs/videos/healthcare-and-life-sciences/README.md
+++ b/docs/videos/healthcare-and-life-sciences/README.md
@@ -1,0 +1,106 @@
+# Deployment Instructions
+
+1. Clone the repo
+
+   ```bash
+   git clone https://github.com/nick-stroud/hpc-toolkit.git
+   cd hpc-toolkit
+   git checkout hcls_mega_blueprint
+   ```
+
+1. Build the HPC Toolkit
+
+   ```bash
+   make
+   ```
+
+1. Generate the deployment folder
+
+   The home and apps server IPs will be populated after the setup stage and are
+   needed for later stage deployments. For now we will populate with a dummy
+   value to allow HPC Toolkit to run.
+
+   ```bash
+   ./ghpc create docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml -w \
+       --vars project_id=<project> \
+       --vars homefs_server_ip=1.1.1.1 \
+       --vars appsfs_server_ip=1.1.1.1 \
+       --vars bucket_name_inputs=<input_bucket_name> \
+       --vars bucket_name_outputs=<output_bucket_name> \
+       --vars bucket_name_software=<software_bucket_name>
+   ```
+
+1. Run terraform commands for `enable_apis` group
+
+   Call the provided `terraform init` and `terraform apply` commands for the
+   `enable_apis` group. This will ensure that all of the needed apis are
+   enabled before deploying the cluster.
+
+1. Deploy the `setup` group
+
+   This group will create a network and file systems to be used by the cluster.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
+   `setup` group.
+
+   > **Note**: You may have to tinker with bucket names to find an unused
+   > namespace. If you get errors that the bucket already exists update the
+   > bucket names in the `ghpc create` command and repeat the sequence.
+   > (`ghpc create ...`, `terraform init`, `terraform apply`).
+
+1. Upload VMD tarball
+
+   VMD is visualization software used by the remote desktop. While the software is
+   free the user must perform a registration before downloading it.
+
+   To download the software, complete the registration
+   [here](https://www.ks.uiuc.edu/Development/Download/download.cgi?PackageName=VMD)
+   and then download the tarball. The blueprint has been tested with the
+   `LINUX_64 OpenGL, CUDA, OptiX, OSPRay` version
+   (`vmd-1.9.3.bin.LINUXAMD64-CUDA8-OptiX4-OSPRay111p1.opengl.tar.gz`) but should
+   work with any compatible 1.9.x version.
+
+   Put the `tar.gz` file in the bucket with the name defined in
+   `bucket_name_software`. The virtual desktop will automatically look for this
+   file when booting up.
+
+1. Re-generate the deployment folder with updated IP addresses
+
+   Now that the file stores have been deployed we need to populate their ip
+   addresses so subsequent deployment groups can use them. Look up the IP
+   addresses for the apps and home file system on the filestore page in the
+   Google Cloud UI and populate them in the command below.
+
+   ```bash
+   ./ghpc create docs/videos/build-your-own-blueprint/hcls-blueprint.yaml -w \
+       --vars project_id=<project_id> \
+       --vars homefs_server_ip=<home_ip> \
+       --vars appsfs_server_ip=<apps_ip> \
+       --vars bucket_name_inputs=<input_bucket_name> \
+       --vars bucket_name_outputs=<output_bucket_name> \
+       --vars bucket_name_software=<software_bucket_name>
+   ```
+
+1. Deploy the `software_installation` group.
+
+   This deployment group will deploy a builder VM that will build gromacs and
+   save the compiled application on the apps filestore.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
+   `software_installation` group.
+
+   This may take several hours to run. You will know that installation has
+   completed successfully when the builder VM has shut itself down.
+
+1. Deploy the `cluster` group
+
+   This deployment group contains the Slurm cluster and the Chrome remote
+   desktop visualization node.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
+   `cluster` group.
+
+1. Set up Chrome Remote Desktop
+
+   - go to https://remotedesktop.google.com/headless and set up crd
+   - open terminal and run `vmd` on the command line
--- a/docs/videos/healthcare-and-life-sciences/README.md
+++ b/docs/videos/healthcare-and-life-sciences/README.md
@@ -3,9 +3,8 @@
 1. Clone the repo
 
    ```bash
-   git clone https://github.com/nick-stroud/hpc-toolkit.git
+   git clone https://github.com/GoogleCloudPlatform/hpc-toolkit.git
    cd hpc-toolkit
-   git checkout hcls_mega_blueprint
    ```
 
 1. Build the HPC Toolkit
@@ -30,18 +29,28 @@
        --vars bucket_name_software=<software_bucket_name>
    ```
 
-1. Run terraform commands for `enable_apis` group
+1. Deploy the `enable_apis` group
 
-   Call the provided `terraform init` and `terraform apply` commands for the
-   `enable_apis` group. This will ensure that all of the needed apis are
-   enabled before deploying the cluster.
+   Call the following terraform commands to deploy the `enable_apis` deployment
+   group.
+
+   ```bash
+   terraform -chdir=hcls-01/enable_apis init && terraform -chdir=hcls-01/enable_apis apply
+   ```
+
+   This will ensure that all of the needed apis are enabled before deploying the
+   cluster.
 
 1. Deploy the `setup` group
 
    This group will create a network and file systems to be used by the cluster.
 
-   Call the provided `terraform init` and `terraform apply` commands for the
-   `setup` group.
+   Call the following terraform commands to deploy the `setup` deployment
+   group.
+
+   ```bash
+   terraform -chdir=hcls-01/setup init && terraform -chdir=hcls-01/setup apply
+   ```
 
    > **Note**: You may have to tinker with bucket names to find an unused
    > namespace. If you get errors that the bucket already exists update the
@@ -51,7 +60,7 @@
 1. Upload VMD tarball
 
    VMD is visualization software used by the remote desktop. While the software is
-   free the user must perform a registration before downloading it.
+   free the user must register before downloading it.
 
    To download the software, complete the registration
    [here](https://www.ks.uiuc.edu/Development/Download/download.cgi?PackageName=VMD)
@@ -66,10 +75,17 @@
 
 1. Re-generate the deployment folder with updated IP addresses
 
-   Now that the file stores have been deployed we need to populate their ip
+   Now that the file stores have been deployed we need to populate their IP
    addresses so subsequent deployment groups can use them. Look up the IP
-   addresses for the apps and home file system on the filestore page in the
-   Google Cloud UI and populate them in the command below.
+   addresses for the `apps` and `home` file system and populate them in the
+   command below.
+
+   > **Note**: You can use the following command to list information about
+   > filestores, including their IP address.
+
+   ```bash
+   gcloud filestore instances list --project <project_id>
+   ```
 
    ```bash
    ./ghpc create docs/videos/build-your-own-blueprint/hcls-blueprint.yaml -w \
@@ -83,24 +99,45 @@
 
 1. Deploy the `software_installation` group.
 
-   This deployment group will deploy a builder VM that will build gromacs and
-   save the compiled application on the apps filestore.
+   This group will deploy a builder VM that will build gromacs and save the
+   compiled application on the apps filestore.
 
-   Call the provided `terraform init` and `terraform apply` commands for the
-   `software_installation` group.
+   Call the following terraform commands to deploy the `software_installation`
+   deployment group.
 
-   This may take several hours to run. You will know that installation has
-   completed successfully when the builder VM has shut itself down.
+   ```bash
+   terraform -chdir=hcls-01/software_installation init && \
+     terraform -chdir=hcls-01/software_installation apply
+   ```
+
+   This may take **several hours** to run. After the software installation is
+   complete the builder VM will automatically shut itself down. This allows you
+   to monitor the status of the builder VM to know when installation has
+   finished.
+
+   You can check the serial port 1 logs and the Spack logs
+   (`/var/log/spack.log`) to check status. If the builder VM never shuts down it
+   may be a sign that something went wrong with the software installation.
+
+   This builder VM can be deleted once the software installation has completed
+   successfully.
 
 1. Deploy the `cluster` group
 
    This deployment group contains the Slurm cluster and the Chrome remote
    desktop visualization node.
 
-   Call the provided `terraform init` and `terraform apply` commands for the
-   `cluster` group.
+   Call the following terraform commands to deploy the `cluster` deployment
+   group.
+
+   ```bash
+   terraform -chdir=hcls-01/cluster init && terraform -chdir=hcls-01/cluster apply
+   ```
 
 1. Set up Chrome Remote Desktop
 
-   - go to https://remotedesktop.google.com/headless and set up crd
-   - open terminal and run `vmd` on the command line
+   - Follow
+     [the instructions](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/develop/community/modules/remote-desktop/chrome-remote-desktop/README.md#setting-up-the-remote-desktop)
+     for setting up the Remote Desktop.
+   - Connect to the remote desktop, open a terminal in the remote session and
+     run `vmd` on the command line
--- a/docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
+++ b/docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
@@ -1,0 +1,356 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+blueprint_name: hcls-cluster
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: hcls-01
+  region: us-central1
+  zone: us-central1-c
+
+  network_name: hcls-cluster-net
+  subnetwork_name: primary-subnet
+  bucket_name_software: user-provided-software
+  bucket_name_inputs: input-bucket
+  bucket_name_outputs: output-bucket
+  use_deployment_name_in_bucket_name: false
+  homefs_server_ip:  # Set on command line
+  appsfs_server_ip:  # Set on command line
+
+deployment_groups:
+- group: enable_apis
+  modules:
+
+  ### Enable APIs ###
+
+  - id: services-api
+    source: community/modules/project/service-enablement
+    settings:
+      gcp_service_list:
+      - file.googleapis.com
+      - iam.googleapis.com
+      - pubsub.googleapis.com
+      - secretmanager.googleapis.com
+      - serviceusage.googleapis.com
+      - compute.googleapis.com
+      - stackdriver.googleapis.com
+
+- group: setup
+  modules:
+
+  ### Network ###
+
+  - id: network-create
+    source: modules/network/vpc
+
+  ### Resource Monitoring ###
+
+  - id: hpc-dash
+    source: modules/monitoring/dashboard
+
+  ### Storage ###
+
+  - id: homefs-create
+    source: modules/file-system/filestore
+    use: [network-create]
+    settings: {name: home}
+
+  - id: appsfs-create
+    source: modules/file-system/filestore
+    use: [network-create]
+    settings: {name: apps}
+
+  - id: bucket-software-create
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_software)
+
+  - id: bucket-input-create
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_inputs)
+
+  - id: bucket-output-create
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_outputs)
+
+- group: software_installation
+  modules:
+
+  - id: network-install
+    source: modules/network/pre-existing-vpc
+
+  - id: appsfs-install
+    source: modules/file-system/pre-existing-network-storage
+    settings:
+      server_ip: $(vars.appsfs_server_ip)
+      remote_mount: nfsshare
+      local_mount: /apps
+      fs_type: nfs
+
+  ### Software ###
+
+  - id: spack
+    source: community/modules/scripts/spack-install
+    settings:
+      install_dir: /apps/spack
+      configs:
+      - type: file
+        scope: defaults
+        content: |
+          config:
+            build_stage:
+              - /apps/spack/spack-stage
+      - type: file
+        scope: defaults
+        content: |
+          modules:
+            default:
+              tcl:
+                hash_length: 0
+                all:
+                  conflict:
+                    - '{name}'
+                projections:
+                  all: '{name}/{version}-{compiler.name}-{compiler.version}'
+      - type: file
+        scope: site
+        content: |
+          packages:
+            slurm:
+              externals:
+                - spec: slurm@21-08-8-2
+                  prefix: /usr/local
+              buildable: False
+      compilers: [gcc@9.3.0 target=x86_64]
+      environments:
+      - name: gromacs
+        type: file
+        content: |
+          spack:
+            definitions:
+            - compilers:
+              - gcc@9.3.0
+            - mpis:
+              - intel-mpi@2018.4.274
+            - mpi_packages:
+              - gromacs@2020.5
+            specs:
+            - matrix:
+              - - $mpis
+              - - $%compilers
+            - matrix:
+              - - $mpi_packages
+              - - $%compilers
+              - - $^mpis
+      # Un-comment and update mirror_url to install from spack cache
+      # spack_cache_url:
+      # - mirror_name: gcs_cache
+      #   mirror_url: gs://...
+
+  - id: spack-builder-startup  # TODO: Update example or make generic
+    source: modules/scripts/startup-script
+    settings:
+      runners:
+      - $(spack.install_spack_deps_runner)
+      - $(spack.install_spack_runner)
+      - type: shell
+        destination: setup_gromacs.sh
+        content: |
+          #!/bin/bash
+          source /apps/spack/share/spack/setup-env.sh
+          spack env activate gromacs
+          chmod -R a+rwX /apps/spack/var/spack/environments/gromacs
+          mkdir -p /apps/gromacs
+          chmod a+rwx /apps/gromacs
+          cd /apps/gromacs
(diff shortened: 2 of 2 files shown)


docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
+# Deployment Instructions
+
+1. Clone the repo
+
+   ```bash
+   git clone https://github.com/nick-stroud/hpc-toolkit.git
+   cd hpc-toolkit
+   git checkout hcls_mega_blueprint
```
cboneti, line 8: You probably want to update the lines above, since this is going to be part of the hpc toolkit repo.

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+       --vars bucket_name_inputs=<input_bucket_name> \
+       --vars bucket_name_outputs=<output_bucket_name> \
+       --vars bucket_name_software=<software_bucket_name>
+   ```
+
+1. Run terraform commands for `enable_apis` group
+
+   Call the provided `terraform init` and `terraform apply` commands for the
+   `enable_apis` group. This will ensure that all of the needed apis are
+   enabled before deploying the cluster.
```
cboneti, line 37: why not provide the commands? these should be known, right? 
AKA:
```
terraform -chdir=hcls-01/enable_apis init &&  terraform -chdir=hcls-01/enable_apis apply
```

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. Deploy the `setup` group
+
+   This group will create a network and file systems to be used by the cluster.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
```
cboneti, line 43: again, why not provide the command? 
```
terraform -chdir=hcls-01/setups init &&  terraform -chdir=hcls-01/setup apply
```

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+   > **Note**: You may have to tinker with bucket names to find an unused
+   > namespace. If you get errors that the bucket already exists update the
+   > bucket names in the `ghpc create` command and repeat the sequence.
+   > (`ghpc create ...`, `terraform init`, `terraform apply`).
+
+1. Upload VMD tarball
+
+   VMD is visualization software used by the remote desktop. While the software is
+   free the user must perform a registration before downloading it.
```
cboneti, line 54: s/the user must perform a registration/the user must register/ ?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  - id: hpc-dash
+    source: modules/monitoring/dashboard
+
+  ### Storage ###
+
+  - id: homefs-create
+    source: modules/file-system/filestore
+    use: [network-create]
+    settings: {name: home}
```
cboneti, line 70: This may create a lot of clashes. Every deployment of this blueprint will create a filestore called `home`. An alternative path is to, instead, set `filestore_share_name`. The instance will be randomply named based on deployment_name and suffix, and the shares will be clear in the UI. What do you think?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  - id: appsfs-create
+    source: modules/file-system/filestore
+    use: [network-create]
+    settings: {name: apps}
```
cboneti, line 75: Same comment as with home.

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+   `bucket_name_software`. The virtual desktop will automatically look for this
+   file when booting up.
+
+1. Re-generate the deployment folder with updated IP addresses
+
+   Now that the file stores have been deployed we need to populate their ip
+   addresses so subsequent deployment groups can use them. Look up the IP
+   addresses for the apps and home file system on the filestore page in the
+   Google Cloud UI and populate them in the command below.
+
```
cboneti, line 73: Do these instances have a specific name? can you also provide a gcloud command that gets that name easily? 
At the very least
```
gcloud filestore instances list
```

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+       --vars bucket_name_outputs=<output_bucket_name> \
+       --vars bucket_name_software=<software_bucket_name>
+   ```
+
+1. Deploy the `software_installation` group.
+
+   This deployment group will deploy a builder VM that will build gromacs and
+   save the compiled application on the apps filestore.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
```
cboneti, line 89: same as the above groups. Why not provide the command line?

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+   This may take several hours to run. You will know that installation has
+   completed successfully when the builder VM has shut itself down.
+
+1. Deploy the `cluster` group
+
+   This deployment group contains the Slurm cluster and the Chrome remote
+   desktop visualization node.
+
+   Call the provided `terraform init` and `terraform apply` commands for the
```
cboneti, line 100: same as the above groups. Why not provide the command line?

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+   `cluster` group.
+
+1. Set up Chrome Remote Desktop
+
+   - go to https://remotedesktop.google.com/headless and set up crd
```
cboneti, line 105: this seems a little incomplete. The instructions here have 7 steps. Maybe link them?
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/develop/community/modules/remote-desktop/chrome-remote-desktop/README.md#setting-up-the-remote-desktop

docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -0,0 +1,106 @@
...
+   - open terminal and run `vmd` on the command line
```
cboneti, line 106: Here you mean, connect to the remote desktop, then open a terminal there and then click vmd?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: hcls-01
+  region: us-central1
+  zone: us-central1-c
+
+  network_name: hcls-cluster-net
+  subnetwork_name: primary-subnet
+  bucket_name_software: user-provided-software
```
cboneti, line 27: I would leave bucket names similar to the `project_id`, `## Set bucket name here ##`, so you *force* people to provide them. Otherwise you may run into the risk of people not reading the instructions on the command line and then having name clashes.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+  bucket_name_inputs: input-bucket
+  bucket_name_outputs: output-bucket
+  use_deployment_name_in_bucket_name: false
```
cboneti, line 30: It is very tempting to set this to true just so you reduce the amount of name clashes you will have.  However, I haven't actually tried it so maybe the result is awkward

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+  - id: bucket-output-create
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_outputs)
+
+- group: software_installation
+  modules:
+
+  - id: network-install
+    source: modules/network/pre-existing-vpc
```
cboneti, line 96: This may look like magic to people. I would add a comment saying that this would be reuse the network created in the previous group, or something like that.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  - id: appsfs-install
+    source: modules/file-system/pre-existing-network-storage
+    settings:
+      server_ip: $(vars.appsfs_server_ip)
+      remote_mount: nfsshare
```
cboneti, line 102: If you change the filestore to have a different sharename, then please don't forget to change it here...

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+              - gromacs@2020.5
+            specs:
+            - matrix:
+              - - $mpis
+              - - $%compilers
+            - matrix:
+              - - $mpi_packages
+              - - $%compilers
+              - - $^mpis
+      # Un-comment and update mirror_url to install from spack cache
```
cboneti, line 161: I would not add these comments until the mirror actually works. If it does, then add it, not as comments. :)

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+    settings:
+      runners:
+      - $(spack.install_spack_deps_runner)
+      - $(spack.install_spack_runner)
+      - type: shell
+        destination: setup_gromacs.sh
+        content: |
+          #!/bin/bash
+          source /apps/spack/share/spack/setup-env.sh
+          spack env activate gromacs
```
cboneti, line 177: I am unsure what is the impact of having lines 176 and 177. It seems that setup-env.sh and spack env activate just load the environment and have no long lasting effect on the next commands.  
Since you are not actually running gromacs here, are these lines needed?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+          # Type can be 'pme' or 'rf'
+
+          source /apps/spack/share/spack/setup-env.sh
+          spack env activate gromacs
+
+          # Check that gmx_mpi exists
+          which gmx_mpi
+          cd $SLURM_SUBMIT_DIR
+          cp /apps/gromacs/water-cut1.0_GMX50_bare/1536/* .
+          scontrol show hostnames ${SLURM_JOB_NODELIST} > hostfile
```
cboneti, line 203: This sounds funny, why do you need to populate a hostfile? I thought slurm would do that for you. What am I missing?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+          gmx_mpi grompp -f pme.mdp -c conf.gro -p topol.top -o input.tpr
+          mpirun -n 60 -hostfile hostfile -ppn 30 gmx_mpi mdrun -notunepme -dlb yes -v -resethway -noconfout -nsteps 4000 -s input.tpr
```
cboneti, line 205: Slurm should be able to populate the hostfile for you, right?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+      - type: shell  # TODO: consider making bool option
```
cboneti, line 207: I would remove the TODO comment.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+                  touch /etc/block_auto_shutdown
+                  shutdown -h +1
+          fi
+
+  - id: spack-builder
+    source: modules/compute/vm-instance
+    use: [network-install, appsfs-install, spack-builder-startup]
+    settings:
+      name_prefix: spack-builder
+      machine_type: c2-standard-16  # TODO Consider bumping up, left low for quota
```
cboneti, line 221: In my experience, Spack will not go any faster with a larger machine. You can validate.
Also, I would remove the TODO comment.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+  modules:
+
+  - id: network1
+    source: modules/network/pre-existing-vpc
+
+  - id: homefs
+    source: modules/file-system/pre-existing-network-storage
+    settings:
+      server_ip: $(vars.homefs_server_ip)
+      remote_mount: nfsshare
```
cboneti, line 233: same comment about the remote_mount if you choose to do that.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+      local_mount: /home
+      fs_type: nfs
+
+  - id: appsfs
+    source: modules/file-system/pre-existing-network-storage
+    settings:
+      server_ip: $(vars.appsfs_server_ip)
+      remote_mount: nfsshare
```
cboneti, line 241: same comment about the remote_mount if you choose to do that.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  - id: bucket-software
+    source: modules/file-system/pre-existing-network-storage
+    settings:
+      remote_mount: $(vars.bucket_name_software)
+      local_mount: /user_provided_software
+      fs_type: gcsfuse
+      mount_options: defaults,_netdev,implicit_dirs
+
+  - id: bucket-input
```
cboneti, line 253: You downloaded the data to the /apps storage. Why adding a bucket-input here? 
I understand you may want to show you could have input and output buckets, but since you are actually not using it, isn't it more confusing?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+      install_nvidia_driver: true
+      startup_script: |
+        sudo apt-get update
+        sudo apt-get -y upgrade
+        find /user_provided_software -name vmd-1.9.*.bin.LINUXAMD64*.tar.gz -exec tar xvzf  '{}' -C . \;
+        cd vmd-1.9.*/
+        ./configure
+        cd src/
+        sudo make install
+        cd ../..
```
cboneti, line 291: `cd ../..` is technically unnecessary since it is the last command of the runner.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  ### Cluster ###
+
+  - id: debug_node_group
```
cboneti, line 295: in the spirit of simplifying this blueprint, do you still need a debug partition / node group?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+    use:
+    - network1
+    - debug_partition
+    - compute_partition
+    - homefs
+    - appsfs
+    - bucket-input
+    - bucket-output
+    settings:
+      disable_controller_public_ips: false
```
cboneti, line 343: Here you enable public IP addresses for the controller and login nodes, but not for the compute nodes. Why?

cboneti (changes requested):
Very good. I left a few comments, most of them suggestions.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,356 @@
...
+
+  - id: slurm_login
+    source: community/modules/scheduler/schedmd-slurm-gcp-v5-login
+    use:
+    - network1
+    - slurm_controller
+    - homefs
+    - appsfs
+    - bucket-input
+    - bucket-output
```
nick-stroud, line 353: Delete login node fs. They are inherited from the controller and are unused.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: This would work much better with IGC. Right now each group has to set up pre-existing-network-storage based on the bucket name. So it is important that it is defined globally and is static. If you include the deployment name then we also have to include it later when we look it up.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I think I see. I originally did this because I could not tell them apart when looking up the ip address. I will try what you suggest.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Removed.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Removed.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I believe this was a bad artifact from some testing. I have removed and will validate.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Removed.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: This is an artifact of our gromacs tutorial. The sbatch was written by @douglasjacobsen. Anything to add?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I have had similar experience when building from cache. Not sure about a true build from source. @douglasjacobsen do you have any input here?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
douglasjacobsen: Probably older versions of slurm. Previously slurm-gcp didn't populate this and you'd have to make a hostfile for it to work correctly with Intel MPI.

It's worth a test (if it's not too much effort) to remove this (and passing in the hostfile below). However, the lines do not really cause any problems (and more generally show people how to get the hostfile if they need it).

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
douglasjacobsen: True build from source might end up being ram limited (unless you're using PD). It's possible to speed up a source build, using parallel builders, but also probably outside the scope of this blueprint.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Added. Leaving it a general lookup since i am going to look at letting the names change.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Done

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Done

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Updated.

docs/videos/healthcare-and-life-sciences/README.md
nick-stroud: Yes. Updated.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: There were a bad artifact. They have been deleted.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I have made this change.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: It works fine without hostname. I have removed it for brevity.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I have reduced to a c2-standard-4.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: We intended to use the input bucket for the demo. I have updated the scripts to stage data to the input bucket.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,330 @@
...
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_software)
+
+  - id: bucket-input-create
+    source: ./community/modules/file-system/cloud-storage-bucket
+    settings:
+      name_prefix: $(vars.bucket_name_inputs)
+
+  - id: bucket-output-create
```
nick-stroud, line 87: @cboneti, we are leaning towards removing the outputs bucket. The demo doesn't use it and it seems likely that an output bucket would be something generated outside of a blueprint on a 1 per user basis with specific permissions.
Thoughts?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
cboneti: makes sense. Thanks!

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
cboneti: funny, I had that thinking for the input, but not for the output. See, the input source (bucket) is typically something you don't change as much, I would say you even do several simulations on the same exact input (with different parameters), whereas you want outputs to be passed on. The reality is that it is likely that both input and output buckets will be reused a lot. 

To simplify, why not create a single bucket with inputs and outputs in it?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
cboneti: Sorry, I was not clear. I have tried building from source with 2, 4, 8, 16, 32, 64 ... cores. Anything up to 16 gives benefits, beyond that point, it's not any faster. Spack only occupies 16 cores (at least the last time I have looked into this). Therefore, I would say 16 is the fastest machine you should pick. I would pick 16 and not try to do any bigger.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,330 @@
...
+    source: modules/scripts/startup-script
+    settings:
+      runners:
+      - $(spack.install_spack_deps_runner)
+      - $(spack.install_spack_runner)
+      - type: shell
+        destination: data_staging.sh
+        content: |
+          #!/bin/bash
+          wget --no-verbose -P /data_input/protien_data_bank/ https://files.rcsb.org/download/1FJS.pdb
```
cboneti, line 180: s/protien_data_bank/protein_data_bank/ (typo)

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,330 @@
...
+          wget --no-verbose -P /data_input/gromacs_inputs https://ftp.gromacs.org/pub/benchmarks/water_GMX50_bare.tar.gz && \
```
cboneti, line 181: this is likely to go *significantly* faster if you wget to /tmp or even local PD and then extract to the gcs bucket.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
cboneti: also I do not see this pdb being used anywhere. Is it?

cboneti (changes requested):
a few more typos and comments

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: This definitely deserves more explanation. The protein is the basis for a much longer example that will be demoed. The sbatch for this workflow will be submitted in a subsequent PR. The decision as of now is to not include this other sbatch as part of the blueprint and instead have it live alongside the blueprint in a separate file that is referenced by a tutorial.

We discussed removing the water example but the thought was that it is useful to have a simple example included in the blueprint. I intend do document all of this as we add meat to the tutorial. With this PR I am really just trying to get a basis from which those involved can collaborate and iterate. LMK if you would prefer to have more complete documentation in this PR.

PS. Will also correct the typo.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Back to c2-standard-16.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I think there are different use cases. There can be a read only inputs library that does not change. I think this makes sense to have as a bucket so it is easily visible and jobs don't have to rely on the public internet.

Then I think there is a valid use case for the output bucket as well (as you describe). I think the concern here is that the outputs bucket would probably not be a shared resource. A scientist would be cautious about writing data somewhere that could be overwritten by anyone else using the cluster. 

My proposal is to leave them both in for now. Write a doc section about this and get consensus before modifying further. I think I am in the camp of it is useful to have the output bucket to show the pattern even if in real life their might be a bucket per user pattern that is not captured in the blueprint.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: The typo is fixed.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I have updated to do as you suggest.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -0,0 +1,331 @@
...
+      - $(spack.install_spack_deps_runner)
+      - $(spack.install_spack_runner)
+      - type: shell
+        destination: data_staging.sh
+        content: |
+          #!/bin/bash
+          wget --no-verbose -P /data_input/protein_data_bank/ https://files.rcsb.org/download/1FJS.pdb
+          wget --no-verbose -P /tmp/ https://ftp.gromacs.org/pub/benchmarks/water_GMX50_bare.tar.gz && \
+            mkdir -p /data_input/gromacs_inputs/ && \
+            tar xzf /tmp/water_GMX50_bare.tar.gz -C $_ && \
```
cboneti, line 183: this is cool, but also looks like magic, why not just type the path here as well?

cboneti (approved):
I all looks good, I have one nit on line 183 that may or may not change.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I removed the magic.

PR #880, opened 2023-02-02
Add instructions to ssh to VM created by vm-instance

nick-stroud:
Ease of use improvement. Adds SSH instructions output to vm-instance.

### Submission Checklist

  install`)


modules/compute/vm-instance/outputs.tf
```diff
@@ -28,3 +28,13 @@ output "internal_ip" {
   description = "Internal IP of the instances"
   value       = google_compute_instance.compute_vm[*].network_interface[0].network_ip
 }
+
+output "instructions" {
+  description = "Instructions on how to SSH into the created VM"
+  value       = <<-EOT
+  Use the following commands to SSH into the first VM created:
+    gcloud compute ssh --zone ${var.zone} ${google_compute_instance.compute_vm[0].name} --project ${var.project_id}
+  Or with tunnel through iap
```
tpdownes, line 37: - use "IAP" or "Identity-Aware Proxy"
- I would present this as less of an "or" and give the reason: something like "if the machine is not accessible from the public internet, an SSH tunnel can be created through IAP" (you might consider something shorter)

tpdownes (changes requested):
Give more of the reason to tunnel through IAP without promising that it will work (because of missing IAM roles, or firewall rules)

modules/compute/vm-instance/outputs.tf
nick-stroud: I have updated the wording to this effect.

modules/compute/vm-instance/README.md
```diff
@@ -209,7 +209,7 @@ limitations under the License.
 | Name | Description |
 |------|-------------|
 | <a name="output_external_ip"></a> [external\_ip](#output\_external\_ip) | External IP of the instances (if enabled) |
-| <a name="output_instructions"></a> [instructions](#output\_instructions) | Instructions on how to SSH into the created VM |
+| <a name="output_instructions"></a> [instructions](#output\_instructions) | Instructions on how to SSH into the created VM. Commands may fail depending on VM configuraiton and IAM permissions. |
```
tpdownes, line 212: Fix misspelling of "configuraiton".

tpdownes (changes requested):
Fix the typo and feel free to squash commit history.

modules/compute/vm-instance/README.md
nick-stroud: Done.

tpdownes (approved):

PR #891, opened 2023-02-08
Update HCLS blueprint examples to run Gromacs w/ GPUs

nick-stroud:
This change:
- Updates Factor Xa example to run on GPUs
- Updates Spack to build GPU enable gromacs
- Adds water benchmark that can be run on GPUs

### Submission Checklist

  install`)

### DIFF
--- a/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/README.md
+++ b/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/README.md
@@ -1,7 +1,8 @@
 # Factor Xa Example
 
 This example demonstrates a real life case of simulating the Factor Xa protein
-in water. It uses the HCLS blueprint to run a multi-step GPU enable gromacs run.
+in water. It uses the HCLS blueprint to run a multi-step GPU enabled gromacs
+run.
 
 This example has been adapted with changes from tutorials by:
 
--- a/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/README.md
+++ b/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/README.md
@@ -1,7 +1,7 @@
 # Factor Xa Example
 
 This example demonstrates a real life case of simulating the Factor Xa protein
-in water.
+in water. It uses the HCLS blueprint to run a multi-step GPU enable gromacs run.
 
 This example has been adapted with changes from tutorials by:
 
--- a/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
+++ b/docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
@@ -7,9 +7,10 @@
 # Creative Commons Attribution 4.0 International (CC BY 4.0) License.
 # See <https://creativecommons.org/licenses/by/4.0/>.
 
-#SBATCH -N 2
-#SBATCH --ntasks-per-node 30
-#SBATCH --partition compute
+#SBATCH -N 1
+#SBATCH --ntasks-per-node 1
+#SBATCH --partition gpu
+#SBATCH --gpus 1
 
 PDB_FILE=1FJS.pdb
 PROTEIN="${PDB_FILE%.*}"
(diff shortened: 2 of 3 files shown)


docs/videos/healthcare-and-life-sciences/Factor-Xa-example/README.md
```diff
@@ -1,7 +1,7 @@
 # Factor Xa Example
 
 This example demonstrates a real life case of simulating the Factor Xa protein
-in water.
+in water. It uses the HCLS blueprint to run a multi-step GPU enable gromacs run.
```
nick-stroud, line 4: s/enable/enabled

docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
```diff
@@ -7,9 +7,10 @@
 # Creative Commons Attribution 4.0 International (CC BY 4.0) License.
 # See <https://creativecommons.org/licenses/by/4.0/>.
 
-#SBATCH -N 2
-#SBATCH --ntasks-per-node 30
-#SBATCH --partition compute
+#SBATCH -N 1
+#SBATCH --ntasks-per-node 1
```
cboneti, line 11: so it doesn't really use the host CPUs?

docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
```diff
@@ -7,9 +7,10 @@
 # Creative Commons Attribution 4.0 International (CC BY 4.0) License.
 # See <https://creativecommons.org/licenses/by/4.0/>.
 
-#SBATCH -N 2
```
cboneti, line 10: Here I am wondering if there would not be value to keep both submit scripts, cpu and gpu, perhaps naming them submit_cpu, submit_gpu? but then, I suppose it would be hard to reconcile the blueprint to support both, right?

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -145,34 +145,51 @@ deployment_groups:
...
         content: |
           spack:
             definitions:
-            - compilers:
-              - gcc@9.3.0
-            - mpis:
-              - intel-mpi@2018.4.274
-            - mpi_packages:
-              - gromacs@2020.5
+              - compilers:
```
cboneti, line 155: here it seems you have indented the whole thing two spaces to the right, was that intentional?

cboneti (approved):
I left a few comments, but otherwise, ok to submit. Thanks!

docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
nick-stroud: I think the main reason to only have 1 is to reduce complexity of maintenance. From preliminary testing, the simulation is cheaper and faster on GPUs. I did leave both cpu and gpu for the water script as I think it is important to at least have 1 example showing both.

Also, since this is git if we really want to dig out the cpu solution then we can don't lose that option.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: I believe you are right. I think both are valid yaml:

```
# 1 
definitions:
  - compilers:
      - gcc@11.3.0
  - cudas:
      - cuda@11.8.0
  - cuda_mpis:
      - openmpi@4.1.4+cuda
  - mpi_cuda_packages:
      - gromacs@2022.3+cuda+mpi

# 2
definitions:
- compilers:
    - gcc@11.3.0
- cudas:
    - cuda@11.8.0
- cuda_mpis:
    - openmpi@4.1.4+cuda
- mpi_cuda_packages:
    - gromacs@2022.3+cuda+mpi
```
But # 2 matches what we have done elsewhere, so I will update.

docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
nick-stroud: I will pass to @danielahlin to answer.

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
nick-stroud: Done.

docs/videos/healthcare-and-life-sciences/Factor-Xa-example/submit.sh
danielahlin: As much as is currently possible is moved to the GPUs but the host CPUs are still used.

PR #897, opened 2023-02-10
Ensure Open Front End dependabot updates target develop

tpdownes:
The choice to restrict this change to the Open Front End code is deliberate as I think it's the only instance of Python code with a properly curated requirements.txt file. There are 2 other instances:

- /tools/cloud-build/requirements.txt
- /docs/hybrid-slurm-cluster/requirements.txt

The cloud-build requirements.txt contains no version constraints at all while the Slurm docs appear to be more of a tutorial without any integration testing. So, for now, I think we are better off configuring what we test.

### Submission Checklist

  install`)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,11 @@ updates:
     time: "03:00"
     timezone: America/Los_Angeles
   target-branch: develop
+- package-ecosystem: pip
+  directory: /community/front-end/ofe/
+  schedule:
+    interval: weekly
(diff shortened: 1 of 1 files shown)


tpdownes:
@heyealex the force push updates the path from `/community/front-end` to `/community/front-end/ofe/` to (A) properly restrict to Open Front End code and (B) include the trailing slash which I think might be more correct.

heyealex (approved):

PR #898, opened 2023-02-10
Fix: Always generate Batch instance template to avoid known at apply time error

nick-stroud:
The conditional creation of the Batch instance template causes "not known at apply time" errors. This change avoids those errors by always creating an instance template even when the user provides an instance template. In the case that the user provides an instance template, the created template will be unused. While this may not be optimal, instance templates do not have a cost to the end user and it is the best solution given the limitations of terraform.

### Submission Checklist

  install`)

### DIFF
--- a/modules/scheduler/batch-job-template/main.tf
+++ b/modules/scheduler/batch-job-template/main.tf
@@ -15,7 +15,7 @@
  */
 
 locals {
-  instance_template = var.instance_template != null ? var.instance_template : one(module.instance_template[*].self_link)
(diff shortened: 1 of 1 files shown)


tpdownes (approved):
I think we both share a dissatisfaction with the approach here, but I think it's good given the current limitations of Terraform and of the Toolkit to manage multiple deployment groups.

For the record, the Terraform-recommended pattern is to always supply the instance template externally and avoid conditional creation inside modules. But, that is not always realistic.

PR #908, opened 2023-02-14
Address terraform_unused_required_providers errors

tpdownes:
- remove unused providers in Chrome Remote Desktop module
- impose minimum Terraform version of 0.12.31 because versions below that are known to have a built-in GPG key that was revoked due to a security incident (HCSEC-2021-12)

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/remote-desktop/chrome-remote-desktop/versions.tf
+++ b/community/modules/remote-desktop/chrome-remote-desktop/versions.tf
@@ -26,5 +26,5 @@ terraform {
       version = ">= 4.12"
     }
   }
-  required_version = ">= 0.12.20"
+  required_version = ">= 0.12.31"
 }
--- a/community/modules/remote-desktop/chrome-remote-desktop/versions.tf
+++ b/community/modules/remote-desktop/chrome-remote-desktop/versions.tf
@@ -15,16 +15,5 @@
 */
 
 terraform {
-  required_providers {
-    google = {
-      source  = "hashicorp/google"
-      version = ">= 3.83"
-    }
-
-    google-beta = {
-      source  = "hashicorp/google-beta"
-      version = ">= 4.12"
-    }
-  }
   required_version = ">= 0.12.31"
 }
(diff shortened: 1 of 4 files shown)


community/modules/remote-desktop/chrome-remote-desktop/versions.tf
```diff
@@ -15,16 +15,5 @@
...
-      source  = "hashicorp/google"
-      version = ">= 3.83"
-    }
-
-    google-beta = {
-      source  = "hashicorp/google-beta"
-      version = ">= 4.12"
-    }
-  }
-  required_version = ">= 0.12.20"
```
nick-stroud, line 29: This same version is also used in the OFE. Should it be addressed there also?
[ref](https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/develop/community/front-end/ofe/infrastructure_files/vpc_tf/GCP#requirement_terraform)

tpdownes: Yes, at low-medium priority. In practice it will replace a loud declaration of a GPG key failure (warning about a man-in-the-middle attack?) with a "this version doesn't work at all" failure. But there probably aren't many people out there with terraform 0.12 in the first place by now.

tpdownes: P.S. This required_version line has no impact on the TFLint errors. Arguably upgrading both from 0.12.20 should be a separate PR. LMK what you prefer.

nick-stroud: Basically, since we know there is a problem with this version we should take some action while we have context. I will leave it to you if that action should be a separate PR with both changes or a task to be discussed.

nick-stroud (approved):

community/modules/remote-desktop/chrome-remote-desktop/versions.tf
tpdownes: Resolved by re-working the branch to include a 2nd commit dedicated to the version issue for both modules.

PR #918, opened 2023-02-15
Fix `cmd/root_test.go` test runs from linked Git worktrees.

mr0re1:
Fix `cmd/root_test.go` test runs from linked Git worktrees.
```
$git worktree add wt && cd wt
$make tests
...
PANIC: root_test.go:112: MySuite.TestCheckGitHashMismatch
```

- Don't use real repo as a subject for inspection, use test ones;
- Use `Check` instead of `Assert`.

**Testing**:
```
$ cd ~/wp/hpc-toolkit && git checkout wt
$ make tests
...
$ git checkout develop && git worktree add wt && cd wt
$ make tests
...
```

**NOTE**:
While running `make tests` from linked worktree, incorrect warning is shown:
```
WARNING: ghpc binary was built from a different commit (wt/7812ced) than the current git branch in ~/wp/hpc-toolkit (develop/edb0ce3). You can rebuild the binary by running 'make'
```
This issue should be addressed in separate PR.

### Submission Checklist

  install`)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -145,7 +145,7 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 			repo, err = git.PlainOpen(dir)
 
 			if err == nil && isHpcToolkitRepo(*repo) {
-				return
+				return repo, dir, nil
 			}
 		}
 	}
@@ -159,10 +159,10 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 
 	repo, err = git.PlainOpen(dir)
 	if err != nil {
-		return
+		return nil, "", err
 	}
 	if isHpcToolkitRepo(*repo) {
-		return
+		return repo, dir, nil
 	}
 	return nil, "", errors.New("unable to find the hpc-toolkit git repo in the file system")
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -164,7 +164,7 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 	if isHpcToolkitRepo(*repo) {
 		return repo, dir, nil
 	}
-	return nil, "", errors.New("unable to find the hpc-toolkit git repo in the file system")
+	return nil, "", errors.New("ghpc executable found in a git repo other than the hpc-toolkit git repo")
 }
 
 // isHpcToolkitRepo will verify that the found git repository has a commit with
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -18,6 +18,7 @@ limitations under the License.
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -157,10 +158,13 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 	dir = filepath.Dir(e)
 
 	repo, err = git.PlainOpen(dir)
-	if err == nil && isHpcToolkitRepo(*repo) {
-		return repo, dir, err
+	if err != nil {
+		return
 	}
-	return nil, "", err
+	if isHpcToolkitRepo(*repo) {
+		return
+	}
+	return nil, "", errors.New("Not the hpc-toolkit repo")
 }
 
 // isHpcToolkitRepo will verify that the found git repository has a commit with
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -164,7 +164,7 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 	if isHpcToolkitRepo(*repo) {
 		return
 	}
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -157,10 +158,13 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
...
-	if err == nil && isHpcToolkitRepo(*repo) {
-		return repo, dir, err
+	if err != nil {
+		return
+	}
+	if isHpcToolkitRepo(*repo) {
+		return
 	}
-	return nil, "", err
+	return nil, "", errors.New("Not the hpc-toolkit repo")
```
cboneti, line 167: This message may not be very clear to an user. Do we not expect it to be ever reached? otherwise, could we try to have something more clear, like, `Unable to find the hpc-toolkit git repo in the file system.` or something of the sort.

mr0re1: Addressed

cmd/root_test.go
```diff
@@ -19,157 +19,196 @@ package cmd
...
+	if err = os.Chdir(path); err != nil {
+		c.Fatal(err)
+	}
+
 	head, _ := repo.Head()
 	hash := head.Hash().String()
 	branch := head.Name().Short()
 
-	// verify current working directory git hash against hash of v1.0.0
-	mockInjectedGitVariables()
```
nick-stroud, line 122: nit: Technically we are now missing this case that was previously covered. 

The case being: 
State: The binary was built from a hash that is actually in the git history
Result: Identify mismatch

This is partially covered by the "Baked commit hash doesn't match" case which is testing
State: The binary is built from a hash not in the git repo (random)
Result: Identify mismatch

I don't know if I really care too much about this case as we know we don't search through the git history, maybe consider how hard it would be to add it. You would have to know a valid commit in the git history that is not the initial commit.

cmd/root_test.go
```diff
@@ -19,157 +19,196 @@ package cmd
...
+		c.Check(mismatch, Equals, false)
+	}
 
-	// verify current working directory git hash against random initial hash
-	mockInjectedGitVariables()
-	GitInitialHash = randomGitHash
-	mismatch, b, h, dir = checkGitHashMismatch()
-	c.Assert(mismatch, Equals, false)
-	c.Assert(b, Equals, "")
-	c.Assert(h, Equals, "")
```
nick-stroud, line 135: nit: Before we verified that in the non-mismatch case we don't emit any spurious data. There is nothing binding us to that contract but I kind of like it in these named return functions which make it easy for random data to slip out.

cmd/root_test.go
```diff
@@ -19,157 +19,196 @@ package cmd
...
 
-	// verify current working directory git hash against hash of v1.0.0
-	mockInjectedGitVariables()
-	mismatch, b, h, dir := checkGitHashMismatch()
-	c.Assert(mismatch, Equals, true)
-	c.Assert(dir, Equals, workDir)
-	c.Assert(b, Equals, branch)
-	c.Assert(h, Equals, hash)
+	{ // Matches
+		GitCommitHash = hash
```
nick-stroud, line 145: For the first 3 of the 4 cases outlined here logic is depending on `GitCommitHash` and `GitInitialHash` as opposed to just setting the one needed for the case I think the tests would be more readable if you set both at the top of each case. So for example on the "Matches" case you would start with:
```
GitCommitHash = hash
GitInitialHash = init.String()
```

cmd/root_test.go
nick-stroud: I see this modality several places, I could be convinced otherwise. Thoughts?

cmd/root_test.go
mr0re1: Added mismatch-case with
```
GitInitialHash = init.String()
GitCommitHash = init.String()
```

cmd/root.go
```diff
@@ -157,10 +158,13 @@ func hpcToolkitRepo() (repo *git.Repository, dir string, err error) {
 	dir = filepath.Dir(e)
 
 	repo, err = git.PlainOpen(dir)
-	if err == nil && isHpcToolkitRepo(*repo) {
-		return repo, dir, err
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
nick-stroud: I see, I missed that.
In that case I would suggest a different error message that is specific to this case. Maybe:

 >_"ghpc executable found in a git repo other than the hpc-toolkit git repo"_

I would guess @cboneti's suggestion was with the assumption that this was a summarizing error message that was meant to be seen by the user., not a specific case that had no system error.

nick-stroud (approved):

cmd/root.go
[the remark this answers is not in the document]
mr0re1: Thank you, I was concerned about message being too wordy/specific, but you have found the right words )

PR #919, opened 2023-02-15
Set minimal scopes for HTCondor IDTOKENs

tpdownes:
The current HTCondor modules creates IDTOKENs with full authorizations on each node within the HTCondor pool. An IDTOKEN is a [JWT](jwt.io) implementing OpenID Connect Standards much like Google Cloud authorizations. This PR reduces the scopes of the token on each node to the values recommended by the HTCondor team.

This PR continues to share the POOL signing key with each node (more or less the "root password" for the pool). It is anticipated that future work will eliminate the dependency on a local copy of the POOL signing key on each node. In the meantime, using the minimally scoped tokens for authentication will be a better practice and hopefully detect any problems.

We will not be able to share the POOL signing key to Windows execute points as we do for Linux execute points. So this change will help us better support Windows on equal footing with Linux.

Additionally, use some top-level variables to set destination filenames, with understanding that the lexical order matters when HTCondor reads its configuration.

- https://github.com/htcondor/htcondor/blob/8bcd2442756564ebbfdb6955fb16483806fda236/build/docker/services/README.md#using-the-execute-node-container
- https://github.com/htcondor/htcondor/blob/8bcd2442756564ebbfdb6955fb16483806fda236/build/docker/services/README.md#using-the-submit-access-point-container

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
+++ b/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
@@ -14,8 +14,16 @@
 
 ---
 - name: Configure HTCondor Role
-  become: true
   hosts: localhost
+  become: true
+  vars:
+    condor_config_root: /etc/condor
+    role_file: 00-role
+    pool_file: 01-pool
+    cm_config_file: 02-central-manager
+    cm_ha_config_file: 02-central-manager-high-availability
+    schedd_config_file: 02-schedd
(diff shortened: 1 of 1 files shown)


tpdownes:
I found a minor linting problem locally (doesn't seem to be caught by PR validation) and force-pushed. I also triggered build 4 to test the solution.

heyealex (approved):

PR #933, opened 2023-02-17
Add retries to cloud ops install

heyealex:
On some images that install software in parallel with the startup script, the apt lock can prevent successful completion of the startup script. Given the lower version of apt in debian 10 and the minimal tooling that comes default with the image, a simple retry loop was the best option.

Other methods tried:
* lsof -> package not available on the base image
* fuser -> package not available on the base image
* `-o DPkg::Lock::Timeout=60` in the apt install command: apt and apt-get are both too old to support this option
* Adding Lock timeout to apt config: same issue as above

### Submission Checklist

  install`)


tpdownes (approved):
It appears that it's easy for the syslog to end up having lots of repeated invocations of failed apt commands. That's a bit annoying but it's better than the alternative of a silent apt failure.

PR #941, opened 2023-02-22
Refactor useModule

tpdownes:
Refactor useModule so that manipulation of list Settings (which require explicit initilization and wrapping function) is handled by a special function with its own unit tests. Additional minor changes to variable names for clarity and avoid use of custom input types in favor of defined structs.

Address offline discussion from #878.

Strictly speaking, cyclomatic complexity for the useModule function increases from 7 to 8, however the clarity of the function is overall improved due to simpler if/else indentation blocks.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -42,6 +42,8 @@ const (
 )
 
 var errorMessages = map[string]string{
+	// general
+	"appendToNonList": "cannot append to a setting whose type is not a list",
 	// config
 	"fileLoadError":      "failed to read the input yaml",
 	"yamlUnmarshalError": "failed to parse the blueprint in %s, check YAML syntax for errors, err=%w",
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -180,20 +180,22 @@ func (mod *Module) addListValue(settingName string, value string) error {
 		errorMessages["appendToNonList"], mod.ID, settingName)
 }
 
-// This function matches input variables in a "using" module to output values
+// useModule matches input variables in a "using" module to output values
 // from a "used" module. It may be used iteratively to successively apply used
 // modules in order of precedence. New input variables are added to the using
 // module as Toolkit variable references (in same format as a blueprint). If
 // the input variable already has a setting, it is ignored, unless the value is
 // a list, in which case output values are appended and flattened using HCL.
-// mod: "using" module as defined above
-// useMod: "used" module as defined above
-// useModGroupID: deployment group ID to which useMod belongs
-// modInputs: input variables as defined by the using module code
-// useOutputs: output values as defined by the used module code
-// settingsToIgnore: a list of module settings not to modify for any reason;
 //
-//	typical usage will be to leave explicit blueprint settings unmodified
+//	mod: "using" module as defined above
+//	useMod: "used" module as defined above
+//	useModGroupID: deployment group ID to which useMod belongs
+//	modInputs: input variables as defined by the using module code
+//	useOutputs: output values as defined by the used module code
+//	settingsToIgnore: a list of module settings not to modify for any reason;
+//	 typical usage will be to leave explicit blueprint settings unmodified
+//
+// returns: a list of variable names that were used during this function call
 func useModule(
 	mod *Module,
 	useMod Module,
@@ -218,16 +220,19 @@ func useModule(
 			continue
 		}
 
-		_, setByUse := mod.Settings[settingName]
-		modVarName := getModuleVarName(useModGroupID, useMod.ID, settingName)
-		isInputList := strings.HasPrefix(inputType, "list")
-
-		if !setByUse && !isInputList {
-			mod.Settings[settingName] = modVarName
-			usedVars = append(usedVars, settingName)
+		// skip settings that are not of list type, but already have a value
+		// these were probably added by a previous call to this function
+		_, alreadySet := mod.Settings[settingName]
+		isList := strings.HasPrefix(inputType, "list")
+		if alreadySet && !isList {
+			continue
 		}
 
-		if isInputList {
+		modVarName := getModuleVarName(useModGroupID, useMod.ID, settingName)
+		if !isList {
+			mod.Settings[settingName] = modVarName
+			usedVars = append(usedVars, settingName)
+		} else {
 			if err := mod.addListValue(settingName, modVarName); err != nil {
 				return nil, err
 			}
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -162,57 +162,79 @@ func getModuleInputMap(inputs []modulereader.VarInfo) map[string]string {
 	return modInputs
 }
 
+// initialize a Toolkit setting that corresponds to a module input of type list
+// create new list if unset, append if already set, error if value not a list
+func (mod *Module) addListValue(settingName string, value string) error {
+	_, found := mod.Settings[settingName]
+	if !found {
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -42,6 +42,8 @@ const (
 )
 
 var errorMessages = map[string]string{
+	// general
+	"appendToNonList": "cannot append to a setting whose type is not a list",
```
mr0re1, line 46: Is there a reason to use lookup-map for error messages? There are few problems with this approach:
- Unused entries can go unnoticed;
- Missing entry can cause runtime error;
- Extra few characters of code, worse readability.

Can we use string constants instead?

tpdownes: I think the intent is to eventually simplify the range of error messages and fail with more custom error types that implement the error interface (`func Error() string`). In particular, I do not like the unit tests "grepping" for information.

For example:

https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/dda341f3e8af56790ca59323364e9519d79448ec/pkg/config/config.go#L686-L693

If you'd like to capture that as a task, it would be a good one. I think it ought to have a small design doc with a well documented approach and example PR replacing one or more of the existing messages.

mr0re1: > I think the intent is to eventually simplify the range of error messages and fail with more custom error types that implement the error interface 
>  I do not like the unit tests "grepping" for information.

I think is this is orthogonal to what I've meant. The only change I suggest is to not add entry to this map, instead declare `const errMsgAppendToNonList string = "....."` and use it. 

>  I think it ought to have a small design doc with a well documented approach

I think this changes is too trivial for a formal design doc.

I understand that you may hesitate to disrupt the "local style" though it's problematic. Feel free to resolve this comment.

pkg/config/expand.go
```diff
@@ -162,57 +162,79 @@ func getModuleInputMap(inputs []modulereader.VarInfo) map[string]string {
...
-		// if the setting has a value, then it was set by "use" because we have
-		// skipped anything explicitly in the blueprint. If the setting is a
-		// list of any type, then we want to concatenate these values in order
-		// they were used; otherwise, the setting should remain set to the
-		// existing used value
 		_, setByUse := mod.Settings[settingName]
 		modVarName := getModuleVarName(useModGroupID, useMod.ID, settingName)
 		isInputList := strings.HasPrefix(inputType, "list")
+
+		if !setByUse && !isInputList {
```
nick-stroud, line 225: The two negative bools make it hard for me to comprehend the case here (maybe just me). Possible suggestion:

```
                shouldSkipAsPreviouslySetByUse := setByUse && !isInputList
                if shouldSkipAsPreviouslySetByUse {
                  continue
                }
		if isInputList {...}
		else {mod.Settings[settingName] = modVarName}
		usedVars = append(usedVars, settingName)
```

Basically there is one more skippable case that you are implicitly skipping. This makes that skip obvious. The bool name is in place of a comment. Also small tweak is to move the `usedVars` assignment out as it now applies to both cases.

pkg/config/expand.go
```diff
@@ -162,57 +162,79 @@ func getModuleInputMap(inputs []modulereader.VarInfo) map[string]string {
...
-				mod.Settings[settingName].([]interface{}), modVarName)
-			usedVars = append(usedVars, settingName)
-		} else if !setByUse {
-			// If input is not a list, set value if not already set and continue
-			mod.Settings[settingName] = modVarName
 			usedVars = append(usedVars, settingName)
 		}
 	}
-	return
+	return usedVars, nil
```
nick-stroud, line 237: I strongly prefer to actually return the values as you are doing here. I think some readability is lost by removing the name from the return value in the signature. I would suggest we reclaim that readability by adding a line in the docstring for the return value:

```
returns: a list of variable names that were used during the process of calling this function
```

pkg/config/config.go
nick-stroud: I agree with @mr0re1, I believe on all accounts. 

I would not require it as part of this PR as I do not want to discourage the good samaritan cleanup that this PR represents. 

@mr0re1, Do you suggest @tpdownes use this pattern only for the new variable added, or clean up all previous usage in this file?

pkg/config/expand.go
mr0re1, line 225: nit. The readability can be improved if we swap those two blocks and make it into if-else-if chain:
```
if isInputList {
  ...
} else if !setByUse {
  ...
}
```

Also `setByUse` is somewhat confusing nmae, removed gave an explanation, but no it rather unclear. We could remove any  assumptions of external interactions (set by another `use`) and simply name `alreadySet`.

pkg/config/config.go
mr0re1: > Do you suggest @tpdownes use this pattern only for the new variable added, or clean up all previous usage in this file?

Only the new message. But, @tpdownes , feel free to not do it.

tpdownes: My strong preference is to create a task that defines the appropriate cleanup approach for all `errorMessages` and then another task that does it. I agree that this approach is not desirable, for the reasons stated and I am sure that there are more.

pkg/config/expand.go
tpdownes: I agree it ought to be in the docstrings and will add it. I think the internal style conventions for named parameters could be reasonably summarized as "there are cases where it is very clearly bad to use them, and there are cases where it is not-too-objectionable to use them."

pkg/config/expand.go
tpdownes: That happens to be the if/else/if block that I am replacing. Although I take the point, combined with above that we can do a few edits and keep the cyclomatic complexity in place. In particular clarifying that the two if blocks in sequence can never be reached in the same run.

tpdownes: I tried a version of this at one point but was not able to massage it to avoid increasing cyclomatic complexity. Combined with comments below, I have managed to keep it constant while integrating a version of this suggestion.

pkg/config/expand.go
```diff
@@ -162,57 +162,79 @@ func getModuleInputMap(inputs []modulereader.VarInfo) map[string]string {
...
 		}
 
-		// if the input exists in the module doing the using, then we modify it
-		// according to these rules:
-		// if the setting has a value, then it was set by "use" because we have
-		// skipped anything explicitly in the blueprint. If the setting is a
-		// list of any type, then we want to concatenate these values in order
-		// they were used; otherwise, the setting should remain set to the
-		// existing used value
 		_, setByUse := mod.Settings[settingName]
```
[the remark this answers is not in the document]
tpdownes: I have added a helpful comment.

pkg/config/expand.go
```diff
@@ -180,20 +180,22 @@ func (mod *Module) addListValue(settingName string, value string) error {
...
 // a list, in which case output values are appended and flattened using HCL.
-// mod: "using" module as defined above
-// useMod: "used" module as defined above
-// useModGroupID: deployment group ID to which useMod belongs
-// modInputs: input variables as defined by the using module code
-// useOutputs: output values as defined by the used module code
-// settingsToIgnore: a list of module settings not to modify for any reason;
 //
-//	typical usage will be to leave explicit blueprint settings unmodified
+//	mod: "using" module as defined above
```
tpdownes, line 190: I am trying to match what I perceive to be [godoc style](https://tip.golang.org/doc/comment) and something that is treated the same by "go fmt" in both 1.18 and 1.20.

nick-stroud (approved):

PR #949, opened 2023-02-24
Allow VM instance name to include prefix and deployment name

nick-stroud:
This change adds the ability for a vm-instance to be named using both a prefix and the deployment name. This is needed to allow multiple deployments into a project when a prefix is also needed on the VM.

This PR preserves the original naming when the new variable is left as its default of `null`.

### Submission Checklist

  install`)


modules/compute/vm-instance/variables.tf
```diff
@@ -73,6 +73,16 @@ variable "name_prefix" {
   default     = null
 }
 
+variable "use_deployment_name_in_resource_name" {
+  description = <<-EOT
```
mr0re1, line 77: nit. Is it common practice for terraform to use ternary logic on optional bool? We could use "enum" (as in https://www.puppeteers.net/blog/terraform-enum-data-type/) for better interfacing:
`use_deployment_name_in_resource_name := "always" | "never" | "only_if_no_prefix"`

modules/compute/vm-instance/main.tf
```diff
@@ -21,7 +21,19 @@ locals {
...
+  # new behavior
+  prefix = var.name_prefix != null ? var.name_prefix : ""
+  deployment = (
+    var.use_deployment_name_in_resource_name == true ? # var could be null
+    var.deployment_name : ""
+  )
+  dash                = local.prefix != "" && local.deployment != "" ? "-" : ""
+  new_resource_prefix = "${local.prefix}${local.dash}${local.deployment}"
+
+  resource_prefix = var.use_deployment_name_in_resource_name == null ? local.old_resource_prefix : local.new_resource_prefix
```
tpdownes, line 36: I find it confusing that we are comparing a bool to `null` rather than just using the bool. Let's find a way to use it directly.

tpdownes (changes requested):
Code opinion that we should resolve: I don't like allowing the bool to default to null because it's causing  `var.x == true ?` and `var.x == null ?` comparisons when the code flows more naturally with a simple `var.x ?`.

Behavior problem to resolve: I use this blueprint

```
---
blueprint_name: simple

vars:
  deployment_name: ansible
  region: us-central1
  zone: us-central1-b

deployment_groups:
- group: first
  modules:
  - id: network1
    source: modules/network/vpc
  - id: simplevm
    settings:
      machine_type: e2-medium
      use_deployment_name_in_resource_name: true
```

And I deploy a machine named "ansible-0". I change the value to `false` and it destroys my original machine/disk and tries to replace them with resources named "-0". This isn't a valid resource name so the API calls fail. This solution breaks when the user does not supply a `var.name_prefix`.

Overall, we are tying ourselves in knots by not having versioned modules or being comfortable breaking things (a bit). I think it might be easier to add an input variable along lines of "use_old_names" with a default of false. I am willing to be argued into true, but I will start with an opinion of false.

Then a modern solution could look like:

```hcl
locals {
  old_resource_prefix = var.name_prefix != null ? var.name_prefix : var.deployment_name
  new_resource_prefix = var.name_prefix != null ? ${var.deployment_name}-${var.name_prefix} : random_pet.prefix.id
  resource_prefix = var.use_old_names ? local.old_resource_prefix : local.new_resource_prefix
}

# https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/pet
resource "random_pet" "prefix" {
  length = 1
  prefix = var.deployment_name
}
```

WDYT?

modules/compute/vm-instance/variables.tf
nick-stroud: I thought about enum. I strayed away because it is a bit clunker to define for user (having to get exact string vs setting bool), but given the rest of the critique, I will move away from bool in favor of enum if the three options are needed.

nick-stroud:
I think all of the concerns around booleans and nullness could be solved using an "enum". This would not solve the issue that @tpdownes raises about naming the resource `-0` in the case that there is no prefix and deployment name is explicitly not included.

I am opposed to the idea of creating a variable named `old_resource_prefix`. 
1. Depending how you set the default you may still be breaking people when there is really a decent option to maintain the original api behavior 
2. What happens when there is a 3rd way to name the resource? `oldest_resource_prefix`. 

I propose the following paths forward:  
1. Similar logic as is currently proposed but use an enum to clearly define the different options. Allow for bad configuration (no-prefix, use_deployment_name: never) to create bad name `-0` which fails on apply.  
1. Same as 1) except in the case of bad configuration add some static stand in, aka: `unnamed-vm-0`. 
1. Change the name of the bool to `add_deployment_name_to_name_prefix` and default to false. As the name implies this bool is only evaluated if prefix is also defined. This would have the odd case of no prefix & `add_deployment_name_to_name_prefix: false` would result in `deployment-name-0`.

nick-stroud:
I have essentially gone with option 3. which turns out to be very close to @tpdownes suggestion, basically use a bool but get rid of the null option. On @tpdownes suggestion I named `always_include_deployment_name`. 

The proposal will not break current blueprints and always treats the bool as a bool.

modules/compute/vm-instance/variables.tf
nick-stroud: I ended up picking a solution that did not need 3 options, so it is still a bool. Closing but please reopen if needed.

modules/compute/vm-instance/main.tf
nick-stroud: I believe this is resolved. Solution no longer relies on nullness of a boolean.

modules/compute/vm-instance/variables.tf
```diff
@@ -73,6 +73,15 @@ variable "name_prefix" {
   default     = null
 }
 
+variable "always_include_deployment_name" {
+  description = <<-EOT
+    If true, the deployment name will always be included as part of the VM name.
```
cboneti, line 78: I think here you could show how the instance will be named if this variable is set to true, for instance: 
If true, the instances will be named with both the prefix and the deployment name. 
"deployment name"-"prefix"-<index>. 
One would argue that, in this case, it is a prefix and not postfix, but I think renaming the variable is a bad idea.

modules/compute/vm-instance/variables.tf
```diff
@@ -73,6 +73,15 @@ variable "name_prefix" {
   default     = null
 }
 
+variable "always_include_deployment_name" {
```
cboneti, line 76: nit (please feel free to ignore):
`always_include_deployment_name` could be more intuitive
I explain:  the default behavior is to have prefix OR deployment name. 
By setting `always_include_deployment_name`, we are now adding deployment name and then the prefix. Would it be simpler to call it something like `add_deployment_name_before_prefix`?

cboneti (commented):
A comment on the variable description and a nit.

modules/compute/vm-instance/variables.tf
mr0re1: The word you're looking for is "affix")

tpdownes: If we want to avoid "see README", I suggest the following description changes...

From

```hcl
variable "always_include_deployment_name" {
  description = <<-EOT
    If true, the deployment name will always be included as part of the VM name. 
    If false, it will only be included if `name_prefix` is not provided.
    EOT
  type        = bool
  default     = false
}
```

to

> If true, the names of VMs and disks will always be prefixed with var.deployment_name to enable uniqueness across deployments. See var.name_prefix for further details on resource naming behavior.

I also suggest changing:

```hcl
variable "deployment_name" {
  description = "Name of the deployment, used to name the cluster"
  type        = string
}
```

to

> Name of the deployment, will optionally be used name resources according to var.name_prefix

and

```hcl
variable "name_prefix" {
  description = "Name Prefix"
  type        = string
  default     = null
}
```

> An optional name for all VM and disk resources. If not supplied, var.deployment_name will be used. When var.name_prefix is supplied, and var.always_include_deployment_name is set, then resources are named by \"var.deployment_name-var.name_prefix\".

tpdownes (changes requested):
Minor changes to descriptions of the related variables to resolve potentially conflicting claims and direct user to one place for the "answer"

modules/compute/vm-instance/variables.tf
nick-stroud: Discussed with @tpdownes who found either name acceptable. Updated to this suggested name.

modules/compute/vm-instance/variables.tf
nick-stroud: Updated descriptions as suggested.

tpdownes (approved):
Feel free to squash (or not)

PR #951, opened 2023-02-24
Use HTCondor Python bindings in autoscaler

tpdownes:
- Replace all shell call-outs and parsing with efficient queries to the scheduler and collector daemons
- Anticipate future work to base auto-scaling on GPU and memory requests by including those job requests in queries
- Alter scaling code to account for the number of CPUs requested by each job. Previous solution implicitly assumes each job requests 1 CPU and will underscale for multi-core jobs
- Do not scale down any execute points that have not been active for less than 150 seconds

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scripts/htcondor-install/files/autoscaler.py
+++ b/community/modules/scripts/htcondor-install/files/autoscaler.py
@@ -185,7 +185,14 @@ class AutoScaler:
         # stewardship. A full list of Job ClassAd attributes can be found at
         # https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html
         schedd = htcondor.Schedd()
-        job_attributes = [ "RequestCpus", "RequestMemory", "RequestGpus" ]
+        REQUEST_CPUS_ATTRIBUTE = "RequestCpus"
+        REQUEST_GPUS_ATTRIBUTE = "RequestGpus"
+        REQUEST_MEMORY_ATTRIBUTE = "RequestMemory"
+        job_attributes = [
+            REQUEST_CPUS_ATTRIBUTE,
+            REQUEST_GPUS_ATTRIBUTE,
+            REQUEST_MEMORY_ATTRIBUTE,
+        ]
 
         # For purpose of scaling a Managed Instance Group, count only jobs that
         # are running or idle ("potentially runnable"). Filter JobStatus by:
@@ -193,8 +200,8 @@ class AutoScaler:
         running_job_ads = schedd.query(constraint="JobStatus==2", projection=job_attributes)
         idle_job_ads = schedd.query(constraint="JobStatus==1", projection=job_attributes)
 
-        total_idle_request_cpus = sum(j["RequestCpus"] for j in idle_job_ads)
-        total_running_request_cpus = sum(j["RequestCpus"] for j in running_job_ads)
+        total_idle_request_cpus = sum(j[REQUEST_CPUS_ATTRIBUTE] for j in idle_job_ads)
+        total_running_request_cpus = sum(j[REQUEST_CPUS_ATTRIBUTE] for j in running_job_ads)
         queue = total_idle_request_cpus + total_running_request_cpus
 
         print(f"Total CPUs requested by running jobs: {total_running_request_cpus}")
--- a/community/modules/scripts/htcondor-install/files/autoscaler.py
+++ b/community/modules/scripts/htcondor-install/files/autoscaler.py
@@ -24,10 +24,10 @@ from pprint import pprint
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
 
+import argparse
 import os
 import math
-import argparse
-import datetime
+import time
 import htcondor
 import classad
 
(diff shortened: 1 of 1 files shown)


community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -180,47 +181,30 @@ def scale(self):
...
-        if len(queue_length_resp) > 1:
-            queue = int(queue_length_resp[0])
-            idle_jobs = int(queue_length_resp[1])
-            on_hold_jobs = int(queue_length_resp[2])
-        else:
-            queue = 0
-            idle_jobs = 0
-            on_hold_jobs = 0
+        total_idle_request_cpus = sum(j["RequestCpus"] for j in idle_job_ads)
+        total_running_request_cpus = sum(j["RequestCpus"] for j in running_job_ads)
```
nick-stroud, line 193: Make "RequestCpus" a constant (all caps var), maybe other request strings as well although less important as they are not duplicated.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -180,47 +181,30 @@ def scale(self):
...
+        queue = total_idle_request_cpus + total_running_request_cpus
 
-        print("Total queue length: " + str(queue))
-        print("Idle jobs: " + str(idle_jobs))
-        print("Jobs on hold: " + str(on_hold_jobs))
```
nick-stroud, line 201: Why no longer report jobs on hold?

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -180,47 +181,30 @@ def scale(self):
...
+        print(f"Running job RequestCpus: {total_running_request_cpus}")
+        print(f"Idle job RequestCpus: {total_idle_request_cpus}")
```
nick-stroud, line 197: Will condor people understand the RequestCpus reference. For me this is less clear than the previous message (even though it is more specific) because it leaves me wondering what the reference to RequestCpus means.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -272,47 +256,37 @@ def scale(self):
...
-                            idle_node_zones[server] = zone
-                        else:
-                            idle_node_zones[server] = None
-                    else:
-                        if status != "Unclaimed":
-                            idle_node_zones[server] = None
+
+            # Find VMs that are unused (no dynamic slots created from
+            # partitionable slots) and have been booted for at least 150 seconds
+            max_daemon_start_time = int(datetime.datetime.now().timestamp()-150)
```
nick-stroud, line 262: I believe `time.time()` could replace `datetime.datetime.now().timestamp()` and would be more idiomatic python.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -180,47 +181,30 @@ def scale(self):
...
-        queue_length_req = (
-            'condor_q -allusers -totals -format "%d " Jobs -format "%d " Idle -format "%d " Held'
-        )
-        queue_length_resp = os.popen(queue_length_req).read().split()
+        # Count jobs that are running or idle ("potentially runnable"); do not
+        # count held jobs or jobs in transitioning states
+        # https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html#JobStatus
+        schedd = htcondor.Schedd()
+        job_attributes = [ "RequestCpus", "RequestMemory", "RequestGpus" ]
+        running_job_ads = schedd.query(constraint="JobStatus==2", projection=job_attributes)
```
nick-stroud, line 189: Silly question: What is `ads`?

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: A few lines above I link to HTCondor Job ClassAd attributes (for `JobStatus` and its defined values).

Would you like me to make that link more generic?

https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html#RequestCpus

Strictly speaking, the names of ClassAd attributes in HTCondor are defined to be case insensitive (though I can't find a link to that). What I can say is that every job gets a RequestCpus (and maybe RequestMemory, RequestGpus, RequestDisk?) with a default value.

So we can do what we want, but convention is CamelCase.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: The prior calculation consumed the stdout of a particular command that added up jobs in the queue. It did so in a manner that required subtracting off held jobs before calculating the autoscaling target. The new calculation computes the right target from the right data in the first place.

Jobs can go on hold deliberately or by HTCondor itself but the autoscaler isn't really the place to log that unless it's useful for autoscaling.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: HTCondor admins will know `RequestCpus` (for reasons noted above) but it's easy for me to alter the text to something more completely correct: "Total CPUs requested by running/idle jobs".

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: I'm anticipating the need for `datetime` in future work, but I agree that `time.time()` is simpler.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: I have added more descriptive comments and links that might help maintainers of this code who are less familiar with HTCondor administration.

tpdownes:
Several changes made to address feedback. PTAL @nick-stroud

community/modules/scripts/htcondor-install/files/autoscaler.py
nick-stroud: My request here was to the following between L187 and L188 (could also go at top of file)

```
REQUEST_CPU = 'RequestCpus'
```

And then use that variable in place of the string literals. This prevents typos and update errors.

tpdownes:
PTAL. Changes have been tested on a running pool.

nick-stroud (approved):

PR #952, opened 2023-02-25
Unify validatorConfig check logic, reduce code duplication

mr0re1:
Currently all validator implementations have same code copied around. Move it into a `func (v *validatorConfig) check`

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/validate.go
+++ b/pkg/config/validate.go
@@ -288,29 +288,6 @@ func (dc *DeploymentConfig) getValidators() map[string]func(validatorConfig) err
 	return allValidators
 }
 
-// check that the keys in inputs and requiredInputs are identical sets of strings
-func testInputList(function string, inputs map[string]interface{}, requiredInputs []string) error {
-	var errored bool
-	for _, requiredInput := range requiredInputs {
-		if _, found := inputs[requiredInput]; !found {
-			log.Printf("a required input %s was not provided to %s!", requiredInput, function)
-			errored = true
-		}
-	}
-
-	if errored {
-		return fmt.Errorf("at least one required input was not provided to %s", function)
-	}
-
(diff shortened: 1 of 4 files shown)


pkg/config/validate.go
```diff
@@ -336,17 +313,8 @@ func mergeBlueprintRequirements(base map[string][]string, new map[string][]strin
...
-	funcName := testApisEnabledName.String()
-	funcErrorMsg := fmt.Sprintf(funcErrorMsgTemplate, funcName)
-
-	if validator.Validator != funcName {
-		return fmt.Errorf("passed wrong validator to %s implementation", funcName)
-	}
-
-	err := testInputList(validator.Validator, validator.Inputs, requiredInputs)
-	if err != nil {
+func (dc *DeploymentConfig) testApisEnabled(c validatorConfig) error {
```
nick-stroud, line 316: nit: I am curious on the decision to go from `validator` to `c`. I know short arg names is idiomatic for go. `validator` could cause confusion because it is so close to `validators` which is something very different. But maybe `config` is more readable than `c`. No strong opinions here.

nick-stroud (approved):

pkg/config/validate.go
tpdownes: Having written much of this code, I'll say that I `+1` disambiguating the several uses of `validate`, `validator`, and `validatorConfigs` here.

mr0re1: My intent is `c` is for `Config`. I would keep it `c` to follow golang code style.
@tpdownes  do you have any concerns? I guess `c` does disambiguate it somewhat.

tpdownes: LGTM

PR #959, opened 2023-02-28
Do not populate subnetwork_project as it is populated from subnetwork_self_link

nick-stroud:
When used with a shared VPC, this subnetwork_project is incorrectly populated with the client project. The [provider docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#subnetwork_project) state that the field is optional and will default to using subnetwork as long as a self link is provided. For the default case we always provide the subnetwork self link. We still need to keep the variable (as null) to allow users to provide the full network_interface block.

### Submission Checklist

  install`)

### DIFF
--- a/modules/compute/vm-instance/main.tf
+++ b/modules/compute/vm-instance/main.tf
@@ -75,7 +75,7 @@ locals {
   default_network_interface = {
     network            = var.network_self_link
     subnetwork         = var.subnetwork_self_link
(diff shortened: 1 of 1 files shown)


tpdownes (approved):
LGTM. The (sub)network self links are generally the safest way to identify networks.

PR #961, opened 2023-02-28
Re-order tasks in HTCondor configuration runner

tpdownes:
- move management of HTCondor service state entirely into handlers (for
  reload and restart notifications) and post_tasks (which runs after
  handlers) for ensuring the service is enabled and started; this has
  the effect of ensuring that the service is only started and harmlessly
  reloaded once on the first boot.
- order the restart handler before the reload handler so that the
  service is reloaded (w/o interruption to services) after the restart
- take a careful eye toward which configuration changes require only
  a reload vs those which require a full restart
- create a dedicated handler for reloading the SystemD unit
  configuration when overriding directives to order services

Before this commit, on first boot, HTCondor would start and immediately
be reloaded, then restarted again. Until the fix to HTCONDOR-1590 is
available in a public release, this ordering resulted in a hot-hot (bad)
configuration of SchedD daemons running in high availability mode.

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
+++ b/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
@@ -93,6 +93,8 @@
         content: |
           COLLECTOR_UPDATE_INTERVAL=60
           NEGOTIATOR_UPDATE_INTERVAL=60
+      notify:
+      - Reload HTCondor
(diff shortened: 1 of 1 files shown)


tpdownes:
@mr0re1 the force push is just a fix to the commit message, no code change. ready for review.

mr0re1 (approved):

PR #962, opened 2023-02-28
Rename filestore network_name to network_id to enable shared VPC via use

nick-stroud:
This change allows for the use of a shared VPC with the Filestore module via the `use` field. When a filestore needs to be provisioned in a client project but use a shared VPC from a host project, filestore must be provided with the `network_id` which includes information about the VPC project, instead of just `network_name`.

## Breaking Change  
This is technically a breaking change. In some edge cases a user will be required to change the key for a setting input from `network_name` to `network_id`. While the name has changed the functionality has not changed. It is expected that the majority of blueprints will be passing `network_name` and `network_id` via the `use` keyword, in which case this will not be a breaking change. 

In the case that a blueprint was explicitly using `network_name` the user will get the following error at create time:

```
found an issue while validating settings for module at ./modules/file-system/filestore: invalid setting provided to a module, cause: a setting was added that is not found in the module
Module ID: homefs Setting: network_name
```

Or in the case that they are not `use`ing a network they will get:

```
failed to apply deployment variables in modules when expanding the config: a required setting is missing from a module: Module ID: homefs Setting: network_id
```

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/fluent-tutorial.yaml
+++ b/community/examples/fluent-tutorial.yaml
@@ -112,7 +112,7 @@ deployment_groups:
     kind: terraform
     id: homefs
     settings:
-      network_name: "default"
+      network_id: "default"
       size_gb: 1024
       filestore_tier: "BASIC_HDD"
       local_mount: /home
@@ -121,7 +121,7 @@ deployment_groups:
     kind: terraform
     id: sharefs
     settings:
-      network_name: "default"
+      network_id: "default"
       size_gb: 1024
       filestore_tier: "BASIC_HDD"
       local_mount: /shared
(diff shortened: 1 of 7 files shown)


community/examples/fluent-tutorial.yaml
```diff
@@ -112,7 +112,7 @@ deployment_groups:
     kind: terraform
     id: homefs
     settings:
-      network_name: "default"
+      network_id: "default"
```
cboneti, line 115: While this change is correct, I am nit picking on the fact that this blueprint does not use a pre-existing-vpc and instead is relying on the arguably incorrect vm instance default, so it is passing the network as a string to all modules that need it. 

Would this not be an opportunity to add a pre-existing-vpc at the top of the deployment group and then just use the network everywhere?

cboneti (approved):
Approved with an optional suggestion.

community/examples/fluent-tutorial.yaml
nick-stroud: Agreed. I will open a separate PR after merging.

PR #964, opened 2023-03-01
Update image building example to use Slurm V5

mr0re1:
- Update `examples/image-builder.yaml`;
- Update `examples/README.md` with increased quota requirements.

### Submission Checklist

  install`)

### DIFF
--- a/examples/image-builder.yaml
+++ b/examples/image-builder.yaml
@@ -23,8 +23,6 @@ vars:
   new_image_family: my-slurm-image
   network_name: image-builder
   subnetwork_name: image-builder-us-central1
-  # You can find size of source image by using following command
-  # gcloud compute images describe-from-family schedmd-v5-slurm-22-05-8-centos-7 --project schedmd-slurm-public
   disk_size: 32
 
 # Documentation for each of the modules used below can be found at
@@ -56,6 +54,9 @@ deployment_groups:
       source_image_project_id: [schedmd-slurm-public]
       # see latest in https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#supported-operating-systems
       source_image_family: schedmd-v5-slurm-22-05-6-hpc-centos-7
+      # You can find size of source image by using following command
+      # gcloud compute images describe-from-family <source_image_family> --project schedmd-slurm-public
+      disk_size: $(vars.disk_size)
       image_family: $(vars.new_image_family)
       state_timeout: 15m
 
@@ -96,7 +97,6 @@ deployment_groups:
     source: community/modules/scheduler/schedmd-slurm-gcp-v5-login
     use: [cluster-network, slurm_controller]
     settings:
-      machine_type: n2-standard-4
       disable_login_public_ips: false
       disk_size_gb: $(vars.disk_size)
       instance_image:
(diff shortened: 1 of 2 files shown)


examples/image-builder.yaml
```diff
@@ -23,6 +23,7 @@ vars:
   new_image_family: my-slurm-image
   network_name: image-builder
   subnetwork_name: image-builder-us-central1
+  disk_size: 100
```
cboneti, line 26: I would try to set it to something smaller. 100GB is pretty large here. 
I believe the schedmd Image is 32GB, so anything larger than that is ok.

examples/image-builder.yaml
```diff
@@ -49,36 +51,52 @@ deployment_groups:
     source: modules/packer/custom-image
     kind: packer
     settings:
-      disk_size: 20
       source_image_project_id: [schedmd-slurm-public]
-      source_image_family: schedmd-slurm-21-08-8-hpc-centos-7
+      # see latest in https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#supported-operating-systems
+      source_image_family: schedmd-v5-slurm-22-05-6-hpc-centos-7
       image_family: $(vars.new_image_family)
+      state_timeout: 15m
```
cboneti, line 58: Is it needed here? (if you decrease the image size)

cboneti (commented):
I added two minor comments. PTAL

examples/image-builder.yaml
mr0re1: One of my tries failed with packer timeout.

cboneti: Sorry, my comment wasn't clear. This is likely needed because the image is 100GB, but if that is reduced, then you might not need this.

examples/image-builder.yaml
mr0re1: Reduced to 32GB

examples/image-builder.yaml
```diff
@@ -49,36 +53,52 @@ deployment_groups:
     source: modules/packer/custom-image
     kind: packer
     settings:
-      disk_size: 20
```
nick-stroud, line 52: nit: I think I prefered when the disk_size was explicitly tied to the packer module. It is easy to miss now that it is being propagated down (not so much for node group since you make an explicit reference).

One option would be to set disk size to 32 here and just let the node group use the default of 50.

examples/image-builder.yaml
```diff
@@ -49,36 +53,52 @@ deployment_groups:
...
       instance_image:
         family: $(vars.new_image_family)
         project: $(vars.project_id)
+
   - id: slurm_login
-    source: community/modules/scheduler/SchedMD-slurm-on-gcp-login-node
+    source: community/modules/scheduler/schedmd-slurm-gcp-v5-login
     use: [cluster-network, slurm_controller]
     settings:
+      machine_type: n2-standard-4
```
nick-stroud, line 99: This looks unneeded, why not just leave with default of `n2-standard-2`?

nick-stroud (approved):

examples/image-builder.yaml
mr0re1: I share your point about being explicit. I'm going to add:
```
kind: packer
...
  disk_size: $(vars.disk_size)
```
But I will keep the same disk_size for other machines to "save customers many"

PR #966, opened 2023-03-01
Mask sleep/suspend targets on chrome-remote-desktop to prevent shutdown

nick-stroud:
Tested: manually tested with and w/o the change to verify that this prevents crd from crashing.

`release-candidate`: I think this change merits inclusion into release candidate as it will probably be this release when the hcls blueprint gets substantial attention. If agreed, I would separately merge to develop as well.

### Submission Checklist

  install`)


tpdownes (approved):
LGTM

PR #967, opened 2023-03-01
Revert "Rename filestore network_name to network_id to enable shared VPC via use"

nick-stroud:
Reason for revert: Caused integration tests to fail.

This reverts commit 0a0230c1d2e5d3da5c823c5885548a0e2490ab5c. (with minor conflict fix)

### Submission Checklist

* [x] Have you installed and run this change against pre-commit? (`pre-commit
  install`)
* [x] Are all tests passing? (`make tests`)
* [x] Have you written unit tests to cover this change?
* [x] Is unit test coverage still above 80%?
* [x] Have you updated all applicable documentation?
* [x] Have you followed the guidelines in our Contributing document?

PR #969, opened 2023-03-02
Roll-forward:"Rename filestore network_name to network_id to enable shared VPC via use"

nick-stroud:
Original PR: #962 
Rollback PR: #967 

Filestore is particular and will store the `network_name` when the network is in the service project and will store `network_id` when it is using a shared VPC. We must discern which case we are in and provide the correct input to avoid destroy/create cycles upon re-apply

### Using Data Source

I tried to use a data source in place of parsing the network id, but it is not possible without doing the same network id parsing that is done in this PR. It would be possible to do the parsing done in this PR and then feed it through a data source and pull values back out the other side (as a sort of validation) but this is just extra steps and results in the same sort of errors that filestore will emit. 

Things tried:
- Try using the `google_compute_network` data source: this requires network name and project id as inputs. The name field will not take network_id or network_self_link. To get network_name and project_id they must be parsed from network_id.
- Try using `google_compute_subnetwork` and `google_compute_network` together: I was hoping to look up `google_compute_subnetwork` with `subnetwork_self_link` and then use the `network` attribute to look up the `google_compute_network` but this is not possible as  `subnetwork_self_link.network` gives off the network_self_link and not the network name.

### Submission Checklist

  install`)

### DIFF
--- a/modules/file-system/filestore/variables.tf
+++ b/modules/file-system/filestore/variables.tf
@@ -36,12 +36,14 @@ variable "region" {
 
 variable "network_id" {
   description = <<-EOT
-    The name of the GCE VPC network to which the instance is connected. 
-    If using shared VPC, full network id must be given with format:
+    The ID of the GCE VPC network to which the instance is connected given in the format:
     `projects/<project_id>/global/networks/<network_name>`"
-    Plain network name is accepted for non-shared VPC case.
     EOT
   type        = string
+  validation {
+    condition     = length(split("/", var.network_id)) == 5
+    error_message = "The network id must be provided in the following format: projects/<project_id>/global/networks/<network_name>."
+  }
 }
 
 variable "name" {
--- a/modules/file-system/filestore/variables.tf
+++ b/modules/file-system/filestore/variables.tf
@@ -34,8 +34,13 @@ variable "region" {
   type        = string
 }
 
-variable "network_name" {
-  description = "The name of the GCE VPC network to which the instance is connected."
+variable "network_id" {
+  description = <<-EOT
+    The name of the GCE VPC network to which the instance is connected. 
+    If using shared VPC, full network id must be given with format:
+    `projects/<project_id>/global/networks/<network_name>`"
+    Plain network name is accepted for non-shared VPC case.
+    EOT
   type        = string
 }
 
--- a/community/examples/hpc-cluster-small-sharedvpc.yaml
+++ b/community/examples/hpc-cluster-small-sharedvpc.yaml
@@ -54,7 +54,6 @@ deployment_groups:
     settings:
       local_mount: /home
       connect_mode: PRIVATE_SERVICE_ACCESS
-      network_name: $(network1.network_id)
 
   # This debug_partition will work out of the box without requesting additional GCP quota.
   - id: debug_partition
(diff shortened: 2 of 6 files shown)


modules/file-system/filestore/variables.tf
```diff
@@ -34,8 +34,13 @@ variable "region" {
...
 }
 
-variable "network_name" {
-  description = "The name of the GCE VPC network to which the instance is connected."
+variable "network_id" {
+  description = <<-EOT
+    The name of the GCE VPC network to which the instance is connected. 
+    If using shared VPC, full network id must be given with format:
+    `projects/<project_id>/global/networks/<network_name>`"
+    Plain network name is accepted for non-shared VPC case.
```
tpdownes, line 42: "Plain network name is accepted" does not appear to be the case in the code. Empirically, this blueprint:

```yaml
blueprint_name: simple

vars:
  deployment_name: simple
  region: us-central1
  zone: us-central1-b

# Documentation for each of the modules used below can be found at
# https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md

deployment_groups:
- group: first
  modules:
  - id: network1
  - id: simpelfs
    settings:
      network_id: default
  - id: simplevm
    use:
    - network1
    settings:
      machine_type: e2-medium
```

yields the following error

```
╷
│ Error: Invalid index
│
│   on modules/filestore/main.tf line 42, in locals:
│   42:   network_name     = local.split_network_id[4]
│     ├────────────────
│     │ local.split_network_id is list of string with 1 element
│
│ The given key does not identify an element in this collection value.
╵
╷
│ Error: Invalid index
│
│   on modules/filestore/main.tf line 43, in locals:
│   43:   network_project  = local.split_network_id[1]
│     ├────────────────
│     │ local.split_network_id is list of string with 1 element
│
│ The given key does not identify an element in this collection value: the given index is greater than or equal to the length of the collection.
```

tpdownes (changes requested):
Big picture: this PR re-creates the problem that network_id takes on two different formats when it has a well-defined format elsewhere in the Toolkit.

This code does appear to avoid permadiff outcomes, however.

The way I read the [provider's Filestore code for the network field within the networks block](https://github.com/hashicorp/terraform-provider-google/blob/34b07a6db2790899bab0777bda29178132c39802/google/resource_filestore_instance.go#L161-L167), it will ignore differences between self_links and names. Might be worth trying simply throwing a self_link at the API directly even if it doesn't appear to be [documented as an allowed input value](https://cloud.google.com/filestore/docs/reference/rest/v1/projects.locations.instances#networkconfig).

I think the value of the Toolkit is in the modules working together effectively. I would rather make it not possible to create a Filestore module with "mynet" as a setting if it means we support reliable project placement of the Filestore instance (for billing) even in Shared VPC environments.

You might be interested to know [how the network ID is manually constructed in the data source](https://github.com/hashicorp/terraform-provider-google/blob/34b07a6db2790899bab0777bda29178132c39802/google/data_source_google_compute_network.go#L76)

modules/file-system/filestore/variables.tf
nick-stroud: This was my intention (as it is how the code is written). I forgot to update the description. Good catch.

I will update description and add validation.

nick-stroud: Done.

tpdownes (approved):

PR #970, opened 2023-03-02
Add validation to prevent usage of variables in backend block.

mr0re1:
- Add validation to prevent usage of variables in backend block.
- Unrelated minor refactoring.

```
$ ./ghpc create examples/hpc-cluster-small.yaml --backend-config "bucket=\$(trenta)"
2023/03/02 06:39:46 can not use variables in backend block, got 'bucket=$(trenta)'
```

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -384,10 +384,6 @@ func importBlueprint(blueprintFilename string) (Blueprint, error) {
 		blueprint.Vars = make(map[string]interface{})
 	}
 
-	if len(blueprint.Vars) == 0 {
-		blueprint.Vars = make(map[string]interface{})
-	}
-
 	// if the validation level has been explicitly set to an invalid value
 	// in YAML blueprint then silently default to validationError
 	if !isValidValidationLevel(blueprint.ValidationLevel) {
@@ -530,6 +526,31 @@ func checkUsedModuleNames(
 	return nil
 }
 
+func checkBackend(b TerraformBackend) error {
+	const errMsg = "can not use variables in backend block, got '%s=%s'"
+	if hasVariable(b.Type) {
+		return fmt.Errorf(errMsg, "type", b.Type)
+	}
+	for k, v := range b.Configuration {
+		if s, ok := v.(string); ok && hasVariable(s) {
+			return fmt.Errorf(errMsg, k, s)
+		}
+	}
+	return nil
+}
+
+func checkBackends(bp Blueprint) error {
+	if err := checkBackend(bp.TerraformBackendDefaults); err != nil {
+		return err
+	}
+	for _, g := range bp.DeploymentGroups {
+		if err := checkBackend(g.TerraformBackend); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // validateConfig runs a set of simple early checks on the imported input YAML
 func (dc *DeploymentConfig) validateConfig() {
 	_, err := dc.Config.DeploymentName()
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -384,10 +384,6 @@ func importBlueprint(blueprintFilename string) (Blueprint, error) {
 		blueprint.Vars = make(map[string]interface{})
 	}
 
-	if len(blueprint.Vars) == 0 {
```
tpdownes, line 387: Good catch.

pkg/config/config_test.go
```diff
@@ -1049,3 +1049,83 @@ func (s *MySuite) TestValidatorConfigCheck(c *C) {
 		c.Check(err, ErrorMatches, "only 3 inputs \\[in0 in1 in2\\] should be provided to test_project_exists")
 	}
 }
+
+func (s *MySuite) TestCheckBackends(c *C) {
+	// Helper to create blueprient with backend blocks only (first one is defaults)
```
tpdownes, line 1054: fix typo in "blueprient"

pkg/config/config_test.go
```diff
@@ -1049,3 +1049,83 @@ func (s *MySuite) TestValidatorConfigCheck(c *C) {
...
+		}
+		return checkBackends(bp)
+	}
+	dummy := TerraformBackend{}
+
+	{ // Absent
+		c.Check(checkBackends(Blueprint{}), IsNil)
+	}
+
+	{ // Dummies
```
tpdownes, line 1073: should probably clarify that empty backends should pass

pkg/config/config_test.go
```diff
@@ -1049,3 +1049,83 @@ func (s *MySuite) TestValidatorConfigCheck(c *C) {
...
+		c.Check(check(dummy, dummy, dummy), IsNil)
+	}
+
+	{ // OK
+		b := TerraformBackend{
+			Type: "gcs",
+			Configuration: map[string]interface{}{
+				"bucket":                      "trenta",
```
tpdownes, line 1081: nit: use something more like foo/bar or "not-a-bucket" etc.

pkg/config/expand.go
```diff
@@ -78,7 +78,9 @@ func (dc *DeploymentConfig) expand() {
 			"failed to apply deployment variables in modules when expanding the config: %v",
 			err)
 	}
-	dc.expandVariables()
+	if err := dc.expandVariables(); err != nil {
```
tpdownes, line 81: expandVariables now returns an error, instead of log.Fatal on its own. Do you believe there should now be an TestExpandVariables unit test? If so, please create task for it in the backlog.

The coverage for hpc-toolkit/pkg/config appears to improve by 0.2% with this commit, so I do not believe the lack of a unit test should block the PR.

tpdownes (changes requested):
Minor changes and a question to address.

pkg/config/config_test.go
mr0re1: Added OK/FAIL mark to each case, modified some comments.

pkg/config/config_test.go
mr0re1: I assume your intentions is to "reduce mental load". I see how it can be applicable for this particular test, but readability gains are too minuscule, I would argue not existent.
On another hand overuse  of "foo/bar/buzz" can hinder both readability and debugability of a test.
So I'm trying to avoid using "foo/bar/buzz" in code and use diverse random relatively short words instead.
If I misunderstood your comment intention, please correct me.

tpdownes (approved):

PR #977, opened 2023-03-03
Add utils for parising and normalizing HCL dtype

mr0re1:
Those utils will be used later for variables dtype validators and checks

### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulereader/hcl_utils_test.go
+++ b/pkg/modulereader/hcl_utils_test.go
@@ -12,12 +12,10 @@ func (s *MySuite) TestNormalizeType(c *C) {
 
 	c.Check(normalizeType("?invalid_type"), Equals, "?invalid_type")
 
-	c.Check(normalizeType("object({b=string,a=number})"), Equals, "object({a=number,b=string})")
-
 	// `any` is special type, check that it works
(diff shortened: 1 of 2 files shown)


pkg/modulereader/hcl_utils_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+func (s *MySuite) TestNormalizeType(c *C) {
+	c.Check(
+		normalizeType("object({count=number,kind=string})"),
+		Equals,
+		normalizeType("object({kind=string,count=number})"))
+
+	c.Check(normalizeType("?invalid_type"), Equals, "?invalid_type")
+
+	c.Check(normalizeType("object({b=string,a=number})"), Equals, "object({a=number,b=string})")
```
nick-stroud, line 15: This seems perhaps over specific. It looks like we are checking that the internal behavior of `hclsyntax` and `typeexpr` return objects in alpha order. Is this behavior documented? My suggestion would be to always use the `normalizeType` on the right side of the comparison as well. I guess it depends if we care just that the type has been normalized or if we care about the specific format it has been normalized to.

nick-stroud (approved):

PR #983, opened 2023-03-06
Add enforcement of minimum # of idle VMs to HTCondor autoscaler

tpdownes:
- Enable the HTCondor autoscaler to maintain a minimum number of idle HTCondor execute points by tracking the machines known to be idle by HTCondor itself and the number of machines still being created by the Managed Instance Group.
- Migrate all per-MIG settings to a SystemD override file that is loaded when the MIG autoscaler is triggered

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scripts/htcondor-install/files/autoscaler.py
+++ b/community/modules/scripts/htcondor-install/files/autoscaler.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 # -*- coding: utf-8 -*-
 
 # Copyright 2018 Google Inc. All Rights Reserved.
@@ -20,6 +20,7 @@
 
 from absl import app
 from absl import flags
+from collections import OrderedDict
 from pprint import pprint
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
@@ -31,7 +32,7 @@ import time
 import htcondor
 import classad
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument("--p", required=True, help="Project id", type=str)
 parser.add_argument(
     "--z",
@@ -53,6 +54,12 @@ parser.add_argument(
 )
 parser.add_argument(
     "--g", required=True, help="Name of the managed instance group", type=str
+)
+parser.add_argument(
+    "--i",
+    default=0,
+    help="Minimum number of idle compute instances",
+    type=int
 )
 parser.add_argument(
     "--c", required=True, help="Maximum number of compute instances", type=int
@@ -87,26 +94,19 @@ class AutoScaler:
         else:
             self.instanceGroupManagers = self.service.instanceGroupManagers()
 
-    # Remove specified instance from MIG and decrease MIG size
-    def deleteFromMig(self, instance, zone):
-        instanceUrl = "https://www.googleapis.com/compute/v1/projects/" + self.project
-        instanceUrl += "/zones/" + zone
-        instanceUrl += "/instances/" + instance
-
-        instances_to_delete = {"instances": [instanceUrl]}
-
+    # Remove specified instances from MIG and decrease MIG size
+    def deleteFromMig(self, node_self_links):
         requestDelInstance = self.instanceGroupManagers.deleteInstances(
             project=self.project,
             **self.zoneargs,
             instanceGroupManager=self.instance_group_manager,
-            body=instances_to_delete,
+            body={ "instances": node_self_links },
         )
 
         # execute if not a dry-run
         if not self.dryrun:
             response = requestDelInstance.execute()
             if self.debug > 0:
-                print("Request to delete instance " + instance)
                 pprint(response)
             return response
         return "Dry Run"
@@ -185,6 +185,10 @@ class AutoScaler:
         # stewardship. A full list of Job ClassAd attributes can be found at
         # https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html
         schedd = htcondor.Schedd()
+        # encourage the job queue to start a new negotiation cycle
+        # there are internal unconfigurable rate limits so not guaranteed
+        schedd.reschedule()
+        time.sleep(3)
         REQUEST_CPUS_ATTRIBUTE = "RequestCpus"
         REQUEST_GPUS_ATTRIBUTE = "RequestGpus"
         REQUEST_MEMORY_ATTRIBUTE = "RequestMemory"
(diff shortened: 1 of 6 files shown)


community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -226,29 +232,21 @@ def scale(self):
             print("Information about the compute instance template")
             pprint(instanceTemplateInfo)
 
-        # Calculate number instances to satisfy current job queue CPU requests
-        if queue > 0:
-            self.size = int(math.ceil(float(queue) / float(self.cores_per_node)))
```
mr0re1, line 231: nit. Are we sill supporting  Python2 ? If no, `float` and `int` casts can be removed.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -226,29 +232,21 @@ def scale(self):
...
-        if self.compute_instance_limit > 0 and self.size > self.compute_instance_limit:
-            self.size = self.compute_instance_limit
-            print(
-                "MIG target size will be limited by " + str(self.compute_instance_limit)
-            )
-
-        print("New MIG target size: " + str(self.size))
+        # Calculate the minimum number of instances that, for fully packed
+        # execute points, could satisfy current job queue
+        min_hosts_for_queue = int(math.ceil(float(queue) / float(self.cores_per_node)))
```
mr0re1, line 237: Rest of code isn't Python2-compatible. In Python3 `/`-op does "true division", also `math.ceil` returns int.
```python
min_hosts_for_queue = math.ceil(queue / self.cores_per_node)
```

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -260,45 +258,59 @@ def scale(self):
...
+            min_hosts_for_queue += self.compute_instance_min_idle - delta - n_idle_total
+            print(f"Adjusting target to maintain idle VMs: {min_hosts_for_queue}")
+
+        if min_hosts_for_queue > self.compute_instance_limit:
+            self.size = self.compute_instance_limit
+            print(f"MIG target size will be limited by {self.compute_instance_limit}")
+        else:
+            self.size = min_hosts_for_queue
+
+        print("New MIG target size: " + str(self.size))
```
mr0re1, line 288: nit. Rest of this code uses f-strings, let's keep it consistent, just to drop `str` cast.

tpdownes:
@nick-stroud I have discovered a bug in the method by which idle nodes are calculated. I'm resolving it and will update this branch when I have the fix.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: This has been captured in a backlog task for modernizing the style and addressing linting issues.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: Where I am using existing code, I've mostly left it alone, but I agree that this merits updating. I have simultaneously updated the shebang to python3.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -260,45 +259,59 @@ def scale(self):
...
+        # overlap with the "verifying" state that will be addressed by
+        # only deleting nodes whose uptime exceeds this period of overlap
+        n_idle = len(idle_node_ads)
+        n_idle_total = n_being_born + n_idle
+        print(f"Idle nodes found: {n_being_born} birthing and {n_idle} idling")
+
+        # always keep size tending toward the minimum idle VMs requested
+        delta = min_hosts_for_queue - currentTarget
+        if delta < self.compute_instance_min_idle:
+            min_hosts_for_queue += self.compute_instance_min_idle - delta - n_idle_total
```
nick-stroud, line 280: This block (L278,279,280) feels is quite confusing. 

I don't understand why we have the **if** statement. What if delta=10 and compute_instance_min_idle=2. I don't understand why in this case we wouldn't want to bump up to account for min idle.

It seems to me like we could just set new_target = min_hosts_for_queue + compute_instance_min_idle. This could be done without regard to what is currently idle.

nick-stroud, line 280: I believe L280 could be substituted on to be re-written as:
```
step1: min_hosts_for_queue += self.compute_instance_min_idle - delta - n_idle_total
step2: min_hosts_for_queue = min_hosts_for_queue + self.compute_instance_min_idle - (min_hosts_for_queue - currentTarget) - n_idle_total
step3: min_hosts_for_queue = min_hosts_for_queue + self.compute_instance_min_idle - min_hosts_for_queue + currentTarget - n_idle_total
step4: min_hosts_for_queue = current_target + self.compute_instance_min_idle - n_idle_total
```

First, I would argue that step 4 as a statement transmits clearer meaning. Second, it seems odd to me that `min_hosts_for_queue` as calculated on L237 is not incorporated in this value.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: This was updated in a subsequent commit. I will resolve the conversation.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: I am going to resolve this conversation while quoting it in the other conversation in same set of lines.

tpdownes: Quote from https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/983#discussion_r1127092373 (resolved and tightly coupled to this conversation)

> This block (L278,279,280) feels is quite confusing.
>
> I don't understand why we have the if statement. What if delta=10 and compute_instance_min_idle=2. I don't understand why in this case we wouldn't want to bump up to account for min idle.
>
> It seems to me like we could just set new_target = min_hosts_for_queue + compute_instance_min_idle. This could be done without regard to what is currently idle.

tpdownes: I did originally write the block as

```python
min_hosts_for_queue = current_target + self.compute_instance_min_idle - n_idle_total
```

and had trouble justifying each term. So both this and "why if?" are good questions. I've given some thought and seen it "in the wild" and have come to the following proposal. In pseudo-Python:

```python
# no if block
new_target = current_target + self.compute_instance_min_idle - n_idle_total + n_request_cpus_in_idle_jobs
```

Each term is individually justifiable, using some basic feedback loop concepts:

* `current_target`: a stable feedback loop ought to naturally stay at its current value in the absence of other signals
* `self.compute_instance_min_idle - n_idle_total`: a feedback signal restoring the target to whatever value is necessary to maintain `min_idle`
* `n_request_cpus_in_idle_jobs`: a feedback signal setting the target to the most compact set of CPUs that will meet the needs of idle jobs

@nick-stroud let me know if the summary above seems (a) better in terms of scaling accuracy (b) better in terms of the equation seeming "natural"

I have also reached a couple other conclusions:

* the counting of idle hosts and idle jobs should account for whether they have participated a negotiation cycle. i.e. an idle host or job that hasn't even attempted to match yet isn't a growth/contraction signal. I believe that this will help account for imperfect job packing onto machines
* right now, there is a bias toward deleting machines that already exist vs machines that are just being created and haven't yet joined the pool. the bias should probably go in the other direction.

mr0re1 (approved):

community/modules/scripts/htcondor-install/files/autoscaler.py
nick-stroud: The new logic seems much clearer to me. Basically everything now focus on idle jobs and idle nodes.

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -257,65 +249,110 @@ def scale(self):
...
+                idle_nodes.move_to_end(self_link)
+            idle_nodes[self_link] = "idle"
+        n_idle = len(idle_nodes)
+
+        print(f"There are {n_idle} VMs being created or idle in the pool")
+        if self.debug > 1:
+            print("Listing idle nodes:")
+            pprint(idle_nodes)
+
+        # always keep size tending toward the minimum idle VMs requested
```
nick-stroud, line 318: nit: This function is 200 lines long. My general rule of thumb is that growing beyond 100 usually means there is an opportunity to split up. To me it feels like the real meat of the scaling starts here and below. Much (but not all) of what is above this line is collecting information to perform the scaling. 

How about creating helper functions to gather each piece of critical information?

community/modules/scripts/htcondor-install/files/autoscaler.py
```diff
@@ -185,6 +185,10 @@ def scale(self):
         # stewardship. A full list of Job ClassAd attributes can be found at
         # https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html
         schedd = htcondor.Schedd()
+        # encourage the job queue to start a new negotiation cycle
+        # there are internal unconfigurable rate limits so not guaranteed
+        schedd.reschedule()
+        time.sleep(3)
```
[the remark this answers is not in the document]
tpdownes: All this really does is increase the odds that the filter on jobs allows very recently submitted jobs to pass. Thus the autoscaling is achieved with lower latency.

That said, the architecture of HTCondor is starting to allow the scheduler to do some of its own negotiating, especially when it rapidly processes a large queue of jobs with identical requests. In this case, the negotiator cycle is only part of the job-matching picture in the first place.

My top bullet for the next meeting with HTCondor team (early April) is how to best identify whether a job has participated in a negotiation cycle or direct scheduling by scheduler. I didn't find an obvious answer.

If we wanted to be very strict about this, we could repeatedly query the negotiator until `LastNegotiationCycleTime0` was recent. I *think* this might be nearly guaranteed of succeeding every 20 seconds, but it's a different form of brute force. In practice, at small scale, I didn't observe the sleep to be necessary to see the timestamp update within the same process.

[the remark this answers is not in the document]
nick-stroud: IIUC a natural negotiation cycle might happen every 20 seconds. You are forcing one to happen and then waiting, hoping that it has concluded. Is that correct? Is this really necessary for your scaling algorithm. It seems like the way you have it now where idle jobs and idle nodes cancel out that it wouldn't matter that there was not a recent negotiation.

To my original question, it seems like you are answering "no". I am just programed to always question a sleep in production code that is not tied to some state event. 

If you think we really need it then I would add a comment as to why 3. Otherwise it just looks like a magic number.

community/modules/scripts/htcondor-install/files/autoscaler.py
tpdownes: ACK

community/modules/scripts/htcondor-install/files/autoscaler.py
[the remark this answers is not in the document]
tpdownes: I am removing the "magic number" (a fair provocation) in favor of a comment that clarifies that the encouragement of the negotiation cycle is nice, but not necessary.

PR #999, opened 2023-03-08
Clean up Filestore regardless of instances presence

mr0re1:
* Skip clean up IFF there are active builds;
    * Remove Filestore instances.

### Submission Checklist

  install`)

### DIFF
--- a/tools/clean-filestore-limit.sh
+++ b/tools/clean-filestore-limit.sh
@@ -23,43 +23,49 @@ if [ -z "$PROJECT_ID" ]; then
 fi
 
 ACTIVE_BUILDS=$(gcloud builds list --project "${PROJECT_ID}" --filter="id!=\"${BUILD_ID}\"" --ongoing 2>/dev/null)
-ACTIVE_FILESTORE=$(gcloud filestore instances list --project "${PROJECT_ID}" --format='value(name)')
-if [[ -z "$ACTIVE_BUILDS" && -z "$ACTIVE_FILESTORE" ]]; then
-	echo "Disabling Filestore API..."
-	gcloud services disable file.googleapis.com --force --project "${PROJECT_ID}"
-
-	echo "Deleting all Filestore peering networks"
-	# the output of this command matches
-	# filestore-peer-426414172628;filestore-peer-646290499454 default
-	peerings=$(gcloud compute networks peerings list --project "${PROJECT_ID}" --format="value(peerings.name,name)")
-	while read -r peering; do
-		# split the output into:
-		# 0: a semi-colon separated list of peerings
-		# 1: the name of a VPC network
-		read -ra parr <<<"$peering"
-		# split the list of peerings into an array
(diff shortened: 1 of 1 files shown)


tools/clean-filestore-limit.sh
```diff
@@ -23,43 +23,49 @@ if [ -z "$PROJECT_ID" ]; then
 fi
 
 ACTIVE_BUILDS=$(gcloud builds list --project "${PROJECT_ID}" --filter="id!=\"${BUILD_ID}\"" --ongoing 2>/dev/null)
-ACTIVE_FILESTORE=$(gcloud filestore instances list --project "${PROJECT_ID}" --format='value(name)')
-if [[ -z "$ACTIVE_BUILDS" && -z "$ACTIVE_FILESTORE" ]]; then
-	echo "Disabling Filestore API..."
-	gcloud services disable file.googleapis.com --force --project "${PROJECT_ID}"
+if [[ -n "$ACTIVE_BUILDS" ]]; then
+	echo "There are active Cloud Build jobs. Skip clean up."
+	exit 0
```
cboneti, line 28: I am wondering if you should not fail the build to clearly signal it did not run.

tools/clean-filestore-limit.sh
```diff
@@ -23,43 +23,49 @@ if [ -z "$PROJECT_ID" ]; then
...
-	while read -r peering; do
-		# split the output into:
-		# 0: a semi-colon separated list of peerings
-		# 1: the name of a VPC network
-		read -ra parr <<<"$peering"
-		# split the list of peerings into an array
-		IFS=";" read -ra peers <<<"${parr[0]}"
-		# capture the VPC network
-		network=${parr[1]}
+ACTIVE_FILESTORE=$(gcloud filestore instances list --project "${PROJECT_ID}" | tail -n +2 2>/dev/null)
```
cboneti, line 31: this needs test, but I am pretty sure that simply disabling the API will delete all existing instances. Some of the following may still be interesting to explicitly show that the instances were deleted...

cboneti (approved):
Two minor comments, you may choose to change nothing and move forward with this. 
If you make any minor changes (return value), feel free to submit without reapproval.

cboneti:
Also, please make sure you manually test this trigger before merging :)

tools/clean-filestore-limit.sh
mr0re1: > this needs test, but I am pretty sure that simply disabling the API will delete all existing instances.

This was my assumption as well, but it still complained about resources.
`--force` helps with deleting APIs that depend on requested one along the way, but it doesn't delete resources.

tools/clean-filestore-limit.sh
mr0re1: Good point, brining attention may be helpful. Can make it silent in the future if it proves to be noisy/useless.

cboneti (approved):

PR #1000, opened 2023-03-08
Specify strict type for labels = map(string)

mr0re1:
### Submission Checklist

  install`)


tpdownes (approved):
I like this a lot. I have a sneaking suspicion that we don't enforce the Go type of `vars.labels` very well and might very well attempt construct a value for labels in `terraform.tfvars` that is an object of mixed types. (e.g. all keys as strings, but some values as non-strings)

That said, I believe the behavior of Terraform in this case would be to convert the object in terreaform.tfvars to a "least common denominator" map, probably of type map(string). If this PR causes an edge case problem, I would argue that it is a problem worth finding and resolving, and it will probably be in how we represent `Vars` inside Go.

PR #1004, opened 2023-03-09
Actually check mount instead of just checking dir exists

nick-stroud:
This change:
- Resolves issue that we were only testing the default partition (but still running it multiple times). Now we test all partitions.
- Checks if a path has been mounted instead of just checking if a directory exists.

### Submission Checklist

  install`)


tpdownes (approved):
Because the check is named "Check mounts on ....". I find it a bit confusing that it appears in the logs as having been "skipped" when things are good.

If you wanted to address this nit, you might rename it "Fail if partitions unmounted". But the functionality is here.

nick-stroud:
Updated with suggestion.

PR #1009, opened 2023-03-09
Prefer local_mount (non-plural) in nfs-server to match project convention

nick-stroud:
This naming mismatch has always bothered me. I expect some discussion. 

Validation is caught on terraform plan or apply (before any resources are created).

Note there is a known bad case:
```
    settings: # the bad case
      local_mounts: [/share]
      local_mount: /data
```

In this case there will be no error and the local mount will be `/share` as we won't recognize that the user is overriding the default with the default. I think this is a reasonable bad case to allow for the benefit of having the default show in the docs.

Additional Work:
- Update examples

### Submission Checklist

  install`)

### DIFF
--- a/community/modules/file-system/nfs-server/variables.tf
+++ b/community/modules/file-system/nfs-server/variables.tf
@@ -100,10 +100,20 @@ variable "scopes" {
   default     = ["https://www.googleapis.com/auth/cloud-platform"]
 }
 
-variable "local_mounts" {
+variable "local_mount" {
   description = "Mountpoint for this NFS compute instance"
+  type        = string
+  default     = "/data"
+  validation {
+    condition     = substr(var.local_mount, 0, 1) == "/"
+    error_message = "Local mountpoint must start with '/'."
+  }
+}
+
+variable "local_mounts" {
+  description = "DEPRECATED: Use `local_mount` instead: Mountpoint for this NFS compute instance"
   type        = list(string)
-  default     = ["/data"]
+  default     = []
 
   validation {
     condition = alltrue([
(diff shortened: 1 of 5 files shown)


community/modules/file-system/nfs-server/variables.tf
```diff
@@ -112,7 +122,7 @@ variable "local_mounts" {
     error_message = "Local mountpoints have to start with '/'."
   }
   validation {
-    condition     = length(var.local_mounts) > 0
-    error_message = "At least one local mount must be specified in var.local_mounts."
+    condition     = var.local_mounts != null
```
mr0re1, line 125: nit. Can we set `nullable = false` instead ?

tpdownes (commented):
Forgive the stupid question: is the `local_mount` the target of a mount operation, or the root of the NSF export?

The name strongly suggests the former, but a read of the code suggests the latter.

community/modules/file-system/nfs-server/variables.tf
tpdownes: `nullable = false` does 2 things:

1. The value can never be `null` _inside the module_ 
2. If a parent module of this module uses this as a child module and supplied `null` to `var.local_mounts`, then `var.local_mounts` takes on its default value

In the context of this PR, 1 is desirable and 2 is probably not desirable. That said, the way I read the [docs] suggests that if you remove the default value, set `nullable = false`, you might have the equivalent of the validation block.

https://developer.hashicorp.com/terraform/language/values/variables#disallowing-null-input-values

Having said _that_, if it works at all, it might not produce a useful error message. I believe that this error message should be improved to guide the user directly to `var.local_mount` or to web-based documentation about migrating to `var.local_mount`.

nick-stroud:
Closing as this has grown stale. I will re-open when I get a chance.

PR #1020, opened 2023-03-13
Double length of time for HTCondor integration test to detect job queue

tpdownes:
### Submission Checklist

  install`)


tpdownes:
The SCHEDD_INTERVAL we configure is 30s (well below the default of 300s) which is only giving the current 6*10=60s roughly 2 intervals of time for the SCHEDD to advertise itself to the collector. This PR doubles that time.

nick-stroud (approved):

PR #1022, opened 2023-03-14
Remove ModuleToGroup from DeploymentConfig

mr0re1:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -298,9 +298,7 @@ func (mc *ModConnection) isEmpty() (isEmpty bool) {
 type DeploymentConfig struct {
 	Config Blueprint
 	// Indexed by Resource Group name and Module Source
-	ModulesInfo map[string]map[string]modulereader.ModuleInfo
-	// Maps module ID to group index
-	ModuleToGroup     map[string]int
+	ModulesInfo       map[string]map[string]modulereader.ModuleInfo
 	expanded          bool
 	moduleConnections []ModConnection
 }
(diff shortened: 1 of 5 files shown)


pkg/config/config_test.go
```diff
@@ -540,13 +540,6 @@ func (s *MySuite) TestHasKind(c *C) {
 
 }
 
-func (s *MySuite) TestCheckModuleAndGroupNames(c *C) {
```
tpdownes, line 543: The overall unit test coverage goes up by 0.1% but I think we are missing some important edge cases in this particular function.

After running `go tool cover -html=coverage.out hpc-toolkit/pkg/config` and `go tool cover -html=coverage.out` I see these fields:

```
                        if _, ok := seen[mod.ID]; ok {
                                return fmt.Errorf("%s: %s used more than once", errorMessages["duplicateID"], mod.ID)
                        }
```

and

```
                        } else if grp.Kind != mod.Kind {
                                return fmt.Errorf(
                                        "%s: deployment group %s, got: %s, wanted: %s",
                                        errorMessages["mixedModule"],
                                        grp.Name, grp.Kind, mod.Kind)
                        }
```

pkg/config/config.go
```diff
@@ -471,47 +469,55 @@ func validateGroupName(name string, usedNames map[string]bool) {
...
 					errorMessages["mixedModule"],
 					grp.Name, grp.Kind, mod.Kind)
 			}
 		}
 	}
-	return moduleToGroup, nil
+	return nil
+}
+
+func modToGrp(groups []DeploymentGroup, modID string) (int, error) {
```
tpdownes, line 498: I see this function fully covered by unit tests, probably indirectly. Do you feel that's sufficient?

pkg/config/config.go
```diff
@@ -471,47 +469,55 @@ func validateGroupName(name string, usedNames map[string]bool) {
...
+	for iGrp, grp := range groups {
+		for _, mod := range grp.Modules {
```
tpdownes, line 500: Should we consider [slices.ContainsFunc](https://pkg.go.dev/golang.org/x/exp/slices#ContainsFunc)? An example is already in pkg/config/expand.go. There is probably a marginal cost in performance, but one less thing.

pkg/config/config.go
```diff
@@ -471,47 +469,55 @@ func validateGroupName(name string, usedNames map[string]bool) {
...
+			if mod.ID == modID {
+				return iGrp, nil
+			}
+		}
+	}
+	return 0, fmt.Errorf("Module %s was not found", modID)
```
tpdownes, line 506: The behavior of [slices.Index](https://pkg.go.dev/golang.org/x/exp/slices#Index) is to return `-1` when it is not found. I think this is probably good behavior to imitate.

pkg/config/config_test.go
tpdownes: This was probably true before, but I'd rather improve this along the way while we are here.

tpdownes (changes requested):
This is a significant improvement by replacing an structure that is passed around from func to func with a single, clear source. The performance penalty is unlikely to be significant and the use of a func simplifies future re-factoring.

There are a few changes I'd like to see or discuss, but this is nearly good to go.

pkg/config/config.go
mr0re1: Thank you! I did forget that Go doesn't support negative indexes )

pkg/config/config.go
mr0re1: Quite the opposite, I'm disturbed by abundance of tests for "private" functions )

pkg/config/config_test.go
mr0re1: Added test for ` used more than once `. But unearthed other problems in this function, fixing + adding test is going out of scope of this PR.

tpdownes (approved):
LGTM. If you believe that a useful test is missing (mismatched Kind?), please create a task for it in the backlog.

Issue #1019, opened 2023-03-13
remote-desktop module should provision remote chromoting tools


proppy:
### Describe the bug

After deploying https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/community/modules/remote-desktop/chrome-remote-desktop and ssh'ing into the vm instance and running the chromoting setup command snippet according to the provided instructions, the chromoting tools don't seem to be installed on the system.

### Steps to reproduce

https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/community/modules/remote-desktop/chrome-remote-desktop/README.md#setting-up-the-remote-desktop

### Expected behavior

Setup process continue and ask the user to enter a pin.

### Actual behavior

The chromoting setup command snippet fails with the following command:
```
-bash: /opt/google/chrome-remote-desktop/start-host: No such file or directory
```

### Version (`ghpc --version`)

```
hpc-toolkit 👺 ./ghpc --version
ghpc version - not built from official release
Built from 'develop' branch.
Commit info: v1.14.0-84-g907d0e7d
```

### Blueprint

```yaml
blueprint_name: remote-desktop

vars:
  project_id: catx-demo-radlab
  deployment_name: radlab-remote-desktop
  region: us-central1
  zone: us-central1-c

deployment_groups:
- group: primary
  modules:
  - id: network1
    source: modules/network/vpc

  - id: remote-desktop
    source: community/modules/remote-desktop/chrome-remote-desktop
    use: [network1]
    settings:
      install_nvidia_driver: true

```

### Expanded Blueprint

If applicable, please attach or paste the expanded blueprint. The expanded blueprint can be obtained by running `ghpc expand your-blueprint.yaml`.

Disregard if the bug occurs when running `ghpc expand ...` as well.

```yaml
blueprint_name: remote-desktop
validators:
  - validator: test_module_not_used
    inputs: {}
  - validator: test_project_exists
    inputs:
      project_id: ((var.project_id))
  - validator: test_apis_enabled
    inputs: {}
  - validator: test_region_exists
    inputs:
      project_id: ((var.project_id))
      region: ((var.region))
  - validator: test_zone_exists
    inputs:
      project_id: ((var.project_id))
      zone: ((var.zone))
  - validator: test_zone_in_region
    inputs:
      project_id: ((var.project_id))
      region: ((var.region))
      zone: ((var.zone))
validation_level: 1
vars:
[...]

PR #1027, opened 2023-03-15
Add an integration test for chrome-remote-desktop module

nick-stroud:
This change adds a simple integration test to confirm the installation and startup of the chrome-remote-desktop module.

This test will help monitor for issues such as the one raised in #1019, and will prevent breakage to the CRD startup scripts as changes are made.

### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/daily-tests/integration-group-4.yaml
+++ b/tools/cloud-build/daily-tests/integration-group-4.yaml
@@ -20,7 +20,7 @@
 #    └── htcondor (group 4)
 #       └── Cloud Batch
 #          └── slurm-gcp-v5-ubuntu2004
-
+#             └── chrome-remote-desktop
 
 timeout: 14400s  # 4hr
 steps:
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/integration-group-4.yaml
```diff
@@ -107,3 +107,26 @@ steps:
     ansible-playbook tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml \
       --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT}" \
       --extra-vars="@tools/cloud-build/daily-tests/tests/slurm-v5-ubuntu.yml"
+
```
cboneti, line 110: We usually also update at line 15, a list of the tests in this group for easy visibility. If that is now off sync, we may want to remove it entirely or update it.

tools/cloud-build/daily-tests/tests/chrome-remote-desktop.yml
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+test_name: chrome-remote-desktop
+deployment_name: chrome-remote-desktop-{{ build }}
+zone: us-central1-c
+workspace: /workspace
+blueprint_yaml: "{{ workspace }}/tools/validate_configs/test_configs/remote-desktop.yaml"
```
cboneti, line 19: I believe the test configs blueprints were a dumping ground for stuff that just literally has to pass validation.

I am wondering if we should enforce a duplicated file under tools/cloud-build/daily-tests/tests/ as that makes it clear this blueprint is used for an integration test and so future changes to it are less liberal.  What do you think?

cboneti (commented):
Thank you for doing this. I added two comments.

tools/cloud-build/daily-tests/tests/chrome-remote-desktop.yml
nick-stroud: `tests/slurm-v5-startup-scripts.yml` & `tests/pbspro.yml` also reference a validation config. I would vote not to do it (duplicate the files) unless it becomes a problem.

tools/cloud-build/daily-tests/integration-group-4.yaml
nick-stroud: Updated.

tools/cloud-build/daily-tests/tests/chrome-remote-desktop.yml
cboneti: if there are precedents, then I'm ok with this. I would say I missed it in the past, but also not a huge deal since tests run every day.

cboneti (approved):

PR #1030, opened 2023-03-15
Fix broken test config use-resources.yaml

mr0re1:
```
$ ./ghpc create tools/validate_configs/test_configs/use-resources.yaml -l ERROR --vars="project_id=X"
module homefs uses module network1, but matching setting and outputs were not found. This may be because the value is set explicitly or set by a prior used module
module slurm_login uses module homefs, but matching setting and outputs were not found. This may be because the value is set explicitly or set by a prior used module
module slurm_login uses module scratchfs, but matching setting and outputs were not found. This may be because the value is set explicitly or set by a prior used module
module slurm_login uses module projectsfs, but matching setting and outputs were not found. This may be because the value is set explicitly or set by a prior used module
One or more used modules could not have their settings and outputs linked.
error: validator test_module_not_used failed
```

### Submission Checklist

  install`)

### DIFF
--- a/tools/validate_configs/test_configs/use-resources.yaml
+++ b/tools/validate_configs/test_configs/use-resources.yaml
@@ -36,7 +36,6 @@ deployment_groups:
     use: [network1]
     settings:
       local_mount: /home
-      network_id: $(network1.network_id)
 
 
   - id: projectsfs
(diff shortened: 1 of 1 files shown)


tools/validate_configs/test_configs/use-resources.yaml
```diff
@@ -71,11 +70,8 @@ deployment_groups:
   - id: slurm_login
     source: ./community/modules/scheduler/SchedMD-slurm-on-gcp-login-node
     use:
-    - homefs
-    - scratchfs
```
nick-stroud, line 75: Why are these being removed (homefs, & scratchfs)? They seem valid.

mr0re1: God spot! Addressed

nick-stroud (approved):

PR #1032, opened 2023-03-15
Check validations during `make tests`; Add flag to skip validators 

mr0re1:
* Add `Skip bool` to `validatorConfig`;
* Add flag `skip-validators` to `create` and `expand` commands;
* Change validation level IGNORE -> ERROR in validate_configs.sh;
* Add default validators even if others are present;

```
$ make && ./ghpc expand examples/hpc-cluster-small.yaml --vars="project_id=GG" --skip-validators="test_project_exists,test_apis_enabled"
```

expanded.yaml:

```
...
validators:
  - validator: test_project_exists
    inputs: {}
    skip: true
  - validator: test_apis_enabled
    inputs: {}
    skip: true
  - validator: test_module_not_used
    inputs: {}
    skip: false
...
```

### Submission Checklist

  install`)

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -41,8 +41,8 @@ func init() {
 		"Sets the output directory where the HPC deployment directory will be created.")
 	createCmd.Flags().StringSliceVar(&cliVariables, "vars", nil, msgCLIVars)
 	createCmd.Flags().StringSliceVar(&cliBEConfigVars, "backend-config", nil, msgCLIBackendConfig)
-	createCmd.Flags().StringVarP(&validationLevel, "validation-level", "l", "WARNING",
-		validationLevelDesc)
+	createCmd.Flags().StringVarP(&validationLevel, "validation-level", "l", "WARNING", validationLevelDesc)
+	createCmd.Flags().StringSliceVar(&validatorsToSkip, "skip-validators", nil, skipValidatorsDesc)
 	createCmd.Flags().BoolVarP(&overwriteDeployment, "overwrite-deployment", "w", false,
 		"If specified, an existing deployment directory is overwritten by the new deployment. \n"+
 			"Note: Terraform state IS preserved. \n"+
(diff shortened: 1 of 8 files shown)


tpdownes:
* What is the intended order of merging explicit and implicit validators?
* Is there a mechanism by which to not use the default validators?

mr0re1:
> The new validators block appears to allow the "skip" setting to default to False if no setting is there (i.e. the status quo). 

Yes default is false for to NOT TO SKIP, and preserves status quo.

> The behavior of blueprint "skip: false" is per-validator while the behavior of CLI "--skip-validators" will skip all validators using the underlying function (i.e. you can implement multiple test_project_exists validators, but the CLI tool will skip all of them)

Technically this take is incorrect, there are three type of objects involved:
* **validator-config** - items in `Blueprint.Validators`;
* **validator** - type referring to implementation (many-to-one);
* **validator-implementation** - actual implementation.

So to rephrase your statement "correctly":
> The behavior of blueprint "skip: false" is per-~validator~-**validor-config** while the behavior of CLI "--skip-validators" will skip validators  ~using the underlying function~ **by type** (i.e. you can ~implement~ **provide inputs for** multiple test_project_exists validators, but the CLI tool will skip all of them)

> What is the intended order of merging explicit and default validators?

Before this PR: add default validators IFF no explicit validators were configured.
After this PR:  add default validators IFF no explicit validators of the same type were configure.

> Is there a mechanism by which to not use the default validators?
Add `--skip-validators="<names of all default validators>"`

tpdownes:
Thank you for the confirmation and clarification for 1 and 2. My conclusions for 3 and 4:

- (3) is a significant improvement over the status quo
- We should have something simpler for (4)

The documented solution for (4) is to set the validator-configs block to the empty list.

https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/blueprint-validation.md#explicit-validators

If I understand (3) correctly, your solution probably breaks the documented approach (`[]`). That said, I never liked setting it at the blueprint level in the first place. 2 thoughts:

1. Would it be simple to add a flag such as `--skip-all-validators` or a special `--skip-validators="all"`?
2. Ensure that the validator failure message includes a link that advises the user how to specify the flag in (1)

I like that we are getting toward "everyone should validate all the time" but the fact of the matter is that validation will fail for inexperienced users who have not properly authenticated and it may also be desirable to skip validation in some other settings (e.g. CI/CD). I would like to make sure that these users are given opportunities to follow the links for support or **easily** "shut all validators up" if they so desire.

```
validators are shown below.

```

mr0re1:
> Would it be simple to add a flag such as --skip-all-validators or a special --skip-validators="all"?
>  easily "shut all validators up" if they so desire.

One can just use `-l IGNORE` for this purpose.

> --skip-validators="all"

Feels like an unneeded sugar that can backfire.

>  to follow the links for support 

Currently we already provide this:
```
```

> your solution probably breaks the documented approach ([]).

This PR also updates documentation.

tpdownes (approved):

PR #1036, opened 2023-03-16
Fix dtype of slurm node-group.preemtible to bool

mr0re1:
The rest of slurm modules have inputs `preemtible` typed as `bool`.
```
name:  preemptible

type:  bool
decalred:
  schedmd-slurm-gcp-v5-controller
  schedmd-slurm-gcp-v5-login

type:  string
decalred:
  schedmd-slurm-gcp-v5-node-group
```

### Submission Checklist

  install`)


tpdownes (approved):
A quick survey of slurm-gcp (at SchedMD) suggests that preemptible is used throughout as a bool.

I am approving, but I'd appreciate if you tested the theory that a customer used "false" or "true" (as strings) in a blueprint, and that after `ghpc create`, Terraform converts the string value to a boolean for you:

https://developer.hashicorp.com/terraform/language/expressions/type-constraints#conversion-of-primitive-types

if it does that, then I consider this a non-breaking change and you should merge. if not, then I'd like to consider it a bit more.

Either way, it's a positive change. I don't like how Terraform treats string and booleans quite so interchangeably.

mr0re1:
> I am approving, but I'd appreciate if you tested the theory that a customer used "false" or "true" (as strings) in a blueprint, and that after ghpc create, Terraform converts the string value to a boolean for you:

The "stringness" is preserved in created TF.

_main.tf_:
```
module "compute_node_group" {
  source      = "./modules/schedmd-slurm-gcp-v5-node-group"
  project_id  = var.project_id
  preemptible = "false"
  labels      = merge(var.labels, { ghpc_role = "compute",})
}
```

But `plan` and `apply` don'y mind it.

PR #1040, opened 2023-03-17
Make readers usable outside of package modulereader

mr0re1:
Exported readers can't be used outside of package because they require initialization of un-exported state.

- Add constructors;
- Refactoring of related tests.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulereader/packerreader.go
+++ b/pkg/modulereader/packerreader.go
@@ -31,6 +31,11 @@ type PackerReader struct {
 	allModInfo map[string]ModuleInfo
 }
 
+// NewPackerReader is a constructor for PackerReader
+func NewPackerReader() PackerReader {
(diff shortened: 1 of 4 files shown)


tpdownes (commented):
Could you describe the functionality that this will enable?

mr0re1:
> Could you describe the functionality that this will enable?

1. It's just a fix for a bug "Exported but useless in deceiving way", it could have been also addressed by un-exporting readers, but I found them useful in (2).
2. I got burned by this while writing module interface inspector tool outside of this package. One could use `GetModuleInfo`, but it's 
    * "less explicit" if intention to read local modules of particular kind only;
    * static with state, so it may have negative side effects.

tpdownes (approved):
LGTM. I agree that we do not have a clear theory of what is/not exported in our packages and this improves matters.

PR #1056, opened 2023-03-20
Upgrade pip in builder image

tpdownes:
### Submission Checklist
Upgrade pip in builder image to enable dry-run installation test in #1054 

* [x] Have you installed and run this change against pre-commit? (`pre-commit
  install`)
* [x] Are all tests passing? (`make tests`)
* [ ] Have you written unit tests to cover this change?
* [x] Is unit test coverage still above 80%?
* [x] Have you updated all applicable documentation?
* [x] Have you followed the guidelines in our Contributing document?

PR #1057, opened 2023-03-20
Add simple sanity test installing the OFE virtual environment

tpdownes:
### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-build/hpc-toolkit-pr-validation.yaml
+++ b/tools/cloud-build/hpc-toolkit-pr-validation.yaml
@@ -53,6 +53,20 @@ steps:
     time addlicense -check -ignore **/lysozyme-example/submit.sh . \
       || { echo "addlicense failed"; exit 1; }
     time make tests
+- id: ofe-virtual-env
+  waitFor:
+  - git-fetch-unshallow
+  name: >-
+    us-central1-docker.pkg.dev/$PROJECT_ID/hpc-toolkit-repo/hpc-toolkit-builder
+  entrypoint: /bin/bash
+  args:
+  - '-c'
+  - |
+    set -e
+    python3 -m venv /opt/ofe
+    source /opt/ofe/bin/activate
+    pip install --upgrade pip
+    pip install --dry-run --no-cache-dir -r community/front-end/ofe/requirements.txt
 timeout: "1200s"
 options:
   machineType: N1_HIGHCPU_8


tools/cloud-build/hpc-toolkit-pr-validation.yaml
```diff
@@ -53,6 +53,20 @@ steps:
...
+    us-central1-docker.pkg.dev/$PROJECT_ID/hpc-toolkit-repo/hpc-toolkit-builder
+  entrypoint: /bin/bash
+  args:
+  - '-c'
+  - |
+    set -e
+    python3 -m venv /opt/ofe
+    source /opt/ofe/bin/activate
+    pip install --upgrade pip
+    pip install --dry-run --no-cache-dir -r community/front-end/ofe/requirements.txt
```
nick-stroud, line 69: IIUC this will crash without a new version of pip. Is that new version of pip implemented in our weekly builder? If not I don't think it is a good idea to add this test knowing it is not compatible with the automated builder.

tpdownes: It is in our weekly builder #1056 and the copy inside the virtual environment is upgraded again in the line above your comment. Doing both may be overkill, of course.

tpdownes: I manually upgrade the weekly image and you can see the results in the PR check ("ofe-virtual-env")

tools/cloud-build/hpc-toolkit-pr-validation.yaml
```diff
@@ -53,6 +53,20 @@ steps:
     time addlicense -check -ignore **/lysozyme-example/submit.sh . \
       || { echo "addlicense failed"; exit 1; }
     time make tests
+- id: ofe-virtual-env
+  waitFor:
+  - git-fetch-unshallow
```
nick-stroud, line 58: nit: To me it feels heavy handed to run this on every PR. Looking at recent builds it does not seem to have increased validation time, which is my reason to not block on this comment. I would propose creating a separate build trigger that is only triggered when this file is changed. You can use the `Included files` field in the trigger. https://cloud.google.com/build/docs/automating-builds/create-manage-triggers

nick-stroud (approved):

tools/cloud-build/hpc-toolkit-pr-validation.yaml
tpdownes: I agree with the claim and the recommendation. That said, the reason to put it in PR validation is that its standard output has been made visible to the public. The balance I chose to strike was to run it in parallel with the other (longer) tests.

PR #1059, opened 2023-03-20
Order settings alphabettically and format main.tf

mr0re1:
* Introduce a stable order to simplyfy comparison of produced deployments;
* Use canonical HCL formatter before writing main.tf down. 

Example of produced module:
```
module "homefs" {
  source          = "./modules/filestore"
  deployment_name = var.deployment_name
  labels          = merge(var.labels, { ghpc_role = "file-system", })
  local_mount     = "/home"
  network_id      = module.network1.network_id
  project_id      = var.project_id
  region          = var.region
  zone            = var.zone
}
```

### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulewriter/tfwriter.go
+++ b/pkg/modulewriter/tfwriter.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2/hclsyntax"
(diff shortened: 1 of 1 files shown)


tpdownes (commented):
It would be outside the norms for the output not to begin with `source` as the first key.

Common next value is often project_id (if present). After that, any patterns may be hard to automate.

mr0re1:
> It would be outside the norms for the output not to begin with source as the first key.

`source` is not a part of settings, so it stays on first place.

> Common next value is often project_id (if present).

We can do it later if we want to.

PR #1077, opened 2023-03-23
Add additional documentation on enable_reconfigure troubleshooting

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
@@ -11,11 +11,13 @@ The [user guide][slurm-ug] provides detailed instructions on customizing and
 enhancing the Slurm on GCP cluster as well as recommendations on configuring the
 controller for optimal performance at different scales.
 
-> **_WARNING:_** The variables [enable\_reconfigure], [enable\_cleanup\_compute] and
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -82,7 +85,19 @@ This option has some additional requirements:
 * The project in your gcloud config must match the project the cluster is being
   deployed onto due to a known issue with the reconfigure scripts. To set your
   default config project, run the following command:
-  `gcloud config set core/<<PROJECT ID>>`
+
+  ```bash
+  gcloud config set core/<<PROJECT ID>>
```
tpdownes, line 90: do you mean `gcloud config set core/project <<PROJECT_ID>>`

I think an alternative might be ``gcloud config set project <<PROJECT_ID>>`

tpdownes (approved):
I may be mistaken, but I think the main recommendation has a mistake in it. If I am correct, please fix it, then squash+merge.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
nick-stroud: Good catch! Both your suggestions worked for me. I will stick with `core/project` as that was the original intention.

PR #1078, opened 2023-03-23
Refactoring to support edge-tracking in graph

tpdownes:
- Re-implement `modReference` and `varReference` structs to implement a new interface named `reference`
- Adopt the `reference` interface in the `ModConnection` struct that tracks edges within the blueprint graph

These changes represent some design elements considered during implementation of references but weren't quite yet necessary to glue together.

Coverage in expand.go drops slightly due to the inclusion of new getter functions that seem too simple to merit unit testing. There appears to be no way to mark code as unreachable/not-meriting-testing for coverage percentages. This is area of active discussion in various Go channels.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -271,11 +271,8 @@ const (
 // ModConnection defines details about connections between modules. Currently,
 // only modules connected with "use" are tracked.
 type ModConnection struct {
-	toID   string
-	fromID string
-	// Currently only supports useConnection
-	kind ConnectionKind
-	// List of variables shared from module `fromID` to module `toID`
+	ref             reference
+	kind            ConnectionKind
 	sharedVariables []string
 }
 
@@ -322,7 +319,8 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
 	unusedModules := make(map[string][]string)
 	for _, conn := range dc.moduleConnections {
 		if conn.isEmpty() {
-			unusedModules[conn.fromID] = append(unusedModules[conn.fromID], conn.toID)
+			fromMod := conn.ref.getFromModuleID()
+			unusedModules[fromMod] = append(unusedModules[fromMod], conn.ref.getToModuleID())
 		}
 	}
 	return unusedModules
@@ -508,8 +506,8 @@ func modToGrp(groups []DeploymentGroup, modID string) (int, error) {
 
 // checkUsedModuleNames verifies that any used modules have valid names and
 // are in the correct group
-func checkUsedModuleNames(depGroups []DeploymentGroup) error {
-	for _, grp := range depGroups {
+func checkUsedModuleNames(bp Blueprint) error {
+	for _, grp := range bp.DeploymentGroups {
 		for _, mod := range grp.Modules {
 			for _, usedMod := range mod.Use {
 				ref, err := identifyModuleByReference(usedMod, grp)
@@ -517,7 +515,7 @@ func checkUsedModuleNames(depGroups []DeploymentGroup) error {
 					return err
 				}
 
-				if err = ref.validate(depGroups); err != nil {
+				if err = ref.validate(bp); err != nil {
 					return err
 				}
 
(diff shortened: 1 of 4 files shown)


pkg/config/expand.go
```diff
@@ -513,20 +512,46 @@ type varContext struct {
 	blueprint  Blueprint
 }
 
+type reference interface {
+	validate(Blueprint) error
+	isIntergroup() bool
+	String() string
+	getFromModuleID() string
+	getToModuleID() string
```
nick-stroud, line 520: I think it is idiomatic for Go to drop the `get` from the getters: https://go.dev/doc/effective_go#Getters 

There is a lack of clarity there because it mentions the lower/upper case being sufficient to tell them apart. You are making these fields private so that difference is not present. Thoughts? It also seems odd to me that all of the underlying data is public but the interface methods are private, should it not be the other way around?

pkg/config/expand.go
```diff
@@ -513,20 +512,46 @@ type varContext struct {
...
+func (ref modReference) isIntergroup() bool {
+	return ref.ToGroupID == ref.FromGroupID
+}
+
+func (ref modReference) getFromModuleID() string {
+	return ref.FromModuleID
+}
+
+func (ref modReference) getToModuleID() string {
+	return ref.FromModuleID
```
nick-stroud, line 554: I believe this is returning the wrong variable.

pkg/config/expand.go
```diff
@@ -567,19 +592,37 @@ func identifyModuleByReference(yamlReference string, dg DeploymentGroup) (modRef
...
+func (ref varReference) isIntergroup() bool {
+	return ref.ToGroupID == ref.FromGroupID
+}
+
+func (ref varReference) getFromModuleID() string {
+	return ref.FromModuleID
+}
+
+func (ref varReference) getToModuleID() string {
+	return ref.FromModuleID
```
nick-stroud, line 625: I believe this is returning the wrong variable.

pkg/config/expand.go
```diff
@@ -275,7 +275,7 @@ func (dc *DeploymentConfig) applyUseModules() error {
 
 				// this module contains information about the target module that
 				// was specified by the user in the blueprint
-				toMod, err := toGroup.getModuleByID(modRef.ID)
+				toMod, err := toGroup.getModuleByID(modRef.ToModuleID)
```
nick-stroud, line 278: nit: I would recommend using the getter now that you have it, here and elsewhere, and consider making the struct variable private.

pkg/config/config.go
```diff
@@ -322,7 +319,8 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
 	unusedModules := make(map[string][]string)
 	for _, conn := range dc.moduleConnections {
 		if conn.isEmpty() {
-			unusedModules[conn.fromID] = append(unusedModules[conn.fromID], conn.toID)
+			fromMod := conn.ref.getFromModuleID()
+			unusedModules[fromMod] = append(unusedModules[fromMod], conn.ref.getToModuleID())
```
nick-stroud, line 323: nit: OCD doesn't like the different treatment of from and to here. One is stored separate and the other is inline. You could initialize both on one line (L322) with something like `x, y := a, b`.

pkg/config/expand.go
tpdownes: Modified on PR branch.

pkg/config/expand.go
tpdownes: Modified on PR branch.

pkg/config/expand.go
tpdownes: Out-of-band discussion resulted in a plan to use getters where necessary and unexported variables within the Go package.

pkg/config/expand.go
nick-stroud: I believe this is done, close if you agree.

pkg/config/expand.go
nick-stroud: I believe this is done, close if you agree.

pkg/config/expand.go
nick-stroud: I believe this is done, close if you agree.

pkg/config/config.go
tpdownes: My thinking here was to avoid 2 func calls rather than cleanliness. If you bump this above nit, glad to address it. Otherwise, "ack".

pkg/config/expand.go
nick-stroud: Discussed offline. Fields are now private. We can switch to using getters later if there is value.

nick-stroud (approved):

PR #1081, opened 2023-03-23
Remove field Module.ModuleName add DeploymentSource instead

mr0re1:
Remove any source-related logic from tfwritter, simply use DeploymentSource.

```
$ make &&  ./ghpc create examples/image-builder.yaml --vars=$TOOLKIT_DEFAULT_VARS --out=img_new
$ diff -r ~/tmp/img_orig/ img_new
$
```

### Submission Checklist

  install`)


pkg/config/config.go
```diff
@@ -226,9 +226,9 @@ func (g DeploymentGroup) HasKind(kind string) bool {
 // Module stores YAML definition of an HPC cluster component defined in a blueprint
 type Module struct {
 	Source           string
+	DeploymentSource string `yaml:"-"` // prevent user from specifying it
```
nick-stroud, line 229: 1. Can we get a doc string with a 1-liner description for each field. Something like [this](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/4787cdeb81237b71db533bede2199d38d772dec5/pkg/config/expand.go#L540-L547).
2. The name `DeploymentSource` is a little misleading for me. I think what you are getting at is, "when we write the deployment, this will be the place from which to source the module." Maybe with some documentation that will be fine. But since we are using it in `CopySource` function it seems easy to get it confused with "the source of the original module (which is Source)". Maybe `DeployDirPath` but I guess that doesn't make sense with git source... Shrug, maybe just start with the doc string and that will be enough.

pkg/modulewriter/modulewriter.go
```diff
@@ -135,35 +135,38 @@ func createGroupDirs(deploymentPath string, deploymentGroups *[]config.Deploymen
 
 func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) error {
 
-	for iGrp, grp := range *deploymentGroups {
+	for iGrp := range *deploymentGroups {
+		grp := &(*deploymentGroups)[iGrp]
```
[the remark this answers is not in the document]
mr0re1: > do something clean
Yes, this was my only motivation to do it this way.

pkg/config/config.go
mr0re1: +1 for doc strings, though I don't want to write it for rest of the fields.
Added a line for `DeploymentSource`.

nick-stroud (approved):

PR #1082, opened 2023-03-24
Lustre example

cboneti:
### Submission Checklist

  install`)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -23,6 +23,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [hpc-cluster-amd-slurmv5.yaml](#hpc-cluster-amd-slurmv5yaml-) ![community-badge]
   * [cloud-batch.yaml](#cloud-batchyaml-) ![core-badge]
   * [batch-mpi.yaml](#batch-mpiyaml-) ![core-badge]
+  * [lustre.yaml](#lustreyaml-) ![core-badge]
   * [spack-gromacs.yaml](#spack-gromacsyaml--) ![community-badge] ![experimental-badge]
   * [omnia-cluster.yaml](#omnia-clusteryaml--) ![community-badge] ![experimental-badge]
   * [hpc-cluster-small-sharedvpc.yaml](#hpc-cluster-small-sharedvpcyaml--) ![community-badge] ![experimental-badge]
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -8,38 +8,41 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
 -->
```
nick-stroud, line 8: nit: I would have suggested doing the reorg in it's own commit.

examples/lustre.yaml
```diff
@@ -0,0 +1,40 @@
...
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: small-lustre
+  region: us-central1
+  zone: us-central1-c
+
+deployment_groups:
+- group: primary
+  modules:
+  - id: network1
```
nick-stroud, line 26: nit: I think blueprints look much cleaner with white space between each module.

examples/README.md
```diff
@@ -606,6 +406,232 @@ The blueprint contains the following:
 
 [batch-mpi.yaml]: ../examples/batch-mpi.yaml
 
+### [lustre.yaml] ![core-badge]
+
+Creates a DDN EXAScaler lustre file-system that is mounted in two client instances.
+
+The [DDN Exascaler Lustre](../community/modules/file-system/DDN-EXAScaler/README.md)
+file system is designed for high IO performance. It has a capacity of ~10TiB and is mounted at `/lustre`.
```
nick-stroud, line 414: nit: `default` capacity? You can set to higher correct?

examples/lustre.yaml
```diff
@@ -0,0 +1,40 @@
...
+    source: modules/network/pre-existing-vpc
+  - id: lustre
+    source: community/modules/file-system/DDN-EXAScaler
+    use: [network1]
+    settings:
+      local_mount: /lustre
+  - source: modules/compute/vm-instance
+    kind: terraform
```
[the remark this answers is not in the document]
cboneti: oh, good catch.

examples/lustre.yaml
cboneti: I went for compact. But I think you are right. Fixed.

examples/lustre.yaml
```diff
@@ -0,0 +1,40 @@
...
+    id: compute_instances
+    use: [network1, lustre]
+    settings:
+      name_prefix: client-vm
+      instance_count: 2
+      machine_type: n2-standard-2
```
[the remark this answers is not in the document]
cboneti: That was my concern. 
Also, should we start thinking about moving wait for startup to core?

examples/README.md
cboneti: I thought it was easier to bundle, since I was already there. It was a mistake and I reverted it now. PTAL

examples/README.md
cboneti: yup, fixed.

examples/README.md
```diff
@@ -8,38 +8,41 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
...
-* [Blueprint Schema](#blueprint-schema)
-* [Writing an HPC Blueprint](#writing-an-hpc-blueprint)
-  * [Blueprint Boilerplate](#blueprint-boilerplate)
-  * [Top Level Parameters](#top-level-parameters)
-  * [Deployment Variables](#deployment-variables)
-  * [Deployment Groups](#deployment-groups)
-* [Variables](#variables)
-  * [Blueprint Variables](#blueprint-variables)
-  * [Literal Variables](#literal-variables)
+* [Example Blueprints](#example-blueprints)
```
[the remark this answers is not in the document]
cboneti: moved to another commit.

examples/README.md
nick-stroud: SG. I was even saying 2 commits 1 PR would have been easier to review. Looks good now I think.

nick-stroud (approved):

examples/lustre.yaml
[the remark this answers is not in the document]
cboneti: I agree, but I think it is adding to the scope here. I think we should improve startup script to automatically add to the message of the day. Will make sure there is a task for that.

PR #1084, opened 2023-03-24
Improve version constraint on batch-job-template module

tpdownes:
This PR does not alter the version of the module that would be installed today but prevents future breaking upgrades.

### Submission Checklist

  install`)

### DIFF
--- a/modules/scheduler/batch-job-template/main.tf
+++ b/modules/scheduler/batch-job-template/main.tf
@@ -66,7 +66,7 @@ locals {
 
 module "instance_template" {
   source  = "terraform-google-modules/vm/google//modules/instance_template"
-  version = "> 7.6.0"
+  version = "~> 8.0"
 
   name_prefix        = var.instance_template == null ? "${local.job_id}-instance-template" : "unused-template"
   project_id         = var.project_id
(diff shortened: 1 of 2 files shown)


modules/scheduler/batch-job-template/main.tf
```diff
@@ -66,7 +66,7 @@ locals {
 
 module "instance_template" {
   source  = "terraform-google-modules/vm/google//modules/instance_template"
-  version = "> 7.6.0"
+  version = "~> 8.0"
```
nick-stroud, line 69: [This](https://developer.hashicorp.com/terraform/language/expressions/version-constraints#terraform-core-and-provider-versions) states to only constrain on a minimum. Thoughts?

I agree it seems kind of crazy to silently allow upgrade on major version without some intervention.

tpdownes: Here is the text I read

- When depending on third-party modules, require specific versions to ensure that updates only happen when convenient to you.
- For modules maintained within your organization, specifying version ranges may be appropriate if semantic versioning is used consistently or if there is a well-defined release process that avoids unwanted updates.

I believe bullet 2 is what applies to us.

nick-stroud (approved):

PR #1085, opened 2023-03-24
Resolve accidental destruction of startup-scripts

tpdownes:
When the startup-script module is used with VM instance templates that have create_before_destroy enabled (a common setting), a change to the script contents will result in the script being created, then destroyed. This commit resolves that by ensuring that all startup-script objects are created with unique, repeatable identifiers. This allows the new object to be created while destroying the old object with a different identifier.

### Submission Checklist

  install`)

### DIFF
--- a/modules/scripts/startup-script/main.tf
+++ b/modules/scripts/startup-script/main.tf
@@ -88,7 +88,7 @@ resource "google_storage_bucket" "configs_bucket" {
 resource "google_storage_bucket_object" "scripts" {
   # this writes all scripts exactly once into GCS
   for_each = local.runners_map
-  name     = "${local.storage_folder_path_prefix}${each.key}-${try(md5(each.value.content), filemd5(each.value.source))}"
+  name     = "${local.storage_folder_path_prefix}${each.key}-${substr(try(md5(each.value.content), filemd5(each.value.source)), 0, 4)}"
   content  = each.value.content
   source   = each.value.source
   bucket   = local.storage_bucket_name
--- a/modules/scripts/startup-script/main.tf
+++ b/modules/scripts/startup-script/main.tf
@@ -64,8 +64,7 @@ locals {
   stdlib = join("", local.stdlib_list)
 
   runners_map = { for runner in local.runners :
-    basename(runner["destination"])
-    => {
+    basename(runner["destination"]) => {
       content = lookup(runner, "content", null)
       source  = lookup(runner, "source", null)
     }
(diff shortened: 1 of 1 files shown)


tpdownes:
This PR will be followed up by another PR to update all internal use of startup-script to the appropriate ref.

When testing, please give extra consideration to runners specified with "source" rather than "content" as those are less covered in tests.

tpdownes:
An additional test that I performed but you should also consider: trigger a change to the template that does not involve the script and observe that nothing changes with the script. e.g. machine_type.

modules/scripts/startup-script/main.tf
```diff
@@ -89,9 +88,9 @@ resource "google_storage_bucket" "configs_bucket" {
 resource "google_storage_bucket_object" "scripts" {
   # this writes all scripts exactly once into GCS
   for_each = local.runners_map
-  name     = "${local.storage_folder_path_prefix}${each.key}"
-  content  = each.value["content"]
-  source   = each.value["source"]
+  name     = "${local.storage_folder_path_prefix}${each.key}-${try(md5(each.value.content), filemd5(each.value.source))}"
```
nick-stroud, line 91: I think it will be more user friendly if we truncate. I am thinking terraform cli output, pantheon ui, etc...

md5 spits out 32 chars. I think 6 or 7 should be more than sufficient (for ref git short sha is 7).

nick-stroud, line 91: This will have the effect of replacing every deployed startup script file on a version update re-apply. This in turn will update the master startup script, which reference the file names in buckets.

I think generally this won't be very impactful. It will result in update to metadata on VMs and on restart will run the new scripts fine. In the case of Batch it will force replacement of instance template, which I also don't think is a big deal.

So overall, let's move forward but just think it is worth pointing out, especially if you can think of some case I am missing. Feel free to close this comment.

tpdownes: If what you mean is "upon the next release every customer will experience a destroy/create" for every object managed by the startup-script module, even when the content has not changed, yes. It's essentially an unavoidable consequence of being chained to a create_before_destroy resource. But, that will propagate correctly as you describe and should be noted clearly in the release notes.

tpdownes: I understand that the git short sha is calculated dynamically based on the repo size.  But 7 is a reasonable guidepost.  I personally think there is value in knowing the md5sum from the object name and the line becomes a bit messy if you add substr inside of it.

But if you feel strongly about it, let's pick the first 8 characters as it divides evenly into 32, and do it once correctly. WDYT?

nick-stroud: If you think others would get value out of having the full md5 then you can add variable that lets you configure this functionality. But I don't think having a longer hash provides any value.

If the line gets too long and cannot be effectively wrapped, you can calculate it in a local block and add it to a dict or create a new dict.

I don't see why dividing evenly is necessary for truncating. I would argue that even 7 is far more than needed since we are really only ever distinguishing between two versions at a time. You could go with 4 digits. Any of these options would be better than 32.

PR #1086, opened 2023-03-24
Copy all embedded modules into deployment, use unique source for locals

mr0re1:
Modify logic of selecting deployment source. Rules are following:
* git source => keep the same source
* packer  => <mod.ID>
* embedded  => ./modules/embedded/<source>
* other => ./modules/<basename(source)>-<hash(abs(source))>

Copy all embedded modules into deployment folder

```
blueprint_name: lucky
vars: { deployment_name: mango }
deployment_groups:
- group: red
  modules:
  - {id: emb, source: modules/green}
  - {id: dev, source: ./modules/green}
  - {id: rel, source: ../../../../../../../tmp/green/}
  - {id: abs, source: /tmp/green}

$ terraform -chdir=mango/red init
Initializing modules...
- abs in modules/green-b6a4
- dev in modules/green-c71e
- emb in modules/embedded/modules/green
- rel in modules/green-b6a4

$ tree mango/red/ -d
mango/red/
└── modules
    ├── embedded
    │   ├── community
    │   │   └── modules
    │   │       ├── compute
    │   │       │   ├── htcondor-execute-point
    │   │       ...
    │   └── modules
    │       ├── compute
    │       │   └── vm-instance
    │       ├── green
    │       ...
    ├── green-b6a4
    └── green-c71e
```
### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulewriter/modulewriter.go
+++ b/pkg/modulewriter/modulewriter.go
@@ -141,11 +141,10 @@ func deploymentSource(mod config.Module) (string, error) {
 	if sourcereader.IsGitPath(mod.Source) {
 		return mod.Source, nil
 	}
-	switch mod.Kind {
-	case "packer":
+	if mod.Kind == "packer" {
 		return mod.ID, nil
-	case "terraform": // see below
-	default:
+	}
+	if mod.Kind != "terraform" {
 		return "", fmt.Errorf("unexpected module kind %#v", mod.Kind)
 	}
 
@@ -189,10 +188,8 @@ func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGrou
 	for iGrp := range *deploymentGroups {
 		grp := &(*deploymentGroups)[iGrp]
 		basePath := filepath.Join(deploymentPath, grp.Name)
-		if err := copyEmbeddedModules(basePath); err != nil {
-			return fmt.Errorf("failed to copy embedded modules: %v", err)
-		}
 
+		var copyEmbedded = false
 		for iMod := range grp.Modules {
 			mod := &grp.Modules[iMod]
 			ds, err := deploymentSource(*mod)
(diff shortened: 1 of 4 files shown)


pkg/modulewriter/modulewriter.go
```diff
@@ -133,38 +127,92 @@ func createGroupDirs(deploymentPath string, deploymentGroups *[]config.Deploymen
...
+
+	}
+	return nil
+}
+
+func copySource(deploymentPath string, deploymentGroups *[]config.DeploymentGroup) error {
 	for iGrp := range *deploymentGroups {
 		grp := &(*deploymentGroups)[iGrp]
 		basePath := filepath.Join(deploymentPath, grp.Name)
+		if err := copyEmbeddedModules(basePath); err != nil {
```
nick-stroud, line 192: We copy in embedded even if they are not using any embedded modules. I think we should only do this if an embedded module is used. Thoughts?

pkg/modulewriter/modulewriter.go
```diff
@@ -133,38 +127,92 @@ func createGroupDirs(deploymentPath string, deploymentGroups *[]config.Deploymen
...
+	if sourcereader.IsGitPath(mod.Source) {
+		return mod.Source, nil
+	}
+	switch mod.Kind {
+	case "packer":
+		return mod.ID, nil
+	case "terraform": // see below
+	default:
+		return "", fmt.Errorf("unexpected module kind %#v", mod.Kind)
+	}
```
nick-stroud, line 150: nit: 😬 I am not a huge fan of this control flow. It strikes me as a bit clever use a switch for return early with a fall through. You also are mixing patterns as you have several if return early and then the switch. My preference would be for two more return early if blocks like you have for the isGitPath.

```
	if sourcereader.IsGitPath(mod.Source) {
		return mod.Source, nil
	}
	if mod.Kind == "packer" {
		return mod.ID, nil
	}
	if mod.Kind != "terraform" {
		return "", fmt.Errorf("unexpected module kind %#v", mod.Kind)
	}
...
```

pkg/sourcereader/embedded_test.go
```diff
@@ -156,3 +156,15 @@ func (s *MySuite) TestGetModule_Embedded(c *C) {
 	expectedErr = "Source is not valid: .*"
 	c.Assert(err, ErrorMatches, expectedErr)
 }
+
+func (s *MySuite) TestGetModule_NilFs(c *C) {
+	ModuleFS = nil
+	r := EmbeddedSourceReader{}
+	c.Assert(r.GetModule("here", "there"), ErrorMatches, "embedded file system is not initialized")
```
nick-stroud, line 163: I think it is generally bad golang practice to directly test for error string. I will share links for alternatives.

nick-stroud (approved):

pkg/sourcereader/embedded_test.go
mr0re1: Acknowledged. Discussed offline.

PR #1094, opened 2023-03-28
Track graph edges from modules to deployment variables

tpdownes:
This PR adds support for tracking connections from all modules to deployment variables. Future PRs will make use of this functionality to

- detect unused deployment variables
- write only input variables that are used by a given deployment group

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -38,7 +38,7 @@ import (
 const (
 	expectedVarFormat        string = "$(vars.var_name) or $(module_id.output_name)"
 	expectedModFormat        string = "$(module_id) or $(group_id.module_id)"
-	unexpectedConnectionKind string = "connectionKind must be useConnection"
+	unexpectedConnectionKind string = "connectionKind must be useConnection or deploymentConnection"
 )
 
 var errorMessages = map[string]string{
@@ -267,12 +267,12 @@ type connectionKind int
 const (
 	undefinedConnection connectionKind = iota
 	useConnection
+	deploymentConnection
 	// explicitConnection
-	// globalConnection
 )
 
 func (c connectionKind) IsValid() bool {
-	return c == useConnection
+	return c == useConnection || c == deploymentConnection
 }
 
 // ModConnection defines details about connections between modules. Currently,
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -30,6 +30,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 	. "gopkg.in/check.v1"
 )
 
@@ -259,7 +261,16 @@ func getMultiGroupDeploymentConfig() DeploymentConfig {
 	matchingName := "test_match"
 
 	testModuleInfo0 := modulereader.ModuleInfo{
-		Inputs: []modulereader.VarInfo{},
+		Inputs: []modulereader.VarInfo{
+			{
+				Name: "deployment_name",
+				Type: "string",
+			},
+			{
+				Name: "host_project_id",
+				Type: "string",
+			},
+		},
 		Outputs: []modulereader.VarInfo{
 			{
 				Name: matchingName,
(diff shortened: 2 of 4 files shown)


pkg/config/config.go
```diff
@@ -285,14 +292,14 @@ type ModConnection struct {
 // Returns true if a connection does not functionally link the outputs and
 // inputs of the modules. This can happen when a module is connected with "use"
 // but none of the outputs of fromID match the inputs of toID.
-func (mc *ModConnection) isEmpty() (isEmpty bool) {
-	isEmpty = false
+func (mc *ModConnection) isUnused() bool {
+	var isUnused bool
```
mr0re1, line 296: Can you one-line it? 
`return mc.Kind == useConnection && len(mc.sharedVariables) == 0`

pkg/config/expand.go
```diff
@@ -504,7 +504,7 @@ type varContext struct {
 	varString  string
 	groupIndex int
 	modIndex   int
-	blueprint  Blueprint
+	dc         DeploymentConfig
```
mr0re1, line 507: I think (still fuzzy on Go var management) you may need to store a reference to DeploymentConfig here.
It only works now because ModuleConnection is a map and shallow copy of `dc` will point to the same map.

pkg/modulewriter/modulewriter.go
```diff
@@ -285,3 +289,29 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+
+func filterVarsByGraph(vars map[string]interface{}, group config.DeploymentGroup, graph map[string][]config.ModConnection) map[string]interface{} {
+	if graph == nil {
+		return vars
+	}
+
+	groupInputs := []string{}
+	for _, mod := range group.Modules {
+		connections, ok := graph[mod.ID]
+		if ok {
```
mr0re1, line 301: nit. Let's reduce number of levels by checking `!ok -> continue` instead

pkg/config/expand.go
tpdownes: Please review and let me know if you believe this is fragile.

pkg/config/config_test.go
```diff
@@ -473,6 +500,58 @@ func (s *MySuite) TestAddKindToModules(c *C) {
...
+	err := dc.applyUseModules()
+	c.Assert(err, IsNil)
+	err = dc.applyGlobalVariables()
+	c.Assert(err, IsNil)
+	err = dc.expandVariables()
+	// TODO: this will become nil once intergroup references are enabled
+	c.Assert(err, NotNil)
+
+	// check that ModuleConnections has map keys for each module ID
+	c.Assert(len(dc.moduleConnections), Equals, 2)
```
mr0re1, line 521: Can we use DeepEquals instead of point-checks?
Say `dc.moduleConnections` contains just one entry.
Your test will fails with `1 != 2` and no additional information.
But if we write it like this:
```
c.Check(dc.moduleConections, DeepEquals, map[string][]ModConnection{
  modID0: []ModConnection{
    {kind: deploymentConnection, sharedVariables: []string{"project_id"}, ...etc},
    {kind: deploymentConnection, sharedVariables: []string{"deployment_name"}, ...etc},
  }
  modID1: []ModConnection{...etc}
})
```
The error message will show you exactly what went wrong.
Pros:
* 1 check instead of dozen
* shows all information
* better test readability
Cons: 
* rely on order of items in slice

pkg/config/expand.go
mr0re1: I think it is, this code relies on tricky "map is shared across copies, because it's a pointer" fact. What if `varContext ` user will attempt to modify other fields in `dc`, or `moduleConnection` will stop being a map.
This concern is applicable for all usages of `varContext` and `applyGlobalVarsInGroup `
See toy example: https://go.dev/play/p/GF3fGiNuvRn

Did I miss something clever that renders my concern void?

tpdownes: Nothing clever missed, you just expressed some uncertainty so I didn't prioritize it while fixing other issues. Should be addressed.

pkg/config/config_test.go
tpdownes: Done on HEAD of PR. Good suggestion.

mr0re1 (approved):
Please don't forget to update `applyGlobalVarsInGroup` signature to pass `dc` by reference.

PR #1100, opened 2023-03-31
Replace startup-srcipt examples with bool inputs

mr0re1:
* Remove examples folder;
* Add bool inputs `install_cloud_ops_agent` and `install_ansible`;
* Deprecate `prepend_ansible_installer` input.

### Submission Checklist

  install`)


modules/scripts/startup-script/files/install_cloud_ops_agent.sh
nick-stroud, line 1: nit: `/scripts/` seems to be a more common pattern for this kind of subdir. We are not totally consistent (htcondor has a files subdir). I don't have a strong preference, but think scripts may be more descriptive.

modules/scripts/startup-script/README.md
```diff
@@ -171,13 +171,8 @@ they are able to do so by using the `gcs_bucket_path` as shown in the below exam
...
   settings:
     gcs_bucket_path: gs://user-test-bucket/folder1/folder2
-    runners:
-      - type: shell
-        source: modules/startup-script/examples/install_ansible.sh
-        destination: install_ansible.sh
-      - type: shell
-        source: modules/startup-script/examples/install_cloud_ops_agent.sh
-        destination: install_cloud_ops_agent.sh
+    install_ansible: true
```
nick-stroud, line 174: nit: I would prefer not to put the `install_ansible` in the example. Most common use case is that users will not define this setting and use the default behavior.

tools/validate_configs/test_configs/2filestore-4instances.yaml
```diff
@@ -53,13 +53,10 @@ deployment_groups:
       - type: shell
         content: $(homefs.install_nfs_client)
         destination: "install-nfs.sh"
```
nick-stroud, line 55: Now that you added the `use: homefs` you can also remove the `install-nfs.sh` runner (and this whole startup script block).

tools/validate_configs/test_configs/apt-collision.yaml
```diff
@@ -39,17 +39,12 @@ deployment_groups:
     kind: terraform
     id: network1
 
-  - source: ./modules/scripts/startup-script
```
nick-stroud, line 42: We likely want to leave at least one local reference in our `test_configs` since this is still a valid use we want coverage over.

tools/validate_configs/test_configs/complex-data.yaml
```diff
@@ -64,13 +64,10 @@ deployment_groups:
       - type: shell
         content: $(homefs.install_nfs_client)
         destination: "install-nfs.sh"
```
nick-stroud, line 66: This whole block can be removed. It happens automatically now.

tools/validate_configs/test_configs/hpc-cluster-slurm-with-startup.yaml
```diff
@@ -40,10 +40,7 @@ deployment_groups:
   - id: startup
     source: modules/scripts/startup-script
     settings:
-      runners:
-      - type: shell
-        source: "modules/startup-script/examples/install_ansible.sh"
-        destination: install_ansible.sh
+      install_ansible: true
```
nick-stroud, line 43: nit: It is confusing to me why we are installing ansible here at all but that pre-existed this change.

tools/validate_configs/test_configs/instance-with-startup.yaml
```diff
@@ -41,9 +41,6 @@ deployment_groups:
       - type: shell
         content: $(homefs.install_nfs_client)
         destination: "install_nfs.sh"
```
nick-stroud, line 43: This whole startup script block can be removed.

tools/validate_configs/test_configs/slurm-two-partitions-workstation.yaml
```diff
@@ -41,9 +41,6 @@ deployment_groups:
       - type: shell
         content: $(homefs.install_nfs_client)
         destination: "install_nfs.sh"
```
nick-stroud, line 43: This whole block can be removed. Happens automatically.

modules/scripts/startup-script/files/install_cloud_ops_agent.sh
mr0re1: Acknowledged. I'm going to keep in `/files`.

nick-stroud (approved):

PR #1101, opened 2023-03-31
Unused deployment variables validator

tpdownes:
Add new validator `test_deployment_variable_not_used` which ensures that all deployment variables (`vars` block in blueprint) are used automatically or explicitly in blueprint settings. Ignore `deployment_name` as it is required by `ghpc` and also ignore `labels` as it is automatically generated and applied.

### Submission Checklist

  install`)
* [x] Have you followed the guidelines in our Contributing document?\

### DIFF
--- a/docs/blueprint-validation.md
+++ b/docs/blueprint-validation.md
@@ -12,32 +12,8 @@ project(s):
 One can [explicitly define validators](#explicit-validators), however, the
 expectation is that the implicit behavior will be useful for most users. When
 implicit, a validator is added if all deployment variables matching its inputs
-is defined. Validators listed below that have no inputs are always enabled
-because they read the entire blueprint and do not require any specific
-deployment variable. If `project_id`, `region`, and `zone` are defined as
-deployment variables, then the following validators are enabled:
-
-```yaml
-validators:
-- validator: test_project_exists
-  inputs:
-    project_id: $(vars.project_id)
-- validator: test_apis_enabled
-  inputs: {}
-- validator: test_region_exists
-  inputs:
-    project_id: $(vars.project_id)
-    region: $(vars.region)
-- validator: test_zone_exists
-  inputs:
-    project_id: $(vars.project_id)
-    zone: $(vars.zone)
-- validator: test_zone_in_region
-  inputs:
-    project_id: $(vars.project_id)
-    zone: $(vars.zone)
-    region: $(vars.region)
-```
+are defined. Validators that have no inputs are always enabled by default
+because they do not require any specific deployment variable.
 
 Each validator is described below:
 
@@ -90,7 +66,35 @@ Each validator is described below:
 ### Explicit validators
 
 Validators can be overwritten and supplied with alternative input values,
-however they are limited to the set of functions defined above.
+however they are limited to the set of functions defined above. As an example,
+the default validators added when `project_id`, `region`, and `zone` are defined
+is:
+
+```yaml
+validators:
+  - validator: test_module_not_used
+    inputs: {}
+  - validator: test_deployment_variable_not_used
+    inputs: {}
+  - validator: test_project_exists
+    inputs:
+      project_id: $(vars.project_id)
+  - validator: test_apis_enabled
+    inputs: {}
+  - validator: test_region_exists
+    inputs:
+      project_id: $(vars.project_id)
+      region: $(vars.region)
+  - validator: test_zone_exists
+    inputs:
+      project_id: $(vars.project_id)
+      zone: $(vars.zone)
+  - validator: test_zone_in_region
+    inputs:
+      project_id: $(vars.project_id)
+      region: $(vars.region)
+      zone: $(vars.zone)
+```
 
 ### Skipping or disabling validators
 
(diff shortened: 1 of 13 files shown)


tpdownes:
This PR depends on #1094.

tpdownes:
> This PR depends on #1094.

The dependent PR was merged and this PR is reviewable.

tools/validate_configs/test_configs/new_project.yaml
```diff
@@ -17,6 +17,7 @@
 blueprint_name: new_project
 
 vars:
+  project_id: test_project
```
cboneti, line 20: won't this fail validation? since the project is not yet created?

cboneti (changes requested):

pkg/config/config.go
```diff
@@ -349,6 +353,39 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
 	return unusedModules
 }
 
+func (dc *DeploymentConfig) listUnusedDeploymentVariables() []string {
+	var usedVars = []string{}
```
mr0re1, line 357: Why not `map[string]bool` ? No need to use `slices.Contains` then.

pkg/config/config.go
```diff
@@ -349,6 +353,39 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
...
+			}
+		}
+	}
+
+	// variables that are either required or automatically constructed and
+	// applied; these should not be returned as unused otherwise no blueprints
+	// could be validated
+	requiredVars := []string{"labels", "deployment_name"}
+	var unusedVars = []string{}
+	for _, v := range maps.Keys(dc.Config.Vars) {
```
mr0re1, line 376: `for v, _ := range dc.Config.Vars {`

pkg/config/config.go
```diff
@@ -349,6 +353,39 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
...
+
+				}
+			}
+		}
+	}
+
+	// variables that are either required or automatically constructed and
+	// applied; these should not be returned as unused otherwise no blueprints
+	// could be validated
+	requiredVars := []string{"labels", "deployment_name"}
```
mr0re1, line 374: same

pkg/validators/validators.go
```diff
@@ -49,6 +51,22 @@ func handleClientError(e error) error {
...
+// TestDeploymentVariablesNotUsed errors if there are any unused deployment
+// variables and prints any to the output for the user
+func TestDeploymentVariablesNotUsed(unusedVariables []string) error {
+	var foundUnused bool
+	for _, v := range unusedVariables {
+		foundUnused = true
+		log.Printf(unusedDeploymentVariableMsg, v)
+	}
+
+	if foundUnused {
```
mr0re1, line 63: No need for `foundUnused`, just use `len(unusedVariables) > 0`

docs/blueprint-validation.md
```diff
@@ -12,10 +12,10 @@ project(s):
 One can [explicitly define validators](#explicit-validators), however, the
 expectation is that the implicit behavior will be useful for most users. When
 implicit, a validator is added if all deployment variables matching its inputs
-is defined. The `test_apis_enabled` validator is always enabled because it reads
-the entire blueprint and does not require any specific deployment variable. If
-`project_id`, `region`, and `zone` are defined as deployment variables, then the
-following validators are enabled:
+is defined. Validators listed below that have no inputs are always enabled
```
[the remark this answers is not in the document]
tpdownes: Addressed.

docs/blueprint-validation.md
```diff
@@ -12,10 +12,10 @@ project(s):
...
+because they read the entire blueprint and do not require any specific
+deployment variable. If `project_id`, `region`, and `zone` are defined as
+deployment variables, then the following validators are enabled:
```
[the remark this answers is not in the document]
tpdownes: Addressed, although I added the list of validators (updated) to the "explicit validators" section because I believe it is useful to have a starting point from which to work.

tools/validate_configs/test_configs/new_project.yaml
tpdownes: Because we do not use a real project ID in make tests, the following validators are disabled during make tests:

- test_project_exists
- test_apis_enabled
- test_region_exists
- test_zone_exists
- test_zone_in_region

tpdownes: With project_id set in the module, it fails validation because project_id is set as a deployment variable at the command line (but is unused).

pkg/config/config.go
tpdownes: Addressed.

pkg/config/config.go
tpdownes: Addressed. This was a relic of a refactor I didn't catch. Thanks.

pkg/config/config.go
tpdownes: Addressed

pkg/validators/validators.go
tpdownes: Addressed.

pkg/config/config.go
```diff
@@ -349,6 +352,34 @@ func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
...
+				for _, v := range conn.sharedVariables {
+					usedVars[v] = true
+				}
+			}
+		}
+	}
+
+	unusedVars := []string{}
+	for k := range dc.Config.Vars {
+		if _, ok := usedVars[k]; !ok {
```
mr0re1, line 375: nit. Due to default `bool == false`, you can write `if !usedVars[k] {`

cboneti (approved):

PR #1107, opened 2023-04-03
Add ability to set disk size for htcondor execute point

thiagosgobe:
### Submission Checklist

  install`)


tpdownes (changes requested):
The rest of the Toolkit uses type "number" for disk_size_gb. I realize that you are matching the input type of the instance template module, but the default behavior of Terraform is to convert numbers/bools to strings when necessary. We are making an effort to align the types of variables commonly used through the Toolkit and I prefer to align with `number` (as it's more intuitive).

https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion

Otherwise, this looks OK. If you make the change, I'll run it against the test that includes HTCondor and merge.

tpdownes (approved):

PR #1109, opened 2023-04-03
Track explicit connections between modules

tpdownes:
- track "explicit" module references made when a module setting is set to `$(module_id.output_name)`
- alter a unit test to cover behavior of `use` when the existing setting is set via deployment variables or explicitly

An alternative implementation could be considered in which the `useModule` function directly sets module settings into Toolkit-internal module format. e.g. `((module_id.output_name))`. This would avoid a brute force search through the graph, however, it would presently require the conversion from user syntax "`$(module_id.output_name)`" to internal format to occur in 2 different sections of the code. I believe it is safer to focus this conversion on a section of code that has access to the `reference` implementation. When references across groups are implemented I anticipate a function similar to:

```
type reference interface {
    HCLstring()
}
```

which will encapsulate a lot of the `fmt.Sprintf()` strewn about this section of code.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -832,6 +832,9 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 		if trackModuleGraph {
 			var found bool
 			for _, conn := range context.dc.moduleConnections[varRef.fromModuleID] {
+				if conn.kind != useConnection {
+					continue
+				}
 				if slices.Contains(conn.sharedVariables, varRef.name) {
 					found = true
 					break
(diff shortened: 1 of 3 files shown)


pkg/config/expand.go
```diff
@@ -829,6 +829,18 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 	case varRef.fromGroupID:
 		// intragroup reference can make direct reference to module output
 		expandedVariable = fmt.Sprintf("((module.%s.%s))", varRef.toModuleID, varRef.name)
+		if trackModuleGraph {
+			var found bool
+			for _, conn := range context.dc.moduleConnections[varRef.fromModuleID] {
+				if slices.Contains(conn.sharedVariables, varRef.name) {
```
mr0re1, line 835: Let's check for `conn.kind != explicit`.

tpdownes: Addressed on PR branch.

PR #1113, opened 2023-04-04
Adds a gke-cluster module to community

nick-stroud:
This change adds a module to create a k8s cluster.

This initial implementation seeks to achieve simplicity by exposing a limited set of functionality which will grow over time where there is clear value. As part of this objective the following functionality has been excluded:
- Autopilot is disabled
- Auto-provisioning of new node pools is disabled
- Network policies are not supported
- General addon configuration is not supported
- Only regional cluster is supported

Follow on change is coming with basic documentation for module and example.

Tested: Manually deployed 2 clusters into the same project using the provided example as a basis.

### Submission Checklist

  install`)

### DIFF
--- a/community/examples/gke.yaml
+++ b/community/examples/gke.yaml
@@ -1,0 +1,41 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+blueprint_name: small-gke
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: cluster-01
+  region: us-central1
+  zone: us-central1-c
+
+deployment_groups:
+- group: primary
+  modules:
+  - id: network1
+    source: modules/network/vpc
+    settings:
+      subnetwork_name: gke-subnet
+      secondary_ranges:
+        gke-subnet:
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gke_cluster
+    source: community/modules/scheduler/gke-cluster
+    use: [network1]
--- a/community/modules/scheduler/gke-cluster/README.md
+++ b/community/modules/scheduler/gke-cluster/README.md
@@ -1,0 +1,84 @@
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+Copyright 2022 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


community/modules/scheduler/k8s-cluster/main.tf
```diff
@@ -0,0 +1,261 @@
...
+    client_certificate_config {
+      issue_client_certificate = false
+    }
+  }
+
+  pod_security_policy_config {
+    enabled = true
+  }
+
+  enable_shielded_nodes = true
```
cboneti, line 66: Does this makes all pods as shielded nodes? or does this just allow for shielded nodes to exist?

community/modules/scheduler/k8s-cluster/main.tf
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  lifecycle {
+    # Ignore all changes to the default node pool. It's being removed after creation.
+    ignore_changes = [
+      node_config
+    ]
+  }
+
+  logging_service    = "logging.googleapis.com/kubernetes"
```
cboneti, line 153: This is likely the first thing you will need to parametrize.

community/modules/scheduler/k8s-cluster/main.tf
```diff
@@ -0,0 +1,261 @@
...
+  monitoring_service = "monitoring.googleapis.com/kubernetes"
+}
+
+# We define explict node pools, so that it can be modified without
+# having to destroy the entire cluster.
+resource "google_container_node_pool" "system-node-pools" {
+  provider = google-beta
```
cboneti, line 160: I am wondering if we could parametrize this provider. Does this need to be beta?

community/modules/scheduler/k8s-cluster/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+      contains(["NO_UPGRADES", "NO_MINOR_UPGRADES", "NO_MINOR_OR_NODE_UPGRADES"], x.exclusion_scope)
+    ])
+    error_message = "`exclusion_scope` must be set to `NO_UPGRADES` OR `NO_MINOR_UPGRADES` OR `NO_MINOR_OR_NODE_UPGRADES`."
+  }
+}
+
+variable "system_node_pool_name" {
+  description = "Name of the system node pool."
+  type        = string
+  default     = "system"
```
cboneti, line 109: The GKE recommendations doc recommends "gke-system".

community/modules/scheduler/k8s-cluster/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+
+variable "prefix_with_deployment_name" {
+  description = "If true, cluster name will be prefixed by `deployment_name` (ex: <deployment_name>-<name_suffix>)."
+  type        = bool
+  default     = true
+}
+
+variable "region" {
+  description = "The region to host the cluster in."
+  type        = string
```
cboneti, line 41: when you specify the region, does it always pick three zones out of this region? how do you make sure it is the right three? maybe that will be a more important decision at the node pool level...

community/modules/scheduler/k8s-cluster/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+variable "authenticator_security_group" {
+  description = "The name of the RBAC security group for use with Google security groups in Kubernetes RBAC. Group name must be in format gke-security-groups@yourdomain.com"
+  type        = string
+  default     = null
+}
+
+variable "enable_istio" {
+  description = "(Beta) Enable Istio addon"
+  type        = bool
+  default     = true
```
cboneti, line 189: Why is Istio enabled by default?

nick-stroud:
In parallel to your review I will update the naming of the module and look into tf-lint failures. It should not affect the discussed topics.

community/modules/scheduler/k8s-cluster/main.tf
cboneti: no, just bringing awareness. Ok to leave it as is.

community/modules/scheduler/k8s-cluster/main.tf
cboneti: then, it's ok as is.

cboneti (approved):
approved pending PR validation passes.

PR #1117, opened 2023-04-04
Add basic documentation for gke-cluster module and example

nick-stroud:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/scheduler/gke-cluster/README.md
+++ b/community/modules/scheduler/gke-cluster/README.md
@@ -1,3 +1,59 @@
+## Description
+
+This module creates a Google Kubernetes Engine
+([GKE](https://cloud.google.com/kubernetes-engine)) cluster.
+
+> **_NOTE:_** This is an experimental module and the functionality and
+> documentation will likely be updated in the near future. This module has only
+> been tested in limited capacity.
+
+### Example
+
+The following example creates a GKE cluster and a VPC designed to work with GKE.
+See [VPC Network](#vpc-network) section for more information about network
+requirements.
+
+```yaml
+  - id: network1
+    source: modules/network/vpc
+    settings:
+      subnetwork_name: gke-subnet
+      secondary_ranges:
+        gke-subnet:
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gke_cluster
+    source: community/modules/scheduler/gke-cluster
+    use: [network1]
+```
+
+Also see a full [GKE example blueprint](../../../examples/gke.yaml).
+
+### VPC Network
+
+This module is configured to create a
+[VPC-native cluster](https://cloud.google.com/kubernetes-engine/docs/concepts/alias-ips).
+This means that alias IPs are used and that the subnetwork requires secondary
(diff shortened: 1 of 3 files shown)


community/modules/scheduler/gke-cluster/README.md
```diff
@@ -1,5 +1,59 @@
+## Description
+
+This module creates a Kubernetes cluster using
```
cboneti, line 3: nit: wondering if you should just say: this module creates a [GKE](https://cloud.google.com/kubernetes-engine) cluster

community/modules/scheduler/gke-cluster/README.md
```diff
@@ -1,5 +1,59 @@
...
+> been tested in limited capacity.
+
+### Example
+
+The following example creates a GKE cluster and a VPC designed to work with GKE.
+See [VPC Network](#vpc-network) section for more information about network
+requirements.
+
+```yaml
+  - id: network1
```
cboneti, line 17: why not provide a full blueprint? It sounds silly you are omitting 4 lines. Why not just embed teh file? 
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet

cboneti (approved):
Two nits, feel free to correct them as needed and not re-request review.

community/modules/scheduler/gke-cluster/README.md
nick-stroud: I wanted to include `Kubernetes` somewhere in case people search for it (I believe this was your idea). I could instead say `This module creates a Google Kubernetes Engine (GKE) cluster.`

community/modules/scheduler/gke-cluster/README.md
nick-stroud: A full blueprint is an additional 9 lines (11 with white space). To provide a full blueprint would be breaking with the pattern we have on every other module documentation page ([example](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/develop/community/modules/file-system/DDN-EXAScaler/README.md#mounting)). 

I think the idea of an embedded code snippet is interesting but it seems it is quick to cut to a scrolling pane (see example below) instead of showing the full text. Also, I expect that blueprint to evolve a lot as we grow the ecosystem (node pool, job, file systems, etc.). Also, since it has to have a commit reference in the permalink, it takes an additional commit to make updates (one to update the example, another to update the link).

I will start by adding a reference to the gke.yaml example. As for the snippet, my preference would be to leave as is, but I would be open to adding content required to make it a valid blueprint (aka, add the 9 lines to this static snippet). What would you prefer?

https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/062e3d0e3f41ca9b9ea0aaceb216dedbe3d617a0/community/examples/gke.yaml#L27-L40

community/modules/scheduler/gke-cluster/README.md
nick-stroud: I have updated to use the `This module creates a Google Kubernetes Engine (GKE) cluster.` wording.

community/modules/scheduler/gke-cluster/README.md
nick-stroud: I added a link to the full example but will otherwise leave as is.

PR #1127, opened 2023-04-06
Improve variable references

tpdownes:
- Create HclString function as single implementation for converting variable references to HCL representation.
- create a const for string used in several related tests

### Submission Checklist

  install`)


pkg/config/expand.go
```diff
@@ -32,6 +32,7 @@ const (
 	blueprintLabel  string = "ghpc_blueprint"
 	deploymentLabel string = "ghpc_deployment"
 	roleLabel       string = "ghpc_role"
+	globalGroupID   string = "deployment"
```
mr0re1, line 35: IIUC we never surface this "deployment" string to the user, at the same time it prevents (or even worse silently allows, but makes mistakes) users from naming they groups "deployment".
This change can be a good opportunity to solve this problem. As a simple solution can we replace this with less usable word e.g. "hpc_toolkit_internal_keyword_for_global_variables".

pkg/config/expand.go
```diff
@@ -630,6 +638,17 @@ func (ref varReference) ToModuleID() string {
...
 
+func (ref varReference) HclString() string {
+	switch ref.toGroupID {
+	case globalGroupID:
+		return "var." + ref.name
+	case ref.fromGroupID:
+		return "module." + ref.toModuleID + "." + ref.name
+	default:
+		return "var." + ref.name + "_" + ref.toModuleID
+	}
```
nick-stroud, line 649: I feel like we lost some context with this function. 1. being what these cases stand for (global, within group, inter group) and 2. why the hcl is the way that it is. Part of me want to see the `IsIntergroup` function used but it does not lend itself to switch and the switch is also nice.

What about just adding some comments:
```
	case globalGroupID:
		return "var." + ref.name
	case ref.fromGroupID: 
		// intragroup reference can make direct reference to module output
		return "module." + ref.toModuleID + "." + ref.name
	default:  
		// for intergroup reference ghpc will create a variable
		return "var." + ref.name + "_" + ref.toModuleID
```

nick-stroud (approved):

pkg/config/expand.go
tpdownes: I think this has some value but I'd rather not (potentially) protract this particular PR. There is a current task to reject deployment groups that are named "deployment" and using a reserved keyword for this value may be a better way to address that.

pkg/config/expand.go
tpdownes: Addressed on HEAD of PR.

PR #1129, opened 2023-04-06
Add support for sensitive output values

tpdownes:
Modify code so that:

- output values have their own struct (they support [a subset of the properties of input variables](https://developer.hashicorp.com/terraform/language/values/outputs))
- support specifying the `sensitive` attribute of output values
- retain support for the simple "name-only" method of identifying output values while adding a new method which enables specification of the 3 primary attributes via a map (name, description, sensitive, depends_on)
- adopt this new struct where appropriate in existing code

If the `name` key is missing, this error is generated:

```
2023/04/06 22:46:53 failed to parse the blueprint in simple-vm.yaml, check YAML syntax for errors, err=line 52: missing required keys: ["name"]
```

If all required keys are present, but invalid keys are added to the `outputs` field, this error is generated:

```
2023/04/06 22:28:30 failed to parse the blueprint in simple-vm.yaml, check YAML syntax for errors, err=line 52: provided invalid keys: ["badkey0" "badkey1"]
```

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -206,7 +206,7 @@ func useModule(
 	useMod Module,
 	useModGroupID string,
 	modInputs []modulereader.VarInfo,
-	useOutputs []modulereader.VarInfo,
+	useOutputs []modulereader.OutputInfo,
 	settingsToIgnore []string,
 ) ([]string, error) {
 	usedVars := []string{}
@@ -809,7 +809,7 @@ func (ref varReference) validate(bp Blueprint) error {
 			"failed to get info for module at %s while expanding variables: %e",
 			refMod.Source, err)
 	}
-	found := slices.ContainsFunc(modInfo.Outputs, func(o modulereader.VarInfo) bool { return o.Name == ref.name })
+	found := slices.ContainsFunc(modInfo.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == ref.name })
 	if !found {
 		return fmt.Errorf("%s: module %s did not have output %s",
 			errorMessages["noOutput"], refMod.ID, ref.name)
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -880,8 +880,8 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 
 		// ensure that the target module outputs the value in the root module
 		// state and not just internally within its deployment group
-		if !slices.Contains(toMod.Outputs, varRef.name) {
-			toMod.Outputs = append(toMod.Outputs, varRef.name)
+		if !slices.ContainsFunc(toMod.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == varRef.name }) {
+			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{Name: varRef.name})
 		}
 
 		// TODO: remove error
--- a/pkg/modulereader/resreader.go
+++ b/pkg/modulereader/resreader.go
@@ -25,18 +25,78 @@ import (
 	"log"
 	"path"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 )
 
 // ModuleFS contains embedded modules (./modules) for use in building
 var ModuleFS embed.FS
 
-// VarInfo stores information about a module's input or output variables
+// VarInfo stores information about a module input variables
 type VarInfo struct {
 	Name        string
 	Type        string
 	Description string
 	Default     interface{}
 	Required    bool
+}
+
+// OutputInfo stores information about module output values
+type OutputInfo struct {
+	Name        string
+	Description string `yaml:",omitempty"`
+	Sensitive   bool   `yaml:",omitempty"`
+	// DependsOn   []string `yaml:"depends_on,omitempty"`
+}
+
+// UnmarshalYAML supports parsing YAML OutputInfo fields as a simple list of
+// strings or as a list of maps directly into OutputInfo struct
+func (mo *OutputInfo) UnmarshalYAML(value *yaml.Node) error {
+	var name string
+	const yamlErrorMsg string = "block beginning at line %d: %s"
+
+	err := value.Decode(&name)
+	if err == nil {
+		mo.Name = name
+		return nil
+	}
+
+	var fields map[string]interface{}
+	err = value.Decode(&fields)
+	if err != nil {
+		return fmt.Errorf(yamlErrorMsg, value.Line, "outputs must each be a string or a map{name: string, description: string, sensitive: bool}")
+	}
+
+	err = enforceMapKeys(fields, map[string]bool{
+		"name": true, "description": false, "sensitive": false},
+	)
+	if err != nil {
+		return fmt.Errorf(yamlErrorMsg, value.Line, err)
+	}
+
+	type rawOutputInfo OutputInfo
+	if err := value.Decode((*rawOutputInfo)(mo)); err != nil {
+		return fmt.Errorf("line %d: %s", value.Line, err)
+	}
+	return nil
+}
+
+// enforceMapKeys ensures the presence of required keys and absence of unallowed
+// keys with a useful error message; input is a map of all allowed keys to a
+// boolean that is true when key is required and false when optional
+func enforceMapKeys(input map[string]interface{}, allowedKeys map[string]bool) error {
+	for key := range input {
+		if _, ok := allowedKeys[key]; !ok {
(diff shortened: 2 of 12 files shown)


pkg/config/expand.go
```diff
@@ -863,8 +863,8 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 
 		// ensure that the target module outputs the value in the root module
 		// state and not just internally within its deployment group
-		if !slices.Contains(toMod.Outputs, varRef.name) {
-			toMod.Outputs = append(toMod.Outputs, varRef.name)
+		if !slices.ContainsFunc(toMod.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == varRef.name }) {
+			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{Name: varRef.name})
```
mr0re1, line 867: Are you intentionally loosing `description` and `sensitive` here?

pkg/modulereader/resreader.go
```diff
@@ -39,16 +41,41 @@ type VarInfo struct {
 	Required    bool
 }
 
+// OutputInfo stores information about module output values
+type OutputInfo struct {
+	Name        string
+	Description string   `yaml:",omitempty"`
+	Sensitive   bool     `yaml:",omitempty"`
+	DependsOn   []string `yaml:"depends_on,omitempty"`
```
mr0re1, line 49: IIUC `DependsOn` isn't functionally supported yet.  Shall we omit it from yaml?

pkg/config/expand.go
tpdownes: Right now this block ends in an error. When we enable intergroup references, `Sensitive` will probably be set to a user-configurable global value for all automatically generated output values. This will probably be `true` by default.

In the TF writer, a default description is generated if the field is its zero value. This is, for example:

```hcl
output "internal_ip_simplevm" {
  description = "Generated output from module 'simplevm'"
  value       = module.simplevm.internal_ip
}
```

I don't think it's necessarily helpful to set a different default value here, but I could be convinced, esp. if we want to specifically call out that it was generated automatically by an intergroup reference.

pkg/modulereader/resreader.go
cboneti: +1

pkg/config/expand.go
mr0re1: That is reasonable, thank you for explanation.

mr0re1 (approved):
LGTM. Please address `DependsOn`  concern before landing.

tpdownes:
Re-assigned to address comments and also some extra code to address lack of "strict" unmarshaling when using custom unmarshaler (https://github.com/go-yaml/yaml/issues/460)

tpdownes:
@mr0re1 I didn't catch that this brings coverage down below 80%. I will address but please review solution in the interim.

pkg/modulereader/resreader.go
```diff
@@ -39,16 +41,87 @@ type VarInfo struct {
...
+	Sensitive   bool   `yaml:",omitempty"`
+	// DependsOn   []string `yaml:"depends_on,omitempty"`
+}
+
+// UnmarshalYAML supports parsing YAML OutputInfo fields as a simple list of
+// strings or as a list of maps directly into OutputInfo struct
+func (mo *OutputInfo) UnmarshalYAML(value *yaml.Node) error {
+	var name string
+
+	switch value.Kind {
```
mr0re1, line 57: We could avoid this complexity (knowing types of yaml.Node) by keeping your old approach:
```
var name string
err := value.Decode(&name)
if err == nil { ... }

var fields map[string]interface{}
err = value.Decode(&fields)

if err != nil {
 return fmt.Errorf("line %d: outputs must each be a string or a map...")
}

// do fields validation.
```

pkg/modulereader/resreader.go
```diff
@@ -39,16 +41,87 @@ type VarInfo struct {
...
+		}
+		return nil
+	default:
+		return fmt.Errorf("line %d: outputs must each be a string or a map{name: string, description: string, sensitive: bool}", value.Line)
+	}
+}
+
+// enforceMapKeys ensures the presence of required keys and absence of unallowed
+// keys with a useful error message
+func enforceMapKeys(input map[string]interface{}, requiredKeys []string, optionalKeys []string) error {
```
mr0re1, line 86: nit. It can be written less verbosly : 
```
satisfied = map[string]bool {"name": false, "description": true, "sensitive": true}
for _, key := range fields {
  if _, ok := satisfied[key]; !ok {
     return fmt.Errorf("provided invalid key: %#v", key)
  }
  satisfied[key] = true
}
for key, sat := range satisfied {
  if !sat {
    return fmt.Errorf("missing required key: %#v", key)
  }
}
```

tpdownes: A similar solution has been adopted on the head of PR. Additionally prevents accidental inclusion of the same key as required and optional.

pkg/modulereader/resreader.go
tpdownes: To summarize an offline discussion: there is a good argument for keeping this simple and as we learn more about custom unmarshal'ing (and limitations in `go-yaml` package), we may revert to the simple approach. For now, I've chosen to refactor the switch as an if/else that tracks both `Kind` and `Tag` to ensure that we strictly adhere to attempting to parse values that are identified as maps or strings and nothing else. If we learn that this has real world bad implications, totally happy to revert to the simple approach.

pkg/modulereader/resreader.go
```diff
@@ -53,62 +53,50 @@ type OutputInfo struct {
 // strings or as a list of maps directly into OutputInfo struct
 func (mo *OutputInfo) UnmarshalYAML(value *yaml.Node) error {
 	var name string
+	const yamlErrorMsg string = "block beginning at line %d: %s"
```
mr0re1, line 56: nit. It seems like you always use the same value of `value.Line`, let's make it a part of this const.

mr0re1 (approved):

pkg/modulereader/resreader_test.go
```diff
@@ -228,6 +229,144 @@ func (s *MySuite) TestGetInfo_MetaReader(c *C) {
 	c.Assert(err, ErrorMatches, expErr)
 }
 
+func (s *MySuite) TestUnmarshalOutputInfo(c *C) {
```
mr0re1, line 232: You could have simplified your tests significantly if used just a string constains instead of `yaml.Node`

pkg/modulereader/resreader.go
mr0re1: go-yaml `Unmarshal` does not differentiate between `"0"` and `0` if recipient field in structure is `string`.
By being strict here we are going to get a behavioral disparity with rest of blueprints yaml.
Let's go easy road )

tpdownes: Done. I don't like how `go-yaml` does things, but I agree we should not be unaligned with the rest of the behavior.

pkg/modulereader/resreader_test.go
tpdownes: Done.

tpdownes:
@mr0re1 given the changes in decisions, I felt the best approach was to squash and force push. All the interesting bits are in ec536336d314cccb14cd510cd715e95ae0fb740e.

pkg/modulereader/resreader.go
```diff
@@ -39,16 +41,74 @@ type VarInfo struct {
...
+	err := value.Decode(&name)
+	if err == nil {
+		mo.Name = name
+		return nil
+	}
+
+	var fields map[string]interface{}
+	err = value.Decode(&fields)
+	if err != nil {
+		return fmt.Errorf(yamlErrorMsg, value.Line, "outputs must each be a string or a map{name: string, description: string, sensitive: bool}")
```
mr0re1, line 67: Here you discarding info from `err`, this may be useful to include it.

mr0re1 (approved):
LGTM. Please address one last comment.

pkg/modulereader/resreader.go
tpdownes: Added a `log` entry. As an example of replacing the map syntax with a list (that should fail):

```
  line 52: cannot unmarshal !!seq into map[string]interface {}
2023/04/10 12:11:17 failed to parse the blueprint in simple-vm.yaml, check YAML syntax for errors, err=block beginning at line 52: outputs must each be a string or a map{name: string, description: string, sensitive: bool}
```

PR #1154, opened 2023-04-11
Pin google terraform provider to latest version

nick-stroud:
Pinning latest terraform provider version post release of v1.16.0.


nick-stroud:
I did not concern myself to update `tools/cloud-build/provision/trigger-schedule/versions.tf` as this does not seem like it is relevant to keep this on the latest. Let me know if you feel differently.

PR #1156, opened 2023-04-11
Check nested values in TerraformBackend for variables

mr0re1:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -591,16 +591,17 @@ func checkUsedModuleNames(bp Blueprint) error {
 }
 
 func checkBackend(b TerraformBackend) error {
-	const errMsg = "can not use variables in backend block, got '%s=%s'"
+	const errMsg = "can not use variables in backend block, got '%s'"
 	if hasVariable(b.Type) {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -591,16 +591,17 @@ func checkUsedModuleNames(bp Blueprint) error {
 }
 
 func checkBackend(b TerraformBackend) error {
-	const errMsg = "can not use variables in backend block, got '%s=%s'"
+	const errMsg = "can not use variables in backend block, got '%s'"
```
nick-stroud, line 594: This seems like a perfect use case to define your own error type. As written it has some smells. We have string formatting on a constant string in multiple different functions. We are doing odd test verification on some specific string syntax which could just be passed as a field instead. This pr shows the tests are just change detectors. 

[Here](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8bac6ecaeb2d50dcc470321194e5bbe6a8c9f7c1/pkg/config/config.go#L807-L814) is an example of where it is done elsewhere in the code base.

PR #1157, opened 2023-04-12
OFE deployment using config file.

ek-nag:
New feature:
* Added support for automated OFE deployment using a config file, allowing for faster and more efficient deployment of OFE.

Changes:
 * Modified the OFE deployment script to automatically read parameters from a config file if the --config option is used.
 * Modified the deployment script to skip the confirmation prompt if deploying from the config file using the --config option.
 * Added a new documentation section on automated OFE deployment using a config file with an example YAML file.

### Submission Checklist

  install`)

### DIFF
--- a/community/front-end/ofe/deploy.sh
+++ b/community/front-end/ofe/deploy.sh
@@ -547,13 +547,16 @@ TFVARS
 		# TODO - upload terraform files to the FE server, so that they can be recovered if ever
 		#        needed
 
-		case $(ask "    Proceed to deploy? [y/N] ") in
-		[Yy]*) ;;
-		*)
-			echo "Exiting."
-			exit 0
-			;;
-		esac
+		if [ -z "$config_file" ]; then
+			# Ask for user confirmation before deploying
+			case $(ask "    Proceed to deploy? [y/N] ") in
+			[Yy]*) ;;
+			*)
+				echo "Exiting."
+				exit 0
+				;;
+			esac
+		fi
 
 		# -- Start the deployment using Terraform.
 		#    Note: Extract $? for terraform using PIPESTATUS, as $? below is
--- a/community/front-end/ofe/deploy.sh
+++ b/community/front-end/ofe/deploy.sh
@@ -105,6 +105,26 @@ HELP1
         (or Service Account Admin)
       - Project IAM Admin
 
+  Usage: ./deploy.sh [--config <path-to-config-file>] [--help]
+
+      --config <path-to-config-file> : path to YAML configuration file containing deployment variables.
+                                       If not specified, script will prompt user for input.
+      --help                         : display this help message
+
+  If --config option is used, all variables required for deployment must be specified in the YAML
+  file. The script will not prompt for any input in this case.
+
+  Example YAML file:
+  deployment_name: MyDeployment
+  project_id: my-project-id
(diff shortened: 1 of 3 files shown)


community/front-end/ofe/deploy.sh
```diff
@@ -105,6 +105,26 @@ HELP1
...
+
+  Example YAML file:
+  deployment_name: MyDeployment
+  project_id: my-project-id
+  zone: us-west1-a
+  subnet_name: my-subnet (optional)
+  dns_hostname: myhostname.com (optiona)
+  ip_address: 1.2.3.4 (optional)
+  django_superuser_username: sysadm
+  django_superuser_password: Passw0rd!
```
cboneti, line 125: Writing a password in plain text here is probably not the best, is there any other mechanism we can think about?

cboneti (commented):
Please find a few comments.

community/front-end/ofe/deploy.sh
ek-nag: One way or another, you will need to provide the password to the deployment script. We can have another argument such as `--django-password` or use environment variable. But I think you would still expose the password in the plain text anyway, just in a different place. I am open for discussion, let me know of any ideas.

community/front-end/ofe/deploy.sh
```diff
@@ -105,6 +105,26 @@ HELP1
...
+
+  If --config option is used, all variables required for deployment must be specified in the YAML
+  file. The script will not prompt for any input in this case.
+
+  Example YAML file:
+  deployment_name: MyDeployment
+  project_id: my-project-id
+  zone: us-west1-a
+  subnet_name: my-subnet (optional)
+  dns_hostname: myhostname.com (optiona)
```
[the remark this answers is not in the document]
ek-nag: fixed

ek-nag:
Interesting that pre-commit failed here, but when I was running it locally it did not for me.
It is complaining about:
```
Step #2 - "pre-commit-run": In community/front-end/ofe/deploy.sh line 940:
Step #2 - "pre-commit-run": 	if [[ -n ${DJANGO_SUPERUSER_PASSWORD+x} ]]; then
Step #2 - "pre-commit-run":                  ^----------------------------^ SC2153: Possible misspelling: DJANGO_SUPERUSER_PASSWORD may not be assigned, but django_superuser_password is.
```
But correct me if I am wrong.. 
In Bash, the +x operator is used with the -n operator to check if a variable is set or not. If the variable is set and not null, the expression is true. If the variable is unset or null, the expression is false.

I am not so sure why it's complaining.

community/front-end/ofe/deploy.sh
cboneti: This is better, thanks. That allows the env variable to be set from a cloud secrets manager and then fed into cloud build, for instance.

cboneti:
I am not sure why the problem is occurring. In my environment (with the latest dev dependencies), I can reproduce it. Alas, I did not get the chance to propose a fix tonight. :(

ek-nag:
What shellcheck version are you using @cboneti? I found some people hit the same issue, https://github.com/koalaman/shellcheck/issues/518#issuecomment-236113928.

cboneti:
We are updating our shellcheck version, so this should pass the tests soon.

PR #1160, opened 2023-04-12
Align internal use of Toolkit GitHub refs

tpdownes:
A user discovered that PBS modules were not using the latest version of vm-instance modules which propagates `var.zone` to the `google_compute_disk` resource. In practice, I think this means we switched from creating regional PDs (which have limited compatibility with VM families) to zonal PDs.

We should separately consider whether we also want to support regional PDs, but this commit aligns our usage with the current state of the vm-instance module.

- keep all startup-script usage at 34bb7250
- align other modules to 6c6b9e0a in which internal use of the startup-script module was updated to 34bb7250

### Submission Checklist

  install`)


tpdownes:
Has secondary impact of speeding up `terraform init` for PBS blueprints in particular.

nick-stroud (approved):

PR #1165, opened 2023-04-12
Add TransformSimpleToHcl for cty.Value

mr0re1:
This is a preparatory step for replacing "simple" variables with HCL ones in Blueprint.

* Add `TransformSimpleToHcl` for cty.Value;
* Add `Reference` struct, as "contextless" `varReference`;
* Minor changes of existing code to use them;
* Change `identifySimpleVariable ` to take `$(xxx)` instead of `xxx`

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expression.go
+++ b/pkg/config/expression.go
@@ -1,0 +1,94 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Reference is data struct that represents a reference to a variable.
+// Neither checks are performed, nor context is captured, just a structural
+// representation of a reference text
+type Reference struct {
+	GlobalVar bool
+	Group     string // should be empty if GlobalVar, otherwise optional
+	Module    string // should be empty if GlobalVar. otherwise required
+	Name      string // required
+}
+
+// SimpleVarToReference takes a string `$(...)` and transforms it to `Reference`
+func SimpleVarToReference(s string) (Reference, error) {
+	if !isSimpleVariable(s) {
+		return Reference{}, fmt.Errorf(
+			"got %s, variables \"$(...)\" within strings are not yet implemented. remove them or add a backslash to render literally", s)
+	}
+	contents := simpleVariableExp.FindStringSubmatch(s)
+	if len(contents) != 2 { // Should always be (match, contents) here
+		return Reference{}, fmt.Errorf("%s %s, failed to extract contents: %v",
+			errorMessages["invalidVar"], s, contents)
+	}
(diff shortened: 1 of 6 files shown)


pkg/config/expression.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+// VariableTranslator is an interface that provides function
+// to translate "simple" variable (`$(...)`) into HCL format
+type VariableTranslator interface {
+	TranlateSimpleToHcl(s string) (string, error)
+}
+
+// DoNotAllowVariablesTranslator does not do any translation, it raises an error if any varaibles are met
```
tpdownes, line 74: "variables"

pkg/config/expression.go
```diff
@@ -0,0 +1,94 @@
...
+type DoNotAllowVariablesTranslator struct {
+	VariableTranslator
+}
+
+// TranlateSimpleToHcl raises an error
```
tpdownes, line 79: "Translate"

pkg/config/expression.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// TransformSimpleToHcl produces a new value from passed one, replacing all occurrence
+// of simple variables `$(xxx)` with HCL ones `((yyy)`, using specified translator.
+func TransformSimpleToHcl(val cty.Value, translator VariableTranslator) (cty.Value, error) {
+	return cty.Transform(val, func(p cty.Path, v cty.Value) (cty.Value, error) {
+		if v.Type() != cty.String || !hasVariable(v.AsString()) {
+			return v, nil
+		}
+		h, err := translator.TranlateSimpleToHcl(v.AsString())
```
tpdownes, line 91: We generally try to report all errors for a given object such that the user does not have to fix them "one by one". It looks to me like this will error once for a backend configuration with multiple blueprint variables. Is that right?

I don't think this is necessarily fatal. You could alter the return of `checkBackend` to append a statement to effect of "variables are not allowed for any field within terraform_backend"

pkg/config/expression.go
```diff
@@ -0,0 +1,94 @@
...
+	GlobalVar bool
+	Group     string // should be empty if GlobalVar, otherwise optional
+	Module    string // should be empty if GlobalVar. otherwise required
+	Name      string // required
+}
+
+// SimpleVarToReference takes a string `$(...)` and transforms it to `Reference`
+func SimpleVarToReference(s string) (Reference, error) {
+	if !isSimpleVariable(s) {
+		return Reference{}, fmt.Errorf(
```
tpdownes, line 37: is this retaining the behavior of reporting all (not yet supported) blueprint variables that are attempting string interpolation? or is it just reporting generically that we do not yet support string interpolation?

I think this is a case where we want to retain the existing behavior: all exact strings should be printed.

tpdownes (commented):
There is a lot of progress being made in this PR although the actual adoption is limited to the terraform backend (a good place to start) and simplification of `identifySimpleVariable`

It appears to me that the long-term objective is for `Reference` to be more directly parsable as it has no context. Is there a plan more module references, e.g. via `use` keyword?

pkg/config/expression.go
mr0re1: > we want to retain the existing behavior: all exact strings should be printed.

There is no such behavior currently. Settings traversal will stop as soon it finds first offender and reports only it.

As for string-interpolation error it will still output the whole (but not all) offending string. The only change is removal of "hint" from error message, as it doesn't provide any value and adds complexity.

pkg/config/expression.go
mr0re1: Given there is no such behavior right now (report all offenders) and already complex state of this PR, I wouldn't address this suggestion yet.

tpdownes: I agree this is not fatal and should be prioritized only if there is any user confusion about it.

tpdownes (approved):

PR #1166, opened 2023-04-13
Don't use hclwrite tokenizer for "literal" strings

mr0re1:
* Don't use hclwrite tokenizer for "literal" strings because it escapes and  breaks variables passthrough;
* Remove "map compactor" because it contains bug.

### Submission Checklist

  install`)


mr0re1:
This PR become a #1167

PR #1167, opened 2023-04-13
Add specialized tokenizer to handle `((HCL literals))`

mr0re1:
#### Problem A: 
`hclwrite.TokensForValue` is aggressively escaping characters, rendering `((HCL literals))` beyond trivial unusable.
E.g. `"(( \"${vars.foo}-bar\" ))"` becomes `"(( \\\"$${vars.foo}-bar\\\" ))"`.
#### Problem B:
To unwrap literals `tfwriter` uses `handleLiteralVariables` function that performs naive find-and-replace on text of **whole** tf-file. This is dangerous and have to be addressed.

#### Solution:
Do not use `hclwrite.TokensForValue` , instead implement specialized tokenizer that has same behavior, except it tokenizes 
`((HCL literals))` "as is". 
Remove `handleLiteralVariables` as all valid cases for using `((HCL literals))` are covered by value-tokenizer.

#### Other changes:
* Remove "map-compactor" as it has a bug and is not required for validity;
* Simplify `simpleTokenFromString `.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulewriter/modulewriter_test.go
+++ b/pkg/modulewriter/modulewriter_test.go
@@ -500,6 +500,7 @@ func (s *MySuite) TestWriteMain(c *C) {
 		ID: "test_module",
 		Settings: map[string]interface{}{
 			"testSetting": "testValue",
+			"passthrough": `(("${vars.deployment_name}-allow\"))`,
 		},
 	}
 	testModules = append(testModules, testModule)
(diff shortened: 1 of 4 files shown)


pkg/modulewriter/modulewriter_test.go
```diff
@@ -500,6 +491,7 @@ func (s *MySuite) TestWriteMain(c *C) {
 		ID: "test_module",
 		Settings: map[string]interface{}{
 			"testSetting": "testValue",
+			"passthrough": `(("${vars.deployment_name}-allow\"))`,
```
tpdownes, line 494: I want to certain what you're testing here. You are testing a Go literal string that includes `\` and `$` ensuring that `hclwrite` (and our layers above it), do not escape them?

pkg/modulewriter/tfwriter.go
```diff
@@ -324,25 +293,19 @@ func writeProviders(vars map[string]cty.Value, dst string) error {
...
-			regToken := simpleTokenFromString("var.region")
-			regTokens := []*hclwrite.Token{&regToken}
-			provBody.SetAttributeRaw("region", regTokens)
+			provBody.SetAttributeRaw("region", simpleTokens("var.region"))
 		}
 		hclBody.AppendNewline()
 	}
 
 	// Write file
-	hclBytes := handleLiteralVariables(hclFile.Bytes())
```
tpdownes, line 345: It appears to me that this func was not used in `packerwriter.go` and therefore no changes are needed in `packerwriter.go` (which only writes `.tfvars` attributes files)

mr0re1: Correct, packerwrite only writes down result of `ResolveVariables` that shouldn't have raw HCL literals.

PR #1169, opened 2023-04-13
Write group metadata to deployment folder

tpdownes:
- for each deployment group, write only the deployment variables blocks and attributes that are used
- in the GHPC hidden directory, write a metadata file that contains minimal information about each group's inputs and outputs

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -364,6 +364,12 @@ func (dc *DeploymentConfig) GetModuleConnections() map[string][]ModConnection {
 	return dc.moduleConnections
 }
 
+// SetModuleConnections directly sets module connection graph (primarily for
+// unit testing where config expansion is not well-supported)
+func (dc *DeploymentConfig) SetModuleConnections(mc map[string][]ModConnection) {
+	dc.moduleConnections = mc
+}
+
 // listUnusedModules provides a mapping of modules to modules that are in the
 // "use" field, but not actually used.
 func (dc *DeploymentConfig) listUnusedModules() map[string][]string {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -292,11 +292,26 @@ type ModConnection struct {
 	sharedVariables []string
 }
 
+// IsDeploymentKind returns true if connection is to a deployment variable
+func (c ModConnection) IsDeploymentKind() bool {
+	return c.kind == deploymentConnection
+}
+
+// IsUseKind returns true if connection is module-to-module via use keyword
+func (c ModConnection) IsUseKind() bool {
+	return c.kind == useConnection
+}
+
+// GetSharedVariables returns variables used in the connection (can be empty!)
+func (c ModConnection) GetSharedVariables() []string {
+	return c.sharedVariables
+}
+
 // Returns true if a connection does not functionally link the outputs and
 // inputs of the modules. This can happen when a module is connected with "use"
 // but none of the outputs of fromID match the inputs of toID.
-func (mc *ModConnection) isUnused() bool {
-	return mc.kind == useConnection && len(mc.sharedVariables) == 0
+func (c *ModConnection) isUnused() bool {
+	return c.kind == useConnection && len(c.sharedVariables) == 0
 }
 
 // DeploymentConfig is a container for the imported YAML data and supporting data for
@@ -341,6 +356,12 @@ func (dc *DeploymentConfig) addModuleConnection(ref reference, kind connectionKi
 	fromModID := ref.FromModuleID()
 	dc.moduleConnections[fromModID] = append(dc.moduleConnections[fromModID], conn)
 	return nil
+}
+
+// GetModuleConnections returns the graph of connections between modules and
+// from modules to deployment variables
+func (dc *DeploymentConfig) GetModuleConnections() map[string][]ModConnection {
+	return dc.moduleConnections
 }
 
 // listUnusedModules provides a mapping of modules to modules that are in the
--- a/pkg/modulewriter/modulewriter.go
+++ b/pkg/modulewriter/modulewriter.go
@@ -18,6 +18,7 @@
 package modulewriter
 
 import (
+	"bytes"
 	"crypto/md5"
 	"embed"
 	"encoding/hex"
@@ -30,12 +31,15 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+
+	"gopkg.in/yaml.v3"
 )
 
 const (
 	hiddenGhpcDirName          = ".ghpc"
 	prevDeploymentGroupDirName = "previous_deployment_groups"
 	gitignoreTemplate          = "deployment.gitignore.tmpl"
+	deploymentMetadataName     = "deployment_metadata.yaml"
 )
 
 // ModuleWriter interface for writing modules to a deployment
@@ -48,6 +52,16 @@ type ModuleWriter interface {
 		deployDir string,
 	) error
 	restoreState(deploymentDir string) error
+}
+
+type deploymentMetadata struct {
+	DeploymentMetadata []groupMetadata `yaml:"deployment_metadata"`
+}
+
+type groupMetadata struct {
+	Name    string
+	Inputs  []string
+	Outputs []string
 }
 
 var kinds = map[string]ModuleWriter{
@@ -90,6 +104,9 @@ func WriteDeployment(blueprint *config.Blueprint, graph map[string][]config.ModC
 		return err
 	}
 
+	metadata := deploymentMetadata{
+		DeploymentMetadata: []groupMetadata{},
+	}
 	for _, grp := range blueprint.DeploymentGroups {
 		writer, ok := kinds[grp.Kind]
 		if !ok {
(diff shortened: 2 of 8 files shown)


pkg/modulewriter/modulewriter.go
```diff
@@ -338,3 +366,68 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+						groupInputs[v] = true
+					}
+				}
+			}
+		}
+	}
+
+	filteredVars := make(map[string]interface{})
+	for key, val := range vars {
+		if _, ok := groupInputs[key]; ok {
```
mr0re1, line 390: nit. `if groupInputs[key] {`

pkg/modulewriter/modulewriter.go
```diff
@@ -338,3 +366,68 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+	ghpcDir := filepath.Join(depDir, hiddenGhpcDirName)
+	if _, err := os.Stat(ghpcDir); os.IsNotExist(err) {
+		return fmt.Errorf(
+			"While trying to update the deployment directory at %s, the '.ghpc/' dir could not be found", depDir)
+	}
+
+	metadataFile := filepath.Join(ghpcDir, deploymentMetadataName)
+	f, err := os.OpenFile(metadataFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		log.Fatal(err)
```
mr0re1, line 407: Why `Fatal` here, how is this error different from one above?

pkg/modulewriter/modulewriter.go
```diff
@@ -338,3 +366,68 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+	}
+
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
```
mr0re1, line 411: nit.
```go
encoder := yaml.NewEncoder(&buf)
defer encoder.Close()
encoder.SetIndent(2)
if err := encoder.Encode(metadata); err != nil {
	return err
}
```

pkg/modulewriter/modulewriter.go
```diff
@@ -338,3 +366,68 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+		}
+	}
+	return filteredVars
+}
+
+func writeDeploymentMetadata(depDir string, metadata deploymentMetadata) error {
+	ghpcDir := filepath.Join(depDir, hiddenGhpcDirName)
+	if _, err := os.Stat(ghpcDir); os.IsNotExist(err) {
+		return fmt.Errorf(
+			"While trying to update the deployment directory at %s, the '.ghpc/' dir could not be found", depDir)
```
mr0re1, line 401: nit. Start with lower case `w`

pkg/modulewriter/modulewriter.go
```diff
@@ -90,19 +104,33 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
...
+	}
 	for _, grp := range blueprint.DeploymentGroups {
 		writer, ok := kinds[grp.Kind]
 		if !ok {
 			return fmt.Errorf(
 				"Invalid kind in deployment group %s, got '%s'", grp.Name, grp.Kind)
 		}
 
-		err := writer.writeDeploymentGroup(grp, blueprint.Vars, deploymentDir)
+		filteredVars := filterVarsByGraph(blueprint.Vars, grp, graph)
```
mr0re1, line 117: nit. Filtering appears to be terraform specific logic, it would be reasonable to move it to `tfwritter`.
But we can do it later, once it easier (no explicit graph is passed around).

pkg/modulewriter/modulewriter.go
tpdownes: Done.

pkg/modulewriter/modulewriter.go
tpdownes: I had logic for this but looking up the stack, it is treated as fatal at an appropriate location. So changed it so simply `return err`.

pkg/modulewriter/modulewriter.go
```diff
@@ -338,3 +366,68 @@ func prepDepDir(depDir string, overwrite bool) error {
...
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+	encoder.SetIndent(2)
+	err = encoder.Encode(metadata)
+	encoder.Close()
+	if err != nil {
+		return err
+	}
+	if _, err := f.Write(buf.Bytes()); err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
tpdownes: I had logic for this but looking up the stack, it is treated as fatal at an appropriate location. So changed it so simply `return err`.

pkg/modulewriter/modulewriter.go
tpdownes: Resolved (along with several other examples in this file)

pkg/modulewriter/modulewriter.go
tpdownes: Done (along with the more important closing of the file)

pkg/modulewriter/modulewriter.go
tpdownes: Done, as I agree.

pkg/modulewriter/modulewriter.go
```diff
@@ -90,19 +104,33 @@ func WriteDeployment(blueprint *config.Blueprint, outputDir string, overwriteFla
 		return err
 	}
 
+	metadata := deploymentMetadata{
+		DeploymentMetadata: []groupMetadata{},
+	}
 	for _, grp := range blueprint.DeploymentGroups {
 		writer, ok := kinds[grp.Kind]
```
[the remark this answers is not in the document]
tpdownes: Done

pkg/modulewriter/modulewriter_test.go
```diff
@@ -695,15 +714,30 @@ func (s *MySuite) TestWriteDeploymentGroup_PackerWriter(c *C) {
...
+	testDC := config.DeploymentConfig{
+		Config: config.Blueprint{
+			BlueprintName:   "",
+			Validators:      nil,
+			ValidationLevel: 0,
+			Vars:            testVars,
+			DeploymentGroups: []config.DeploymentGroup{
+				testDeploymentGroup,
+			},
+			TerraformBackendDefaults: config.TerraformBackend{},
```
mr0re1, line 727: nit. You can omit this declaration of `TerraformBackend` , it has no efffect, default value is the same.

tpdownes: Done

pkg/modulewriter/tfwriter.go
```diff
@@ -439,3 +447,43 @@ func orderKeys[T any](settings map[string]T) []string {
 	sort.Strings(keys)
 	return keys
 }
+
+func filterVarsByGraph(vars map[string]cty.Value, group config.DeploymentGroup, graph map[string][]config.ModConnection) map[string]cty.Value {
+	if graph == nil {
```
mr0re1, line 452: IIUC this for testing only? It's kind of ugly, but we can keep it.

pkg/modulewriter/packerwriter.go
```diff
@@ -58,36 +59,45 @@ func writePackerAutovars(vars map[string]cty.Value, dst string) error {
...
 			"error converting deployment vars to cty for writing: %w", err)
 	}
+
+	depGroup := dc.Config.DeploymentGroups[grpIdx]
 	groupPath := filepath.Join(deployDir, depGroup.Name)
+	allInputs := []string{}
 	for _, mod := range depGroup.Modules {
 
 		ctySettings, err := config.ConvertMapToCty(mod.Settings)
+		allInputs = append(allInputs, orderKeys(ctySettings)...)
```
mr0re1, line 78: Is there a chance for collision? Is it possible to have multiple modules in packer group?

pkg/modulewriter/tfwriter.go
tpdownes: Removed.

pkg/modulewriter/packerwriter.go
tpdownes: When enabling IGC, we plan to impose a restriction to have a single module limit for Packer groups, but nevertheless it would be safer to do it right so I've changed to the `map[string]bool` approach used elsewhere.

pkg/config/config.go
```diff
@@ -343,6 +358,18 @@ func (dc *DeploymentConfig) addModuleConnection(ref reference, kind connectionKi
 	return nil
 }
 
+// GetModuleConnections returns the graph of connections between modules and
+// from modules to deployment variables
+func (dc *DeploymentConfig) GetModuleConnections() map[string][]ModConnection {
+	return dc.moduleConnections
```
mr0re1, line 364: Is `moduleConnection` always initialized? it will be nice to have a guarantee that this function never returns `nil`.

mr0re1 (approved):

pkg/config/config.go
tpdownes: Yes, it is part of `config.NewDeploymentConfig` which called by both `create` and `expand` directly.

PR #1170, opened 2023-04-13
Instructions to run the toolkit from Cloud Workstations

cboneti:
### Submission Checklist

  install`)

### DIFF
--- a/tools/cloud-workstations/030_configure-hpc-toolkit.sh
+++ b/tools/cloud-workstations/030_configure-hpc-toolkit.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+sudo -i -u user bash /bin/configure-hpc-toolkit.sh
--- a/tools/cloud-workstations/Dockerfile
+++ b/tools/cloud-workstations/Dockerfile
@@ -1,0 +1,50 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+
+# Getting Terraform and Packer
+FROM us-central1-docker.pkg.dev/cloud-workstations-images/predefined/code-oss:latest
+
+ARG TFLINT_VERSION
+
+RUN curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -  && \
+    apt-get -y update && apt-get -y install \
+    software-properties-common \
+    keychain \
+    dnsutils \
+    shellcheck && \
+    apt-add-repository "deb [arch=$(dpkg --print-architecture)] https://apt.releases.hashicorp.com bullseye main" && \
+    apt-get -y update && apt-get install -y unzip python3-pip python3-venv terraform packer jq && \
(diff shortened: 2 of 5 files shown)


tools/cloud-workstations/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    go install golang.org/x/lint/golint@latest                      && \
+    go install github.com/fzipp/gocyclo/cmd/gocyclo@latest          && \
+    go install github.com/go-critic/go-critic/cmd/gocritic@latest   && \
+    go install github.com/google/addlicense@latest                  && \
+    go install mvdan.cc/sh/v3/cmd/shfmt@latest                      && \
+    go install golang.org/x/tools/cmd/goimports@latest
+
+RUN pip install --no-cache-dir --upgrade pip && \
+    pip install --no-cache-dir -r https://raw.githubusercontent.com/SchedMD/slurm-gcp/5.6.2/scripts/requirements.txt && \
+    pip install --no-cache-dir -r tools/cloud-build/requirements.txt && \
```
nick-stroud, line 49: I get a failure on this line. We are not in the toolkit repo so it makes sense. Consider changing to a url also.

tools/cloud-workstations/README.md
```diff
@@ -0,0 +1,29 @@
+# Configuring Cloud Workstations for usage with the Cloud HPC Toolkit
+
+> **_NOTE:_** If you want to redirect container registry repos to artifact registry, please see:
+> https://cloud.google.com/artifact-registry/docs/transition/setup-gcr-repo?&_ga=2.33584865.-1391632029.1681343137#redirect-enable
```
nick-stroud, line 4: nit: it might be nice to hide this ugly url with the `[]()` notation.

tools/cloud-workstations/Dockerfile
cboneti: good catch. Thanks!

tools/cloud-workstations/README.md
cboneti: agreed.

tools/cloud-workstations/configure-hpc-toolkit.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+# Check if the hpc-toolkit folder exists in the user home directory
+if [ ! -d "$HOME/hpc-toolkit" ]; then
+	# If not, clone the repository
+	git clone https://github.com/GoogleCloudPlatform/hpc-toolkit.git "$HOME/hpc-toolkit"
+	cd "$HOME/hpc-toolkit" || exit
+	make install-dev-deps
+	pre-commit install
```
nick-stroud, line 24: Why not also `make` & `make install`?

tools/cloud-workstations/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+      | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
+    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg \
+      | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && \
+    apt-get -y update && apt-get -y install google-cloud-sdk && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+RUN curl -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash
+
+RUN pip install --no-cache-dir --upgrade pip && \
+    pip install --no-cache-dir -r https://raw.githubusercontent.com/SchedMD/slurm-gcp/5.6.2/scripts/requirements.txt && \
```
nick-stroud, line 44: I would suggest also using main here. One less thing we have to worry about updating.

tools/cloud-workstations/030_configure-hpc-toolkit.sh
nick-stroud, line 1: why name this `030`?

tools/cloud-workstations/configure-hpc-toolkit.sh
cboneti: I was trying to strike a balance between time an instance takes to create and how much of it would be needed anyway since people maybe do a git pull. 
I think that since this is created by the workstation, it is very likely someone would need a make. 
personally, I dislike make install, so I will leave it out.

cboneti: PTAL

tools/cloud-workstations/Dockerfile
cboneti: good point. PTAL

tools/cloud-workstations/030_configure-hpc-toolkit.sh
cboneti: files under `/etc/workstation-startup.d/` automatically run in lexicographical order at container startup, so you can prefix your filename to run it at the appropriate time during workstation startup. There are already 010, and 020 tasks that should probably happen before this. 030, seemed [to make sense](https://cloud.google.com/workstations/docs/customize-container-images#extend-cloud-workstations-base-images).

nick-stroud: Makes sense. It seems like a magic number that would be good to document somewhere, but not sure where. Maybe it is just good enough to have this comment as a paper trail.

nick-stroud (approved):

PR #1171, opened 2023-04-13
Update SchedMD modules to 5.6.3 (from 5.6.2)

SkylerMalinowski:
### Submission Checklist

  install`)

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
@@ -15,7 +15,7 @@
  */
 
 # Most variables have been sourced and modified from the SchedMD/slurm-gcp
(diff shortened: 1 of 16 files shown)


cboneti (changes requested):
Shouldn't the controller expose its newly added output? (https://github.com/SchedMD/slurm-gcp/commit/be610be8c4b459fbe8be32f968997f914ab874be#diff-204625a3602feb0099fa10345ddb53424263341577cc43fd51ffef6ebfe66d0c).

SkylerMalinowski:
Exposing it on the controller would make it available for toolkit to use. For this PR there is nothing that can be done as the requested change would have to be released in a subsequent version.

cboneti (approved):

cboneti:
Ok, I will merge it as is while the outputs are not available at the controller.

PR #1178, opened 2023-04-14
Bump minimal Terraform version 1.0 -> 1.2

mr0re1:
### Submission Checklist

  install`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # PREAMBLE
 MIN_PACKER_VERSION=1.6 # for building images
-MIN_TERRAFORM_VERSION=1.0 # for deploying modules
+MIN_TERRAFORM_VERSION=1.2 # for deploying modules
 MIN_GOLANG_VERSION=1.18 # for building ghpc
 
 .PHONY: install install-user tests format add-google-license install-dev-deps \
--- a/community/modules/compute/gke-node-pool/README.md
+++ b/community/modules/compute/gke-node-pool/README.md
@@ -17,7 +17,7 @@ limitations under the License.
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.2 |
 | <a name="requirement_google"></a> [google](#requirement\_google) | >= 4.60.0, < 5.0 |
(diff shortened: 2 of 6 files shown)


tpdownes (dismissed):
I understand that we are generally bumping our recommended Terraform to 1.2.

I believe the gke-node-pool module requires 1.2+ for interpolation within the error_message. That said, I do not see why the gke-cluster module is being bumped. If there's a good reason, please merge. If there is not, please remove the change and then merge. 
At the same time, I don't see any features in the two GKE modules that require the upgrade. Am I not seeing something?

tpdownes:
> I understand that we are generally bumping our recommended Terraform to 1.2.
> 
> I believe the gke-node-pool module requires 1.2+ for interpolation within the error_message. That said, I do not see why the gke-cluster module is being bumped. If there's a good reason, please merge. If there is not, please remove the change and then merge. At the same time, I don't see any features in the two GKE modules that require the upgrade. Am I not seeing something?

I may have reversed the module names in the above comment. My point is that I only see the justification for going to 1.2 for one of these modules.

tpdownes:
A by-eye search suggests the following modules may also need 1.2+

- vm-instance
- batch-login-node
- batch-job-template
- dashboard

ankit-gautam23 (approved):
LGTM

mr0re1:
Reverted changes on modules, making this PR into "update ghpc-wide required terraform version".
Module-specifc changes can be resolved independently.

Makefile
```diff
@@ -1,6 +1,6 @@
 # PREAMBLE
```
cboneti, line 1: Nit: don't you want to already move Packer to 1.8 here?

Makefile
```diff
@@ -1,6 +1,6 @@
...
 MIN_PACKER_VERSION=1.6 # for building images
-MIN_TERRAFORM_VERSION=1.0 # for deploying modules
+MIN_TERRAFORM_VERSION=1.2 # for deploying modules
```
cboneti, line 3: This is ok, but you should probably update https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/pkg/modulewriter/tfversions.go#L19 as well.

cboneti (changes requested):

Makefile
mr0re1: I would bump it to `1.7.9` instead to match requirements in `modules/packer/custom-image/versions.pkr.hcl`, but I would do it separately and after better version comparator is deliverd.

cboneti (approved):

PR #1180, opened 2023-04-15
Add HclExpression struct

mr0re1:
* Implement HclExpression to represent HCL-literals;
* Transform both "simple" variables and HCL-literals into HclExpression during Dict YAML unmarshalling;
* Modify hcl writter to temporarily handle both cases (plain HCL literal string and special HclExpression value).
* Remove `Group` from `Reference`, store "explicit group name" as a mark on value.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -618,11 +618,19 @@ func checkUsedModuleNames(bp Blueprint) error {
 
 func checkBackend(b TerraformBackend) error {
 	const errMsg = "can not use variables in backend block, got '%s=%s'"
+	// TerraformBackend.Type is typed as string, "simple" variables and HCL literals stay "as is".
 	if hasVariable(b.Type) {
 		return fmt.Errorf(errMsg, "type", b.Type)
 	}
-	_, err := TransformSimpleToHcl(b.Configuration.AsObject(), DoNotAllowVariablesTranslator{})
-	return err
+	if _, is := IsRawHclLiteral(cty.StringVal(b.Type)); is {
+		return fmt.Errorf(errMsg, "type", b.Type)
+	}
+	return cty.Walk(b.Configuration.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
+		if _, is := IsHclValue(v); is {
+			return false, fmt.Errorf("can not use variables in backend block")
+		}
+		return true, nil
+	})
 }
 
 func checkBackends(bp Blueprint) error {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -617,7 +617,7 @@ func checkUsedModuleNames(bp Blueprint) error {
 }
 
 func checkBackend(b TerraformBackend) error {
-	const errMsg = "can not use variables in backend block, got '%s=%s'"
+	const errMsg = "can not use variables in terraform_backend block, got '%s=%s'"
 	// TerraformBackend.Type is typed as string, "simple" variables and HCL literals stay "as is".
 	if hasVariable(b.Type) {
 		return fmt.Errorf(errMsg, "type", b.Type)
@@ -627,7 +627,7 @@ func checkBackend(b TerraformBackend) error {
 	}
 	return cty.Walk(b.Configuration.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
 		if _, is := IsHclValue(v); is {
-			return false, fmt.Errorf("can not use variables in backend block")
+			return false, fmt.Errorf("can not use variables in terraform_backend block")
 		}
 		return true, nil
 	})
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -622,7 +622,7 @@ func checkBackend(b TerraformBackend) error {
 	if hasVariable(b.Type) {
 		return fmt.Errorf(errMsg, "type", b.Type)
 	}
-	if _, is := IsRawHclLiteral(cty.StringVal(b.Type)); is {
+	if _, is := IsYamlHclLiteral(cty.StringVal(b.Type)); is {
 		return fmt.Errorf(errMsg, "type", b.Type)
 	}
 	return cty.Walk(b.Configuration.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
--- a/pkg/config/expression.go
+++ b/pkg/config/expression.go
@@ -141,10 +141,10 @@ func TraversalToReference(t hcl.Traversal) (Reference, error) {
 	}
 }
 
-// IsRawHclLiteral checks if passed value of type cty.String
+// IsYamlHclLiteral checks if passed value of type cty.String
 // and its content starts with "((" and ends with "))".
 // Returns trimmed string and result of test.
-func IsRawHclLiteral(v cty.Value) (string, bool) {
+func IsYamlHclLiteral(v cty.Value) (string, bool) {
 	if v.Type() != cty.String {
 		return "", false
 	}
@@ -204,11 +204,11 @@ func (e HclExpression) References() []Reference {
 	return c
 }
 
-// AsRawValue returns a cty.Value, that is rendered as
+// makeYamlLiteralValue returns a cty.Value, that is rendered as
 // HCL literal in Blueprint syntax. Returned value isn't functional,
 // as it doesn't reference HclExpression.
 // This method should only be used for marshaling Blueprint YAML.
-func (e HclExpression) AsRawValue() cty.Value {
+func (e HclExpression) makeYamlLiteralValue() cty.Value {
 	return cty.StringVal("((" + e.s + "))")
 }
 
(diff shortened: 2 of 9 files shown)


pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
+	}
+	s := v.AsString()
+	if len(s) < 4 || s[:2] != "((" || s[len(s)-2:] != "))" {
+		return "", false
+	}
+	return s[2 : len(s)-2], true
+}
+
+// HclExpression is a representation of HCL literals in Blueprint
+type HclExpression struct {
```
tpdownes, line 156: I realize that these are private fields, but I think it is worth adding a comment that `s` is a string in HCL syntax. e.g. a static value, a "var.x" reference or a "module.x.y" reference.

If these fields are used directly anywhere in this package, I believe they should have longer, meaningful variable names. If they are uniformly accessed through getters, I am less concerned.

pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
+// TraversalToReference takes HCL traversal and returns `Reference`
+func TraversalToReference(t hcl.Traversal) (Reference, error) {
+	if t.IsRelative() {
+		return Reference{}, fmt.Errorf("got relative traversal")
+	}
+	getAttrName := func(i int) (string, error) {
+		if i >= len(t) {
+			return "", fmt.Errorf("traversal does not have enough components")
+		}
+		if a, ok := t[i].(hcl.TraverseAttr); ok {
```
tpdownes, line 114: what cases do we expect this cast to fail and error "got unexpected..."? Any realistic ones?

pkg/config/expression.go
mr0re1: Those fields meant to be "private"

pkg/config/expression.go
mr0re1: Any of those: `var`, `module`, `module.module_name`

pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
+}
+
+// Tokenize returns Tokens to be used for marshalling HCL
+func (e HclExpression) Tokenize() hclwrite.Tokens {
+	return hclwrite.TokensForIdentifier(e.s)
+}
+
+// References return Reference for all variables used in the expression
+func (e HclExpression) References() []Reference {
+	c := make([]Reference, len(e.rs))
```
tpdownes, line 200: could we just return `slices.Clone(e.rs)`?

I realize we might get satisfaction from using `make`, but I don't think we are setting capacity here, so I bet behavior is the same.

pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
 }
 
-// DoNotAllowVariablesTranslator does not do any translation, it raises an error if any variables are met
-type DoNotAllowVariablesTranslator struct {
-	VariableTranslator
+// AsRawValue returns a cty.Value, that is rendered as
+// HCL literal in Blueprint syntax. Returned value isn't functional,
+// as it doesn't reference HclExpression.
+// This method should only be used for marshaling Blueprint YAML.
+func (e HclExpression) AsRawValue() cty.Value {
```
tpdownes, line 211: This comment is very good but it's also worth getting name of this right. What do we mean by "Raw"? Is "Value" the right word?

- "For the blueprint"
- "has (( and ))"?

I have experienced a lot of confusion over time trying to figure out what we mean in various contexts.

Would it be helpful to name it "AsBlueprintLiteral"?

https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/examples#literal-variables

I have never liked calling them "literal variables" and prefer "literals".

pkg/config/expression.go
```diff
@@ -49,13 +51,14 @@ func MakeStringInterpolationError(s string) error {
 }
 
 // SimpleVarToReference takes a string `$(...)` and transforms it to `Reference`
-func SimpleVarToReference(s string) (Reference, error) {
+// Returns `Reference` and name of explicit group if one was set.
+func SimpleVarToReference(s string) (Reference, string, error) {
 	if !isSimpleVariable(s) {
-		return Reference{}, MakeStringInterpolationError(s)
+		return Reference{}, "", MakeStringInterpolationError(s)
```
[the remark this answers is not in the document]
mr0re1: Good catch, we should probably, handle this case (`!hasVariable`) separately.

pkg/config/expression.go
tpdownes: OK: I am comfortable with the variables names, but do please add comment. I realize we are getting toward a simpler world, but the multiple string representations of things can make it complicated to know "which is which and where".

pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
+// See: https://pkg.go.dev/github.com/zclconf/go-cty/cty#Value.Mark
+// "Marks" should be of hashable type, sadly HclExpression isn't one.
+// We work it around by using `hclExpressionKey` - unique identifier
+// of HclExpression and global map of used expressions `globalHclExpressions`.
+// There are two guarantees to hold:
+// * `ex1.key() == ex2.key()` => `ex1` and `ex2` are identical.
+// It's achieved by using HCL literal as a key.
+// * Every "mark" is in agreement with `globalHclExpressions`.
+// Achieved by declaring `HclExpression.AsValue()` as the ONLY way to produce "marks".
+type hclExpressionKey struct {
```
tpdownes, line 225: this is a useful comment. Should I be concerned that two expressions with identical HCL literals but different Marks may collide?

e.g. one set by the user explicitly with `$(vars.value)` and another with `use` keyword?

pkg/config/expression.go
```diff
@@ -49,13 +51,14 @@ func MakeStringInterpolationError(s string) error {
 }
 
 // SimpleVarToReference takes a string `$(...)` and transforms it to `Reference`
-func SimpleVarToReference(s string) (Reference, error) {
+// Returns `Reference` and name of explicit group if one was set.
```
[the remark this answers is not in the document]
mr0re1: Current comment still has `$(...)` in it.
```
// SimpleVarToReference takes a string `$(...)` and transforms it to `Reference`
// Returns `Reference` and name of explicit group if one was set.
func SimpleVarToReference(s string) (Reference, string, error)
```

pkg/config/expression.go
```diff
@@ -64,47 +67,216 @@ func SimpleVarToReference(s string) (Reference, error) {
...
+// `ex1.key() == ex2.key()` => `ex1` and `ex2` are identical.
+func (e HclExpression) key() hclExpressionKey {
+	return hclExpressionKey{k: e.s}
+}
+
+// AsValue returns a cty.Value that represents the expression.
+// This function should be the ONLY way to get HCL expression as a cty.Value.
+func (e HclExpression) AsValue() cty.Value {
+	k := e.key()
+	// we don't care if ot overrides as expressions are identical
```
tpdownes, line 241: there's a typo in this comment that makes me not sure what you're saying

pkg/config/config.go
```diff
@@ -618,11 +618,19 @@ func checkUsedModuleNames(bp Blueprint) error {
...
 		return fmt.Errorf(errMsg, "type", b.Type)
 	}
-	_, err := TransformSimpleToHcl(b.Configuration.AsObject(), DoNotAllowVariablesTranslator{})
-	return err
+	if _, is := IsRawHclLiteral(cty.StringVal(b.Type)); is {
+		return fmt.Errorf(errMsg, "type", b.Type)
+	}
+	return cty.Walk(b.Configuration.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
+		if _, is := IsHclValue(v); is {
+			return false, fmt.Errorf("can not use variables in backend block")
```
tpdownes, line 630: I suggest we use "terraform_backend block" here while we are it.

pkg/config/expression.go
mr0re1: Marks are attributes of `cty.Value` and not has no relation to `HclExpression`.
So one CAN NOT answer question "Is this expression result of `use` keyword", instead one should ask "Is this value in module settings is  result of `use` keyword"

pkg/config/expression.go
mr0re1: > I don't think we are setting capacity here
We do set capacity, second argument is `len(e.rs)` is exact capacity we want.

> just return `slices.Clone`
I'm hesitant to take dependency on `x/exp` packages with no good reason.

pkg/config/expression.go
mr0re1: Renamed to `makeYamlLiteralValue`.

tpdownes (approved):

PR #1186, opened 2023-04-18
Allow Packer groups to resolve deployment variables in structural types (list, map)

tpdownes:
Enable deployment variables nested inside structural values to properly resolve in Packer groups when written to `defaults.auto.pkrvars.hcl`. The status quo is for nested values to be skipped and for this file to have "blueprint literals" populated without erroring to the user. Unit test coverage is improved from 84.2% to 85.8% for the config package.

Example blueprint (with vars.labels set to a value that demonstrates the functionality without intending to be interpreted as valid GCP labels):

```yaml
blueprint_name: image-builder

vars:
  deployment_name: image-builder-001
  zone: us-central1-c
  new_image_family: my-slurm-image
  subnetwork_name: image-builder-us-central1
  disk_size: 32

deployment_groups:
- group: packer
  modules:
  - id: custom-image
    kind: packer
    settings:
      source_image_project_id: [schedmd-slurm-public]
      source_image_family: schedmd-v5-slurm-22-05-8-hpc-centos-7
      disk_size: $(vars.disk_size)
      image_family: $(vars.new_image_family)
      state_timeout: 15m
      labels:
        hello: $(vars.zone)
        tata:
          fornow: $(vars.zone)
        toodaloo:
        - $(vars.disk_size)
      tags:
      - $(vars.zone)
```

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -799,28 +799,67 @@ func ResolveVariables(
 		Variables: map[string]cty.Value{"var": cty.ObjectVal(origin)},
 	}
 	for key, val := range ctyMap {
-		if val.Type() == cty.String {
-			var valString string
-			if err := gocty.FromCtyValue(val, &valString); err != nil {
-				return err
-			}
-			ctx, varName, found := IdentifyLiteralVariable(valString)
-			// only attempt resolution on global literal variables
-			// leave all other strings alone (including non-global)
-			if found && ctx == "var" {
-				varTraversal := hcl.Traversal{
-					hcl.TraverseRoot{Name: ctx},
-					hcl.TraverseAttr{Name: varName},
-				}
-				newVal, diags := varTraversal.TraverseAbs(evalCtx)
-				if diags.HasErrors() {
-					return diags
-				}
-				ctyMap[key] = newVal
-			}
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -799,28 +799,67 @@ func ResolveVariables(
...
+		if err != nil {
+			return err
+		}
+		ctyMap[key] = newVal
+	}
+	return nil
+}
+
+func resolveValue(val cty.Value, evalCtx *hcl.EvalContext) (cty.Value, error) {
+	t := val.Type()
```
mr0re1, line 812: You can use `cty.Transform` to automate "walk" and don't do "is it map?, is it list?"

pkg/config/config.go
```diff
@@ -799,28 +799,67 @@ func ResolveVariables(
...
+	if t == cty.String {
+		var valString string
```
mr0re1, line 814: Just use `cty.Value.AsString()`

pkg/config/config.go
tpdownes: @nick-stroud this is a good suggestion so I'm removing your assignment for the moment.

PR #1187, opened 2023-04-18
Make Blueprint.Vars a Dict instead of map[string]interface{}

mr0re1:
* move `validateVars` before `expand`.

### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/dict.go
+++ b/pkg/config/dict.go
@@ -30,6 +30,15 @@ type Dict struct {
 	m map[string]cty.Value
 }
 
+// NewDict constructor
+func NewDict(m map[string]cty.Value) Dict {
+	d := Dict{}
+	for k, v := range m {
+		d.Set(k, v)
+	}
+	return d
+}
+
 // Get returns stored value or cty.NilVal.
 func (d *Dict) Get(k string) cty.Value {
 	if d.m == nil {
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -442,9 +442,15 @@ func combineModuleLabels(dc *DeploymentConfig, iGrp int, iMod int) error {
 	return nil
 }
 
-func mergeLabels[V interface{}](to map[string]V, from map[string]V) map[string]V {
-	r := maps.Clone(from)
-	maps.Copy(r, to)
+// mergeLabels returns a new map with the keys from both maps. If a key exists in both maps,
+// the value from the first map is used.
+func mergeLabels[V interface{}](a map[string]V, b map[string]V) map[string]V {
+	r := maps.Clone(a)
+	for k, v := range b {
+		if _, exists := a[k]; !exists {
+			r[k] = v
+		}
+	}
 	return r
 }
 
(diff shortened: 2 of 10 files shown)


pkg/config/expand.go
```diff
@@ -103,8 +103,7 @@ func (dc *DeploymentConfig) addMetadataToModules() error {
 	for iGrp, grp := range dc.Config.DeploymentGroups {
 		for iMod, mod := range grp.Modules {
 			if mod.RequiredApis == nil {
-				_, projectIDExists := dc.Config.Vars["project_id"].(string)
-				if !projectIDExists {
+				if dc.Config.Vars.Get("project_id").Type() != cty.String {
```
tpdownes, line 106: When I wrote this code, I inserted this validation logic here because (going by memory), one could choose to construct RequiredAPIs in such a way that one didn't need to define `project_id`. I think it may make more sense to  change the logic slightly:

1. in `validateVars`, enforce logic similar to `vars.labels`. if `project_id` exists, it must be a `cty.String`
2. At this point in the code, switch to `! Has("project_id")`.

pkg/config/expand.go
```diff
@@ -429,24 +433,24 @@ func combineModuleLabels(dc *DeploymentConfig, iGrp int, iMod int) error {
...
 	return nil
 }
 
-func mergeInLabels[V interface{}](to map[string]V, from map[string]V) {
-	for k, v := range from {
-		if _, exists := to[k]; !exists {
-			to[k] = v
-		}
-	}
+func mergeLabels[V interface{}](to map[string]V, from map[string]V) map[string]V {
```
tpdownes, line 445: I would add a short comment explaining the merge behavior of maps.Copy ("from" and "to" are hints, but not quite 100% of information I'd want at a glance)

pkg/config/dict.go
```diff
@@ -30,6 +30,15 @@ type Dict struct {
 	m map[string]cty.Value
 }
 
+// NewDict constructor
+func NewDict(m map[string]cty.Value) Dict {
```
tpdownes, line 34: It looks to me like the code (e.g. `Set`) is safe against a blueprint with empty `vars` block. But I'd like to raise possibility here of using `NewDict(empty_map)` at some point if there is any reason.

If you believe there is no reason to do so, please close this comment.

tpdownes (approved):
I am suggesting 2 minor changes and requesting you address a question. I believe the changes should be minor so if you agree to make them, please do so and merge.

pkg/config/expand.go
mr0re1: I would like to keep this PR free of behavior changes.

pkg/config/expand.go
mr0re1: Rewrote it, added a comment.

pkg/config/dict.go
mr0re1: We are good.
```go
NewDict(nil) == NewDict(map[string]cty.Value{}) == Dict{}
```

PR #1190, opened 2023-04-19
Ensure that outputs are written as HCL references

tpdownes:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/modulewriter/tfwriter.go
+++ b/pkg/modulewriter/tfwriter.go
@@ -103,8 +103,8 @@ func writeOutputs(
 				desc = fmt.Sprintf("Generated output from module '%s'", mod.ID)
 			}
 			blockBody.SetAttributeValue("description", cty.StringVal(desc))
-			value := fmt.Sprintf("((module.%s.%s))", mod.ID, output.Name)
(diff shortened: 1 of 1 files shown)


mr0re1 (approved):

pkg/modulewriter/tfwriter.go
```diff
@@ -103,8 +103,8 @@ func writeOutputs(
 				desc = fmt.Sprintf("Generated output from module '%s'", mod.ID)
 			}
 			blockBody.SetAttributeValue("description", cty.StringVal(desc))
-			value := fmt.Sprintf("((module.%s.%s))", mod.ID, output.Name)
-			blockBody.SetAttributeValue("value", cty.StringVal(value))
```
mr0re1, line 107: To be aligned with rest of the code we could write:
```
value := cty.StringVal(fmt.Sprintf("((module.%s.%s))", mod.ID, output.Name))
blockBody.SetAttributeRaw("value", TokensForValue(value))
```
But this code will be likely changed soon, I will add tests when I touch it next time.

tpdownes: Offline discussion: this works and may be modified when unit test added to test for this regression

PR #1198, opened 2023-04-20
Limiting line length to 2K to avoid bufio.Scanner: token too long issues

cboneti:
Under rare circumstances, the call to ansible-playbook omnia.yaml inside of our startup script will create a line that is longer than the bufio.Scanner limit. 

This PR will limit this specific call outputs to 2048 characters to avoid these issues. 

### Submission Checklist

  install`)


tpdownes (changes requested):
It seems to me that `tail -c 2048` is more likely to contain actionable information. Do you agree?

Another alternative might be something like:

```shell
cmd | split -b2000
for j in x[a-z][a-z]*; do cat $j; done
```

We do not have to completely over-engineer this hack but I'd rather not lose the information we want in the first place.

tpdownes (approved):

PR #1199, opened 2023-04-20
Ensure IGC outputs are marked sensitive to reduce output verbosity

tpdownes:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -856,7 +856,11 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 		// ensure that the target module outputs the value in the root module
 		// state and not just internally within its deployment group
 		if !slices.ContainsFunc(toMod.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == varRef.name }) {
-			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{Name: varRef.name})
+			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{
+				Name:        varRef.name,
+				Description: "Automatically-generated output exported for use by later deployment groups",
+				Sensitive:   true,
+			})
 		}
 	}
 	return fmt.Sprintf("((%s))", varRef.HclString()), nil
--- a/pkg/modulewriter/tfwriter.go
+++ b/pkg/modulewriter/tfwriter.go
@@ -96,6 +96,7 @@ func writeOutputs(
 		for _, output := range mod.Outputs {
 			// Create output block
 			outputName := config.AutomaticOutputName(output.Name, mod.ID)
+			hclBody.AppendNewline()
 			hclBlock := hclBody.AppendNewBlock("output", []string{outputName})
 			blockBody := hclBlock.Body()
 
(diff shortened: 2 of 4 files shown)


pkg/config/expand.go
```diff
@@ -868,7 +868,11 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 		// ensure that the target module outputs the value in the root module
 		// state and not just internally within its deployment group
 		if !slices.ContainsFunc(toMod.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == varRef.name }) {
-			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{Name: varRef.name})
+			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{
+				Name:        varRef.name,
+				Description: fmt.Sprintf("Automatically-generated output from module '%s'", toMod.ID),
+				Sensitive:   true,
```
nick-stroud, line 874: Does this happen after we add outputs requested by the user? So that if a user requested output and an IGC output align it will not be marked sensitive.

pkg/config/expand.go
```diff
@@ -868,7 +868,11 @@ func expandSimpleVariable(context varContext, trackModuleGraph bool) (string, er
 		// ensure that the target module outputs the value in the root module
 		// state and not just internally within its deployment group
 		if !slices.ContainsFunc(toMod.Outputs, func(o modulereader.OutputInfo) bool { return o.Name == varRef.name }) {
-			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{Name: varRef.name})
+			toMod.Outputs = append(toMod.Outputs, modulereader.OutputInfo{
+				Name:        varRef.name,
+				Description: fmt.Sprintf("Automatically-generated output from module '%s'", toMod.ID),
```
nick-stroud, line 873: This is just slightly different that the [message that is written](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8bac6ecaeb2d50dcc470321194e5bbe6a8c9f7c1/pkg/modulewriter/tfwriter.go#L107) for user generated outputs. I am wondering if we shouldn't explain a bit more that this was generated by IGC requirements. Thoughts?

nick-stroud (approved):
Please make description more specific to IGC.

pkg/config/expand.go
tpdownes: I would like us to discuss what lexicon/terminology we will present to users and align our messaging on that. I could imagine "for later deployment groups" instead of "from...". Would you prefer that?

This value doesn't show up anywhere except in outputs.tf so it's of limited utility.

pkg/config/expand.go
tpdownes: The outputs specified by the user (which can be sensitive true/false) are not altered by the automatic addition of outputs. Additionally, the description will be set to the default value rather than the one for IGC references contained in this PR.

pkg/config/expand.go
tpdownes: Offline discussion resulted in:

```
"Automatically-generated output exported for use by later deployment groups"
```

PR #1205, opened 2023-04-21
Use Dict for Module.Settings, derive connectivity from it

mr0re1:
### Submission Checklist

  install`)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,8 +17,6 @@ package config
 
 import (
 	"bytes"
-	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -26,9 +24,8 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
-	ctyJson "github.com/zclconf/go-cty/cty/json"
+	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v3"
 
@@ -297,7 +294,7 @@ type Module struct {
 	Use              []string
 	WrapSettingsWith map[string][]string
 	Outputs          []modulereader.OutputInfo `yaml:"outputs,omitempty"`
-	Settings         map[string]interface{}
+	Settings         Dict
 	RequiredApis     map[string][]string `yaml:"required_apis"`
 }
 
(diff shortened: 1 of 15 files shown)


pkg/config/config.go
```diff
@@ -408,51 +341,46 @@ func (dc *DeploymentConfig) ExpandConfig() error {
...
-// GetModuleConnections returns the graph of connections between modules and
-// from modules to deployment variables
-func (dc *DeploymentConfig) GetModuleConnections() map[string][]ModConnection {
-	return dc.moduleConnections
-}
+// listUnusedModules provides a list modules that are in the
+// "use" field, but not actually used.
+func (m Module) listUnusedModules() []string {
+	used := map[string]bool{}
+	cty.Walk(m.Settings.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
```
tpdownes, line 354: I think it is worth leaving a comment here and at other invocations of cty.Walk guiding future readers as to what is happening. 1-liner is good. Something like"

```go
// Recurse through objects/maps/lists evaluating each element for references marked by Use keyword
```

tpdownes: I think you are familiar enough with the idea that cty.Value abstracts many types of values, but future readers will not be.

pkg/config/config.go
```diff
@@ -408,51 +341,46 @@ func (dc *DeploymentConfig) ExpandConfig() error {
...
-	unusedModules := make(map[string][]string)
-	for _, connections := range dc.moduleConnections {
-		for _, conn := range connections {
-			if conn.isUnused() {
-				fromMod := conn.ref.FromModuleID()
-				unusedModules[fromMod] = append(unusedModules[fromMod], conn.ref.ToModuleID())
+// GetUsedDeploymentVars returns a list of deployment vars used in the given value
+func GetUsedDeploymentVars(val cty.Value) []string {
+	res := map[string]bool{}
+	cty.Walk(val, func(path cty.Path, val cty.Value) (bool, error) {
```
tpdownes, line 373: Similar 1-liner here

pkg/config/config.go
```diff
@@ -1015,3 +818,19 @@ func (b *Blueprint) WalkModules(walker func(*Module) error) error {
 	}
 	return nil
 }
+
+func checkModuleSettings(bp Blueprint) error {
```
tpdownes, line 822: I think a comment here similar to:

```go
// validate every module setting in the blueprint containing a Reference
```

pkg/config/expand.go
```diff
@@ -737,147 +529,53 @@ func (ref modReference) validate(bp Blueprint) error {
...
-// this function validates every field within a varReference struct and that
-// the reference must be to the same or earlier group.
-// ref.GroupID: this group must exist or be the value "deployment"
-// ref.ID: must be an existing module ID or "vars" (if groupID is "deployment")
-// ref.name: must match a module output name or deployment variable name
-// ref.explicitInterGroup: intergroup references must explicitly identify the
-// target group ID and intragroup references cannot have an incorrect explicit
-// group ID
-func (ref varReference) validate(bp Blueprint) error {
+// validateModuleSettingReference checks that a reference within module settings is valid
```
tpdownes, line 532: I think this comment is tautological. I would like to see a few big failures clearly stated.

1. References to deployment variables must exist
2. References to the outputs of other modules must exist (either intra- or inter- group)
  * Packer is a sub-case of this and doesn't really need to be called out in the comment 
4. Intergroup references must be in the right order

If I'm missing more, please add them.

pkg/config/expand.go
```diff
@@ -1048,3 +666,52 @@ func (dc *DeploymentConfig) addDefaultValidators() error {
 
 	return nil
 }
+
+// FindIntergroupReferences finds all references to other groups used it the given value
```
tpdownes, line 670: "it" -> "in"?

tpdownes (changes requested):
Consider this an initial review, but I am not seeing giant warning signs. This eliminates a lot of duplicated functionality in favor of a consistent approach to parsing areas of blueprints that contain "unknown quantities" potentially including references that must be evaluated for the underlying value (e.g. deployment variable) or for correctness when chaining modules to one another.

The widespread use of `cty.Value` is an improvement. What I think may be obvious to you (and to me after some experience) is that there are many kinds of `cty.Value` and I think some commenting here/there would improve future readers to understand why one would Walk/Transform a Value. Additionally, I'd like the intended logic of Settings validation to be clearly commented in the code, although the new solution is simpler to read.

pkg/config/expand.go
```diff
@@ -408,43 +347,45 @@ func combineModuleLabels(dc *DeploymentConfig, iGrp int, iMod int) error {
...
+		mod.Settings.Set(labels, cty.ObjectVal(mergeLabels(modLabels, g)))
 	}
 	return nil
 }
 
 // mergeLabels returns a new map with the keys from both maps. If a key exists in both maps,
 // the value from the first map is used.
-func mergeLabels[V interface{}](a map[string]V, b map[string]V) map[string]V {
-	r := maps.Clone(a)
+func mergeLabels(a map[string]cty.Value, b map[string]cty.Value) map[string]cty.Value {
```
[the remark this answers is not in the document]
mr0re1: Niether `maps.Copy` nor `maps.Clone` don't work nicely with `nil-map`

pkg/modulewriter/tfwriter.go
```diff
@@ -459,40 +458,73 @@ func orderKeys[T any](settings map[string]T) []string {
...
+		if !is {
+			return v, nil
+		}
+		ue := e.RenderHclAsString()
+		for _, r := range e.References() {
+			oi, exists := igcRefs[r]
+			if !exists {
+				continue
+			}
+			s := fmt.Sprintf("module.%s.%s", r.Module, r.Name)
```
tpdownes, line 503: This "last minute" change from a HCL module reference to a variable reference seems non-ideal. The issues:

1. There are points in the code where the reference appears to be an outwardly valid "intragroup" `module.x.y` reference. This can be a source of confusion for future development.
2. We're treating intergroup references as the ones to treat special rather than converting all Expression to correct HCL representations simultaneously.

That said, I see a comment at `RenderHclAsString` which suggests you see issues here, and different (worse) versions of this problem already exist. I'd like to merge this but discuss offline how you plan to address RenderHclAsString.

tpdownes (approved):

PR #1207, opened 2023-04-21
Remove unused `DeploymentConfig.expanded`

mr0re1:


pkg/config/config.go
nick-stroud, line 1: Please clean up pr description (delete checklist if you are not using it).  

nit: Add one line of context to description of you have background on this. Ex: "No longer used now that we do ..."

nick-stroud (approved):

pkg/config/config.go
mr0re1: "No longer used now that we do ..."
It appears to be not used for a while, not because of some recent change.

PR #1214, opened 2023-04-25
Use optimize utilization autoscaling profile

nick-stroud:
The optimize utilization profile is recommended for batch workloads. On original implementation, I had thought this related to auto-provisioning (automatic creation of node pools), which is not the case. It applies to horizontal scaling of node pools.

### DIFF
--- a/community/modules/scheduler/gke-cluster/main.tf
+++ b/community/modules/scheduler/gke-cluster/main.tf
@@ -68,9 +68,12 @@ resource "google_container_cluster" "gke_cluster" {
 
   enable_shielded_nodes = true
 
-  cluster_autoscaling { # Auto provisioning of node-pools
-    enabled             = false
-    autoscaling_profile = "OPTIMIZE_UTILIZATION"
+  cluster_autoscaling {
+    # Controls auto provisioning of node-pools
+    enabled = false
+
+    # Controls autoscaling algorithm of node-pools
+    autoscaling_profile = var.autoscaling_profile
   }
 
   datapath_provider = var.enable_dataplane_v2 ? "ADVANCED_DATAPATH" : "LEGACY_DATAPATH"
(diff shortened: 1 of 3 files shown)


community/modules/scheduler/gke-cluster/main.tf
```diff
@@ -69,9 +69,8 @@ resource "google_container_cluster" "gke_cluster" {
   enable_shielded_nodes = true
 
   cluster_autoscaling { # Auto provisioning of node-pools
-    enabled = false
-    # Recomended profile if we ever turn on
-    # autoscaling_profile = "OPTIMIZE_UTILIZATION"
+    enabled             = false
```
cboneti, line 72: These settings are applied to 

Here my understanding is that you are still maintaining the autoscaler off. Is this a default setting for all node_pools? 

I see no problem leaving it on for the system cluster, by the way. 

Can people enable this for the node_pools individually? otherwise, it stands to reason you should immediately expose these two as configurable variables.

cboneti (approved):
I left a comment, if you review it and deem it does not need action, feel free to move ahead.

community/modules/scheduler/gke-cluster/main.tf
nick-stroud: It is a poor api (hence my original confusion) but these two settings apply to entirely different features. 

`enable` controls if `auto-provisioning` is turned on. This is a very complex feature that allows the cluster to automatically create new node pools. We do not want to turn this on. 

`autoscaling_profile` selects the algorithm that is used to scale down a given node pool. This setting is cluster wide and applies to all node pools. It cannot be overridden at the node pool level. With that said, each node pool can turn auto scaling on or off, but once it is on it will use this algorythom.

To address your comment:
- I will move some comments around to make the above more clear
- I will expose `autoscaling_profile` as a var. I don't really think this is necessary but it also won't hurt. 
- I will wait to address if system pool should use auto scaling until after I discuss system pool requirements with gke team.

nick-stroud: Done.

cboneti (approved):
Thanks for the clarification. This looks good.

PR #1215, opened 2023-04-25
Add better version comparator for Makefile

mr0re1:
Our current checks are naive and use `expr >=` as a comparator that performs string comparison:
```shell
$ expr 1.10.1 \>= 1.7.3
0 # - incorrect
```

Add helper script that is more robust:
```shell
check "1.2" "1.3" # no
check "v1.2.9" "v1.3" # no
check "go1.20.9" "go1.21"   # no
check "go1.21" "1.22"  # no
check "go1.21-20230317-RC01" "1.22" # no

check "1.3" "1.2" # yes
check "1.2.0" "1.2" # yes
check "10.1" "9.9" # yes
check "v1.3" "v1.2.9" # yes
check "go1.21" "go1.20.9"  # yes
check "go1.21" "1.20"  # yes
check "go1.21-20230317-RC01" "1.3" # yes
```


Makefile
nick-stroud, line 1: I am just starting review, but I would appreciate a description on the PR.

In what ways is the version comparator "better"? What is the problem this PR is trying to solve?

mr0re1: Our current checks are naive and use `expr >=` as a comparator that performs string comparison:
```shell
$ expr 1.10.1 \>= 1.7.3
0
```

tools/version_check.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
nick-stroud, line 1: I could be mistaken, but does this constitute us adding python3 as a dependency to build the toolkit. 
I believe that technically this is already a transitive requirement as we require gcloud and gcloud requires python3, but python is not listed in our dependencies: https://cloud.google.com/hpc-toolkit/docs/setup/install-dependencies

Since we already have the transitive dependency, I would not block on this, but it may be good to have @cboneti weigh in.

tools/version_check.py
```diff
@@ -0,0 +1,48 @@
...
+    match = re.match(r"\s*v?(?P<release>\d+(?:\.\d+)*)\s*", s, re.VERBOSE | re.IGNORECASE)
+    if not match:
+        raise ValueError(f"Invalid version: '{s}'")
+    parts = [int(p) for p in match.group('release').split('.')][:3]
+    return tuple(parts + [0] * (3 - len(parts)))
+
+def meet(version, requirement):
+    return parse(version) >= parse(requirement)
+
+if __name__ == "__main__":
```
nick-stroud, line 29: Add a 1 line description and 1 example usage.

cboneti (changes requested):
Please make sure that the makefile still builds (even with the version warning) in an environment where Python3 is not available. 
Please make sure you or someone in the team, test this in: 
1) cloud shell
2) a mac

Please add a description to this PR explaining what we are doing and why.

mr0re1:
Remove python-comparator, implemented bash one.

cboneti (approved):
Thank you for these changes!

PR #1216, opened 2023-04-25
Use deadline instead of retries in wait-for-startup

mr0re1:
```shell
$ TIMEOUT=50 INSTANCE_NAME=A ZONE=A PROJECT_ID=A time community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
could not detect end of startup script. Sleeping.
...
could not detect end of startup script. Sleeping.
startup-script timed out after 50 seconds
to inspect the startup script output, please run:
gcloud compute instances get-serial-port-output A --port 1 --zone A --project A
Command exited with non-zero status 1
7.84user 1.78system 0:55.45elapsed
```

### DIFF
--- a/community/modules/scripts/wait-for-startup/main.tf
+++ b/community/modules/scripts/wait-for-startup/main.tf
@@ -13,10 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-locals {
-  retries = var.timeout / 5
-}
-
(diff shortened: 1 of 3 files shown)


tpdownes (changes requested):
I would like you to make changes that guarantee at least one execution of the loop. My suggestion:

- a `validation` block on var.timeout to guarantee it is non-negative
- modify the until loop with `gt` instead of `ge`

PR #1218, opened 2023-04-25
Silence daily cleanup notifications and enable retries for other builds

tpdownes:
- remove the "notify chat" tag from the daily cleanup trigger to reduce potential noise from 5 execution attempts per night
- enable a single retry for all other builds

The [defined behavior of the API](https://cloud.google.com/scheduler/docs/reference/rpc/google.cloud.scheduler.v1#google.cloud.scheduler.v1.RetryConfig) when setting both `max_retry_duration` and `retry_count` does not seem intuitive. In particular, "until both limits are reached." I have chosen to interpret this as a reason to remove `max_retry_duration`. Please review with fresh eyes.

### DIFF
--- a/tools/cloud-build/provision/trigger-schedule/main.tf
+++ b/tools/cloud-build/provision/trigger-schedule/main.tf
@@ -19,10 +19,9 @@ resource "google_cloud_scheduler_job" "schedule" {
 
   attempt_deadline = "180s"
   retry_config {
-    max_backoff_duration = "0s"
-    max_doublings        = 5
-    max_retry_duration   = "3600s"
-    min_backoff_duration = "1m"
+    max_backoff_duration = "1200s"
+    max_doublings        = 2
+    min_backoff_duration = "300s"
     retry_count          = var.retry_count
   }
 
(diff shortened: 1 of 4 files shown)


tools/cloud-build/provision/trigger-schedule/main.tf
```diff
@@ -19,10 +19,9 @@ resource "google_cloud_scheduler_job" "schedule" {
 
   attempt_deadline = "180s"
   retry_config {
-    max_backoff_duration = "0s"
-    max_doublings        = 5
-    max_retry_duration   = "3600s"
-    min_backoff_duration = "1m"
+    max_backoff_duration = "1200s"
+    max_doublings        = 2
+    min_backoff_duration = "300s"
```
nick-stroud, line 24: So my understanding of time between tests
1st - 5m - 2nd - 10m - 3rd - 20m - 4th - 20m - 5th

close if you agree.

nick-stroud (approved):

tools/cloud-build/provision/trigger-schedule/main.tf
tpdownes: Yes, but writing down a bit differently to include the initial failure.

FAIL -> 5m -> Retry 1 (FAIL) -> 10m -> Retry 2 (FAIL) -> 20m -> Retry 3 (FAIL) -> 20m -> Retry 4 (FAIL) -> 20m -> Retry 5 (FAIL) -> Done

PR #1219, opened 2023-04-25
Implement stub export-outputs command

tpdownes:
Implements the stub interface for the `export-outputs` command and demonstrates the first error one would get if executing `export-outputs` without a copy of Terraform available in the execution PATH.

### Submission Checklist

  install`)


cmd/export.go
```diff
@@ -0,0 +1,86 @@
...
+	metadataFlag := "metadata"
+	artifactsFlag := "artifacts"
+	exportCmd.Flags().StringVarP(&artifactsDir, artifactsFlag, "a", "", "Alternative artifacts output directory (automatically configured if unset)")
+	exportCmd.Flags().StringVarP(&metadataFile, metadataFlag, "m", "", "Deployment metadata YAML file (automatically configured if unset)")
+	exportCmd.MarkFlagDirname(artifactsFlag)
+	exportCmd.MarkFlagFilename(metadataFlag, "yaml", "yml")
+	rootCmd.AddCommand(exportCmd)
+}
+
+const defaultMetadataFile string = "../.ghpc/deployment_metadata.yaml"
```
nick-stroud, line 38: Just to confirm, this is just the expanded blueprint? Part of me wants to name it as such, but naming it as you have it might discourage people from messing with it, where naming it something like `expanded_blueprint` might make people think they could move it.

cmd/export.go
```diff
@@ -0,0 +1,86 @@
...
+	"path"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	metadataFlag := "metadata"
+	artifactsFlag := "artifacts"
+	exportCmd.Flags().StringVarP(&artifactsDir, artifactsFlag, "a", "", "Alternative artifacts output directory (automatically configured if unset)")
+	exportCmd.Flags().StringVarP(&metadataFile, metadataFlag, "m", "", "Deployment metadata YAML file (automatically configured if unset)")
```
nick-stroud, line 32: What is the best case for having the metadata file as an argument? I am struggling to think of a good case and it just opens the door to more user journeys that we will have to support later. Thoughts on removing it until it is asked for? Also I don't believe it is not in your design doc.

cmd/export.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const defaultMetadataFile string = "../.ghpc/deployment_metadata.yaml"
+
+var (
+	artifactsDir string
+	metadataFile string
+	exportCmd    = &cobra.Command{
+		Use:               "export-outputs DEPLOYMENT_DIRECTORY",
+		Aliases:           []string{"output"},
```
nick-stroud, line 45: Consider not providing an alias to start. I feel like the less options we give the simpler the product will be to use. We can always add an option but harder to take away. Also `ghpc output my/dir` just does not seem very clear to me. Thoughts?

What are we planning for the inverse command?

I like the following options:
```
# you could add hyphen, I like that propagate could apply to both to make clear they go together
ghpc propagate outputs my/dir
ghpc propagate inputs my/dir

# This is also clear. I don't expect people to try to remember the commands
ghpc export-outputs my/dir
ghpc collect-inputs my/dir
```

cmd/export.go
tpdownes: It is not yet the expanded blueprint, but will be. When #1205 is merged the deployment metadata file will continue to be written without modification, so these are not in conflict in the branch sense or functionality sense. However, #1205 will make it much easier to get the information directly from the expanded blueprint without potential mismatch. Additionally, it will enable proper/elegant support for writing inputs to Packer modules that will not be present in the follow-up to this PR.

I could imagine:

- changing the flag to `--blueprint-metadata` (or similar) if we are worried about exposing the name now on develop branch.
- putting big giant "DO_NOT_DELETE_THIS_DIRECTORY" type file into the ".ghpc" directory when writing the deployment, along with the expanded blueprint when we write it

cmd/export.go
tpdownes: I am generally wary of assuming that the directory structure is exactly right else the only thing we can do is fail.

In particular, I believe CI/CD pipelines are likely to exist where the blueprint and the deployment groups "wander off" from one another. Given that the flag is optional and the help indicates "automatically configured" I do not see much pain to supporting it.

cmd/export.go
tpdownes: I can remove the alias. I have simplified the "import" command to address only the use case where the export command has identified an alternative artifacts directory. In this case, it will look for appropriately named files in the artifacts directory and stage them into the right group directory. If they are not found, import fails.

In other words, our "simple" case for deploying will only require the exports-outputs command.

Context: in a CI/CD pipeline, one is likely to find it easy to chain stages together in which all artifacts are tarballed from stage 0 and copied into a single folder in stage N.

nick-stroud (approved):

PR #1220, opened 2023-04-25
Fail wait-for-startup fast if log can not be fetched

mr0re1:


community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -28,11 +28,15 @@ fi
 
 now=$(date +%s)
 deadline=$(("${now}" + "${TIMEOUT}"))
+error_file=$(mktemp)
 
 until [ "${now}" -gt "${deadline}" ]; do
-	GCLOUD="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} --port 1 --zone ${ZONE} --project ${PROJECT_ID}"
+	ser_log=$(gcloud compute instances get-serial-port-output "${INSTANCE_NAME}" --port 1 --zone "${ZONE}" --project "${PROJECT_ID}" 2>"${error_file}") || {
```
nick-stroud, line 34: We are now storing the full output, instead of immediately feeding it to grep. IDK anything about shell variable buffers. The serial port output can be very long. Should we be at all concerned about storing this in a variable?

If this concern is valid then ideally we would want to do the grep at this point. I think you could do it by performing the grep up front and then just checking the error file instead of using the OR. If you know this to not be a concern then respond and close.

nick-stroud, line 34: I am wondering if it is possible for this to fail early on, such as right after the instance is created. My observations is that terraform says the VM has started, but for several seconds after this the serial port log is empty. I cannot remember/ do not know what is returned from gcloud at this time. 

If this was a legitimate concern then maybe we should ignore the first few errors. I also don't want to add complexity for no reason. If you have observed this working, or have reason to believe that my concern is not possible then we can ignore.

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
nick-stroud, line 1: Please clean up the pr description.

Also either put in pr description what manual testing you have done, or run an integration test that exercises this module.

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
mr0re1: Good point, added grep to fetch

nick-stroud: I think your comment here is actually in response to my other comment (about grep). So I am closing that one.

I think my concern on this comment was seeing early failures on startup, but seeing tests past make me conclude this is not a concern.

PR #1224, opened 2023-04-26
Add ghpc version to expanded blueprint

mr0re1:
```
$ head expanded.yaml
...
blueprint_name: hpc-cluster-small
ghpc_version: v1.16.0-185-g1b863f09
validators:
...
```

### DIFF
--- a/cmd/expand.go
+++ b/cmd/expand.go
@@ -78,6 +78,7 @@ func runExpandCmd(cmd *cobra.Command, args []string) {
 	if err := deploymentConfig.ExpandConfig(); err != nil {
 		log.Fatal(err)
 	}
+	deploymentConfig.Config.GhpcVersion = GitCommitInfo
 	deploymentConfig.ExportBlueprint(outputFilename)
 	fmt.Printf(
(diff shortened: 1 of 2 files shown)


cmd/expand.go
```diff
@@ -78,6 +78,7 @@ func runExpandCmd(cmd *cobra.Command, args []string) {
 	if err := deploymentConfig.ExpandConfig(); err != nil {
 		log.Fatal(err)
 	}
+	deploymentConfig.Config.GhpcVersion = GitCommitInfo
```
tpdownes, line 81: I would like to see a warning that informs the user that any value they set was ignored.

tpdownes (changes requested):

cmd/expand.go
tpdownes: In fact, what occurs with this value when ingesting an expanded blueprint?

mr0re1: >  a warning 

Done
 
> what occurs with this value when ingesting an expanded blueprint?

This property is intentionally inert  so far.

tpdownes (approved):

PR #1225, opened 2023-04-26
Initial implementation of export-outputs command

tpdownes:
- obtain outputs from a Terraform deployment group where "terraform apply" has already been executed
- verify that the outputs are not out-of-sync with cloud state by checking that "terraform plan" does not propose any changes
- do not automatically apply any changes to cloud infrastructure
- write outputs in standard HCL format to configurable artifacts directory that defaults to the hidden .ghpc deployment directory


pkg/shell/common.go
```diff
@@ -0,0 +1,100 @@
...
+	"golang.org/x/sys/unix"
+	"gopkg.in/yaml.v3"
+)
+
+// GetDeploymentKinds performs a basic sanity check of metadata file and returns
+// the module kinds for the deployment
+func GetDeploymentKinds(metadataFile string, deploymentRoot string) (map[string]config.ModuleKind, error) {
+	md, err := loadMetadata(metadataFile)
+	if err != nil {
+		return map[string]config.ModuleKind{}, err
```
mr0re1, line 35: `return nil, err`

pkg/shell/common.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	groupKinds := make(map[string]config.ModuleKind)
+	for _, gm := range md {
+		groupPath := path.Join(deploymentRoot, gm.Name)
+		if isDir, _ := DirInfo(groupPath); !isDir {
+			return map[string]config.ModuleKind{},
```
mr0re1, line 42: same

mr0re1 (approved):

PR #1227, opened 2023-04-26
Fix whitespace in deployment directories

tpdownes:
- Do not write outputs.tf if empty
- Improve whitespace in deployments
    - ensure 1 line of whitespace after Apache license
    - eliminate trailing whitespace in TF files
    - add spacing between HCL attributes (.tfvars)


pkg/modulewriter/modulewriter_test.go
```diff
@@ -555,11 +555,7 @@ func (s *MySuite) TestWriteOutputs(c *C) {
 	testModules := []config.Module{}
 	outputs, err := writeOutputs(testModules, testOutputsDir)
 	c.Assert(err, IsNil)
-	c.Check(outputs, DeepEquals, []string{})
-
-	// Failure: Bad path
-	_, err = writeOutputs(testModules, "not/a/real/path")
-	c.Assert(err, ErrorMatches, "error creating outputs.tf file: .*")
+	c.Check(outputs, IsNil)
```
mr0re1, line 558: Aren't this check should fail? outputs of no-modules is empty list, not nil

pkg/modulewriter/tfwriter.go
```diff
@@ -119,10 +111,17 @@ func writeOutputs(
 		}
 	}
 
-	// Write file
 	hclBytes := hclFile.Bytes()
+	if len(hclBytes) == 0 {
```
mr0re1, line 115: Let's rely on number of outputs, not byte-size of tokenized HCL.
And return empty list, not nil.

mr0re1 (approved):
LGTM, please address one comment left.

PR #1231, opened 2023-04-26
Update GKE settings to match recommendations from GKE team

nick-stroud:
This change:
- Updates our GKE module settings with revised recommendations from the GKE team
  - Add taint to node pools meant for user workloads
  - Expose GKE cluster version
  - Disable istio (and remove support)
  - Increase the minimum system pool node count to 2
- It also increases the maximum system pool node count to 10 just to provide more headroom

Manually tested by running job against deployed cluster.


community/modules/compute/gke-node-pool/README.md
```diff
@@ -19,6 +19,15 @@ The following example creates a GKE node group.
 
 Also see a full [GKE example blueprint](../../../examples/gke.yaml).
 
+### Taints and Tolerations
+
+By default node pools created with this module will be tainted with
+`user-workload=true:NoSchedule` to prevent system pods from being scheduled.
```
cboneti, line 25: I may be missing it here, but should this not be "PREFER_NO_SCHEDULE" ?

cboneti (approved):

PR #1235, opened 2023-04-27
Ensure that terraform cleanup always runs

tpdownes:
The current position of "terraform destroy" within the rescue block positions it to not run when previous steps in the rescue block fail. This will leak infrastructure into our build project. Additionally, this PR reduces code duplication by splitting out the terraform cleanup into its own task file and include that in another place cleanup is necessary.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -150,13 +150,14 @@
 
     ## Cleanup and fail gracefully
     rescue:
+    - name: Capture terraform stderr
+      ansible.builtin.set_fact:
+        terraform_apply_stderr_one_line: "{{ terraform_output.results.1.stderr | replace('\n',' ') }}"
     - name: Gather logs
       ansible.builtin.include_tasks:
         file: tasks/gather_startup_script_logs.yml
         apply:
           delegate_to: localhost
-      vars:
-        terraform_apply_stderr: "{{ terraform_output.results.1.stderr }}"
     - name: Cleanup firewall and infrastructure
       ansible.builtin.include_tasks:
         file: tasks/rescue_terraform_failure.yml
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -150,12 +150,25 @@
 
     ## Cleanup and fail gracefully
     rescue:
-    - name: Include rescue from terraform failure
-      ansible.builtin.include_tasks: "tasks/rescue_terraform_failure.yml"
+    - name: Gather logs
+      ansible.builtin.include_tasks:
+        file: tasks/gather_startup_script_logs.yml
+        apply:
+          delegate_to: localhost
+      vars:
+        terraform_apply_stderr: "{{ terraform_output.results.1.stderr }}"
+    - name: Cleanup firewall and infrastructure
+      ansible.builtin.include_tasks:
+        file: tasks/rescue_terraform_failure.yml
+        apply:
+          delegate_to: localhost
       vars:
         deployment_name: "{{ deployment_name }}"
         workspace: "{{ workspace }}"
-        terraform_apply_stderr: "{{ terraform_output.results.1.stderr }}"
+    - name: Trigger failure (rescue blocks otherwise revert failures)
+      ansible.builtin.fail:
+        msg: "Failed while setting up test infrastructure"
+      when: true
 
 - name: Run Integration Tests
   hosts: remote_host
@@ -178,28 +191,12 @@
       loop: "{{ post_deploy_tests }}"
       loop_control:
         loop_var: test
-
-    ## Always cleanup, even on failure
     always:
-    - name: Delete Firewall Rule
-      register: fw_deleted
-      changed_when: fw_deleted.rc == 0
-      failed_when: false
-      run_once: true
-      delegate_to: localhost
-      ansible.builtin.command:
-        argv:
-        - gcloud
-        - compute
-        - firewall-rules
-        - delete
-        - "{{ deployment_name }}"
-    - name: Tear Down Cluster
-      run_once: true
-      changed_when: true # assume something got destroyed
-      delegate_to: localhost
-      environment:
-        TF_IN_AUTOMATION: "TRUE"
-      ansible.builtin.command:
-        cmd: terraform destroy -auto-approve
-        chdir: "{{ workspace }}/{{ deployment_name }}/primary"
+    - name: Cleanup firewall and infrastructure
+      ansible.builtin.include_tasks:
+        file: tasks/rescue_terraform_failure.yml
+        apply:
+          delegate_to: localhost
+      vars:
+        deployment_name: "{{ deployment_name }}"
+        workspace: "{{ workspace }}"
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
```diff
@@ -150,12 +150,19 @@
 
     ## Cleanup and fail gracefully
     rescue:
-    - name: Include rescue from terraform failure
+    - name: Cleanup firewall and infrastructure
       ansible.builtin.include_tasks: "tasks/rescue_terraform_failure.yml"
       vars:
         deployment_name: "{{ deployment_name }}"
         workspace: "{{ workspace }}"
+    - name: Gather logs
```
nick-stroud, line 158: With this ordering we will never get the serial port logs because we have already destroyed the machine. We need to gather logs while they the machine is still up (using the current methodology).

tools/cloud-build/daily-tests/ansible_playbooks/tasks/gather_logs.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Assert variables are defined
+  ansible.builtin.assert:
+    that:
+    - terraform_apply_stderr is defined
+
+- name: Remove New Lines From Terraform Apply STDERR
+  ansible.builtin.set_fact:
+    terraform_apply_stderr_one_line: "{{ terraform_apply_stderr | replace('\n',' ') }}"
+- name: Get Startup Script Logs
+  ansible.builtin.command: "{{ terraform_apply_stderr_one_line | regex_search('please run: (.+)', '\\1') | first }}"
```
nick-stroud, line 24: We are still prone to hitting the same error if for some reason wait for startup does not spit out the gcloud command. In the current ordering this error will happen after destruction but as pointed out above, this ordering means we will never get the logs.

When looking at your example builds it seems to me that the real issue is in the output generated by wait-for-startup. In the failure case wait-for-startup is generating:

```
could not detect end of startup script. Sleeping.
invalid return status '2'
to inspect the startup script output, please run:


MSG:
```

Why is it not printing the command after `please run:`?

Ideally we would answer that question and solve that problem as that is a user facing issue.

We we are not able to resolve that then instead I would suggest making this jinja expression more resilient. Ideally it would not fail if the regex does not find anything, although again this is surfacing a user facing error with wait-for-startup so I kind of want it to fail.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/tasks/gather_logs.yml
tpdownes: This PR is deliberately not intended to address the specific issue in the gcloud command but to solve the "prone-to-leak" problem of the current test implementation. A follow-up PR is planned to address the stdout issue that just happened to trigger the problem.

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
tpdownes: Re-shuffling things but ensuring that no step in log gathering is ever allowed to fail (and stop the rescue block).

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
```diff
@@ -150,12 +150,27 @@
...
-    - name: Include rescue from terraform failure
-      ansible.builtin.include_tasks: "tasks/rescue_terraform_failure.yml"
+    - name: Gather logs
+      ansible.builtin.include_tasks:
+        file: tasks/gather_logs.yml
+        apply:
+          delegate_to: localhost
+      vars:
+        project: "{{ project }}"
+        remote_node: "{{ remote_node }}"
```
nick-stroud, line 160: Using the remote_node is not sufficient in all cases. There is no guarantee that wait-for-startup is waiting on the remote_node. The original implementation pulls the gcloud command from the output reason, so that we pull logs from the machine where the failure occured. What machine we are waiting on is embedded in the blueprint and not really knowable until terraform apply.

Example: Batch mpi where remote node is the login node but wait-for-startup waits on builder.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
tpdownes: Point taken. Will document this as a task. I believe the right solution is to ensure that the relevant terraform modules have an output which includes the resource name and allow us to construct instructions like the gcloud command directly instead of via regex searches.

Re-worked PR to address only guarantee that terraform destroy is reached without addressing when Ansible crashes and burns if jinja can't resolve the command to invoke gcloud for startup scripts.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
```diff
@@ -155,12 +155,21 @@
...
     rescue:
+    - name: Gather logs
+      ansible.builtin.include_tasks:
+        file: tasks/gather_startup_script_logs.yml
+        apply:
+          delegate_to: localhost
+      vars:
+        terraform_apply_stderr: "{{ terraform_output.results.1.stderr }}"
     - name: Include rescue from terraform failure
-      ansible.builtin.include_tasks: "tasks/rescue_terraform_failure.yml"
```
nick-stroud, line 159: You have not added back in the fail out functionality. that was originally contained in this file. IIUC We will now show slurm tests passing that have actually failed. Please fix.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
nick-stroud, line 1: Can we reuse rescue_terraform_failure to clean up cluster at end of test here also?

tools/cloud-build/daily-tests/ansible_playbooks/tasks/rescue_terraform_failure.yml
nick-stroud, line 1: Please rename this file. Now that we are using it to tear down in other places the cluster the name no longer matches the use.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
tpdownes: Added

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
tpdownes: Done

tools/cloud-build/daily-tests/ansible_playbooks/tasks/gather_logs.yml
nick-stroud: As a bread crumb. Root cause was addressed in: https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1238

tools/cloud-build/daily-tests/ansible_playbooks/tasks/rescue_terraform_failure.yml
nick-stroud: As discussed, we might not get to this on this pr.

nick-stroud (approved):

PR #1239, opened 2023-04-27
Handle "wrong-type-of-packer" in `make warn-packer-missing`

mr0re1:
On some distributions, there may be another tool named `packer`, simple `which packer` doesn't provide correct signal.
Inspect `packer -h` for containing `"build image(s) from template"`.

### DIFF
--- a/tools/detect_packer.sh
+++ b/tools/detect_packer.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tools/detect_packer.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [[ -z $(which packer) ]]; then
+	echo "no"
+	exit 1
+fi
+
+# On some distributions, there may be another tool named `packer` installed by default.
+# https://developer.hashicorp.com/packer/tutorials/docker-get-started/get-started-install-cli#troubleshooting
+# Check if the right tool is installed by greep "help page"
+if ! packer -h 2>&1 | grep "build image(s) from template" >/dev/null; then
```
cboneti, line 24: Did you test the other tool? My understanding is that it would just wait on user input (simiar to `cat`).

cboneti (changes requested):
Please provide a description to this PR.

tools/detect_packer.sh
mr0re1: Yes, tested on CentOS with wrong type of packer installed , it hangs on `packer version` but `packer -h`, prints out usage.

cboneti (approved):
Awesome, thanks!

PR #1241, opened 2023-04-28
Address security alert CVE-2023-30608

tpdownes:
- upgrade to sqlparse 0.4.4 in Open Front End


cboneti:
I am assuming you did not get to test this, so I will deploy with this and try creating a cluster. I would feel more comfortable if we merged to develop first, but I don't know if that would complicate the release candidate process (if you kept the branch and merged to rc later).

cboneti (approved):
I have desployed OFE with this branch and verified that it is possible to create vpcs, file systems and clusters without problems.  
It is important to merge this into develop as well...

PR #1245, opened 2023-04-28
Add option to select zones for gke-node-pool

nick-stroud:
Adds option to select zones used by the gke-node-pool. 

This is a little tricky because it can take a list of zones, which we want to support. This does not play well with the normal HPC Toolkit paradigm of defining zone as a deployment variable. If the variable is named `zone` then we receive an error that the deployment variable type `string` does not match input type of `list(string)`. In short we do not do the same list flattening for deployment variables that we do for `use` (and I think this is WAI).

I thought about having a `zone` and a `zones` variable with some sort of merging but feel this is adding more complexity before we know that we need it. 

Tested the following cases:
- omitted zone
- multiple hard coded zones: `zones: [us-central1-c, us-central1-f]`
- reference deployment var: `zones: [$(vars.zone), us-central1-a]`


community/modules/compute/gke-node-pool/variables.tf
```diff
@@ -24,6 +24,12 @@ variable "cluster_id" {
   type        = string
 }
 
+variable "zones" {
+  description = "A list of zones to be used. Zones must be in region of cluster. If null, cluster zones will be picked. Note `zones` not `zone`; does not work with `zone` deployment variable."
```
cboneti, line 28: It would be good to document the default behavior. What happens if you do not specify zones?

cboneti (changes requested):
One small suggestion.

cboneti (approved):

PR #1247, opened 2023-04-29
Remove modReference

mr0re1:
* Remove modReference;
* Refactor modReference.validate into separate function;
* Use consistent name for Blueprint reciever `bp` (over `b`).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -92,17 +92,17 @@ type DeploymentGroup struct {
 
 // Module return the module with the given ID
 func (bp *Blueprint) Module(id string) (*Module, error) {
-	var found *Module
+	var mod *Module
 	bp.WalkModules(func(m *Module) error {
 		if m.ID == id {
-			found = m
+			mod = m
 		}
 		return nil
 	})
-	if found == nil {
+	if mod == nil {
 		return nil, fmt.Errorf("%s: %s", errorMessages["invalidMod"], id)
 	}
-	return found, nil
+	return mod, nil
 }
 
 // ModuleGroup returns the group containing the module
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/zclconf/go-cty/cty"
 	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v3"
 
 	"hpc-toolkit/pkg/modulereader"
@@ -91,26 +90,24 @@ type DeploymentGroup struct {
 	Kind             ModuleKind
 }
 
-func (g DeploymentGroup) getModuleByID(modID string) (Module, error) {
-	idx := slices.IndexFunc(g.Modules, func(m Module) bool { return m.ID == modID })
-	if idx == -1 {
-		return Module{}, fmt.Errorf("%s: %s", errorMessages["invalidMod"], modID)
+// Module return the module with the given ID
+func (bp *Blueprint) Module(id string) (*Module, error) {
+	var found *Module
+	bp.WalkModules(func(m *Module) error {
+		if m.ID == id {
+			found = m
+		}
+		return nil
+	})
+	if found == nil {
+		return nil, fmt.Errorf("%s: %s", errorMessages["invalidMod"], id)
 	}
-	return g.Modules[idx], nil
-}
-
-func (dc DeploymentConfig) getGroupByID(groupID string) (DeploymentGroup, error) {
-	groupIndex := slices.IndexFunc(dc.Config.DeploymentGroups, func(d DeploymentGroup) bool { return d.Name == groupID })
-	if groupIndex == -1 {
-		return DeploymentGroup{}, fmt.Errorf("%s: %s", errorMessages["groupNotFound"], groupID)
-	}
-	group := dc.Config.DeploymentGroups[groupIndex]
-	return group, nil
+	return found, nil
 }
 
 // ModuleGroup returns the group containing the module
-func (b Blueprint) ModuleGroup(mod string) (DeploymentGroup, error) {
-	for _, g := range b.DeploymentGroups {
+func (bp Blueprint) ModuleGroup(mod string) (DeploymentGroup, error) {
+	for _, g := range bp.DeploymentGroups {
 		for _, m := range g.Modules {
 			if m.ID == mod {
 				return g, nil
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -91,26 +90,24 @@ type DeploymentGroup struct {
...
-	return g.Modules[idx], nil
-}
-
-func (dc DeploymentConfig) getGroupByID(groupID string) (DeploymentGroup, error) {
-	groupIndex := slices.IndexFunc(dc.Config.DeploymentGroups, func(d DeploymentGroup) bool { return d.Name == groupID })
-	if groupIndex == -1 {
-		return DeploymentGroup{}, fmt.Errorf("%s: %s", errorMessages["groupNotFound"], groupID)
+// Module return the module with the given ID
+func (bp *Blueprint) Module(id string) (*Module, error) {
+	var found *Module
```
tpdownes, line 95: Elsewhere, we typically use `found` as a bool indicating success of a search. I'd rather this remain a name like `mod` or similar to indicate its type.

pkg/config/config.go
```diff
@@ -121,14 +118,23 @@ func (b Blueprint) ModuleGroup(mod string) (DeploymentGroup, error) {
...
-	g, err := b.ModuleGroup(mod)
+func (bp Blueprint) ModuleGroupOrDie(mod string) DeploymentGroup {
+	g, err := bp.ModuleGroup(mod)
 	if err != nil {
 		panic(fmt.Errorf("module %s not found in blueprint: %s", mod, err))
 	}
 	return g
 }
 
+func (bp *Blueprint) groupOrder(g DeploymentGroup) int {
```
tpdownes, line 129: The name `groupOrder` suggests a "is it sorted?" `bool` to me. I think this would be better as `groupIndex`.

pkg/config/config.go
tpdownes: I think `GetModule()` would be a better func name here. The current name suggests a constructor.

tpdownes (changes requested):
This is an initial review:

1. Please break out the `b`/`bp` changes into a separate commit.
2. Please address the func/variable name changes.

While you do that, I will review the logic of validation changes more thoroughly.

pkg/config/expand.go
```diff
@@ -510,39 +411,29 @@ func validateModuleSettingReference(bp Blueprint, mod Module, r Reference) error
...
+	tg := bp.ModuleGroupOrDie(tm.ID)
 
 	// references must refer to the same or an earlier group;
-	if targetModuleGroupIndex > callingModuleGroupIndex {
-		return fmt.Errorf("%s: %s is in the later group %s", errorMessages["intergroupOrder"], r.Module, targetModuleGroup.Name)
+	if bp.groupOrder(tg) > bp.groupOrder(g) {
+		return fmt.Errorf("%s: %s is in the later group %s", errorMessages["intergroupOrder"], r.Module, tg.Name)
 	}
 
 	// at this point, we have a valid intragroup or intergroup references to a
```
tpdownes, line 426: This comment now belongs to the code block below. The Packer error is self-explanatory and does not need its own comment. In principle, the Packer error is a subset of the "output not found" error, but I like making it explicit.

pkg/config/expand.go
```diff
@@ -510,39 +411,29 @@ func validateModuleSettingReference(bp Blueprint, mod Module, r Reference) error
 		return nil
 	}
 	g := bp.ModuleGroupOrDie(mod.ID)
-	callingModuleGroupIndex := slices.IndexFunc(bp.DeploymentGroups, func(d DeploymentGroup) bool { return d.Name == g.Name })
 
-	targetModuleGroupIndex, err := modToGrp(bp.DeploymentGroups, r.Module)
+	tm, err := bp.Module(r.Module)
```
tpdownes, line 415: I believe this block will become self-explanatory (no comment needed), if you change the func name to `GetModule` as advised elsewhere. Otherwise, this block needs a comment like:

```go
// error if reference target module not found
```

pkg/config/expand.go
```diff
@@ -405,95 +375,26 @@ func (dc *DeploymentConfig) applyGlobalVariables() error {
...
-		return fmt.Errorf("%s: %s.%s",
-			errorMessages["referenceWrongGroup"], ref.toGroupID, ref.toModuleID)
+	fg := bp.ModuleGroupOrDie(from.ID)
+	tg := bp.ModuleGroupOrDie(to.ID)
+	if bp.groupOrder(tg) > bp.groupOrder(fg) {
+		return fmt.Errorf("%s: %s is in a later group", errorMessages["intergroupOrder"], to.ID)
 	}
-
 	return nil
 }
```
tpdownes, line 400: Update  the comment below to read

```
* referenced module and module output exist
```

pkg/config/expand.go
```diff
@@ -405,95 +375,26 @@ func (dc *DeploymentConfig) applyGlobalVariables() error {
...
 
-func (ref modReference) validate(bp Blueprint) error {
-	callingModuleGroupIndex := slices.IndexFunc(bp.DeploymentGroups, func(d DeploymentGroup) bool { return d.Name == ref.fromGroupID })
-	if callingModuleGroupIndex == -1 {
-		return fmt.Errorf("%s: %s", errorMessages["groupNotFound"], ref.fromGroupID)
-	}
-
-	targetModuleGroupIndex, err := modToGrp(bp.DeploymentGroups, ref.toModuleID)
+func validateModuleReference(bp Blueprint, from Module, toID string) error {
```
tpdownes, line 383: Please add a small, bulleted list of conditions that this function is testing for.

pkg/config/expand.go
```diff
@@ -405,95 +375,26 @@ func (dc *DeploymentConfig) applyGlobalVariables() error {
...
+	to, err := bp.Module(toID)
 	if err != nil {
 		return err
```
tpdownes, line 386: I believe this block will become self-explanatory (no comment needed), if you change the func name to `GetModule` as advised elsewhere. Otherwise, this block needs a comment like:

```go
// error if reference target module not found
```

tpdownes (changes requested):
Review complete, please address these requests in addition to the previous requests.

mr0re1:
> Please break out the b/bp changes into a separate commit.

I can't, this change is forced by `go vet`

pkg/config/config.go
mr0re1: > Elsewhere, we typically use found ...

done

> I think GetModule() would be a better func name here. 

https://go.dev/doc/effective_go#Getters

pkg/config/expand.go
mr0re1: Made it less cumbersome by ignoring error.

pkg/config/expand.go
mr0re1: Removed all comments

PR #1255, opened 2023-05-01
Skip TestFindTerraform if no terraform is installed

mr0re1:
To prevent making terraform a build dependency
Lower required coverage for `pkg/shell` 20 -> 15%


tpdownes (approved):
A reasonable solution to allowing the error to be tested when we know it should be raised.

PR #1257, opened 2023-05-01
Remove ansible-lint to unblock PRs

mr0re1:


tpdownes (approved):
For posterity: we have never obtained reliably useful behavior from the pre-commit hook once ansible-lint made the transition to require repos be organized like "proper" playbooks. Going by memory, we can't use versions that are significantly earlier due to other blocking issues. We should revisit this when we support recent copies of Ansible in VMs.

PR #1258, opened 2023-05-02
Refactor create_deployment.sh

nick-stroud:
The end goal of this change was to be able to supply an arbitrary dict of deployment variables to be overwritten by the cli at blueprint creation. While working on this problem it seemed cleaner to bring some parts of the script into ansible. In the end it seemed easiest to write the full script in ansible.

This change:
- Removes `sed` for `network_name` opting to explicitly write deployment variables.
- Rewrite create_deployment.sh into ansible (create_deployment_directory.yml).
- Remove use of ALWAYS_RECOMPILE it was only used by condor and is not needed there.
- Remove `sed` to overwrite `max_node_count` as it did not seem to be serving a purpose.


nick-stroud:
All integration tests are passing. PR validation will fail until synced past #1257

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Create cli flag for extra deployment variables
+  ansible.builtin.set_fact:
+    deployment_vars_str: "--vars {{ cli_deployment_vars.items() | map('join', '=') | join(',') }}"
+  when: cli_deployment_vars is defined
+
+- name: Create Blueprint
+  ansible.builtin.command: |
+      ./ghpc create -l ERROR "{{ blueprint_yaml }}" \
+      --backend-config bucket=daily-tests-tf-state \
```
mr0re1, line 32: IIUC, we don't expose this bucket currently

community/examples/htcondor-pool.yaml
```diff
@@ -20,6 +20,8 @@ vars:
   deployment_name: htcondor-001
   region: us-central1
   zone: us-central1-c
+  network_name: htcondor-pool
```
tpdownes, line 23: I believe that you're "promoting" (sub)network_name to deployment variable to enable being overridden at CLI, correct?

We have previously made choice to attach these variables at module level because they are not used by non-VPC modules except v4 Slurm and Packer (which I intend to address). i.e. if it's doesn't need to be a deployment variable, let's not make it one.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
```diff
@@ -0,0 +1,52 @@
...
+    - blueprint_yaml is defined
+    - project is defined
+    - workspace is defined
+    - deployment_name is defined
+    - test_name is defined
+
+- name: Create cli flag for extra deployment variables
+  ansible.builtin.set_fact:
+    deployment_vars_str: "--vars {{ cli_deployment_vars.items() | map('join', '=') | join(',') }}"
+  when: cli_deployment_vars is defined
```
tpdownes, line 27: Add `and cli_deployment_vars is mapping` (after double-checking my syntax below)

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html#type-tests

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Create Blueprint
+  ansible.builtin.command: |
+      ./ghpc create -l ERROR "{{ blueprint_yaml }}" \
+      --backend-config bucket=daily-tests-tf-state \
+      {{ deployment_vars_str if deployment_vars_str is defined else '' }} \
+      --vars project_id={{ project }} \
+      --vars deployment_name={{ deployment_name }}
```
tpdownes, line 35: Just to state, this positioning of project_id and deployment_name should "win" over values in the mapping. If this is desirable (it probably is), please close this comment.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
tpdownes, line 32: This is exposed via the `GHPC_DEV_BUCKET` variable in the existing script and while that appears to go unused, I think it would be better to avoid duplication and facilitate setting this via secret manager or similar.

I suggest something like

```yaml
- name: Set bucket
  ansible.builtin.set_fact:
    state_bucket: "{{ state_bucket | default(daily-tests-tf-state) }}"
```

This enables `state_bucket` to be set all the way back at `daily-tests/builds/*`.

You might benefit from single quotes around the whole thing and double quotes around the bucket default value.

tpdownes (changes requested):

community/examples/htcondor-pool.yaml
tpdownes: Intended as a point of discussion - enabling CLI override may merit the change, but I'd like to understand the intent.

nick-stroud: You are correct on the motivation. There are three options I can think of:
1. Move to deployment vars so they can be overwritten, as I have done
2. Remove hard-coded naming of networks and just allow default to take place: named after deployment which in all cases is very close to actual name **(prefered now that I write these out)**. I would still leave them in the case of the packer example as there are already in deployment vars and needed there.
3. In cases where names are hardcoded as settings, use sed in the test to pre-modify the blueprint, [example](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/5bc086a1239519b1b27255c6915abc6d41a7ff11/tools/cloud-build/daily-tests/builds/batch-mpi.yaml#L59-L67)

nick-stroud: Discussed offline, went with case 2.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
nick-stroud: Added.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
nick-stroud: Good point. I think if test explicitly sets that it should take precedence. I switched the order to put `deployment_vars_str` last. I also tested to make sure having final line as empty string does not cause failure of command.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/create_deployment_directory.yml
nick-stroud: Discussed offline. My opinion is that we have not used this option in all the time it has been available. It is very easy to add this functionality if we transition this value to secret manager.

I do agree with your point that the value should be defined in a single place. I have set is as a fact and used the fact in subsequent tasks.

tpdownes (approved):

PR #1261, opened 2023-05-02
Add a simple example using Google Cloud Storage buckets

nick-stroud:
Creates an example focused on the usage of Google Cloud Storage with HPC Toolkit.

It was suggested to do a multi-group example (first group creates, second group references) but I opted to use a external bucket instead. This was guided by a few factors:

1. It is simpler to explain without the multi-group steps
2. The multigroup flow will change dramatically once import and export are implemented
3. Using the multi-group case restricts the usage of `random_suffix` which means that users are likely to get bucket naming clashes on deployment.


examples/README.md
```diff
@@ -665,6 +666,46 @@ python /var/tmp/qsim-example.py
...
+
+To use the blueprint you must supply the project id and the name of an existing
+bucket:
+
+```shell
+./ghpc create community/examples/google-cloud-storage.yaml \
+  --vars project_id=<project_id> \
+  --vars existing_bucket_name=<name_of_existing_bucket>
+```
+
```
cboneti, line 697: Isn't it standard for examples to show how much quota they require?

cboneti (changes requested):
I would prefer to show a data staging example, but I agree that this was the scope of what was defined earlier.  

Please take a look at the quota question.

examples/README.md
nick-stroud: It is not standard. None of the other community experimental examples have quota statements. Also this example only requires 2 x vCPU of e2 which I would bet is included in every starter project. I would propose that it is not worth the risk of drift to include it in this case.

cboneti (approved):

PR #1262, opened 2023-05-02
Address usability suggestions for multi-group deployments

tpdownes:
- do not write inputs file for group if there are no matching inputs
- silence usage printout after command errors
- create a .ghpc/artifacts directory where deployment metadata and export-outputs artifacts are stored (instead of .ghpc)
- - flush .ghpc/outputs during "create -w" (incl. a file warning user not to manually modify the artifacts directory)
- replace "executing terraform..." logging with descriptive messages
- improve description of automatically generated input variables in Terraform groups

TODO:
- replace "advanced usage" warning with export-outputs/import-inputs descriptions (will be included with Packer inputs support)


pkg/shell/terraform.go
```diff
@@ -76,14 +76,14 @@ func needsInit(tf *tfexec.Terraform) bool {
 
 func initModule(tf *tfexec.Terraform) error {
 	var err error
-	log.Printf("executing \"terraform -chdir=%s init\"\n", tf.WorkingDir())
 	if needsInit(tf) {
+		log.Printf("initializing terraform directory %s", tf.WorkingDir())
```
nick-stroud, line 80: You lost the new line on move. Typo or intentional?

pkg/modulewriter/modulewriter.go
```diff
@@ -373,14 +386,26 @@ func prepDepDir(depDir string, overwrite bool) error {
 	return nil
 }
 
+func prepArtifactsDir(deploymentio deploymentio.Deploymentio, artifactsDir string) error {
+	err := os.Mkdir(artifactsDir, 0700)
+	if err != nil && !os.IsExist(err) {
```
nick-stroud, line 391: I would change to MkdirAll so you are not dependent on creation of prevGroupDir which happens outside this function. I also would not allow the case where the dir already exists. We just deleted it so it should not be there.

pkg/modulewriter/modulewriter.go
```diff
@@ -355,6 +364,10 @@ func prepDepDir(depDir string, overwrite bool) error {
 		return fmt.Errorf("failed to create directory to save previous deployment groups at %s: %w", prevGroupDir, err)
 	}
 
+	if err := prepArtifactsDir(deploymentio, artifactsDir); err != nil {
```
nick-stroud, line 367: I would put this call before L361 to not split up processing of pervGroupDir. Also now that this is growing all of the prevGroupDir code should also go in its own method `prepPrevGroupDir`.

nick-stroud (changes requested):

pkg/shell/terraform.go
tpdownes: Unintentional but caused me to read the documentation for log more carefully for its newline behavior which I've never quite understood. It [states](https://pkg.go.dev/log#pkg-overview):

>  Every log message is output on a separate line: if the message being printed does not end in a newline, the logger will add one.

So, I'm removing all trailing `\n` from every log message in this package.

pkg/modulewriter/modulewriter.go
tpdownes: I misread the above block as being the last step of `.ghpc` processing. `prepArtifactsDir` moved up accordingly.

pkg/modulewriter/modulewriter.go
```diff
@@ -333,11 +337,16 @@ func prepDepDir(depDir string, overwrite bool) error {
...
 		}
 
-		// Confirm we have a previously written deployment dir before overwritting.
+		// Confirm we have a previously written deployment dir before overwriting.
 		if _, err := os.Stat(ghpcDir); os.IsNotExist(err) {
 			return fmt.Errorf(
 				"while trying to update the deployment directory at %s, the '.ghpc/' dir could not be found", depDir)
 		}
+
+		if err := os.RemoveAll(artifactsDir); err != nil {
```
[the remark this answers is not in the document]
tpdownes: Done, (2) is useful for the next reader.

pkg/modulewriter/modulewriter.go
tpdownes: Due to the shuffling of "RemoveAll", I agree it makes sense not to ignore any non-nil err. I'm not entirely sold on MkdirAll but will add it.

Seeing this code with fresh eyes, I think the passing of `deploymentio.Deploymentio` is probably over-complicating what ought to be a simple file write and I've removed that in favor of simple `os` calls.

cmd/export.go
```diff
@@ -31,7 +31,7 @@ func init() {
 	rootCmd.AddCommand(exportCmd)
 }
 
-const defaultArtifactsDir string = ".ghpc"
+const defaultArtifactsDir string = ".ghpc/artifacts"
```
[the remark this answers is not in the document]
tpdownes: Done. There is a bit of a rat's nest around `modulereader` and `modulewriter` that we need to disentangle some day so that all our packages source a common source of strings like this, but this one works.

pkg/modulewriter/modulewriter.go
```diff
@@ -348,20 +353,24 @@ func prepDepDir(depDir string, overwrite bool) error {
 		}
 	}
 
-	// clean up old dirs
+	if err := prepArtifactsDir(artifactsDir); err != nil {
+		return err
+	}
+
+	// remove any existing backups of deployment group
 	prevGroupDir := filepath.Join(ghpcDir, prevDeploymentGroupDirName)
```
nick-stroud, line 361: nit: I would put all the prevGroupDir in its own function (L360-L381) similar to what you have done with prepArtifactsDir. Something like prepPrevGroupDir.

nick-stroud (approved):

pkg/modulewriter/modulewriter.go
tpdownes: ack

PR #1263, opened 2023-05-02
Add "golden copy" test for text escaping

mr0re1:


tools/validate_configs/golden_copies/text_escape/.ghpc/deployment_metadata.yaml
tpdownes, line 1: Please be aware that #1262 will move the location of this file to `.ghpc/artifacts`. We will want to merge/rebase to avoid subsequent PR validation tests.

tpdownes: #1262 is merged. I advise rebase before merging this.

tools/validate_configs/test_configs/text_escape.yaml
```diff
@@ -0,0 +1,42 @@
...
+vars:
+  project_id:  #
+  deployment_name: text_escape
+  region: us-east4
+  zone: us-east4-c
+  labels:
+    red: \((blue))
+
+deployment_groups:
+- group: zero
```
nick-stroud, line 27: I don't believe having multiple groups is related to text escaping and we already have a golden test for igc. I would prefer to keep the scope smaller to only test what we claim to be testing: text escaping.

Please simplify the blueprint to have a single group and focus test around text escaping.

tools/validate_configs/test_configs/text_escape.yaml
nick-stroud, line 1: Please move golden test configs into their own folder. As they are currently intermixed with the rest of the test configs they are hard to identify. I would suggest:
`tools/validate_configs/golden_test_configs`

Or you could update the structure as a whole to be
`tools/validate_configs/golden/outputs` and `tools/validate_configs/golden/test_configs`

If we still want to run these through validator, which I think we do, then it should also be added to the [validate configs script](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8bac6ecaeb2d50dcc470321194e5bbe6a8c9f7c1/tools/validate_configs/validate_configs.sh#L119).

nick-stroud (changes requested):

PR #1264, opened 2023-05-03
Use dedicated dtype ModuleID and GroupName instead of string

mr0re1:
Rationale:
* To bolster interfaces;
* To enable move of validations to unmarshaling stage;

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
 	"golang.org/x/exp/maps"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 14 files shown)


tpdownes:
@mr0re1 #1262 created a merge conflict that needs to be addressed. #1265 (once merged) might, too.

pkg/config/config.go
```diff
@@ -82,16 +83,30 @@ var movedModules = map[string]string{
 	"community/modules/scheduler/cloud-batch-login-node": "modules/scheduler/batch-login-node",
 }
 
+// GroupName is the name of a deployment group
+type GroupName string
+
+// Validate checks that the group name is valid
+func (n GroupName) Validate() error {
+	if n == "" {
+		return errors.New(errorMessages["emptyGroupName"])
```
nick-stroud, line 92: nit: Previously you had suggested we start moving these to constants. I would recommend we start using the new pattern as we touch these. Would you move this particular message to a constant? Same for "duplicateGroup" message in this PR.

pkg/config/config.go
```diff
@@ -253,14 +268,17 @@ func (v *validatorConfig) check(name validatorName, requiredInputs []string) err
 	return nil
 }
 
+// ModuleID is a unique identifier for a module in a blueprint
+type ModuleID string
```
nick-stroud, line 272: nit: Should we validate this string is not empty?

pkg/config/config.go
nick-stroud, line 1: Make sure to rebase as Tom suggests.

nick-stroud (approved):

pkg/config/config.go
mr0re1: We should, but it's out of scope

pkg/config/config.go
mr0re1: Agree, but I would rather do it in separate change.

PR #1268, opened 2023-05-03
Write Packer intergroup input values

tpdownes:
- ensure Packer deployment groups are always 1-module long
- modify location to match Packer module destination directory
- evaluate the Packer settings using deployment variables and intergroup output values from prior groups

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -490,7 +490,8 @@ func (bp *Blueprint) addKindToModules() {
 	})
 }
 
-// setModulesInfo populates needed information from modules
+// checkModulesInfo ensures each module in the blueprint has known detailed
+// metadata (inputs, outputs)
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -141,6 +141,17 @@ func (bp Blueprint) ModuleGroupOrDie(mod ModuleID) DeploymentGroup {
 	return g
 }
 
+// GroupIndex returns the index of the input group in the blueprint
+// return -1 if not found
+func (bp *Blueprint) GroupIndex(groupName GroupName) int {
```
mr0re1, line 146: No need for a reference, just take a value.

mr0re1 (approved):

pkg/config/config.go
tpdownes: Done

PR #1270, opened 2023-05-03
Add `add_deployment_name_before_prefix` to blueprints used in test

mr0re1:
* Add `add_deployment_name_before_prefix` to blueprints used in test;
* Remove sed from omnia test.

### DIFF
--- a/community/examples/omnia-cluster.yaml
+++ b/community/examples/omnia-cluster.yaml
@@ -74,6 +74,7 @@ deployment_groups:
     - startup-manager
     settings:
       name_prefix: omnia-manager
+      add_deployment_name_before_prefix: true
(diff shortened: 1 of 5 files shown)


community/examples/omnia-cluster.yaml
nick-stroud, line 1: Please run integration tests associated with the change. My concern is that one of these is being used as the remote host and we need to update the name. Anyway, it won't hurt to run related tests.

nick-stroud (approved):

community/examples/omnia-cluster.yaml
mr0re1: All tests green (required additional changes to make omnia-test work)

PR #1272, opened 2023-05-03
Filter blueprint autocompletion on YAML files

tpdownes:
Cause tab completion when the user has typed `ghpc create` to filter on directories and files with the extension `yaml` or `yml`. For bash, this requires that the user install bash-completion and generate a ghpc-specific autocompletion configuration following steps in `ghpc completion --help`

The `fish` shell does not support this feature.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -66,11 +66,12 @@ var (
 	skipValidatorsDesc  = "Validators to skip"
 
 	createCmd = &cobra.Command{
-		Use:   "create BLUEPRINT_NAME",
-		Short: "Create a new deployment.",
-		Long:  "Create a new deployment based on a provided blueprint.",
-		Run:   runCreateCmd,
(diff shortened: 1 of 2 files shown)


tpdownes:
See [cobra docs](https://github.com/spf13/cobra/blob/284f4101043c4b1faebde411caec35a7b6e36494/shell_completions.md#dynamic-completion-of-nouns) for discussion/example.

cmd/create.go
nick-stroud, line 1: LGTM. I am assuming you have manually verified functionality.

nick-stroud (approved):

cmd/create.go
tpdownes: Yes, the only quibble I have is that it will include hidden directories in tab-completion. I filed https://github.com/spf13/cobra/issues/1950

PR #1288, opened 2023-05-05
Add `automatic_restart` to `vm-instance`

mr0re1:
Keep current logic around `compact_placement || spot`, but treat it as a default value, respecting user choise if `automatic_restart` was set explicitly.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#automatic_restart https://cloud.google.com/compute/docs/instances/setting-vm-host-options#settingoptions


modules/compute/vm-instance/variables.tf
```diff
@@ -245,6 +245,12 @@ variable "guest_accelerator" {
   default = null
 }
 
+variable "automatic_restart" {
+  description = "Specifies if the instance should be restarted if it was terminated by Compute Engine (not a user). Defaults to true."
+  type        = bool
+  default     = null
```
nick-stroud, line 251: This raises the same concerns as in #949. "Bool should not default to null; Bool should not be compared to null". I need to think on it some more as this case is somewhat simpler, but I wanted to raise this early so you could think about it as well.

modules/compute/vm-instance/variables.tf
```diff
@@ -245,6 +245,12 @@ variable "guest_accelerator" {
   default = null
 }
 
+variable "automatic_restart" {
+  description = "Specifies if the instance should be restarted if it was terminated by Compute Engine (not a user). Defaults to true."
```
nick-stroud, line 249: I think default to true may be misleading. More like, by default we will turn off when appropriate and otherwise use the platform default.

modules/compute/vm-instance/main.tf
```diff
@@ -38,9 +38,15 @@ locals {
 
   # both of these must be false if either compact placement or preemptible/spot instances are used
   # automatic restart is tolerant of GPUs while on host maintenance is not
-  automatic_restart           = local.compact_placement || var.spot ? false : null
+  automatic_restart_default   = local.compact_placement || var.spot ? false : null
```
tpdownes, line 41: There is an existing problem with `local.compact_placement` that we should address:

```
  compact_placement = try(var.placement_policy.collocation, null) != null
```

This assumes that any value at all for `var.placement_policy.collocation` implies that the user wants compact placement. There are 2 alternatives:

1. They have asked for a "spread" placement policy (the opposite of compact)
2. They have asked for an invalid placement policy.

This could probably with some validation blocks on `var.placement_policy` (something like `var.placement_policy == null || contains(var.placement_policy.collocation, ["valid", "values"]`.

I will also add this PR to a tracker for reasons we might consider upgrading to 1.3 (not obvious it helps, but it might).

tpdownes: The documentation is suggestive (though not definitive) that a spread placement policy should have not require specifying a restart policy. The [section for applying a compact policy][compact] to an existing instance includes the text:

> For compact placement policies, you must include the --maintenance-policy=TERMINATE and --no-restart-on-failure flags.

The [section for spread policies][spread] does not include this text.

[compact]: https://cloud.google.com/compute/docs/instances/define-instance-placement#apply-new-instance
[spread]: https://cloud.google.com/compute/docs/instances/define-instance-placement#apply-existing-instance

modules/compute/vm-instance/variables.tf
mr0re1: > default we will turn off when appropriate and otherwise use the platform default.

I'm not sure if it adds any clarity, I would just remove the mention of "default" all together.

modules/compute/vm-instance/variables.tf
mr0re1: `add_deployment_name_before_prefix`  case is indeed simpler, because we have full understanding of its behaviour and doesn't leave the scope of the module.

`automatic_restart`  is different story, I would let to provider to fall to it's default behaviour while allowing user to be explicit if they want to.

We need `optional[bool]`, not `bool`, and `default = null` achieves it.

modules/compute/vm-instance/main.tf
mr0re1: Discussed offline, the issue is not blocking for this PR. Tracked in separate issue.

modules/compute/vm-instance/main.tf
nick-stroud, line 1: Resolve failing pr validation.

nick-stroud (approved):

PR #1292, opened 2023-05-05
Don't perform file-wide text substitutions

mr0re1:
Don't perform file-wide text substitutions, limit scope to string literals in variables values

### DIFF
--- a/pkg/modulewriter/hcl_utils.go
+++ b/pkg/modulewriter/hcl_utils.go
@@ -26,16 +26,16 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-func escapeBlueprintVariables(hclBytes []byte) []byte {
+func escapeBlueprintVariables(s string) string {
 	// Convert \$(not.variable) to $(not.variable)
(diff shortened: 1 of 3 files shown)


pkg/modulewriter/hcl_utils.go
```diff
@@ -71,6 +71,12 @@ func TokensForValue(val cty.Value) hclwrite.Tokens {
 	}
 
 	ty := val.Type()
+	if ty == cty.String {
+		s := val.AsString()
+		s = escapeBlueprintVariables(escapeLiteralVariables(s))
```
tpdownes, line 76: Is this "call inside a call" necessary or is it just a way of converting 2 lines to 1 because most calls are "NO-OP"? Could it be done in the other order?

I personally prefer the 2-line approach but in any case, do not merge without a comment indicating that 2 things are happening here and indicate whether the order matters.

tpdownes (approved):
Approved on condition that a comment is added to the 2 calls to "escapeX". 2 lines vs 1 is a "nit"

PR #1300, opened 2023-05-09
Creates GPU GKE example, waiting for operations module

nick-stroud:
This change:
- Adds a GKE GPU example
- Documents the `kubernetes-operations` module
- Adds `enable_private_endpoint: false` to GKE examples.
  - For best security this variable should be true but it restricts access from the public internet and provides a poor first time user experience as the cluster is unreachable in many cases. Decision leave default to true but show it off in examples so user is aware.

### Submission Checklist

### DIFF
--- a/community/modules/compute/gke-job-template/README.md
+++ b/community/modules/compute/gke-job-template/README.md
@@ -89,6 +89,7 @@ No modules.
 | <a name="input_allocatable_cpu_per_node"></a> [allocatable\_cpu\_per\_node](#input\_allocatable\_cpu\_per\_node) | The allocatable cpu per node. Used to claim whole nodes. Generally populated from gke-node-pool via `use` field. | `list(number)` | <pre>[<br>  -1<br>]</pre> | no |
 | <a name="input_backoff_limit"></a> [backoff\_limit](#input\_backoff\_limit) | Controls the number of retries before considering a Job as failed. Set to zero for shared fate. | `number` | `0` | no |
 | <a name="input_command"></a> [command](#input\_command) | The command and arguments for the container that run in the Pod. The command field corresponds to entrypoint in some container runtimes. | `list(string)` | <pre>[<br>  "hostname"<br>]</pre> | no |
+| <a name="input_has_gpu"></a> [has\_gpu](#input\_has\_gpu) | Do nodes have GPUs attached. Generally populated from gke-node-pool via `use` field. | `list(bool)` | <pre>[<br>  false<br>]</pre> | no |
 | <a name="input_image"></a> [image](#input\_image) | The container image the job should use. | `string` | `"debian"` | no |
 | <a name="input_machine_family"></a> [machine\_family](#input\_machine\_family) | The machine family to use in the node selector (example: `n2`). If null then machine family will not be used as selector criteria. | `string` | `null` | no |
 | <a name="input_name"></a> [name](#input\_name) | The name of the job. | `string` | `"my-job"` | no |
--- a/community/modules/compute/gke-job-template/README.md
+++ b/community/modules/compute/gke-job-template/README.md
@@ -6,10 +6,8 @@ The job template file can be submitted as is or used as a template for further
 customization. Add the `instructions` output to a blueprint (as shown below) to
 get instructions on how to use `kubectl` to submit the job.
 
-This module is designed to`use` one or more `gke-node-pool` modules.
-
-that can be submitted to a GKE cluster
-using `kubectl` and will run on the specified node pool.
+This module is designed to `use` one or more `gke-node-pool` modules. The job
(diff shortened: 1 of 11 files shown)


community/modules/compute/gke-job-template/variables.tf
```diff
@@ -50,6 +50,12 @@ variable "allocatable_cpu_per_node" {
   default     = [-1]
 }
 
+variable "has_gpu" {
+  description = "Do nodes have GPUs attached. Generally populated from gke-node-pool via `use` field."
```
tpdownes, line 54: There are a few issues I see here:

1. I believe the point is to produce a job template that require GPUs, not whether nodes have GPUs (even if that is the logic of the Toolkit) so the description should say that.
2. I don't quite like crossing the line into having our module variable descriptions adopt Toolkit terms in part because the behavior of "use" takes some explaining.

Please change to something more like "Indicates that the job should request nodes with GPUs. Matches output value from gke-node-pool."

I think this can develop and be discussed with time, but I think a job or a deployment variable should express the request for GPUs and then node pools should be created to match the job. The approach we're taking here is backwards, in a sense, from how the job submitter thinks.

I also foresee difficulty surrounding `use` given that you appear to be producing a single template yet potentially consuming multiple node pools. The way I read the code, if you produce a set of node pools with at least one having GPUs, then supply that set to this module with `use`, then the job will always request GPUs.

I realize this is a difficult problem but we should probably flesh out a target or maybe reject CPU/GPU of heterogeneity in creating a template. This can develop with time and I'm not requesting changes other than the variable description.

community/modules/compute/gke-job-template/README.md
tpdownes, line 1: The top of this file (as it exists on develop) has some broken text. Fix the spacing before "`use`" and the whitespace problems.

```md
This module is designed to`use` one or more `gke-node-pool` modules.

that can be submitted to a GKE cluster
using `kubectl` and will run on the specified node pool.
```

tpdownes (changes requested):

examples/README.md
```diff
@@ -820,6 +821,32 @@ to the cluster using `kubectl` and will run on the specified node pool.
 
 [gke.yaml]: ../community/examples/gke.yaml
 
+### [ml-gke.yaml] ![community-badge] ![experimental-badge]
```
tpdownes, line 824: Please make it bloody obvious that you absolutely must set `vars.authorized_cidr` or the module won't work as presently designed.

nick-stroud: I think as designed it is obvious as you are required to override at the command line.

But since our validation does not support leaving vars empty and I will have to populate with something, I agree it is dangerous that someone will miss this newance.  

I have added a warning at the top of the documentation. I have added additional comments in the blueprint. I also tested that terraform apply will fail if the user does not update the value.

community/modules/compute/gke-job-template/README.md
```diff
@@ -6,10 +6,8 @@ The job template file can be submitted as is or used as a template for further
 customization. Add the `instructions` output to a blueprint (as shown below) to
 get instructions on how to use `kubectl` to submit the job.
 
-This module is designed to`use` one or more `gke-node-pool` modules.
-
-that can be submitted to a GKE cluster
-using `kubectl` and will run on the specified node pool.
+This module is designed to `use` one or more `gke-node-pool` modules. The job
+will be configured to run on the specified `gke-node-pool`.
```
tpdownes, line 10: Change to "on any of the ... pools".

tpdownes (approved):
Please address the very minor text change (assuming my interpretation is correct).

"alltrue" seems like better behavior for the goals of this module. In medium term, I think we might consider a validation/condition check of some sort that says you have to pick or choose. But we should do that after having more experience.

Feel free to squash the extra commits onto the original 2.

community/modules/compute/gke-job-template/variables.tf
tpdownes: Ultimately by "backwards" what I mean is (A) ideally the variable names would be different as you note and (B) the role of a Job (in the YAML k8s sense) is to describe its runtime requirements.

If you take a long-term reproducibility perspective, a node selector on pool name is not a great way of doing that. You'd rather say something like "I need an Intel processor above this generation" or so on. But this is a good practical solution for a module with the aim of introducing someone to running a (successful) job.

examples/README.md
tpdownes: You might choose a strong word than "updated" and might even link to something like "whatismyip.com"? your decision before merging.

community/modules/compute/gke-job-template/README.md
nick-stroud: Done.

examples/README.md
nick-stroud: Done.

PR #1304, opened 2023-05-10
Initial implementation of deploy command

tpdownes:
Support workflows for deploying single- or multi-group blueprints via the following sequence of commands:

```text
ghpc create blueprint.yaml
ghpc deploy deployment_root_directory
```

No direct user execution of Terraform or Packer is necessary. If the optional `--auto-approve` flag is supplied to `deploy`, then changes to cloud infrastructure are automatically applied. If not supplied, the user must approve every deployment group separately.

In this initial implementation, `ghpc` blindly iterates over all deployment groups in-order. It will only attempt to run "terraform apply" if "terraform plan -detailed-exitcode" returns code 2. It will always attempt to build Packer images. For each group, it will always attempt to invoke "import-inputs" and "export-outputs". These are operationally harmless on groups without inputs or exports, but may print log entries to stdout.

Example output:

```text
❯❯❯ ghpc create -w --vars project_id=project examples/hpc-cluster-small.yaml
Terraform group 'primary' was successfully created in directory hpc-small/primary
To deploy, run the following commands:

terraform -chdir=hpc-small/primary init
terraform -chdir=hpc-small/primary validate
terraform -chdir=hpc-small/primary apply

❯❯❯ ghpc deploy hpc-small
2023/05/10 14:31:05 initializing terraform directory hpc-small/primary
2023/05/10 14:31:23 testing if terraform state of hpc-small/primary is in sync with cloud infrastructure
2023/05/10 14:31:30 cloud infrastructure requires changes
2023/05/10 14:31:30 Summary of proposed changes: Plan: 13 to add, 0 to change, 0 to destroy.
Display full proposed changes, Apply proposed changes, Stop and exit, Continue without applying? [d,a,s,c]: a
2023/05/10 14:31:35 running terraform apply on group hpc-small/primary
module.slurm_login.module.slurm_cluster_login_node.data.google_compute_default_service_account.default: Reading...
##
## MANY LINES OF TERRAFORM APPLY OUTPUT
##
Apply complete! Resources: 13 added, 0 changed, 0 destroyed.
2023/05/10 14:36:22 collecting terraform outputs from hpc-small/primary
2023/05/10 14:36:23 group primary contains no artifacts to export
❯❯❯ ghpc deploy hpc-small --auto-approve
2023/05/10 14:45:04 testing if terraform state of hpc-small/primary is in sync with cloud infrastructure
2023/05/10 14:45:08 cloud infrastructure requires no changes
2023/05/10 14:45:08 collecting terraform outputs from hpc-small/primary
2023/05/10 14:45:08 group primary contains no artifacts to export
```

### Submission Checklist

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -1,0 +1,144 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cmd defines command line utilities for ghpc
+package cmd
+
+import (
+	"fmt"
+	"hpc-toolkit/pkg/config"
+	"hpc-toolkit/pkg/shell"
+	"log"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	artifactsFlag := "artifacts"
+
+	deployCmd.Flags().StringVarP(&artifactsDir, artifactsFlag, "a", "", "Artifacts output directory (automatically configured if unset)")
+	deployCmd.MarkFlagDirname(artifactsFlag)
+
+	autoApproveFlag := "auto-approve"
+	deployCmd.Flags().BoolVarP(&autoApprove, autoApproveFlag, "", false, "Automatically approve proposed changes")
+
+	// anticipate mutually exclusive flags from-directory, from-group, from-blueprint
+	deploymentFlag := "from-directory"
+	deployCmd.Flags().StringVarP(&deploymentRoot, deploymentFlag, "d", "", "Deployment root directory")
+	deployCmd.MarkFlagDirname(deploymentFlag)
+	deployCmd.MarkFlagRequired(deploymentFlag)
+	rootCmd.AddCommand(deployCmd)
+}
+
+var (
+	deploymentRoot string
+	autoApprove    bool
+	applyBehavior  shell.ApplyBehavior
+	deployCmd      = &cobra.Command{
+		Use:          "deploy -d DEPLOYMENT_DIRECTORY",
+		Short:        "deploy all resources in a Toolkit deployment directory.",
+		Long:         "deploy all resources in a Toolkit deployment directory.",
+		Args:         cobra.ExactArgs(0),
+		PreRun:       setApplyBehavior,
+		RunE:         runDeployCmd,
+		SilenceUsage: true,
+	}
+)
+
+func setApplyBehavior(cmd *cobra.Command, args []string) {
+	if autoApprove {
+		applyBehavior = shell.AutomaticApply
+	} else {
+		applyBehavior = shell.PromptBeforeApply
+	}
+}
+
+func runDeployCmd(cmd *cobra.Command, args []string) error {
+	if artifactsDir == "" {
+		artifactsDir = filepath.Clean(filepath.Join(deploymentRoot, defaultArtifactsDir))
+	}
+
+	if err := shell.CheckWritableDir(artifactsDir); err != nil {
+		return err
+	}
+
+	expandedBlueprintFile := filepath.Join(artifactsDir, expandedBlueprintFilename)
+	dc, err := config.NewDeploymentConfig(expandedBlueprintFile)
+	if err != nil {
+		return err
+	}
+
+	if err := shell.ValidateDeploymentDirectory(dc.Config.DeploymentGroups, deploymentRoot); err != nil {
+		return err
+	}
+
+	for _, group := range dc.Config.DeploymentGroups {
+		groupDir := filepath.Join(deploymentRoot, string(group.Name))
+		if err = shell.ImportInputs(groupDir, artifactsDir, expandedBlueprintFile); err != nil {
+			return err
+		}
+
+		var err error
+		switch group.Kind {
+		case config.PackerKind:
+			// Packer groups are enforced to have length 1
+			moduleDir := filepath.Join(groupDir, string(group.Modules[0].ID))
+			err = deployPackerGroup(moduleDir)
+		case config.TerraformKind:
+			err = deployTerraformGroup(groupDir)
+		default:
+			err = fmt.Errorf("group %s is an unsupported kind %s", groupDir, group.Kind.String())
+		}
+		if err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+func deployPackerGroup(moduleDir string) error {
+	if err := shell.TestPacker(); err != nil {
+		return err
+	}
+	buildImage := applyBehavior == shell.AutomaticApply || shell.AskForConfirmation("Build Packer image?")
+	if buildImage {
+		log.Printf("initializing packer module at %s", moduleDir)
+		if err := shell.ExecPackerCmd(moduleDir, false, "init", "."); err != nil {
+			return err
+		}
+		log.Printf("validating packer module at %s", moduleDir)
+		if err := shell.ExecPackerCmd(moduleDir, false, "validate", "."); err != nil {
+			return err
+		}
+		log.Printf("building image using packer module at %s", moduleDir)
+		if err := shell.ExecPackerCmd(moduleDir, true, "build", "."); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func deployTerraformGroup(groupDir string) error {
+	tf, err := shell.ConfigureTerraform(groupDir)
+	if err != nil {
+		return err
+	}
+
+	if err = shell.ExportOutputs(tf, artifactsDir, applyBehavior); err != nil {
+		return err
+	}
+	return nil
+}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -34,11 +34,6 @@ func init() {
 	autoApproveFlag := "auto-approve"
 	deployCmd.Flags().BoolVarP(&autoApprove, autoApproveFlag, "", false, "Automatically approve proposed changes")
 
-	// anticipate mutually exclusive flags from-directory, from-group, from-blueprint
-	deploymentFlag := "from-directory"
-	deployCmd.Flags().StringVarP(&deploymentRoot, deploymentFlag, "d", "", "Deployment root directory")
-	deployCmd.MarkFlagDirname(deploymentFlag)
-	deployCmd.MarkFlagRequired(deploymentFlag)
 	rootCmd.AddCommand(deployCmd)
 }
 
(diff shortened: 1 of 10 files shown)


pkg/shell/terraform.go
```diff
@@ -136,9 +148,29 @@ func getOutputs(tf *tfexec.Terraform) (map[string]cty.Value, error) {
...
+	var apply bool
 	if wantsChange {
-		return map[string]cty.Value{},
-			fmt.Errorf("cloud infrastructure requires changes; please run \"terraform -chdir=%s apply\"", tf.WorkingDir())
+		log.Println("cloud infrastructure requires changes")
+		switch applyBehavior {
+		case AutomaticApply:
+			apply = true
+		case PromptBeforeApply:
+			apply = AskForConfirmation(fmt.Sprintf("Do you want to deploy group %s", tf.WorkingDir()))
```
nick-stroud, line 158: I think `no` should stop the process. If we want a skip function as was discussed that should be a third option.

Currently if I say not it continue to try to deploy the next groups:
```
2023/05/10 23:58:17 initializing terraform directory igc-01/one
2023/05/10 23:58:22 testing if terraform state of igc-01/one is in sync with cloud infrastructure
2023/05/10 23:58:24 cloud infrastructure requires changes
Do you want to deploy group igc-01/one [y/n]: n
2023/05/10 23:58:28 collecting terraform outputs from igc-01/one
2023/05/10 23:58:28 group one contains no artifacts to export
2023/05/10 23:58:28 collecting outputs for group two from group one
Error: consider running "ghpc export-outputs igc-01/one" (detailed error below)
The configuration file "igc-01/.ghpc/artifacts/one_outputs.tfvars" could not be read.
```

I think it would be reasonable to spit out the same exit as from export when changes are needed:

```
2023/05/11 00:01:12 testing if terraform state of igc-01/one is in sync with cloud infrastructure
2023/05/11 00:01:14 cloud infrastructure requires changes
Error: cloud infrastructure requires changes; please run "terraform -chdir=igc-01/one apply"
```

nick-stroud, line 158: I feel there would be a lot of value to print the plan before asking if we should apply. If there is a way for us to add this, even if it is not efficient to start, I think we should as it totally changes the user experience.

My recommendation to not overburden this initial implementation would be to call plan again, now to output, under the circumstances that changes are needed and we need user approval. 

I think you could simply add the following before asking the user to approve:

```
		log.Printf("generating plan for group %s", tf.WorkingDir())
		tf.SetStdout(os.Stdout)
		tf.SetStderr(os.Stderr)
		tf.Plan(context.Background()) // guessing on syntax
		tf.SetStdout(nil)
		tf.SetStderr(nil)
```

User output would look like:

```
initializing terraform directory igc-01/one
testing if terraform state of igc-01/one is in sync with cloud infrastructure
cloud infrastructure requires changes
generating plan for group igc-01/one
...
... many lines of terraform plan
...
Do you want to deploy group igc-01/one [y/n]:
```

Then we can optimize so we are not running plan twice in future pr.

pkg/shell/terraform.go
```diff
@@ -136,9 +148,29 @@ func getOutputs(tf *tfexec.Terraform) (map[string]cty.Value, error) {
 		}
 	}
 
+	var apply bool
 	if wantsChange {
-		return map[string]cty.Value{},
-			fmt.Errorf("cloud infrastructure requires changes; please run \"terraform -chdir=%s apply\"", tf.WorkingDir())
+		log.Println("cloud infrastructure requires changes")
```
nick-stroud, line 153: nit: Remove timestamp from output lines. If you do not do this in this PR, please open a task.

cmd/deploy.go
```diff
@@ -0,0 +1,144 @@
...
+			err = deployTerraformGroup(groupDir)
+		default:
+			err = fmt.Errorf("group %s is an unsupported kind %s", groupDir, group.Kind.String())
+		}
+		if err != nil {
+			return err
+		}
+
+	}
+	return nil
```
nick-stroud, line 109: nit: Probably for future PR. We should print out destroy commands (and sequence) at end of deploy. We are now doing so much for the user, I suspect it will be hard for them to figure out what to do to destroy. Open task to close this.

cmd/deploy.go
nick-stroud, line 1: nit: for `^C` behavior, I think the right thing to do would be to at least try to pass kill to terraform. Not needed for this PR but please open a task.

cmd/deploy.go
```diff
@@ -0,0 +1,144 @@
...
+
+	deployCmd.Flags().StringVarP(&artifactsDir, artifactsFlag, "a", "", "Artifacts output directory (automatically configured if unset)")
+	deployCmd.MarkFlagDirname(artifactsFlag)
+
+	autoApproveFlag := "auto-approve"
+	deployCmd.Flags().BoolVarP(&autoApprove, autoApproveFlag, "", false, "Automatically approve proposed changes")
+
+	// anticipate mutually exclusive flags from-directory, from-group, from-blueprint
+	deploymentFlag := "from-directory"
+	deployCmd.Flags().StringVarP(&deploymentRoot, deploymentFlag, "d", "", "Deployment root directory")
```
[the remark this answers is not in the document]
tpdownes: I think it is likely that future iterations of this command are going to need some significant custom validation logic no matter what. This is exposed through "PreRun/PreRunE". There is already an example of a simple PreRun hook here.

e.g. I think it's easy to say "these flags are mutually exclusive" but my quick read does not suggest an easy way of saying "but you must supply 1".

So, from a technical perspective of validating the inputs, I think all options are about the same amount of work. Different choices (esp. your suggestion) would probably require us to write custom usage/help functions.

cmd/export.go
```diff
@@ -75,28 +75,6 @@ func setArtifactsDir(cmd *cobra.Command, args []string) {
...
-	}
-
-	groupKinds, err := shell.GetDeploymentKinds(dc)
-	if err != nil {
-		return config.UnknownKind, err
-	}
-
-	kind, ok := groupKinds[group]
-	if !ok {
-		return config.UnknownKind, fmt.Errorf("deployment group %s not found in expanded blueprint", group)
```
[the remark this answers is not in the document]
tpdownes: It's now the responsibility of the caller to validate the existence of a group that has been specified actually exists. Some commands target a group (e.g. export-outputs, import-inputs, future versions of this command) while others (e.g. this one) target a deployment directory and then iterate over groups. So this `func` was doing more than is generally useful and I found it better to have the blueprint available to the callers anyways.

I anticipate extending `shell.ValidateDeploymentDirectory` to verify kinds (terraform/packer) based on its best guess of the kind after scanning the directory.

pkg/shell/terraform.go
tpdownes: I will alter this to have a 3d skip-and-continue option.

tpdownes: I can do this, but will probably choose the "optimal" approach of directing the plan to a file and running apply on the file. Lots of good reasons and not obviously much more effort than running twice.

cmd/deploy.go
tpdownes: I agree with the problem and think the answer is to create a destroy command. Packer images an obvious edge case.

tpdownes: Creating task.

cmd/deploy.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+func init() {
+	artifactsFlag := "artifacts"
+
+	deployCmd.Flags().StringVarP(&artifactsDir, artifactsFlag, "a", "", "Artifacts output directory (automatically configured if unset)")
+	deployCmd.MarkFlagDirname(artifactsFlag)
+
+	autoApproveFlag := "auto-approve"
+	deployCmd.Flags().BoolVarP(&autoApprove, autoApproveFlag, "", false, "Automatically approve proposed changes")
```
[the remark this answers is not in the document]
tpdownes: Resolved offline to avoid introducing a short option. Reasoning:

1. Explicitness / alignment with terraform.
2. Easier to add in future than to take away.

pkg/config/config.go
```diff
@@ -152,6 +152,15 @@ func (bp Blueprint) GroupIndex(groupName GroupName) int {
 	return -1
 }
 
+// Group returns the deployment group with a given name
+func (bp Blueprint) Group(groupName GroupName) (DeploymentGroup, error) {
```
mr0re1, line 156: `s/groupName/n/` it's too verbose and "ungoolangy"

pkg/shell/terraform.go
tpdownes: I've chosen the following approach:

- capture the stdout of plan (in human-readable format)
- prompt the user with 4 options: `Display proposed changes, Apply proposed changes, Stop and exit, Continue without applying? [d,a,s,c]`
- Loop forever until user picks 1 of the 4 options.

Logic behind making display of the plan optional is that it is very long and causes me to become disoriented if printed immediately to screen.

There is a sub-optimality in that there does not appear to be an existing func in tfexec that will capture both human and machine readable format for the plan. So there is no guarantee that the apply stage will attempt _exactly_ the changes being proposed. I will address this with a task for future work.

pkg/config/config.go
tpdownes: Addressed here and in `GroupIndex` since the type now conveys the argument sufficiently.

cmd/deploy.go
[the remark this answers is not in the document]
tpdownes: Resolved in favor of not requiring a flag. Future likely to look like:

```
ghpc deploy [--group group_name] deployment_directory_or_blueprint_yaml
```

Where group_name specifies the group up to which you are willing to deploy.

pkg/shell/terraform.go
tpdownes: Continuing this conversation at https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1304#discussion_r1191575429 and resolving this conversation.

nick-stroud: I disagree with requiring the user to take action before they are told what is about to change (in this case the `d` option). In addition we are now providing the user with 4 options here.

Terraform's native behavior is to always print the delta. I don't believe we should deviate from this behavior without good reason. 

I could see at some point in the future only printing the summary line (`x added, x modified, x deleted`) and having the user ask for the full output, but I don't think that should be our starting point.

nick-stroud (changes requested):

pkg/shell/terraform.go
nick-stroud: @cboneti could you help weight in here as a tiebreaker.

cboneti: both of you  are bringing important points here. 
I believe it would be important to give a sense of the changes, so this is what I propose, do one of the two options as a starting point. Pick what you prefer or whatever is easier: 
1) Print, as Nick pointed, a summary line as described below: 
```
❯❯❯ ghpc deploy hpc-small
2023/05/10 14:31:05 initializing terraform directory hpc-small/primary
2023/05/10 14:31:23 testing if terraform state of hpc-small/primary is in sync with cloud infrastructure
2023/05/10 14:31:30 cloud infrastructure requires changes:

Plan: 13 to add, 0 to change, 0 to destroy.

Display proposed changes, Apply proposed changes, Stop and exit, Continue without applying? [d,a,s,c]:
...
```

or, if it is easier this way: 
2) always print terraform plan output and then adjust the prompt accordingly (notice changes after tf plan): 
```
❯❯❯ ghpc deploy hpc-small
2023/05/10 14:31:05 initializing terraform directory hpc-small/primary
2023/05/10 14:31:23 testing if terraform state of hpc-small/primary is in sync with cloud infrastructure
##
## MANY LINES OF TERRAFORM PLAN OUTPUT
##
2023/05/10 14:31:30 cloud infrastructure requires changes for hpc-small/primary
Apply proposed changes, Stop and exit, Continue without applying? [a,s,c]: a
2023/05/10 14:31:35 running terraform apply on group hpc-small/primary
module.slurm_login.module.slurm_cluster_login_node.data.google_compute_default_service_account.default: Reading...
...
```

tpdownes: Offline discussion resulted in choosing (1). 3c598f29 includes:

- printing a summary of changes immediately upon prompt
- continue to allow an optional full summary by entering `d`

Because it was simple within this effort of work, I added the functionality of applying the plan actually being shown to the user rather than "crossing your fingers" that nothing has changed by the time apply is run.

Edge cases addressed:

- Packer full and summary of changes are the same
- Future task will be created to use `PlanJSON` and parse the output for the [change_summary](https://developer.hashicorp.com/terraform/internals/machine-readable-ui#change-summary) field instead of regexp for `Plan:`. If no match is found, an advisory to print the full proposed changes is written to screen.

cboneti: LGTM, please assign to Nick

nick-stroud: LGTM. Just want to test then will approve.

nick-stroud (approved):

PR #1310, opened 2023-05-12
Add a ramble execute module

douglasjacobsen:
This merge adds a new ramble-execute module to create a startup script to execute ramble commands on a VM.


nick-stroud (changes requested):

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
```diff
@@ -0,0 +1,34 @@
...
+  - name: Execute ramble commands
+    ansible.builtin.shell: |
+      . {{ spack_path }}/share/spack/setup-env.sh
+      . {{ ramble_path }}/share/ramble/setup-env.sh
+      echo "" >> {{ log_file }}
+      echo " === Starting ramble commands ===" >> {{ log_file }}
+%{if length(COMMANDS) > 0 ~}
+%{for c in COMMANDS ~}
+      echo " === Executing command: ${c} === " >> {{ log_file }}
+      ${c} >> {{ log_file }}
```
tpdownes, line 31: one of the frustrating bits about the Spack module is that it generates log entries _only_ in its own file and without the timestamps I need for debugging / alignment with other startup steps. Please do either of the following:

- stop redirecting output entirely and allow it to be captured with Ansible
- pipe to `tee -a`

tpdownes (changes requested):

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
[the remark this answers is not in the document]
nick-stroud: My current recommendation would be:
- Take a string as an input and let user define multi-line string in blueprint
- Document in description shell runtime and that scrip will be pre/post pended (to set expectations that #! python won't work)
- Perform any script modification (such as adding env activation and logging in main.tf
- pass full script to ansible template using the following syntax:

```yaml
- name: Execute Ramble Commands
  hosts: localhost
  vars:
    ...
    commands: ${ indent(4, yamlencode( "${commands}" ))}
```

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
douglasjacobsen: I've tested both of these solutions, and `ansible.builtin.shell` does not appear to print out stdout or stderr.

Supposedly, `ansible.builtin.command` can do this, but last time when we were looking at using command it was determined that it was strictly worse in terms of usability than shell.

community/modules/scripts/ramble-execute/README.md
```diff
@@ -0,0 +1,91 @@
+## Description
+
+This module will create startup-script runner that will execute Ramble commands.
+
+Ramble is a multi-platform experimentation framework capable of driving
```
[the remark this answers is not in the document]
douglasjacobsen: I'd prefer to not replicate the text too much. Currently this text is replicated in both the ramble-setup and ramble-execute modules, and in Ramble's repo itself. I'd rather just direct people to the ramble repo and have them learn about it there instead.

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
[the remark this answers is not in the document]
douglasjacobsen: Just FYI, doing this makes it not possible to inject redirection of the spack / ramble commands.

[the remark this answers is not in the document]
douglasjacobsen: I've played with this quite a bit over the last few days. It seems `ansible.builtin.shell` doesn't output stdout / stderr even if a redirection is not used on the individual commands.

I added `set -e` so that errors in ramble commands properly cause the ansible task to error out, and provide a reasonable error code.

I also added debugging for the command output, so it would print in the output of `journalctl -u google-startup-scripts.service`.

I'm not sure if the debug output is in line with your expectations, but currently this is the only way I've been successful in getting output into the serial port 1.

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
douglasjacobsen: I added a comment about this above, but the solution I've pushed right now is to register the output and add `debug` to the ramble execute task.

Outside of this, I have been unable to get the output written to serial port 1 using `ansible.builtin.shell`.

community/modules/scripts/ramble-execute/templates/ramble_execute.yml.tpl
```diff
@@ -28,20 +28,18 @@
 
     - name: Execute ramble commands
       ansible.builtin.shell: |
-        set -e
         . {{ spack_path }}/share/spack/setup-env.sh
         . {{ ramble_path }}/share/ramble/setup-env.sh
+
+        set -eo pipefail
```
nick-stroud, line 34: nit: I would put this at the top of the script.

community/modules/scripts/ramble-execute/templates/ramble_execute.yml.tpl
```diff
@@ -28,20 +28,18 @@
...
         echo " === Starting ramble commands ==="
+        {
         {{ commands }}
+        } | tee -a {{ log_file }}
```
nick-stroud, line 38: nit: I would wrap everything in the tee, including the setup-env and the echo commands.

community/modules/scripts/ramble-execute/README.md
```diff
@@ -0,0 +1,101 @@
...
+> been tested in limited capacity.
+
+# Examples
+
+## Basic Example
+
+Below is a basic example of using this module.
+
+```yaml
+- id: spack
```
nick-stroud, line 28: nit: indent the example by 2 spaces. In a blueprint they will be indented so it makes it easier to copy paste.

community/modules/scripts/ramble-execute/variables.tf
```diff
@@ -0,0 +1,52 @@
...
+
+variable "log_file" {
+  description = "Log file to write output from Ramble execute steps into"
+  default     = "/var/log/ramble-execute.log"
+  type        = string
+}
+
+variable "spack_path" {
+  description = "Path to the Spack installation"
+  default     = null
```
nick-stroud, line 31: IIUC, spack path and ramble path are required. There should be no default, not even null, for required variables.

community/modules/scripts/ramble-execute/main.tf
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  commands_content = indent(4, yamlencode(var.commands))
```
nick-stroud, line 18: The null default needs to be handled with a conditional here.

community/modules/scripts/ramble-execute/main.tf
```diff
@@ -0,0 +1,49 @@
...
+    "${path.module}/templates/ramble_execute.yml.tpl",
+    {
+      spack_path  = var.spack_path
+      ramble_path = var.ramble_path
+      log_file    = var.log_file
+      commands    = local.commands_content
+    }
+  )
+
+  previous_ramble_runner_content = var.ramble_runner == null ? "" : var.ramble_runner["content"]
```
nick-stroud, line 30: We should add validation that:
1. The previous runner is ansible
2. That the previous runner has content (runners can also have a file instead)

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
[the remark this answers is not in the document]
nick-stroud: Lots of back and forth offline, this is now resolved.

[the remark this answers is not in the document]
nick-stroud: We resolved this by teeing the whole block.

community/modules/scripts/ramble-execute/templates/ramble_execute.tpl
nick-stroud: I believe this has been resolved.

community/modules/scripts/ramble-execute/README.md
nick-stroud, line 1: Nit: Please add a PR description above the checklist. Something minimal is fine.

nick-stroud (approved):

community/modules/scripts/ramble-execute/README.md
nick-stroud: NVM.

tpdownes (approved):
Thank you for adding the debug output in a70a77d33244c07cd084e6b9afa6c54c523e4481

PR #1322, opened 2023-05-13
Update docs on enable_reconfigure now that gcloud bug is fixed

nick-stroud:
Tested:
- Add partition
- Resize partition
- Add network storage
- Delete partition
- gcloud project config mismatch

### Submission Checklist

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
@@ -67,6 +67,18 @@ activated through the `enable_reconfigure` setting:
     enable_reconfigure: true
 ```
 
+To reconfigure a running cluster:
+
+1. Edit the blueprint with the desired configuration changes
+1. Call `ghpc create <blueprint> -w` to overwrite the deployment directory
+1. Follow instructions in terminal to deploy
+
+The following are examples of updates that can be made to a running cluster:
+
+* Add or remove a partition to the cluster
+* Resize an existing partition
+* Attach new network storage to an existing partition
+
 This option has some additional requirements:
 
 * The Pub/Sub API must be activated in the target project:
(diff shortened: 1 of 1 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -78,27 +90,10 @@ This option has some additional requirements:
   development environment deploying the cluster. One can use following commands:
 
   ```bash
-  wget https://raw.githubusercontent.com/SchedMD/slurm-gcp/5.7.2/scripts/requirements.txt
-  pip3 install -r requirements.txt --user
+  pip3 install -r https://raw.githubusercontent.com/SchedMD/slurm-gcp/5.7.2/scripts/requirements.txt --user
```
mr0re1, line 93: `--user` will throw if is run from virtualenv. Do we want to discourage usage of virtualenv?

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
```diff
@@ -67,6 +67,18 @@ activated through the `enable_reconfigure` setting:
...
+1. Edit the blueprint with the desired configuration changes
+1. Call `ghpc create <blueprint> -w` to overwrite the deployment directory
+1. Follow instructions in terminal to deploy
+
+The following are examples of updates that can be made to a running cluster:
+
+* Add or remove a partition to the cluster
+* Resize an existing partition
+* Attach new network storage to an existing partition
+
```
cboneti, line 81: Optional: maybe add that  changing the VM type of a partition will likely not work. (I tried with a gpu vm today and it did not work -- as it fails to update the instance template). It seems like a common case.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
cboneti: good point, also, it doesn't seem necessary if ran by non-root. It default to --user, right?

cboneti (approved):
I left two comments, consider it approved after you address them.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
nick-stroud: I am reading that "pip defaults to installing Python packages to a system directory". I went through this process when testing and indeed the `--user` argument is denied when installing into virtualenv, but the error message is very clear. I would opt to leave it as we do not mention anything about a virtual environment. I also don't feel strongly so if you still disagree I can update.

nick-stroud: In fact I just ended up removing `--user` as it does not make much difference to me and I want to move forward with this PR without waiting for re-review.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/README.md
nick-stroud: Done.

PR #1344, opened 2023-05-17
Add pre-commit to check for ghpc_module label

nick-stroud:
This change:
- Adds a stand alone python tool to validate consistency of `ghpc_module` label that can be used for billing filtering
- Adds a check that provider-meta block is consistent with module naming
- Adds pre-commit to run the above check

```
$ time python3 tools/label-check.py
real    0m0.072s
user    0m0.052s
sys     0m0.026s
```

### Submission Checklist

### DIFF
--- a/tools/label-check.py
+++ b/tools/label-check.py
@@ -1,0 +1,182 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import glob
+import sys
+
+
+LOCALS_TEMPLATE = '''
+locals {{
+  # This label allows for billing report tracking based on module.
+  labels = merge(var.labels, {{ ghpc_module = \"{module_label}\" }})
+}}
+'''
+
+
+class ModulePath:
+    """Convenience class to get various path related information about a module"""
+    MAIN_FILENAME = "main.tf"
+    VARS_FILENAME = "variables.tf"
+    VERSIONS_FILENAME = "versions.tf"
+    OUTPUT_FILENAME = "outputs.tf"
+
+    def __init__(self, module_path: str):
+        self.module_path = module_path
+
+    def has_main(self) -> str:
+        return os.path.isfile(os.path.join(self.module_path, self.MAIN_FILENAME))
+
+    def has_vars(self) -> str:
+        return os.path.isfile(os.path.join(self.module_path, self.VARS_FILENAME))
+
+    def has_outputs(self) -> str:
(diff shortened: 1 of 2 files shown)


tools/label-check.py
```diff
@@ -0,0 +1,182 @@
...
+    MAIN_FILENAME = "main.tf"
+    VARS_FILENAME = "variables.tf"
+    VERSIONS_FILENAME = "versions.tf"
+    OUTPUT_FILENAME = "outputs.tf"
+
+    def __init__(self, module_path: str):
+        self.module_path = module_path
+
+    def has_main(self) -> str:
+        return os.path.isfile(os.path.join(self.module_path, self.MAIN_FILENAME))
```
mr0re1, line 39: `return os.path.isfile(self.main())`

tools/label-check.py
```diff
@@ -0,0 +1,182 @@
...
+locals {{
+  # This label allows for billing report tracking based on module.
+  labels = merge(var.labels, {{ ghpc_module = \"{module_label}\" }})
+}}
+'''
+
+
+class ModulePath:
+    """Convenience class to get various path related information about a module"""
+    MAIN_FILENAME = "main.tf"
```
mr0re1, line 30: No need for constance if `MAIN` is only used in the function `main`.

tools/label-check.py
```diff
@@ -0,0 +1,182 @@
...
+        return os.path.basename(self.module_path)
+
+    def name_label(self) -> str:
+        return self.name().lower()
+
+    def _filepath(self, name: str) -> str:
+        return os.path.join(self.module_path, name)
+
+
+def get_module_paths(root_dir="./") -> list[str]:
```
mr0re1, line 73: nit. What type check did you use? IIUC, some (namely Pyre) doesn't like `list` and requires `typing.List` used instead.

mr0re1, line 73: nit. This function name begs to return `List[ModulePath]` instead of `str`

tools/label-check.py
```diff
@@ -0,0 +1,182 @@
...
+    This function will check that every module with a var.labels is merging in a
+    `ghpc_module` label. If missing, the locals block will be added. It will
+    check that all other references to labels points to the merged local.labels.
+    
+    This function also checks that the provider meta name matches the module
+    name
+    
+    Returns: True if checks passed
+    """
+
```
mr0re1, line 172: `return all(check_module, get_modules_paths())`

tools/label-check.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            print('{} provider meta does not match module name'.format(
+                version_file_path))
+            return False
+
+
+def check_module(module_path: ModulePath) -> bool:
+    passed = True
+    if not has_labels_variable(module_path):
+        return passed
```
mr0re1, line 152: `return False`

tools/label-check.py
nick-stroud: 1. I had not enabled type checking. Since enabled and found some additional misses.
2. Using pylance which does not seem to care about list vs List
3. Updated to use List anyway as it is correct

nick-stroud: Done

tools/label-check.py
nick-stroud: I see, but realized I was short circuiting the check for provider meta if labels were not in the module. Logic was updated.

tools/label-check.py
nick-stroud: Done.

mr0re1 (approved):

PR #1347, opened 2023-05-17
Add `config.Path` to point to a particular part of Blueprint

mr0re1:
To be used in error messages:
* To lookup Line&Column info;
* As is if no YAML file were used.


mr0re1:
Close, due to other version landed.

PR #1348, opened 2023-05-18
Add example for the lustre file system.

rohitramu:
This PR adds an example of how to use the pre-existing-network-storage module to create a lustre file system.

### DIFF
--- a/modules/file-system/pre-existing-network-storage/README.md
+++ b/modules/file-system/pre-existing-network-storage/README.md
@@ -54,11 +54,11 @@ filesystem:
   settings:
     fs_type: lustre
     server_ip: 192.168.227.11@tcp
-    local_mount: /data
-    remote_mount: /lustre/mgt
+    local_mount: /scratch
+    remote_mount: /exacloud
 ```
 
-Note the use of the MGS NID (Network ID) in the `server_ip` field.
+Note the use of the MGS NID (Network ID) in the `server_ip` field - in particular, note the `@tcp` suffix.
 
 ### Mounting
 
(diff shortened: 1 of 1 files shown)


modules/file-system/pre-existing-network-storage/README.md
```diff
@@ -45,6 +45,21 @@ were directories. This is important when working with files that were created by
...
+The following is an example of using `pre-existing-network-storage` with the `lustre`
+filesystem:
+
+```yaml
+- id: lustrefs
+  source: modules/file-system/pre-existing-network-storage
+  settings:
+    fs_type: lustre
+    server_ip: 192.168.227.11@tcp
+    local_mount: /data
```
nick-stroud, line 57: nit: Generally we tend to use lustre as a scratch filesystem. It might be a better example to use `/scratch` as a local mount.

modules/file-system/pre-existing-network-storage/README.md
```diff
@@ -45,6 +45,21 @@ were directories. This is important when working with files that were created by
...
+    remote_mount: /lustre/mgt
```
nick-stroud, line 58: Where does this remote mount come from? I believe I have had to use `remote_mount: /exacloud`. Maybe `/lustre/mgt` works also but it is unfamiliar to me.

modules/file-system/pre-existing-network-storage/README.md
```diff
@@ -45,6 +45,21 @@ were directories. This is important when working with files that were created by
...
+```
+
+Note the use of the MGS NID (Network ID) in the `server_ip` field.
```
nick-stroud, line 61: nit: I assume this is referring to the `@tcp` (I am just unfamiliar with the `MSG NID` terminology). Maybe just go one step farther to make this clear:

_... in the `server_ip` field, in this case `@tcp`._

modules/file-system/pre-existing-network-storage/README.md
rohitramu: Thanks for the tip, I've updated it to use `/scratch` for the `local_mount` variable.

nick-stroud (approved):

PR #1352, opened 2023-05-18
Simplify adoption of Spack build caches in Google Cloud Storage

tpdownes:
- do not install google-cloud-storage in the GHPC virtual environment since it is not a necessary dependency of GHPC startup scripts
- explicitly configure Spack to use a dedicated Python interpreter which has google-cloud-storage available to it (no extra steps for users)

This has been tested on `community/examples/ramble.yaml` and executing, for example:

```shell
spack mirror add gcs gs://BUCKET_WITH_BUILDCACHE
spack buildcache list
```

One can intentionally break this by executing

```shell
$ unset SPACK_PYTHON
$ spack buildcache list
==> Error: No module named 'google', google-cloud-storage python module is missing. Please install to use the gs:// backend.
```

This PR anticipates future work need to install a supported copy of Ansible on CentOS 7 and Rocky Linux 8 platforms. This is blocked by Python 3.6 (EOL and dropped by Ansible). Most solutions to that problem require an installation of Python 3.8+ to a non-standard location. This approach allows us to flexibly move past old Python releases without effectively hard-coding to /usr/bin/python3 and being stuck on an unsupported copy of Ansible.

See [RHEL8 list of Python3 packages](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_introduction-to-python_configuring-basic-system-settings) for a sense of how we can move past the base "platform" installation of Python 3.6 through the creation of virtual environments that specifically identify the desired Python interpreter.

### Submission Checklist


douglasjacobsen (approved):

community/modules/scripts/spack-install/scripts/install_spack_deps.yml
```diff
@@ -14,25 +14,30 @@
...
-      manager: auto
-  - name: Install protobuf for old releases of Python
-    when: ansible_facts.packages["python3"][0].version is version("3.7", "<") and ansible_facts.packages["python3"][0].version is version("3.5", ">=")
+    register: package
+    changed_when: package.changed
+    retries: 5
+    delay: 10
+    until: package is success
+  - name: Create virtualenv for Spack
+    # this declares our minimum support for Python 3.6+ along with ensuring the
```
nick-stroud, line 34: IIUC, the `declaration of 3.6 support` comes from the minimum version of pip? or something else?

Also clarifying, `dependency resolution of real-world problems` means that this version of pip will make sure protobuf is installed if needed?

community/modules/scripts/spack-install/templates/install_spack.tpl
```diff
@@ -155,8 +153,13 @@ echo "$PREFIX Populating defined buildcaches"
 %{endfor ~}
 
 if [ ! -f /etc/profile.d/spack.sh ]; then
-        echo "source ${INSTALL_DIR}/share/spack/setup-env.sh" > /etc/profile.d/spack.sh
-        chmod a+rx /etc/profile.d/spack.sh
+        cat <<EOF > /etc/profile.d/spack.sh
```
nick-stroud, line 156: I am not used to seeing text after the <<EOF and it looks odd to me. How does this render. Does the `> /etc/profile.d/spack.sh` end up at the end? It must for this to work. If it does then close this comment.

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -172,7 +172,19 @@ main() {
...
+	# when Ansible creates virtual environments, it defaults to virtualenv
+	# these steps install virtualenv in the Ansible virtual environment and
+	# ensures that it will be found
+	${venv_python_path} -m pip install virtualenv
+	if [ ! -f /etc/ansible/ansible.cfg ]; then
+		mkdir /etc/ansible
+		cat <<-EOF >/etc/ansible/ansible.cfg
+			[defaults]
+			interpreter_python=${venv_python_path}
+		EOF
```
nick-stroud, line 186: nit: Any chance you could throw the following in here while you are at it and improve the readability of our ansible logging.

```
stdout_callback=debug
stderr_callback=debug
```

nick-stroud (approved):

community/modules/scripts/spack-install/templates/install_spack.tpl
douglasjacobsen: This cats the block from until EOF into the file `/etc/profile.d/spack.sh`.

community/modules/scripts/spack-install/scripts/install_spack_deps.yml
tpdownes: I've modified this comment for a bit more direct clarity:

- We want to support Python 3.6 because it's easy to install on common distros of Linux used in HPC (RHEL 7/8 and derivatives)
- pip 21.3.1 is the maximum pip supported by Python 3.6
- The biggest "real world" problem with pip is that the version that comes with RHEL7 (not sure about 8) is so ancient it doesn't even properly recognize package limits on Python version. So it will blindly install the most recent copy of google-cloud-storage which doesn't support Python 3.6
- The second-biggest real world problem is that before 20.3, pip would gladly break environments (including the system-wide environment), to satisfy a `pip install` command (yes, really it would do this). Now it will just fail. So, it's almost always better to create a virtual environment with minimally-scoped dependencies
- The third-biggest real world problem is that, for several days, all GCP python packages were broken when the Python release of protobuf version bumped with a breaking change to 4.21.0. TL;DR: packages should have depended on `~ 3.x` but instead had no upper bound. This unblocks `pip` to safely install the latest compatible release (3.20.3).

community/modules/scripts/spack-install/templates/install_spack.tpl
tpdownes: `<< .. >` is just combined input heredoc-style redirection with output redirection you're probably familiar with

I would use `<<-` but it requires the use of tabs and I don't want to mix whitespace styles in this file.

PR #1353, opened 2023-05-18
Eliminate need to activate virtual environment to run Ansible

tpdownes:
Combined with #1352, the impact of this PR is that the Ansible installation script will:

- install Python3 at the system level
- ensures that the copy of pip in system and in GHPC virtual environment is, at least, 21.3.1 (maximum allowed by Python 3.6)
- by default, when Ansible installs a pip package, it will use pip **from** the GHPC virtual environment, but the package will be installed system-wide (unless actively configured to install into a virtualenv)

Consequently, Ansible management of pip packages is predictable and intuitive.

If one uses the `community/examples/ramble.yaml` example, one can `pip list` a long list of packages installed for Ramble at the system level.

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -15,7 +15,7 @@
 
 REQ_ANSIBLE_VERSION=2.11
 REQ_ANSIBLE_PIP_VERSION=4.10.0
-REQ_PIP_MINOR_VERSION=18
+REQ_PIP_MAJOR_VERSION=21
 REQ_PYTHON3_VERSION=6
 
 apt_wait() {
@@ -165,7 +165,7 @@ main() {
 	fi
 	pip_version=$(${python_path} -m pip --version | sed -nr 's/^pip ([0-9]+\.[0-9]+).*$/\1/p')
 	pip_major_version=$(echo "${pip_version}" | cut -d '.' -f 1)
-	if [ "${pip_major_version}" -lt "${REQ_PIP_MINOR_VERSION}" ]; then
+	if [ "${pip_major_version}" -lt "${REQ_PIP_MAJOR_VERSION}" ]; then
 		${python_path} -m pip install --upgrade pip
 	fi
 
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -47,10 +47,10 @@ install_python_deps() {
 # checking python. Sets the variable to an empty string if neither are found.
 get_python_path() {
 	python_path=""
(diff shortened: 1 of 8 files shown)


modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -15,7 +15,7 @@
 
 REQ_ANSIBLE_VERSION=2.11
 REQ_ANSIBLE_PIP_VERSION=4.10.0
-REQ_PIP_MINOR_VERSION=18
+REQ_PIP_MAJOR_VERSION=21
```
cboneti, line 18: 18.1 was required for Debian 10 support. I am not sure this will be satisfiable in Debian 10 now. See [go/hpc-toolkit-dependency-install-design](http://goto.google.com/hpc-toolkit-dependency-install-design)

cboneti (commented):
Please address the concern about Debian 10 and send this back to me.

modules/scripts/startup-script/files/install_ansible.sh
tpdownes: Debian 10 comes with pip 18.1 in its standard repositories. So I believe the script was written in mind with "that's probably good enough and we don't need to run pip install --upgrade".

It's perfectly easy to upgrade to `pip >=21.3.1` on Debian 10 because it includes Python 3.7. RedHat 7 and 8 impose our hardest constraints on `pip` release because they come with Python 3.6. For that the maximum `pip` is `21.3.1`.

In practice, we _never, ever_ want `pip < 20.3` because [until 20.3 pip didn't fail when a user requests installation of conflicting packages](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020).

So, we could write `> 20.3` and probably be OK. But we might as well write the maximum version allowed by Python 3.6.

tpdownes: In these release notes, you will also find why running `pip install` as root is sufficiently bad that the command itself tells you not to do it while you are doing it:

> This also means that, when you run a pip install command, pip only considers the packages you are installing in that command, and may break already-installed packages. It will not guarantee that your environment will be consistent all the time. If you pip install x and then pip install y, it’s possible that the version of y you get will be different than it would be if you had run pip install x y in a single command. We are considering changing this behavior (per [#7744](https://github.com/pypa/pip/issues/7744)) and would like your thoughts on what pip’s behavior should be; please answer [our survey on upgrades that create conflicts](https://docs.google.com/forms/d/e/1FAIpQLSeBkbhuIlSofXqCyhi3kGkLmtrpPOEBwr6iJA6SzHdxWKfqdA/viewform).

cboneti (approved):

PR #1363, opened 2023-05-22
Remove outputs from `file-system` mods that dup `network_storage`

mr0re1:


cboneti:
I would like Nick to review this as he was the last person to take a closer look at the storage modules design.

mr0re1:
All integration tests (except PR-validation) have passed before latest push

modules/scripts/startup-script/README.md
```diff
@@ -138,7 +138,7 @@ sudo journalctl -u google-startup-scripts.service
   settings:
     runners:
       # Some modules such as filestore have runners as outputs for convenience:
-      - $(homefs.install_nfs_client_runner)
```
nick-stroud, line 141: I am not convinced that trading
```
$(homefs.install_nfs_client_runner)

for

((module.homefs.network_storage.client_install_runner))
```

is better. Is there any way we could use `$(homefs.network_storage.client_install_runner)`. We generally not favored users using the `(())` syntax. Can you add some justification as to why this change is improving the user experience.

docs/network_storage.md
nick-stroud, line 1: The PR description says "Add unit test to ensure that all file-system mod ouptut network_storage."
1. I don't see any diffs including unit tests in this pr.
2. I had thought you already added that change in a different PR I reviewed.

nick-stroud (changes requested):

docs/network_storage.md
mr0re1: Stale description, test where merged in as a separate PR. Updated description.

modules/scripts/startup-script/README.md
mr0re1: Addressed by using newly supported $() functionality.

community/modules/file-system/cloud-storage-bucket/README.md
```diff
@@ -80,8 +80,8 @@ following example:
     source: modules/scripts/startup-script
     settings:
       runners:
-      - $(bucket.client_install_runner)
-      - $(bucket.mount_runner)
+      - $(bucket.network_storage.client_install_runner)
+      - $(bucket.network_storage.mount_runner))
```
nick-stroud, line 84: Typo: extra parens.

community/modules/file-system/nfs-server/main.tf
```diff
@@ -43,11 +43,6 @@ locals {
       "destination" = "mount${replace(mount, "/", "_")}.sh"
     }
   ]
-  ansible_mount_runner = {
-    "type"        = "ansible-local"
-    "source"      = "${path.module}/scripts/mount.yaml"
```
nick-stroud, line 48: If this is the only place it is used we should also delete the source file.

community/modules/file-system/nfs-server/README.md
nick-stroud, line 1: I think this change is healthy but will probably break some users. I think we need to do more to help ease that pain. My recommendations:

- Update title to make deprecation clear and call out deprecated variables. Something like: 
  > Deprecated outputs: `client_install_runner`, `mount_runner` & related outputs removed in favor of `network_storage`
- Clearly document in description how to migrate from the old functionality to replacement functionality. This way if someone finds this PR when looking for why variables went missing they will have a clear path forward.
- Add a file to the repo named `docs/deprecations.md` and add a section in there describing the deprecation and how to migrate to the new format (can be same as the description). We can point to this documentation from release notes.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/blueprints/lustre-with-new-vpc.yaml
```diff
@@ -49,8 +49,8 @@ deployment_groups:
     source: modules/scripts/startup-script
     settings:
       runners:
-      - $(scratchfs.install_ddn_lustre_client_runner)
-      - $(scratchfs.mount_runner)
+      - $(scratchfs.network_storage.client_install_runner)
```
nick-stroud, line 52: You can get rid of the use of startup script completely here and just use the file system downstream. They were kept here to test the runners that you are deleting. See #1429 for example.

nick-stroud (changes requested):

PR #1365, opened 2023-05-22
Remove `modulereader.ModuleFS` use `sourcereader.ModuleFS` instead.

mr0re1:
Motivation: reduce duplication in initialization logic.

### DIFF
--- a/pkg/modulereader/hcl_utils.go
+++ b/pkg/modulereader/hcl_utils.go
@@ -36,10 +36,11 @@ func getHCLInfo(source string) (ModuleInfo, error) {
 	ret := ModuleInfo{}
 
 	if sourcereader.IsEmbeddedPath(source) {
-		if !tfconfig.IsModuleDirOnFilesystem(tfconfig.WrapFS(ModuleFS), source) {
+		wrapFS := tfconfig.WrapFS(sourcereader.ModuleFS)
+		if !tfconfig.IsModuleDirOnFilesystem(wrapFS, source) {
 			return ret, fmt.Errorf("Source is not a terraform or packer module: %s", source)
 		}
-		module, _ = tfconfig.LoadModuleFromFilesystem(tfconfig.WrapFS(ModuleFS), source)
+		module, _ = tfconfig.LoadModuleFromFilesystem(wrapFS, source)
 	} else {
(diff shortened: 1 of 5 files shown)


pkg/modulereader/hcl_utils.go
```diff
@@ -36,10 +36,11 @@ func getHCLInfo(source string) (ModuleInfo, error) {
 	ret := ModuleInfo{}
 
 	if sourcereader.IsEmbeddedPath(source) {
-		if !tfconfig.IsModuleDirOnFilesystem(tfconfig.WrapFS(ModuleFS), source) {
+		wrapFS := tfconfig.WrapFS(sourcereader.ModuleFS)
```
tpdownes, line 39: It appears to me that this has been refactored to avoid introducing the problem from https://github.com/hashicorp/terraform-config-inspect/issues/68. If action is required, please take it.

pkg/sourcereader/embedded.go
```diff
@@ -28,9 +28,10 @@ import (
 // hpc-toolkit/modules are not accessible at the package level.
 var ModuleFS BaseFS
 
-// BaseFS is an extension of the io.fs interface with the functionality needed
+// BaseFS is an extension of the fs.FS interface with the functionality needed
 // in CopyDirFromModules. Works with embed.FS and afero.FS
 type BaseFS interface {
+	fs.FS
```
tpdownes, line 34: This looks like the only "real" change in this PR. The old comment appears to be wrong in 2 ways:

- it confuses `io/fs` and `io.fs`
- it doesn't appear to extend (embed) any interface

The use of `fs BaseFS` in function signatures is potentially confusing due to overlap with `io/fs`.  I suggest we fix this now by picking a different name for the variable. `basefs`, `bfs`?

tpdownes (approved):
Please fix the confusing types for `fs` / `fs` in `embedded.go`.  I realize Go has scoping which avoid direct conflict, but it may be confusing for developers.

Please also confirm that you are not re-introducing the issue in `hcl_utils.go`.

pkg/modulereader/hcl_utils.go
mr0re1: That doesn't seem to be an issue `WrapFS` does solve this issue.

PR #1366, opened 2023-05-22
Resolve destroy/create race in vm-instance

tpdownes:
If one changes a VM parameter that requires destruction and replacement of the VM (e.g. `local_ssd_count: 0` &rarr; `1`) but does not change the zone of the VM, then there is a race condition which causes the disk to be deleted along with the VM (when `var.auto_delete_boot_disk` is its default value of `true`) but not re-created because the explicit Terraform resource was not deleted. Thus the new VM cannot be created until a 2nd Terraform apply refreshes state with the now-deleted disk.

Resolve by using the initialize_params block instead of an explicit google_compute_disk resource. Preserves behavior that setting `var.auto_delete_boot_disk=false` allows a disk to be re-used when the VM is destroyed and replaced (e.g. changing `local_ssd_count`). Additionally, this PR preserves the behavior of labeling disks at creation for cost-tracking.

Example blueprint:

```yaml
---
blueprint_name: simple

vars:
  deployment_name: simple
  region: us-central1
  zone: us-central1-c

deployment_groups:
- group: primary
  modules:
  - id: network1
  - id: simplevm
    use:
    - network1
    settings:
      auto_delete_boot_disk: false
      machine_type: c2-standard-4
      local_ssd_count: 0
```

After this PR, there is a new edge case. If one simultaneously changes `local_ssd_count:0` &rarr; `1` AND changes `zone: us-central1-c` &rarr; `-b`, then a re-apply succeeds although a new disk is created in `-b` and the disk in `-c` is orphaned. I consider this behavior acceptable given that the user is opting into a behavior which suggests retaining the disk (and cost). The user should also not expect a zonal disk to follow the VM to a new zone.

### Submission Checklist

### DIFF
--- a/modules/compute/vm-instance/main.tf
+++ b/modules/compute/vm-instance/main.tf
@@ -85,19 +85,6 @@ data "google_compute_image" "compute_image" {
   project = var.instance_image.project
 }
 
-resource "google_compute_disk" "boot_disk" {
-  project = var.project_id
-
-  count = var.instance_count
-
-  name   = "${local.resource_prefix}-boot-disk-${count.index}"
-  image  = data.google_compute_image.compute_image.self_link
-  type   = var.disk_type
-  size   = var.disk_size_gb
-  labels = local.labels
-  zone   = var.zone
-}
-
 resource "google_compute_resource_policy" "placement_policy" {
   project = var.project_id
 
@@ -128,9 +115,13 @@ resource "google_compute_instance" "compute_vm" {
   labels = local.labels
 
   boot_disk {
-    source      = google_compute_disk.boot_disk[count.index].self_link
-    device_name = google_compute_disk.boot_disk[count.index].name
     auto_delete = var.auto_delete_boot_disk
+    initialize_params {
+      size   = var.disk_size_gb
+      type   = var.disk_type
+      image  = data.google_compute_image.compute_image.self_link
+      labels = local.labels
+    }
   }
 
   dynamic "scratch_disk" {
--- a/modules/compute/vm-instance/README.md
+++ b/modules/compute/vm-instance/README.md
@@ -165,7 +165,6 @@ limitations under the License.
 | Name | Type |
 |------|------|
 | [google-beta_google_compute_instance.compute_vm](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_compute_instance) | resource |
-| [google_compute_disk.boot_disk](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk) | resource |
 | [google_compute_resource_policy.placement_policy](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_resource_policy) | resource |
 | [google_compute_image.compute_image](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_image) | data source |
 


modules/compute/vm-instance/main.tf
```diff
@@ -128,9 +115,13 @@ resource "google_compute_instance" "compute_vm" {
   labels = local.labels
 
   boot_disk {
-    source      = google_compute_disk.boot_disk[count.index].self_link
-    device_name = google_compute_disk.boot_disk[count.index].name
```
mr0re1, line 132: Shall we keep it `"${local.resource_prefix}-boot-disk-${count.index}"` ?

tpdownes: - newly created VMs will have disks with the same name as the VM (similar, but not identical)
- if one deploys the blueprint above using develop and then re-deploys off this PR, it will fail to destroy the attached disk:

  ```text
  Display full proposed changes, Apply proposed changes, Stop and exit, Continue without applying? [d,a,s,c]: a
  2023/05/23 10:43:33 running terraform apply on group simple/primary
  module.simplevm.google_compute_disk.boot_disk[0]: Destroying... [id=projects/toolkit-demo-zero-e913/zones/us-central1-b/disks/simple-boot-disk-0]

  Error: Error detaching disk simple-boot-disk-0 from instance toolkit-demo-zero-e913/us-central1-b/simple-0: googleapi: Error 400: Invalid resource usage: 'To detach the boot disk, the instance must be in TERMINATED state.'., invalidResourceUsage
  ```

So, it will fail but in a way that does not destroy existing infrastructure. [moved] blocks cannot address this because we are intentionally no longer managing a separate `google_compute_disk` resource. My opinion is that separate management of the disk is sufficiently counterproductive that removing it is worth it and that we should advise users to run:

```shell
❯❯❯ terraform state rm module.simplevm.google_compute_disk.boot_disk
Removed module.simplevm.google_compute_disk.boot_disk[0]
Successfully removed 1 resource instance(s).
❯❯❯ terraform plan
module.simplevm.module.netstorage_startup_script.random_id.resource_name_suffix: Refreshing state... [id=1SOGGg]
module.simplevm.data.google_compute_image.compute_image: Reading...
module.network1.data.google_compute_subnetwork.primary_subnetwork: Reading...
module.network1.data.google_compute_network.vpc: Reading...
module.simplevm.module.netstorage_startup_script.google_storage_bucket.configs_bucket[0]: Refreshing state... [id=simple-startup-scripts-d523861a]
module.simplevm.module.netstorage_startup_script.google_storage_bucket_object.scripts["passed_startup_script.sh"]: Refreshing state... [id=simple-startup-scripts-d523861a-passed_startup_script.sh-5617]
module.network1.data.google_compute_subnetwork.primary_subnetwork: Read complete after 0s [id=projects/toolkit-demo-zero-e913/regions/us-central1/subnetworks/default]
module.network1.data.google_compute_network.vpc: Read complete after 0s [id=projects/toolkit-demo-zero-e913/global/networks/default]
module.simplevm.data.google_compute_image.compute_image: Read complete after 1s [id=projects/cloud-hpc-image-public/global/images/hpc-centos-7-v20230419]
module.simplevm.google_compute_instance.compute_vm[0]: Refreshing state... [id=projects/toolkit-demo-zero-e913/zones/us-central1-b/instances/simple-0]

No changes. Your infrastructure matches the configuration.
```

In this case, the `google_compute_instance` is "unharmed" because it detects the boot disk attached to it already (even if it has the "wrong" name) and does not see a need to re-create the resource.

I'll investigate a bit to see if there is anything more automatic that might help us here (e.g. precondition/postcondition).

[moved]: https://developer.hashicorp.com/terraform/language/modules/develop/refactoring

modules/compute/vm-instance/main.tf
nick-stroud, line 1: Just to clarify, you say:

> If one simultaneously changes `local_ssd_count:0` &rarr; `1` AND changes `zone: us-central1-c` &rarr; `-b`, then a re-apply succeeds although a new disk is created in `-b` and the disk in `-c` is orphaned.

This is only the case when `auto_delete_boot_disk` is explicitly set to false. 

I think this is fine.

modules/compute/vm-instance/main.tf
nick-stroud: @tpdownes, under what conditions does the user get the error you describe: `Error detaching disk simple-boot-disk-0 from instance`? 

Does this happen even if I don't change anything on the vm-instance settings?  
Does this happen whether or not `auto_delete_boot_disk` is set to `true` or `false`?

I suspect the answer is yes to both.

tpdownes: `yes` to both on HEAD of PR.

modules/compute/vm-instance/main.tf
tpdownes: Yes, this only occurs under those conditions.

tpdownes:
Closing PR due to discovery that both of the following options trigger user experience problems

- the proposed PR (must manually remove google_compute_disk from state)
- setting `var.auto_delete_boot_disk` to false on a VM where it has previously been set to true (destroy/create cycle)

The 2nd bullet is a bug that has been filed at https://github.com/hashicorp/terraform-provider-google/issues/14707

PR #1373, opened 2023-05-23
Add test that all `file-system` mods output `network_storage`

mr0re1:


pkg/inspect/modules_test.go
```diff
@@ -134,4 +159,8 @@ func TestNetworkStorage(t *testing.T) {
 	for _, mod := range notEmpty(query(hasInput("network_storage")), t) {
 		checkInputType(t, mod, "network_storage", expected)
 	}
+
+	for _, mod := range query(all(ofRole("file-system"), not(hasOutput("network_storage")))) {
+		t.Errorf("%q does not output 'network_storage'", mod.Source)
```
nick-stroud, line 164: Mostly just for context: Obviously this is not checking DAOS (a fully remote module) as DAOS does not have a network_storage output. Lack of automation around DAOS is documented [here](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/develop/docs/network_storage.md). Feel free to close if this is expected.

nick-stroud (approved):

PR #1375, opened 2023-05-24
Update ramble-setup module to be idempotent

douglasjacobsen:
This merge updates the ramble-setup module to be idempotent. It also adds a more complicated example of using ramble with a slurm cluster.

### Submission Checklist


community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -36,32 +22,38 @@
     chown_owner: ${chown_owner}
     chgrp_group: ${chgrp_group}
   tasks:
+  - name: Acquire lock
+    ansible.builtin.shell:
+      mkdir -p "{{ install_dir | dirname }}/.ramble_lock"
```
nick-stroud, line 27: Install_dir is already a dir. Have you tested behavior of dirname if passed a dir. My worry is that if I set `/apps/my/ramble` it will remove the `ramble` where as mkdir would tolerate this.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -36,32 +22,38 @@
     chown_owner: ${chown_owner}
     chgrp_group: ${chgrp_group}
   tasks:
+  - name: Acquire lock
+    ansible.builtin.shell:
```
nick-stroud, line 26: If now shell functionality is used then the `command` module is prefered to `shell`. Ansible lint will error on this although it is temporarily disabled.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
nick-stroud, line 27: I believe the `-p` will basically never fail. My recommendation would be to have a first task that creates install_dir using a -p, then a second task that creates .ramble_lock without the -p and use this second task to capture the return behavior with the lock.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -36,32 +22,38 @@
     chown_owner: ${chown_owner}
     chgrp_group: ${chgrp_group}
   tasks:
+  - name: Acquire lock
```
nick-stroud, line 25: I believe this task will have to have some `failed_when` behavior defined otherwise failure will kill the setup instead of just skipping the tasks. Something like `failed_when: false`.

community/examples/hpc-slurm-ramble-gromacs.yaml
nick-stroud, line 1: Add documentation to /examples

nick-stroud, line 1: nit: simple ramble example is not listed in toc in /examples

nick-stroud (changes requested):

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
douglasjacobsen: Per our discussion, I left the dirname filter.

community/examples/hpc-slurm-ramble-gromacs.yaml
douglasjacobsen: After discussing it with @cboneti I just removed the simple example.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -22,16 +22,24 @@
     chown_owner: ${chown_owner}
     chgrp_group: ${chgrp_group}
   tasks:
+  - name: Create install directory
```
nick-stroud, line 25: Using `ansible.builtin.file` would be more correct as it would accurately reflect the changed property if the file system was modified.

```
- name: Create a directory if it does not exist
  ansible.builtin.file:
    path: /etc/some_directory
    state: directory
```

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -22,16 +22,24 @@
...
     register: lock_out
+    changed_when: lock_out.rc == 0
+    failed_when: false
 
   - name: Clones ramble into installation directory
     ansible.builtin.git:
       repo: "{{ ramble_url }}"
       dest: "{{ install_dir }}"
       version: "{{ ramble_ref }}"
+      force: true
```
nick-stroud, line 42: I had thought you said this was not necessary.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -55,6 +63,12 @@
       recurse: true
     when: chmod_mode != "" and lock_out.rc == 0
 
+  - name: check if ramble profile exists
+    ansible.builtin.stat:
+      path: /etc/profile.d/ramble.sh
+    register: profile_check
+
   - name: Add ramble to profile
```
nick-stroud, line 71: I think the correct way to do this would by with:

```
  - name: Creating a file with content
    copy:
      dest: "/etc/profile.d/ramble.sh"
      content: ". {{ install_dir }}/share/ramble/setup-env.sh"
```

I think you could leave the stat check in place.

community/examples/hpc-slurm-ramble-gromacs.yaml
nick-stroud: SG but it looks to me like the simple example was not removed (the actual yaml).

You could also consider moving it to validate_configs if it would be useful to keep a super simple example around for testing purposes. Your call.

douglasjacobsen: I moved it for now. :)

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
douglasjacobsen: Good catch.

douglasjacobsen:
OK, everything should be resolved. I rebased it because it said (oddly) there was a merge conflict in one of template files, but that seems unlikey...

nick-stroud (approved):

PR #1377, opened 2023-05-26
Add support for indexing to "simple blueprint expressions"

mr0re1:
* Enable usage such `$(instance.name[0])`;
* Replace `((...))` usages with `$(...)`.

### DIFF
--- a/community/examples/client-google-cloud-storage.yaml
+++ b/community/examples/client-google-cloud-storage.yaml
@@ -61,4 +61,4 @@ deployment_groups:
   - id: wait
     source: community/modules/scripts/wait-for-startup
     settings:
-      instance_name: ((module.workstation.name[0]))
+      instance_name: $(workstation.name[0])
--- a/community/examples/omnia-cluster.yaml
+++ b/community/examples/omnia-cluster.yaml
@@ -92,5 +92,5 @@ deployment_groups:
   - id: wait
     source: community/modules/scripts/wait-for-startup
     settings:
-      instance_name: ((module.manager.name[0]))
+      instance_name: $(manager.name[0])
       timeout: 2100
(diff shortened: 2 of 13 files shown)


nick-stroud:
I did some testing with the snippet below. Everything is as I would expect. 

My one area of concern is the last line: `$(homefs.network_storage.nothing)`. In ideal world this would fail on create and inform the user they are making a bad reference. I assume that this is impossible to enforce as outputs do not have typing. 

My concern is as follows:  
Before this change we can validate all uses of $() references on create (except not for typing I guess but at least that they exist). Partially we do this by greatly limiting the allowed usage of $(). After this change, there are more cases where this will fail on terraform apply.

Personally my vote would be to accept this PR and extend the functionality with the risk is worth being about to make `$(compute_1.name[0])` references but I don't think we should unilaterally decide this core functionality of the blueprint syntax. 

@cboneti can you weight in?

```
  - id: homefs
    source: modules/file-system/pre-existing-network-storage
    settings:
      server_ip: 10.0.0.0
      remote_mount: nfsshare
      local_mount: /home
      fs_type: nfs
  
  - id: startup
    settings:
      runners:
        - $(homefs.client_install_runner) # work before and after

        - $(homefs.nothing) # fails before and after

        - $(homefs.network_storage.client_install_runner) # fails before works after

        - type: shell # fails before works after
          content: $(homefs.network_storage.client_install_runner.content) 
          destination: $(homefs.network_storage.client_install_runner.destination)

        - $(homefs.network_storage.nothing) # fails before but passes after (although I would prefer if it failed)
```

cboneti:
> @cboneti can you weight in?

I see the added risk here, as before, we had a "closed" environment (everything was guaranteed to be testable), and now, we do not.  Still, it seems to me that the benefits outweigh the additional risk.

cboneti (approved):

PR #1378, opened 2023-05-26
add config-ssh as a startup script option

cboneti:
With this change, a user could specify `configure_ssh_host_prefix` on a startup script and that would configure SSH for users in the deployed system. Example: 

```
  - id: script
    settings:
      configure_ssh_host_patterns: ['10.182.0.*', 'clustername*']
``` 

It will automate ssh configuration by:
- Defining a Host block for every element of this variable and setting StrictHostKeyChecking to 'No' for the specified block.
  Ex: "10.182.0.*", "hpc*", "hpc01*", "ml*"
- The first time users log-in, it will create ssh keys that are added to the authorized keys list
  This requires a shared /home filesystem and relies on specifying the right prefix.
  
### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/configure-ssh.yml
+++ b/modules/scripts/startup-script/files/configure-ssh.yml
@@ -1,0 +1,35 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+- name: Configure ssh between nodes
+  become: true
+  hosts: localhost
+  vars:
+    ssh_config_path: "/etc/ssh/ssh_config"
+    bashrc: "{{ '/etc/bashrc' if ansible_facts['os_family'] == 'RedHat' else '/etc/bash.bashrc' }}"
+    setup_ssh_script: "/bin/bash /usr/local/ghpc/setup-ssh-keys.sh"
+  tasks:
+  - name: "Set StrictHostKeyChecking to no"
+    ansible.builtin.blockinfile:
+      path: "{{ ssh_config_path }}"
+      block: |
+        Host {{ host_name_prefix }}*
+          StrictHostKeyChecking no
+  - name: "Create ssh keys in .bashrc if not alrady done"
+    ansible.builtin.lineinfile:
+      path: "{{ bashrc }}"
+      regexp: '^{{ setup_ssh_script }}'
+      line: "{{ setup_ssh_script }}"
--- a/modules/scripts/startup-script/files/configure-ssh.yml
+++ b/modules/scripts/startup-script/files/configure-ssh.yml
@@ -26,8 +26,10 @@
     ansible.builtin.blockinfile:
       path: "{{ ssh_config_path }}"
       block: |
-        Host {{ host_name_prefix }}*
+        Host "{{ item }}"
           StrictHostKeyChecking no
+      marker: "# {mark} ANSIBLE MANAGED BLOCK {{item}}"
+    loop: "{{ host_name_prefix }}"
   - name: "Create ssh keys in .bashrc if not alrady done"
     ansible.builtin.lineinfile:
       path: "{{ bashrc }}"
--- a/modules/scripts/startup-script/main.tf
+++ b/modules/scripts/startup-script/main.tf
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


nick-stroud (changes requested):

modules/scripts/startup-script/variables.tf
```diff
@@ -112,6 +112,17 @@ variable "install_ansible" {
   default     = null
 }
 
+variable "configure_ssh" {
```
nick-stroud, line 115: Batch calls this `permissiveSsh`. I would suggest `configure_permissive_ssh` or `enable_permissive_ssh` for this var name.

modules/scripts/startup-script/files/configure-ssh.yml
```diff
@@ -0,0 +1,35 @@
...
+  vars:
+    ssh_config_path: "/etc/ssh/ssh_config"
+    bashrc: "{{ '/etc/bashrc' if ansible_facts['os_family'] == 'RedHat' else '/etc/bash.bashrc' }}"
+    setup_ssh_script: "/bin/bash /usr/local/ghpc/setup-ssh-keys.sh"
+  tasks:
+  - name: "Set StrictHostKeyChecking to no"
+    ansible.builtin.blockinfile:
+      path: "{{ ssh_config_path }}"
+      block: |
+        Host {{ host_name_prefix }}*
```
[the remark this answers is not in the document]
cboneti: I have now addressed this in the following way:
  - id: script
    settings:
      configure_ssh_host_prefix: ['10.182.0.*', 'clustername*']

modules/scripts/startup-script/variables.tf
cboneti: It is not quite permissive, I think we are not introducing any security issue here. So it is really more about configuring ssh.

modules/scripts/startup-script/main.tf
```diff
@@ -26,11 +26,15 @@ locals {
...
   }] : []
 
-  ssh_args = join("", [
-    "-e host_name_prefix=${var.deployment_name}"
-  ])
+  configure_ssh = length(var.configure_ssh_host_prefix) > 0
+  host_args = {
+    host_name_prefix = var.configure_ssh_host_prefix
+  }
+  #  host_args = "{host_name_prefix:${jsonencode(var.configure_ssh_host_prefix)}}"
```
nick-stroud, line 33: I suspect you meant to delete this.

modules/scripts/startup-script/main.tf
```diff
@@ -26,11 +26,15 @@ locals {
...
+
+  ssh_args = "-e  @/tmp/prefix_file.json"
```
nick-stroud, line 35: nit: This local is not really earning its keep. You could make the local just `/tmp/prefix_file.json` and then use it on line 51 and 57 to avoid specifying the path twice.

modules/scripts/startup-script/files/configure-ssh.yml
```diff
@@ -22,12 +22,16 @@
     bashrc: "{{ '/etc/bashrc' if ansible_facts['os_family'] == 'RedHat' else '/etc/bash.bashrc' }}"
     setup_ssh_script: "/bin/bash /usr/local/ghpc/setup-ssh-keys.sh"
   tasks:
+  - debug:
```
nick-stroud, line 25: nit: Ansible lint would complain here that 1. the task does not have a name and 2. you are not using the fully qualified task (`ansible.builtin.debug`). I think a name would be helpful if you are intending to leave this here as it will print in the logs.

modules/scripts/startup-script/variables.tf
```diff
@@ -112,15 +112,16 @@ variable "install_ansible" {
   default     = null
 }
 
-variable "configure_ssh" {
+variable "configure_ssh_host_prefix" {
```
nick-stroud, line 115: Prefix seems like incorrect name now that you are not adding anything to the end. Maybe `configure_ssh_host_patterns`.

nick-stroud (changes requested):

modules/scripts/startup-script/main.tf
cboneti: good catch, thanks!

modules/scripts/startup-script/files/configure-ssh.yml
cboneti: I removed it. Thanks

modules/scripts/startup-script/variables.tf
cboneti: Yes, this is much better indeed.

modules/scripts/startup-script/main.tf
cboneti: good point, I restructured the code.

nick-stroud (approved):

tools/validate_configs/test_configs/config-ssh.yaml
```diff
@@ -0,0 +1,99 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+blueprint_name: hpc-slurm
+
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: hpc-small
```
nick-stroud, line 21: nit: deployment name and blueprint name are not in line with purpose of file.

tools/validate_configs/test_configs/config-ssh.yaml
```diff
@@ -0,0 +1,99 @@
...
+      node_count_dynamic_max: 20
+
+  - id: compute_partition
+    source: community/modules/compute/schedmd-slurm-gcp-v5-partition
+    use:
+    - network1
+    - homefs
+    - compute_node_group
+    settings:
+      partition_name: compute
```
nick-stroud, line 78: nit: no need to have 2 partitions just for the config test of the ssh option. Maybe also delete login.

tools/validate_configs/test_configs/config-ssh.yaml
```diff
@@ -0,0 +1,99 @@
+# Copyright 2022 Google LLC
```
nick-stroud, line 1: nit: 2023

nick-stroud (approved):

tools/validate_configs/test_configs/config-ssh.yaml
cboneti: fixed

tools/validate_configs/test_configs/config-ssh.yaml
cboneti: I kept this purposefully like the hpc-small, as it can serve as an actual useful example. I also kept the deployment_name the same as I set the right cluster name as a pattern. This is to maximize the usefulness of this file.

tools/validate_configs/test_configs/config-ssh.yaml
cboneti: I changed the blueprint name, I left the deployment name because it aligns with hpc-slurm blueprint and so someone could use this blueprint without changes.

PR #1390, opened 2023-05-30
Add validation for deprecated input variables

rohitramu:


modules/network/vpc/variables.tf
```diff
@@ -118,44 +122,52 @@ variable "primary_subnetwork" {
   type        = map(string)
   default     = null
   validation {
-    condition     = var.primary_subnetwork == null || can(var.primary_subnetwork["subnet_name"])
+    condition     = var.primary_subnetwork == null ? true : can(var.primary_subnetwork["subnet_name"])
```
mr0re1, line 125: Why this change? Aren't both of those lines are semantically identical?

modules/network/vpc/variables.tf
```diff
@@ -118,44 +122,52 @@ variable "primary_subnetwork" {
...
     error_message = "Primary subnetwork must define \"subnet_name\"."
   }
   validation {
-    condition     = var.primary_subnetwork == null || can(var.primary_subnetwork["subnet_region"])
```
mr0re1, line 125: I guess all those old validation blocks can be simply removed.

modules/network/vpc/variables.tf
rohitramu: Unfortunately, HCL doesn't use short-circuiting in conditionals: https://github.com/hashicorp/terraform/issues/24128#issuecomment-1155241478

However, it does for ternary statements (i.e. `<condition> ? <true_action> : <false_action>`).

tests.log
```diff
@@ -0,0 +1,111 @@
+**************** vetting go code **********************
```
mr0re1, line 1: Please clean up added files. `tests.log` & `~/Documents`

rohitramu: I have no idea how files outside the repo got added!  Thanks for flagging this, I'll clean it up.

modules/network/vpc/variables.tf
mr0re1: How does lack of short-circuiting changes anything here?
IIUC `can(expr)` will return `false` if `expr` throws an error, so `var.primary_subnetwork == null` shouldn't be a problem here.

rohitramu: That's a good point, I'll revert the changes which involve `can()`.

modules/network/vpc/variables.tf
rohitramu: Done!

modules/network/vpc/variables.tf
```diff
@@ -118,44 +122,24 @@ variable "primary_subnetwork" {
...
-      for s in var.additional_subnetworks : can(s["subnet_region"])
-    ])
-    error_message = "All additional subnetworks must define \"subnet_region\"."
-  }
+  default     = null
   validation {
-    condition     = alltrue([for s in var.additional_subnetworks : can(s["new_bits"])])
-    error_message = "All additional subnetworks must define \"new_bits\"."
+    condition     = var.additional_subnetworks == null
+    error_message = "additional_subnetworks is deprecated."
```
mr0re1, line 136: Add `Please see https://goo.gle/hpc-toolkit-vpc-deprecation for migration instructions.`

mr0re1 (approved):

modules/network/vpc/variables.tf
rohitramu: Good catch - I've added it now.

PR #1391, opened 2023-05-30
Update READMEs about "literal variables"

mr0re1:
Support for indexing was added in #1377

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1142,23 +1142,17 @@ The variable is referred to by the source, either vars for deploment variables
 or the module ID for module variables, followed by the name of the value being
 referenced. The entire variable is then wrapped in “$()”.
 
-Currently, references to variable attributes and string operations with
-variables are not supported.
+Currently, string interpolation with variables is not supported.
 
 ### Literal Variables
 
-Literal variables are not interpreted by `ghpc` directly, but rather embedded in the
-underlying module. Literal variables should only be used by those familiar
-with the underlying module technology (Terraform or Packer); no validation
-will be done before deployment to ensure that they are referencing
-something that exists.
-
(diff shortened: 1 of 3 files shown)


examples/README.md
```diff
@@ -1142,23 +1142,17 @@ The variable is referred to by the source, either vars for deploment variables
...
-Literal variables are not interpreted by `ghpc` directly, but rather embedded in the
-underlying module. Literal variables should only be used by those familiar
-with the underlying module technology (Terraform or Packer); no validation
-will be done before deployment to ensure that they are referencing
-something that exists.
-
-Literal variables are occasionally needed when referring to the data structure
-of the underlying module. For example, to refer to the subnetwork self link from
-a vpc module through terraform itself:
+Literal variables should only be used by those familiar
```
nick-stroud, line 1149: You have removed some content that is still applicable and useful in my opinion such as:
> Literal variables are not interpreted by `ghpc` directly, but rather embedded in the underlying module.

> no validation will be done before deployment to ensure that they are referencing something that exists.

examples/README.md
nick-stroud, line 1: Without context, it is not clear why this change is occuring. Please update the description. I would suggest adding a link to the PR that made the actual change in the description.

nick-stroud (changes requested):

examples/README.md
mr0re1: It's not true anymore, we do interpret them and validate that they are "referencing something that exists"

mr0re1: It's not true anymore, we do interpret them and validate that they are "referencing something that exists"

examples/README.md
mr0re1: PR description was updated

nick-stroud (approved):

PR #1394, opened 2023-05-31
Validate a blueprint's top-level "labels" variable

rohitramu:


cboneti:
PR Validation is failing. PTAL
Also, please assign Tom as a reviewer since he knows this part of the code better.

tpdownes (changes requested):

pkg/config/validate.go
```diff
@@ -148,10 +148,24 @@ func (dc DeploymentConfig) validateVars() error {
 		if !ty.IsObjectType() && !ty.IsMapType() {
 			return errors.New("vars.labels must be a map of strings")
 		}
-		for _, v := range labels.AsValueMap() {
+		for labelName, v := range labels.AsValueMap() {
```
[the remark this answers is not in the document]
rohitramu: That makes sense, thanks for the suggestions!  I've update this PR to address both points.  I didn't think there was any need to use const strings when the `errorMessages` map was already available, but let me know if you think otherwise.

tools/validate_configs/golden_copies/configs/text_escape.yaml
```diff
@@ -20,7 +20,7 @@ vars:
   deployment_name: text_escape
   zone: us-east4-c
   labels:
-    red: \((blue))
+    red: blue
```
tpdownes, line 23: @mr0re1 I think the intent of this test was to ensure that we could escape strings that look like literals. So, "red: blue" isn't helping us much here and we should probably delete it. WDYT?

tpdownes (changes requested):

pkg/config/validate.go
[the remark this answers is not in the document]
tpdownes: We'll eventually move away from the giant `errorMessages` map, but as you can see it's already giant and 2 more isn't a big deal one way or the other.

tools/validate_configs/golden_copies/configs/text_escape.yaml
rohitramu: I've changed the test to have international characters in the label name and label value, to ensure that works as expected.

There might still be some value in keeping the test for escaping chars, because ghpc validation does not currently cover labels passed directly to modules.  Still interested to see what @mr0re1 has to say though.

pkg/config/config_test.go
```diff
@@ -617,6 +620,41 @@ func (s *MySuite) TestImportBlueprint(c *C) {
 		Equals, expectedSimpleBlueprint.DeploymentGroups[0].Modules[0].ID)
 }
 
+func (s *MySuite) TestImportBlueprint_LabelValidation(c *C) {
```
tpdownes, line 623: Must have missed this earlier. I don't actually see a test for success here. Please add one. Also consider a test for edge cases like empty string or a capital "N~".

Approving on assumption that you address this concern.

tpdownes (approved):
Make the change requested and move forward with merging.

tpdownes (approved):
Make the change requested and move forward with merging.

pkg/config/config_test.go
rohitramu: Good idea - done.

PR #1400, opened 2023-06-01
Simplify service-account module

tpdownes:
The service-account module is marked experimental; this PR intends to facilitate increasing its adoption by simplifying its interface.

- make Toolkit-native by using deployment name in resource naming and descriptions
- instead of creating N service accounts with identical project-wide roles, focus on common use case of a single service account
- update hpc-enterprise-slurm example to use minimally-scoped service accounts

### Submission Checklist

### DIFF
--- a/community/modules/project/service-account/outputs.tf
+++ b/community/modules/project/service-account/outputs.tf
@@ -14,47 +14,23 @@
  * limitations under the License.
  */
 
-output "email" {
-  description = "Service account email (for single use)."
-  value       = module.service_accounts.email
-}
-output "emails" {
-  description = "Service account emails by name."
-  value       = module.service_accounts.emails
-}
-output "emails_list" {
-  description = "Service account emails s list."
-  value       = module.service_accounts.emails_list
-}
-output "iam_email" {
-  description = "IAM-format service account email (for single use)."
-  value       = module.service_accounts.iam_email
-}
-output "iam_emails" {
-  description = "IAM-format service account emails by name."
-  value       = module.service_accounts.iam_emails
-}
-output "iam_emails_list" {
-  description = "IAM-format service account emails s list."
-  value       = module.service_accounts.iam_emails_list
-}
 output "key" {
-  description = "Service account key (for single use)."
-  value       = module.service_accounts.key
+  description = "Service account key (if creation was requested)"
+  value       = module.service_account.key
 }
-output "keys" {
-  description = "Map of service account keys."
-  value       = module.service_accounts.keys
+
+output "service_account_email" {
+  description = "Service account e-mail address"
+  value       = module.service_account.email
+  depends_on = [
+    module.service_account,
+  ]
 }
-output "service_account" {
-  description = "Service account resource (for single use)."
-  value       = module.service_accounts.service_account
-}
-output "service_accounts" {
-  description = "Service account resources as list."
-  value       = module.service_accounts.service_accounts
-}
-output "service_accounts_map" {
-  description = "Service account resources by name."
-  value       = module.service_accounts.service_accounts_map
+
+output "service_account_iam_email" {
+  description = "Service account IAM binding format (serviceAccount:name@example.com)"
+  value       = module.service_account.iam_email
+  depends_on = [
+    module.service_account,
+  ]
 }
--- a/community/modules/project/service-account/variables.tf
+++ b/community/modules/project/service-account/variables.tf
@@ -20,26 +20,36 @@ variable "billing_account_id" {
   default     = ""
 }
 
-variable "description" {
-  description = "Default description of the created service accounts (defaults to no description)."
+variable "deployment_name" {
+  description = "Name of the deployment (will be prepended to service account name)"
   type        = string
-  default     = ""
+}
+
+variable "description" {
+  description = "Description of the created service account."
+  type        = string
+  default     = "Service Account"
 }
 
 variable "descriptions" {
-  description = "List of descriptions of the created service accounts (elements default to the value of description)."
+  description = "Deprecated; create single service accounts using var.description."
   type        = list(string)
-  default     = []
+  default     = null
+
+  validation {
+    condition     = var.descriptions == null
+    error_message = "var.descriptions has been deprecated in favor of creating single accounts with var.description"
+  }
 }
 
 variable "display_name" {
-  description = "display names of the created service accounts."
+  description = "Display name of the created service account."
   type        = string
-  default     = ""
+  default     = "Service Account"
 }
 
 variable "generate_keys" {
-  description = "Generate keys for service accounts."
+  description = "Generate keys for service account."
   type        = bool
   default     = false
 }
@@ -56,10 +66,20 @@ variable "grant_xpn_roles" {
   default     = true
 }
 
+variable "name" {
+  description = "Name of the service account to create."
+  type        = string
+}
+
 variable "names" {
-  description = "Names of the services accounts to create."
+  description = "Deprecated; create single service accounts using var.name."
   type        = list(string)
-  default     = []
+  default     = null
+
+  validation {
+    condition     = var.names == null
+    error_message = "var.names has been deprecated in favor of creating single accounts with var.name"
+  }
 }
 
 variable "org_id" {
@@ -69,9 +89,14 @@ variable "org_id" {
 }
 
 variable "prefix" {
-  description = "prefix applied to service account names"
+  description = "Deprecated; prefix now set using var.deployment_name"
   type        = string
-  default     = ""
+  default     = null
+
+  validation {
+    condition     = var.prefix == null
+    error_message = "var.prefix has been deprecated in favor of setting prefix with var.deployment_name"
+  }
 }
 
 variable "project_id" {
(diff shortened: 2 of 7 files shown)


community/modules/project/service-account/variables.tf
```diff
@@ -20,26 +20,25 @@ variable "billing_account_id" {
...
-  default     = []
+variable "description" {
+  description = "Description of the created service account."
+  type        = string
+  default     = "Service Account"
 }
 
 variable "display_name" {
-  description = "display names of the created service accounts."
+  description = "Display names of the created service account."
```
mr0re1, line 35: s/names/name/

community/modules/project/service-account/variables.tf
```diff
@@ -20,26 +20,25 @@ variable "billing_account_id" {
   default     = ""
 }
 
-variable "description" {
-  description = "Default description of the created service accounts (defaults to no description)."
+variable "deployment_name" {
+  description = "Name of the deployment (will be prepended to service account names)"
```
mr0re1, line 24: s/names/name/

community/modules/project/service-account/variables.tf
```diff
@@ -68,18 +66,20 @@ variable "org_id" {
...
-}
-
 variable "project_id" {
   description = "ID of the project"
   type        = string
 }
 
 variable "project_roles" {
-  description = "list of roles to apply to created service accounts"
+  description = "List of roles to apply to created service accounts"
```
mr0re1, line 75: s/accounts/account/

community/modules/project/service-account/outputs.tf
```diff
@@ -14,47 +14,20 @@
...
+  description = "Service account key (if creation was requested)."
   value       = module.service_accounts.key
 }
-output "keys" {
-  description = "Map of service account keys."
-  value       = module.service_accounts.keys
-}
+
 output "service_account" {
   description = "Service account resource (for single use)."
```
mr0re1, line 28: s/(for single use)//

community/modules/project/service-account/variables.tf
tpdownes: Done

community/modules/project/service-account/variables.tf
tpdownes: Done

community/modules/project/service-account/variables.tf
tpdownes: Done

community/modules/project/service-account/outputs.tf
tpdownes: Done

community/modules/project/service-account/outputs.tf
```diff
@@ -14,47 +14,20 @@
  * limitations under the License.
  */
 
-output "email" {
```
cboneti, line 17: could you comment on why you are removing these vars?

examples/hpc-enterprise-slurm.yaml
```diff
@@ -46,6 +46,36 @@ deployment_groups:
   - id: network1
     source: modules/network/pre-existing-vpc
 
+  - id: controller_sa
```
cboneti, line 49: Don't you need to create these in a previous group? Or is terraform trustworthy here in terms of lifecycle?

cboneti (commented):
Thank you for doing this, this is a much needed improvement, especially with respect to the hpc-slurm-enterprise.yaml changes. 
I left a few comments.

community/modules/project/service-account/outputs.tf
tpdownes: I am removing:

- all variables intended for the use case of creating N>1 service accounts
- anything that is duplicative of the service_account output object (e.g. the e-mail address at line 17)

examples/hpc-enterprise-slurm.yaml
tpdownes: There are 2 potential problems I see:

1. Terraform is satisfied once the SA is created (a global resource) but there is a subsequent latency in propagation of IAM permissions across regions. I cannot find the link but we addressed some of this concern in #1073
2. There are conditions in which the SA can be destroyed and re-created with the same e-mail address but a new unique ID (a large number). In this case, a VM in a subsequent group may be satisfied in the Terraform sense (nothing to do on apply) but actually have a non-existent SA attached to it.

2 is not a problem within a single group (because a single Terraform apply sees the connection).  It can be resolved by:

- rebooting the VM 
- recreating the VM

A thing we could do is also pass along the unique_id and trigger recreation based upon the unique ID. I'd want @mr0re1 to think about adding the unique_id attribute to this object.

community/modules/project/service-account/outputs.tf
```diff
@@ -14,47 +14,20 @@
...
-  description = "Service account key (for single use)."
+  description = "Service account key (if creation was requested)"
   value       = module.service_accounts.key
 }
-output "keys" {
-  description = "Map of service account keys."
-  value       = module.service_accounts.keys
-}
+
 output "service_account" {
```
mr0re1, line 27: Nothing to address now, more like things to consider. 
I think this unnecessary nesting is ugly, it would be great if module outputs were just:
```hcl
.../service-account/outputs.tf
output "email" { ... }
output "scopes" { ... }
```
And module itself could be used as an object
```yaml
- id: sa
  source: .../service-account
  ...
- id: vm
  settings:
    service_account: $(sa)
```
Or simply:
```yaml
- id: vm
  use: [sa]
```

Terraform (1.2) does support it, but it would require some changes in toolkit code.

community/modules/project/service-account/outputs.tf
cboneti: got it, thanks!

examples/hpc-enterprise-slurm.yaml
cboneti: I am worried about permissions to access services.. For instance, a VM being created without GCS scope. Wouldn't that fail *before* getting to our startup scripts?

community/modules/project/service-account/variables.tf
```diff
@@ -56,10 +55,9 @@ variable "grant_xpn_roles" {
   default     = true
 }
 
-variable "names" {
-  description = "Names of the services accounts to create."
-  type        = list(string)
-  default     = []
+variable "name" {
+  description = "Name of the services account to create."
```
mr0re1, line 59: s/services/service/

tpdownes: Done

examples/hpc-enterprise-slurm.yaml
tpdownes: What is possible is that a VM will be created with a valid service account whose IAM roles have not yet propagated to the region where the VM is being created. Scopes are a property of the VM, not the service account, so I don't think you can accidentally not have the scopes. Additionally, the underlying message of this PR is that "cloud-platform" is the right scope for custom service accounts in the first place.

The options here are:

1. Enforce a delay upon creation of a new service account hat achieves X 9s of reliability for attaching it a to GCP resource ([SLA guidance](https://cloud.google.com/iam/docs/access-change-propagation))
2. Make solutions that are resilient against transient failures

(1) makes initial deployment slower for no good reason in most cases. (2) is the right thing to do for a long list of reasons. Thus, I would focus on (2) as in #1073

community/modules/project/service-account/outputs.tf
tpdownes: Would this translate in HCL to `service_account = module.sa`?

cboneti (approved):

community/modules/project/service-account/variables.tf
```diff
@@ -20,26 +20,25 @@ variable "billing_account_id" {
...
 
-variable "description" {
-  description = "Default description of the created service accounts (defaults to no description)."
+variable "deployment_name" {
+  description = "Name of the deployment (will be prepended to service account name)"
   type        = string
-  default     = ""
 }
 
-variable "descriptions" {
```
nick-stroud, line 29: I feel like the correct thing to do with variables that no longer exist would be to leave an unused variable in place that does a validation, which would fail for anything other than null. The error message would direct users to a good course of action to update their blueprint.

examples/hpc-enterprise-slurm.yaml
```diff
@@ -78,6 +108,10 @@ deployment_groups:
       instance_image:
         family: $(vars.family)
         project: $(vars.project)
+      service_account:
```
nick-stroud, line 111: IIUC this PR is stating an opinion that service account email should not be bundled with scopes. Can we create a task to update API throughout HPC Toolkit to reflect this opinion.

nick-stroud (changes requested):

community/modules/project/service-account/variables.tf
tpdownes: Done

examples/hpc-enterprise-slurm.yaml
tpdownes: Done

nick-stroud (approved):

PR #1406, opened 2023-06-02
Fix nfs-server attached disk mounting.

mr0re1:
* Don't set `google_compute_disk.image` to avoid confused mounting;
* Remove faulty and unnecessary `yum -y install nfs-utils`;
* `set -ex` on `install-nfs-server.sh`;
* format and mount attached disk explicitly.

### DIFF
--- a/community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
+++ b/community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
@@ -14,17 +14,20 @@
 # limitations under the License.
 set -ex
 
-systemctl start nfs-server rpcbind
-systemctl enable nfs-server 
+# format, add to fstab, and mount the disk. See https://cloud.google.com/compute/docs/disks/add-persistent-disk
+uuid=$(uuidgen)
+mkfs.ext4 -F -m 0 -U "$uuid" -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/disk/by-id/google-attached_disk
 
-# format and mount the disk. See https://cloud.google.com/compute/docs/disks/add-persistent-disk
-mkfs.ext4 -F -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/disk/by-id/google-attached_disk
 mkdir /exports
-mount -o discard,defaults /dev/disk/by-id/google-attached_disk /exports
+echo "UUID=$uuid /exports ext4 discard,defaults 0 0" >> /etc/fstab
+mount --target /exports/
 
 %{ for mount in local_mounts ~}
-mkdir /exports${mount}
+mkdir -p /exports${mount}
 chmod 755 /exports${mount}
 echo '/exports${mount} *(rw,sync,no_root_squash)' >> "/etc/exports"    
 %{ endfor ~}
+
+systemctl start nfs-server rpcbind
+systemctl enable nfs-server 
 exportfs -r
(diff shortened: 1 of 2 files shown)


community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
```diff
@@ -12,14 +12,18 @@
...
+set -ex
 
-yum -y install nfs-utils
 systemctl start nfs-server rpcbind
 systemctl enable nfs-server 
+
+# format and mount the disk. See https://cloud.google.com/compute/docs/disks/add-persistent-disk
+mkfs.ext4 -F -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/disk/by-id/google-attached_disk
 mkdir /exports
+mount -o discard,defaults /dev/disk/by-id/google-attached_disk /exports
```
cboneti, line 23: Since you are already changing this file, I would have made sure that this can run twice in a row (what happens on the next reboot? here you would loose all data, right?)

Also, I would prefer that the device is added to fstab.

This is easy to achieve in ansible, but there is code that checks for the presence in [fstab in here](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/file-system/pre-existing-network-storage/scripts/mount.sh)

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
```diff
@@ -12,14 +12,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+set -ex
 
-yum -y install nfs-utils
 systemctl start nfs-server rpcbind
```
cboneti, line 17: shouldn't these not start after the exports are configured and the folder mounted?

cboneti (changes requested):
I have left a few suggestions that I would like to see there.

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
mr0re1: It seems to work just fine, also there is `exportfs -r` as a last line in the script.
Do you think it's worth to move it down?

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
mr0re1: Added to fstab

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
```diff
@@ -12,15 +12,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+set -ex
+
+# format, add to fstab, and mount the disk. See https://cloud.google.com/compute/docs/disks/add-persistent-disk
+uuid=$(uuidgen)
+mkfs.ext4 -F -m 0 -U "$uuid" -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/disk/by-id/google-attached_disk
```
cboneti, line 19: won't this still format the device if you reboot the machine? how do you prevent the startup script from wiping the data on the second reboot?

cboneti (changes requested):
one more queston

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
mr0re1: Addressed, added if-condition

cboneti (approved):

cboneti (approved):

PR #1411, opened 2023-06-05
Adopt latest release of startup-script modules

tpdownes:
Adopt latest release of startup-script module; the latest module supports add service accounts as bucket viewers explicitly. Do not (yet) adopt this functionality.

### Submission Checklist

### DIFF
--- a/modules/compute/vm-instance/startup_from_network_storage.tf
+++ b/modules/compute/vm-instance/startup_from_network_storage.tf
@@ -55,7 +55,7 @@ locals {
 }
 
 module "netstorage_startup_script" {
-  source = "github.com/GoogleCloudPlatform/hpc-toolkit//modules/scripts/startup-script?ref=34bb7250"
+  source = "github.com/GoogleCloudPlatform/hpc-toolkit//modules/scripts/startup-script?ref=185837b5"
 
   labels          = local.labels
   project_id      = var.project_id
--- a/community/modules/compute/pbspro-execution/README.md
+++ b/community/modules/compute/pbspro-execution/README.md
@@ -74,7 +74,7 @@ No providers.
 
 | Name | Source | Version |
 |------|--------|---------|
-| <a name="module_execution_startup_script"></a> [execution\_startup\_script](#module\_execution\_startup\_script) | github.com/GoogleCloudPlatform/hpc-toolkit//modules/scripts/startup-script | 34bb7250 |
+| <a name="module_execution_startup_script"></a> [execution\_startup\_script](#module\_execution\_startup\_script) | github.com/GoogleCloudPlatform/hpc-toolkit//modules/scripts/startup-script | 185837b5 |
 | <a name="module_pbs_execution"></a> [pbs\_execution](#module\_pbs\_execution) | github.com/GoogleCloudPlatform/hpc-toolkit//modules/compute/vm-instance | 6c6b9e0a |
 | <a name="module_pbs_install"></a> [pbs\_install](#module\_pbs\_install) | github.com/GoogleCloudPlatform/hpc-toolkit//community/modules/scripts/pbspro-install | 6c6b9e0a |
 
--- a/community/modules/compute/pbspro-execution/main.tf
+++ b/community/modules/compute/pbspro-execution/main.tf
@@ -53,7 +53,7 @@ module "pbs_install" {
 }
 
 module "execution_startup_script" {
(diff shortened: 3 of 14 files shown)


cboneti:
I am not going to have time to review this appropriately, so I am re-assigning to @nick-stroud.

modules/compute/vm-instance/startup_from_network_storage.tf
```diff
@@ -52,11 +52,15 @@ locals {
   )
 
   startup_from_network_storage = module.netstorage_startup_script.startup_script
+
+  service_account_email = lookup(var.service_account, "email", null)
+  bucket_viewers        = local.service_account_email != null && var.set_bucket_viewer ? ["serviceAccount:${local.service_account_email}"] : []
```
nick-stroud, line 57: If it is null, we can be pretty sure that they are using the default compute SA. The default compute SA may not have permission to the bucket. What are your thoughts on granting the role to the default compute SA in this case? I think it can be retrieved via a data block.

modules/compute/vm-instance/variables.tf
```diff
@@ -135,6 +135,12 @@ variable "service_account" {
   }
 }
 
+variable "set_bucket_viewer" {
```
nick-stroud, line 138: I am feeling that this variable needs to indicate some reference to startup_script. It is likely not clear that vm-instance, or batch-job-template, or others are creating a bucket. What do you think about something like `grant_startup_script_bucket_viewer`. I guess as I am brainstorming here, `set` also seems like an odd term in the name. I take your meaning to be "set this property on the sub module", but users are probably not thinking about the submodule. The actually effect is granting a permission or binding a role or what have you.

modules/compute/vm-instance/startup_from_network_storage.tf
tpdownes: My preference is to move toward a model similar to `pre-existing-vpc` in which we have a "default-compute-service-account" module which wraps the data source you're thinking of. Then `null` should not be an allowed value for `var.service_account_email` in the first place.

We have previously identified it as a problem that (sub)network_self_link taking on null in the vm-instance module will "default to the default network". We should identify service account null-ness similarly.

In this model, the default SA is "just another SA" and there are no code/documentation loops to understand.

modules/compute/vm-instance/variables.tf
tpdownes: I share a bit of the concern and did not have great answer myself. I chose to align `set` with terminology in the CFT module:

https://github.com/terraform-google-modules/terraform-google-cloud-storage/blob/master/variables.tf#L155

It also had advantage of landing near `service_account_email` in the README. WDYT of `set_startup_script_bucket_viewer`?

I am anticipating that, eventually, most usage of startup-scripts module will be behind the scenes (as it is here) so I agree this is an important issue.

community/modules/scheduler/htcondor-configure/outputs.tf
```diff
@@ -41,6 +59,15 @@ output "execute_point_service_account" {
   ]
 }
 
+output "execute_point_service_account_iam_email" {
```
nick-stroud, line 62: Is it really necessary to expose this parallel channel. It seems like it will just cause user confusion and is complexity we can handle in our modules. It feels like we are pushing the complexity into the interface.

modules/compute/vm-instance/startup_from_network_storage.tf
tpdownes: Resolving due to lower scope of PR

community/modules/scheduler/htcondor-configure/variables.tf
```diff
@@ -66,7 +64,6 @@ variable "execute_point_roles" {
   default = [
     "roles/monitoring.metricWriter",
     "roles/logging.logWriter",
-    "roles/storage.objectViewer",
```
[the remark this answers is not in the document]
tpdownes: Resolving due to lower scope of PR

modules/compute/vm-instance/variables.tf
tpdownes: Resolving due to lower scope of PR

community/modules/scheduler/htcondor-configure/outputs.tf
tpdownes: Resolving due to lower scope of PR

nick-stroud (approved):

PR #1423, opened 2023-06-08
Add `docs/module-guidelines.md`

mr0re1:


cboneti (changes requested):
Thanks for submitting this, this is awesome. I made a few suggestions.

docs/module-guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+  given defaults. This forces the calling module to provide meaningful values.
+* Variables should only have zero-value defaults (like null or empty strings)
+  where leaving the variable empty is a valid preference which will not be
+  rejected by the underlying API(s).
+* Set good defaults wherever possible. Be opinionated about HPC use cases.
+* Follow common variable [naming conventions](#use-common-names-and-types-for-common-variables).
+* If there are common hpc-toolkit variables already defined, then do not set defaults (`region`, `zone`, `project_id`, `deployment_name`, etc.)
+* All files should contain a license header. Headers can be added automatically
+  using `make add-google-license`.
+
```
tpdownes, line 28: No `provider` blocks should be defined in re-usable modules. It is OK to impose a range of acceptable provider versions. In the case on conflicts, the root module will configure all providers and pass alternatives as an alias. See:

https://developer.hashicorp.com/terraform/language/modules/develop/providers#implicit-provider-inheritance

docs/module-guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+File systems, networks, service accounts, GCS buckets can all be passed into the module and should not be created within the module.
+
+## Prefer FQN over ambiguous formats
+
+For instance, in networks `network_self_link` over `network_name`, `subnetwork_self_link` over `subnetwork_name`, as these immediately work with shared VPCs, and already specify `region/zone/project_ids`.
+
+## All resources should be labeled
+
+The module should take a variable called labels and apply it to every resource.
```
tpdownes, line 66: Add an exception for modules that do not create resources that can be labeled.

docs/module-guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+* `variables.tf` file defining the variables used.
+* (Optional) `outputs.tf` file defining any exported outputs used (if any).
+* (Optional) `modules/` sub-directory pointing to submodules needed to create the
+  top level module.
+
+## General Best Practices
+
+* Variables for environment-specific values (like `project_id`) should not be
+  given defaults. This forces the calling module to provide meaningful values.
+* Variables should only have zero-value defaults (like null or empty strings)
```
nick-stroud, line 20: nit: I strongly prefer null as an empty value to empty string. Perhaps just adding

s/(like null or empty strings)/, such as null (prefered) or empty strings,

cboneti: +1

docs/module-guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+  where leaving the variable empty is a valid preference which will not be
+  rejected by the underlying API(s).
+* Set good defaults wherever possible. Be opinionated about HPC use cases.
+* Follow common variable [naming conventions](#use-common-names-and-types-for-common-variables).
+* If there are common hpc-toolkit variables already defined, then do not set defaults (`region`, `zone`, `project_id`, `deployment_name`, etc.)
+* All files should contain a license header. Headers can be added automatically
+  using `make add-google-license`.
```
cboneti, line 27: That only works if you add a google license.  Instead: headers can be added automatically using [addlicense](https://github.com/google/addlicense), or make add-google-license if adding a Google License.

docs/module-guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Terraform Coding Standards
+
+Any Terraform based modules in the HPC Toolkit should implement the following
+standards:
+
+* `terraform-docs` is used to generate `README` files for each module.
+* The first parameter listed under a module should be [`source`](../modules/README.md#source-required)
```
cboneti, line 35:
```
* The first parameter listed under a module should be [`source`](../modules/README.md#source-required)
```

I just realized this is in the yaml, whereas the rest of the bullets refer to module writing. I think this is out of place

cboneti (changes requested):
I left a few more comments. Also, please address Tom's and Nick's comments.

docs/module-guidelines.md
tpdownes: The value `null` is special and we are probably not thinking about it quite right. In particular, the `nullable=false` attribute on a module input variable will cause the module to use the default (not null) value if a user intentionally supplies null. This is perfect for module wrappers to respect the default values of the underlying module.

Easiest summary I could find:

https://blog.nuvotex.de/terraform-default-values-in-nested-modules/

So a default value of `null` in a wrapper is great, perfect, etc. In the module being wrapped, I think we need more careful consideration about what `null` conveys. If the variable is of type list and one iterates over it, _it's_ extremely nice to have a default empty list rather than null.

Additionally, I think a lot of our modules are being written with functional language approaches: e.g. `var.x != null` instead of `coalesce(var.x, default_value)`.

All that said, empty string isn't great. I suggest that, for now, we replace "string" with "list".

cboneti (approved):

PR #1424, opened 2023-06-09
Add option to gke-node-pool for static node count

nick-stroud:
Manually tested:
- deployed with static nodes
- trigger pre-condition by setting autoscaling and static

### Submission Checklist

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cloud HPC Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #1426, opened 2023-06-09
Add options for create and update timeouts to gke modules

nick-stroud:
After #1424, node pools may take longer to create if they are also creating a large number of static nodes. This change allows users to adjust the timeout.

Manually tested with default of null and setting to value.

### Submission Checklist


cboneti (approved):
The PR has nothing wrong, so approved, but please comment on why this is needed.

community/modules/compute/gke-node-pool/main.tf
```diff
@@ -109,6 +109,11 @@ resource "google_container_node_pool" "node_pool" {
     }
   }
 
+  timeouts {
+    create = var.timeout_create
```
soumyapani, line 113: for customer case, since there are 96 nodes being created per nodepool, we have seen GKE taking longer than default value to create and update the nodepools.

soumyapani (approved):

community/modules/compute/gke-node-pool/main.tf
nick-stroud: To add more context, I think it is relevant in combination with #1424 where node creation can now be part of node pool creation (as opposed to nodes being autoscalled after node pool-creation).

nick-stroud: I added some context to the description.

PR #1427, opened 2023-06-09
Update gke service account variable to separate email and scopes

nick-stroud:
This change seeks to follow the pattern laid out in #1400.

### Submission Checklist

### DIFF
--- a/community/modules/compute/gke-node-pool/variables.tf
+++ b/community/modules/compute/gke-node-pool/variables.tf
@@ -139,16 +139,16 @@ variable "compact_placement" {
   default     = false
 }
 
-variable "service_account" {
-  description = "Service account to use with the system node pool"
-  type = object({
-    email  = string,
-    scopes = set(string)
-  })
-  default = {
-    email  = null
-    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
-  }
+variable "service_account_email" {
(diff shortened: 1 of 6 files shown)


community/modules/compute/gke-node-pool/variables.tf
```diff
@@ -134,16 +134,16 @@ variable "compact_placement" {
...
-    email  = null
-    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
-  }
+variable "service_account_email" {
+  description = "Service account e-mail address to use with the node pool"
+  type        = string
+  default     = null
+}
+
+variable "scopes" {
```
tpdownes, line 143: I think my only concern here is that my instinct would be to name it "service_account_scopes" in part to ensure that they always land in the documentation right next to one another (and in email/scopes order). WDYT?

tpdownes (commented):
I'd like group to settle on the one point raised prior to moving forward. Otherwise, thanks for aligning the GKE module with the new service-account module.

community/modules/compute/gke-node-pool/variables.tf
nick-stroud: SGTM and done.

tpdownes (approved):

PR #1439, opened 2023-06-13
Update develop with release-candidate: Fix Ansible installation upon re-run

rohitramu:


modules/scripts/startup-script/files/install_ansible.sh
nick-stroud, line 1: In reality this merge has more to do with including Tom's change than v1.18.1 merge back. Technically the crypto change is already on develop, even though the diff incorrectly shows that is is not. Please update the title to:

`Update develop with release-candidate: Fix Ansible installation upon re-run`

nick-stroud (approved):

PR #1441, opened 2023-06-13
Conditionally exclude nodeSelector when not needed

nick-stroud:
Prior to this change there would be an empty entry in the yaml when no node selectors were provided.

### Submission Checklist

### DIFF
--- a/community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
+++ b/community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
@@ -25,13 +25,10 @@ spec:
                 - ${node_pool}
                 %{~ endfor ~}
       %{~ endif ~}
-      %{~ if has_node_selector ~}
+      %{~ if length(node_selectors) > 0 ~}
       nodeSelector:
-      %{~ if machine_family != null ~}
-        cloud.google.com/machine-family: ${machine_family}
-      %{~ endif ~}
-      %{~ for key, val in node_selectors ~}
-        ${key}: ${val}
+      %{~ for selector in node_selectors ~}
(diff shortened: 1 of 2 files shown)


community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
```diff
@@ -25,13 +25,15 @@ spec:
                 - ${node_pool}
                 %{~ endfor ~}
       %{~ endif ~}
+      %{~ if has_node_selector ~}
```
tpdownes, line 28: I suggest you take the 1 minute to test whether you can remove the extra local variable (which reproduced information you already need anyhow):

```hcl
%{~ if machine_family != null || length(node_selectors) > 0 ~}
```

tpdownes (approved):
Please take a small amount of the time to attempt the suggested change (i.e. just above "nit")

community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
tpdownes: I find it less clear while reading the template since it (outwardly) appears to contain independent information. I won't block the PR on it, though, so effectively this may be a `nit`.

nick-stroud: I have refactored to not pass dup information while still keeping logic in the template to a minimum.

This was time well spent as I also found that `nodeSelectors` were generally broken.

tpdownes (approved):
I'll trust that the output does something useful in a k8s `Job` sense, but the TF side LGTM.

PR #1442, opened 2023-06-13
Adds gke-persistent-volume module

nick-stroud:
The module added in this change will use a filestore and a gke-cluster and will add the necessary persistent volume and persistent volume claim kubernetes objects to the cluster to allow a job to mount the filestore.

See [example blueprint in subsequent PR](https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1443/commits/9ec066d38caf91dbcd8f1178d5846e9f22d12921).

> **Note**: In the future this same module will also support using a GCS bucket with GCSFuse. It should be acknowledged that some of the logic in this module will have to change to support GCS but this will be saved for a future pr to try to minimize the complexity of one.

> **Note**: I will call out the use of the kubectl provider. I would prefer to use the official kubernetes provider but have selected to not do so as it is not able to operate in the same terraform apply as the creation of the gke cluster (aka, the gke-persistent-volume would be required to be in a separate group from cluster creation). While this is doable and perhaps desirable for unrelated reasons, I did not like arbitrary limiting the usage of this module when an alternative exists. I have opened a task to work the issue with the official provider.

Must be rebased after #1441 is merged.

To be completed in future PR:
- Documentation of module

### Submission Checklist


tpdownes:
as you point out in the description, @nick-stroud, I suggest you rebase off develop and remove a36d78fbfb02fae4823c1f73eb0de0db81cd4682 entirely.

community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
```diff
@@ -61,5 +61,20 @@ spec:
             cpu: ${millicpu_request}
           %{~ endif ~}
         %{~ endif ~}
+        %{~ if length(volumes) != 0 ~}
```
tpdownes, line 64: This is the nittiest of nits, but I prefer `> 0` for this comparison despite the impossibility of negative values.

community/modules/compute/gke-job-template/variables.tf
```diff
@@ -113,6 +113,16 @@ variable "random_name_sufix" {
   default     = true
 }
 
+variable "persistent_volume_claim" {
```
tpdownes, line 116: Need this be singular in order to match output values? It would most naturally be plural.

If so, I would probably change the input to plural and the outputs to match.

modules/file-system/filestore/outputs.tf
```diff
@@ -50,3 +50,13 @@ output "mount_runner" {
   EOT
   value       = local.mount_runner
 }
+
+output "filestore_id" {
```
tpdownes, line 54: The way I read the the [filestore] resource docs, the `id` attributes should be exported only after creation. This should cause Terraform to order resources correctly.

[filestore]: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#attributes-reference

modules/file-system/filestore/outputs.tf
```diff
@@ -50,3 +50,13 @@ output "mount_runner" {
...
+  description = "An identifier for the resource with format `projects/{{project}}/locations/{{location}}/instances/{{name}}`"
+  value       = google_filestore_instance.filestore_instance.id
+}
+
+output "capacity_gb" {
```
tpdownes, line 59: The way I read the the [filestore] resource docs, the `file_shares` arguments are part of creation. I am not 100% certain, but this may cause Terraform to order resources incorrectly unless they explicitly depend upon the filestore resource itself. This might need a `depends_on` block.

[filestore]: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#file_shares

community/modules/file-system/gke-persistent-volume/variables.tf
```diff
@@ -14,6 +14,38 @@
...
+    remote_mount          = string,
+    local_mount           = string,
+    fs_type               = string,
+    mount_options         = string,
+    client_install_runner = map(string)
+    mount_runner          = map(string)
+  })
+}
+
+variable "filestore_id" {
```
tpdownes, line 35: As this module develops, is your intent to require that the user supply exactly 1 Filestore ID _or_ 1 bucket _or_ 1 filesystem-to-be-named-later?

community/modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -18,3 +18,71 @@ locals {
   # This label allows for billing report tracking based on module.
   labels = merge(var.labels, { ghpc_module = "gke-persistent-volume" })
 }
+
+locals {
+  split_filestore_id   = var.filestore_id != null ? split("/", var.filestore_id) : [null, null, null, null, null, null]
+  location             = local.split_filestore_id[3]
```
tpdownes, line 24: I think these are more naturally written as:

```hcl
location = try(split("/", var.filestore_id)[3], null)
filestore_name = try(split("/", var.filestore_id)[5], null)
```

community/modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -18,3 +18,71 @@ locals {
...
+  filestore_name       = local.split_filestore_id[5]
+  filestore_share_name = trimprefix(var.network_storage.remote_mount, "/")
+
+  pv_name  = "${local.filestore_name}-pv"
```
tpdownes, line 28: It appears to me that you are doing nothing to address `var.filestore_id == null` here. Do you have a plan? It seems to me that, in this initial commit, `var.filestore_id == null` probably shouldn't be allowed.

My plan forward (assuming the intent is to also support Cloud Storage buckets, etc.) would be:

1. Refactor each type of storage as a submodule of this module (directly in the filesystem path)
2. Instantiate each submodule only if the matching input variable is defined
3. Merge the outputs of each submodule into a set of PVs, PVCs
4. Pluralize the output values

community/modules/compute/gke-job-template/variables.tf
nick-stroud: You are correct about input-output matching. I don't have any strong feelings, just mating the pattern of a singular `network_storage`. 

I think plural is probably better because users are more likely to look at inputs than outputs.

modules/file-system/filestore/outputs.tf
nick-stroud: That is my understanding. Is there any action to be taken with respect to this comment?

community/modules/file-system/gke-persistent-volume/variables.tf
nick-stroud: Correct. There is a 1:1 between filesystem and gke-persistent-volume. A future version would have filestore_id, and some other variable for bucket, they would both default to null and a validation would be done to ensure one of them was set. Probably it is technically incorrect for this version to allow null as in input in validation.

community/modules/file-system/gke-persistent-volume/main.tf
nick-stroud: Good suggestion using submodules. I had more or less considered doing this but all in locals, submodules may be cleaner. Do you want to see these changes as part of this PR or can it be follow on. 

I could imagine, this pr, then separate pr to move to submodule, then pr adding gcs.

modules/file-system/filestore/outputs.tf
tpdownes: No, but I'm contrasting it with the capacity output value (where I have a concern) below.

community/modules/file-system/gke-persistent-volume/variables.tf
tpdownes: The reason I think it merits consideration now is that I did some quick analyses of how you might ensure that only one of the inputs values was set and could not find a good model for using `precondition` to prevent it.  I did have the thought that you might eliminate the problem by simply allowing multiple PVs/PVCs per instance of this module:

1. add an `s` to the names
2. allow the module to combine buckets and Filestore and X/Y/Z into a set of PVs/PVCs

Might avoid a lot of user and developer frustration. I also came across this today:

https://developer.hashicorp.com/terraform/language/functions/try

Look at "We can also use try to deal..." to see how you might treat input Filestore IDs as a string or a list so that you can have N>1 Filestore/Buckets per module, too.

tpdownes: I think the thing to do right now is:

- disallow `var.filestore_id == null`; that appears to generate genuinely broken inputs
- add the `s`

Consider the comments above to evolve the design.

community/modules/file-system/gke-persistent-volume/main.tf
tpdownes: See https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1442#discussion_r1230123258 for the "right now" suggestions but I think you'll find the suggestion potentially simplifying.

modules/file-system/filestore/outputs.tf
nick-stroud: I tested this and the `capacity_gb` output does not release until the filestore is created. 

I used the following blueprint to test:

```
- group: primary
  modules:
  - id: network1

  - id: homefs
    source: ./modules/file-system/filestore
    use: [network1]
    settings:
      local_mount: /home

  - id: wait1
    settings:
      instance_name: ((tostring(module.homefs.capacity_gb)))

  - id: wait2
    settings:
      instance_name: ((tostring(module.homefs.test)))
```

Where `test` is an output I added to filestore:

```
output "test" {
  value       = "a-string"
}
```

Terraform output:

```
module.homefs.random_id.resource_name_suffix: Creating...
module.homefs.random_id.resource_name_suffix: Creation complete after 0s [id=bredWg]
module.wait2.null_resource.wait_for_startup: Creating...
module.wait2.null_resource.wait_for_startup: Provisioning with 'local-exec'...
module.wait2.null_resource.wait_for_startup (local-exec): Executing: ["/bin/sh" "-c" "/bin/bash modules/embedded/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh"]
module.homefs.google_filestore_instance.filestore_instance: Creating...
module.wait2.null_resource.wait_for_startup (local-exec): ERROR: (gcloud.compute.instances.get-serial-port-output) Could not fetch serial port output: The resource ... was not found
module.homefs.google_filestore_instance.filestore_instance: Still creating... [10s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [20s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [30s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [40s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [50s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m0s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m10s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m20s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m30s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m40s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [1m50s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [2m0s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Still creating... [2m10s elapsed]
module.homefs.google_filestore_instance.filestore_instance: Creation complete after 2m14s [id=...]
module.wait1.null_resource.wait_for_startup: Creating...
module.wait1.null_resource.wait_for_startup: Provisioning with 'local-exec'...
module.wait1.null_resource.wait_for_startup (local-exec): Executing: ["/bin/sh" "-c" "/bin/bash modules/embedded/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh"]
module.wait1.null_resource.wait_for_startup (local-exec): ERROR: (gcloud.compute.instances.get-serial-port-output) Could not fetch serial port output: The resource ...' was not found
```

You can see that the string is released immediately but capacity does not release until filestore creation is complete.

community/modules/compute/gke-job-template/templates/gke-job-base.yaml.tftpl
nick-stroud: Done.

community/modules/compute/gke-job-template/variables.tf
nick-stroud: Updated with your suggestion.

community/modules/file-system/gke-persistent-volume/main.tf
nick-stroud: Done / null is no longer valid for this PR

community/modules/file-system/gke-persistent-volume/variables.tf
nick-stroud: I think I have satisfied your short term requests. I will also look at how to control the "choose one" issue as it is clearly not super straight forward.

tpdownes (approved):

PR #1444, opened 2023-06-13
Do not use `log.Fatal` in `pkg/config`

mr0re1:
Motivation:
* `pkg/config` is a library, it shouldn't control lifecycle of the application;
* Improve tests failure investigation.


tpdownes (approved):
This is line with a "better testability" document that we wrote but didn't go anywhere. Thanks!

PR #1448, opened 2023-06-14
Add rich error messages with position and snippet

mr0re1:
* Add `Path` & `BpError` to attribute errors to the BP parts;
* Add `ErrorRenderer` to control rendiring of error messages;
* Add `YamlErrorRenderer` that attributes `Path` to the `Pos`;
* Update a few errors to use `BpError`.

NOTE: The crude implementation of `Path` is improved in experimental #1347 and can be addressed after/before this PR is landed/considered worthy.

```sh
$ make && ./ghpc create tst.yaml
Error: deployment_name input error, cause: value was not of type string
on line 6, column 22:
6:   "deployment_name": []
```

### DIFF
--- a/pkg/config/yaml.go
+++ b/pkg/config/yaml.go
@@ -1,0 +1,110 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Path points to concrete location in the blueprint file.
+type Path struct {
+	v string
+}
+
+func (p Path) String() string {
+	return p.v
+}
+
+// At is a builder method for a path of a child in a sequence.
+func (p Path) At(i int) Path {
+	return Path{fmt.Sprintf("%s[%d]", p.v, i)}
+}
+
+// Dot is a builder method for a path of a child in a mapping.
+func (p Path) Dot(k string) Path {
+	if p.v == "" {
+		return Path{k}
+	}
+	return Path{fmt.Sprintf("%s.%s", p.v, k)}
+}
+
+// Pos is a position in the blueprint file.
+type Pos struct {
+	Line   int
+	Column int
+}
+
+func importBlueprint(f string) (Blueprint, YamlCtx, error) {
+	data, err := os.ReadFile(f)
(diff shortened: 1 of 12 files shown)


cboneti (commented):
Directionally, I like this a lot, but I would like someone else to read the code in detail. Maybe Tom?

mr0re1:
Changed `YamlErrorRenderer -> YamlCtx`. Moved rendering logic to `cmd`.

pkg/config/yaml.go
```diff
@@ -0,0 +1,110 @@
...
+	return bp, newYamlCtx(data), nil
+}
+
+// YamlCtx is a contextual information to render errors.
+type YamlCtx struct {
+	PathToPos map[Path]Pos
+	Lines     []string
+}
+
+func newYamlCtx(data []byte) YamlCtx {
```
nick-stroud, line 75: Please add a few tests for this new function.

pkg/config/yaml.go
```diff
@@ -0,0 +1,110 @@
...
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Path points to concrete location in the blueprint file.
+type Path struct {
```
nick-stroud, line 27: Please add some minimal testing of Path functionality. You may achieve the same coverage via testing of `newYamlCtx`. So perhaps write that first and make a judgement call if dedicated Path testing is needed.

nick-stroud (changes requested):

pkg/config/yaml.go
```diff
@@ -0,0 +1,110 @@
...
+	sc := bufio.NewScanner(bytes.NewReader(data))
+	for sc.Scan() {
+		lines = append(lines, sc.Text())
+	}
+	return YamlCtx{m, lines}
+}
+
+type nodeCapturer struct{ n *yaml.Node }
+
+func (c *nodeCapturer) UnmarshalYAML(n *yaml.Node) error {
```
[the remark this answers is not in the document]
mr0re1: `UnmarshalYAML` is called by `yaml.v3`

pkg/config/yaml_test.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
+)
+
+func TestYamlCtx(t *testing.T) {
+	data := `
```
nick-stroud, line 26: nit: 
1. It takes a little cognitive overhead to confirm line numbers since the data does not start on line 1. 
2. It seems odd to me that `blueprint_name` is reported as being on line 3 instead of 2. I assume it counts L26 as the first line with no characters. 

Possible ideas to reduce cognitive overhead:
- offset this data definition so that the last digit of line number matches. aka, add 5 lines of white space or comments before `data` so that L33 of the test file contains `blueprint_name` which aligns with Line 3 in mapping.
- Put data in separate file
- Do nothing

nick-stroud (approved):

pkg/config/yaml_test.go
mr0re1: > offset this data definition so that the last digit of line number matches.

The alignment wil break soon, as new imports will be added above;

> Put data in separate file

Data is rather small to justify the move

Added `# line D0` comments to improve navigation

PR #1449, opened 2023-06-14
Fix regex for validating GroupName

mr0re1:


pkg/config/config.go
```diff
@@ -92,8 +86,9 @@ func (n GroupName) Validate() error {
 	if n == "" {
 		return errors.New(errorMessages["emptyGroupName"])
 	}
-	if hasIllegalChars(string(n)) {
-		return fmt.Errorf("%s %s", errorMessages["illegalChars"], n)
+
+	if !regexp.MustCompile(`^[\w]+[\w-]*$`).MatchString(string(n)) {
```
tpdownes, line 90: I think it might be wise to disallow ending in a dash. WDYT?

mr0re1: Sounds right, though I can't formulate why ?)

tpdownes: Probably harder to detect as a character vs operator? In any case, it's a rule with GCP resource naming.

https://cloud.google.com/compute/docs/naming-resources#resource-name-format

tpdownes (approved):

PR #1454, opened 2023-06-14
Update all references to golang base image

rohitramu:
The integration tests currently use "golang:latest" as the base image, while the hpc-toolkit-builder uses "golang:bullseye" as the base image.  This means that the tests will build using the system packages available in "golang:latest", and run the built binary in an environment based on "golang:bullseye".  The "golang:latest" image tends to get updated before the "golang:bullseye" image, potentially causing integration test failures when they are mismatched.

This PR makes sure that all Docker image builds will use "golang:latest" as the base image, preventing these kinds of build failures in the future.


nick-stroud:
From side discussion I believe we want to target bullseye and merge to the release-candidate branch.

PR #1464, opened 2023-06-15
Update "google", "google-beta" providers to 4.69.1

rohitramu:


community/front-end/ofe/infrastructure_files/workbench_tf/google/README.md
```diff
@@ -35,7 +35,7 @@ limitations under the License.
 | Name | Source | Version |
 |------|--------|---------|
 | <a name="module_project_radlab_ds_analytics"></a> [project\_radlab\_ds\_analytics](#module\_project\_radlab\_ds\_analytics) | terraform-google-modules/project-factory/google | ~> 11.0 |
-| <a name="module_vpc_ai_notebook"></a> [vpc\_ai\_notebook](#module\_vpc\_ai\_notebook) | terraform-google-modules/network/google | ~> 3.0 |
+| <a name="module_vpc_ai_notebook"></a> [vpc\_ai\_notebook](#module\_vpc\_ai\_notebook) | terraform-google-modules/network/google | ~> 7.0 |
```
cboneti, line 38: What is this update?

community/front-end/ofe/infrastructure_files/workbench_tf/google/main.tf
```diff
@@ -110,7 +110,7 @@ data "google_compute_subnetwork" "default" {
 module "vpc_ai_notebook" {
   count   = var.create_network ? 1 : 0
   source  = "terraform-google-modules/network/google"
-  version = "~> 3.0"
+  version = "~> 7.0"
```
cboneti, line 113: same here, this seems like a pretty big jump, do we know if it works for sure?

community/front-end/ofe/tf/main.tf
```diff
@@ -64,7 +64,7 @@ module "service_account" {
 
 module "control_bucket" {
   source  = "terraform-google-modules/cloud-storage/google"
-  version = "~> 2.2"
+  version = "~> 4.0"
```
cboneti, line 67: same here, are we sure we want to update these?

community/modules/compute/gke-node-pool/README.md
```diff
@@ -142,15 +142,15 @@ limitations under the License.
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
-| <a name="requirement_google"></a> [google](#requirement\_google) | >= 4.60.0, < 5.0 |
-| <a name="requirement_google-beta"></a> [google-beta](#requirement\_google-beta) | >= 4.60.0, < 5.0 |
+| <a name="requirement_google"></a> [google](#requirement\_google) | ~> 4.69.1 |
```
cboneti, line 145: I would talk to Nick here, I assume he left a large range like this by design. Also, you are not actually upgrading it here, if anything you are constraining it...

cboneti (commented):
I am a bit puzzled by the extent of this change and I would like to discuss this with the team tomorrow.

community/front-end/ofe/infrastructure_files/workbench_tf/google/README.md
rohitramu: This is one of the modules which explicitly specify a requirement on a lower version of the google provider.  Once I upgraded the provider version in our modules to the latest (v4.69.1), they conflicted with the requirements of the modules we reference, resulting in the following error message when running `terraform init`:

```
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/google: no available releases match the given
│ constraints >= 3.53.0, ~> 3.53, ~> 4.69.1, < 5.0.0
```

Upgrading the modules so that there's no conflict in the required version of the "hashicorp/google" provider allowed me to run `terraform init` without hitting the above error.

community/modules/compute/gke-node-pool/README.md
rohitramu: I assumed we should update this since we are [pinning to minor versions](https://cloud.google.com/docs/terraform/best-practices-for-terraform#minor-provider-versions) everywhere else.  But I agree, getting Nick's input is probably a good idea here.

community/front-end/ofe/infrastructure_files/workbench_tf/google/main.tf
rohitramu: I've run all of the integration tests, but I'm not sure about how much of this module it covers.  Maybe after talking to the team tomorrow, I'll have a better idea of how to test this.

rohitramu:
After speaking with the team, I've modified this PR to only update provider versions in the "golden_copies" and "modulewriter".

nick-stroud (approved):

PR #1466, opened 2023-06-16
Deprecated `WrapSettingsWith`

mr0re1:
* Construct `merge` expression instead of using `WrapSettingsWith`;
* Move `TokensForValue` into `pkg/config`;
* Add `modulewriter_test` to compensate for missed coverage;
* Mark `WrapSettingsWith` as deprecated, remove any usage;
* Change `addListValue` to construct an expression;
* Add support for multi-module `ProductOfModuleUse`;
* Add "golden copy" test for `flatten` and `merge` usage.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -290,23 +290,15 @@ type ModuleID string
 
 // Module stores YAML definition of an HPC cluster component defined in a blueprint
 type Module struct {
-	Source           string
-	Kind             ModuleKind
-	ID               ModuleID
-	Use              []ModuleID                `yaml:"use,omitempty"`
-	WrapSettingsWith map[string][]string       `yaml:"wrapsettingswith,omitempty"`
-	Outputs          []modulereader.OutputInfo `yaml:"outputs,omitempty"`
-	Settings         Dict                      `yaml:"settings,omitempty"`
+	Source   string
+	Kind     ModuleKind
+	ID       ModuleID
+	Use      []ModuleID                `yaml:"use,omitempty"`
+	Outputs  []modulereader.OutputInfo `yaml:"outputs,omitempty"`
+	Settings Dict                      `yaml:"settings,omitempty"`
 	// DEPRECATED fields, keep in the struct for backwards compatibility
-	RequiredApis interface{} `yaml:"required_apis,omitempty"`
-}
-
-// createWrapSettingsWith ensures WrapSettingsWith field is not nil, if it is
-// a new map is created.
-func (m *Module) createWrapSettingsWith() {
-	if m.WrapSettingsWith == nil {
-		m.WrapSettingsWith = make(map[string][]string)
-	}
+	RequiredApis     interface{} `yaml:"required_apis,omitempty"`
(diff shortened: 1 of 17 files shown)


pkg/config/config.go
```diff
@@ -369,8 +361,8 @@ func (m Module) listUnusedModules() []ModuleID {
 	used := map[ModuleID]bool{}
 	// Recurse through objects/maps/lists checking each element for having `ProductOfModuleUse` mark.
 	cty.Walk(m.Settings.AsObject(), func(p cty.Path, v cty.Value) (bool, error) {
-		if mark, has := HasMark[ProductOfModuleUse](v); has {
-			used[mark.Module] = true
+		for _, mod := range IsProductOfModuleUse(v) {
```
tpdownes, line 364: The way I read the code you are migrating from a list of Marks to a single Mark that contains a string that is parsable into a list?

pkg/config/expand.go
```diff
@@ -247,60 +246,60 @@ func (dc *DeploymentConfig) combineLabels() error {
...
-	}
-	return nil
+	return cty.ObjectVal(mergeMaps(global, extra, modLabels)), nil
 }
 
-// mergeLabels returns a new map with the keys from both maps. If a key exists in both maps,
-// the value from the first map is used.
-func mergeLabels(a map[string]cty.Value, b map[string]cty.Value) map[string]cty.Value {
+// see https://developer.hashicorp.com/terraform/language/functions/merge
+func mergeMaps(ms ...map[string]cty.Value) map[string]cty.Value {
```
tpdownes, line 299: Please re-create the part of the comment that explicitly states which elements win in case of conflict (i.e. first, last)

tools/validate_configs/golden_copies/configs/merge_flatten.yaml
```diff
@@ -0,0 +1,35 @@
...
+  
+  - id: first-vm
+    source: modules/compute/vm-instance
+    use: [first-fs]
+    settings:
+      labels: {"green": "sleeves"}
+    
+  - id: second-vm
+    source: modules/compute/vm-instance
+    use: [first-fs, second-fs]
```
tpdownes, line 35: Please add a trailing newline here.

tpdownes (approved):
1. Please check whether it is possible `FunctionCallExpression` to verify that a function exists in Terraform or HCL "base" language. If it is easy, please add that functionality so that we anticipate the possibility of adding HCL functions into blueprints more directly. My quick look suggests that all they are defined in an [internal package](https://github.com/hashicorp/terraform/blob/5ed38eb3fa6d44f167a48cea80a1bf2941011e3e/internal/lang/funcs/descriptions.go#L284).
1. Address minor comments. I'm glad to see Terraform functions adopted more directly.
1. You need to rebase. If it's easy, go ahead and merge after doing (1) and (2). If there are real conflicts, please alert me.

pkg/config/config.go
mr0re1: Previously there was at most one module per value, now it's a list of modules.
I've opted to a parsable list (hidden implementation detail) because the `Mark` struct has to be `hashable` and `slices` aren't hashable.

PR #1470, opened 2023-06-17
Update "google" provider in OFE from 3.x to 4.x

rohitramu:


rohitramu:
Need to wait for #1464 before merging this PR.

community/front-end/ofe/tf/main.tf
```diff
@@ -64,7 +64,7 @@ module "service_account" {
 
 module "control_bucket" {
   source  = "terraform-google-modules/cloud-storage/google"
-  version = "~> 2.2"
+  version = "~> 3.0"
```
tpdownes, line 67: Please remove this as (A) it's not relevant to the purpose of the PR (B) the release notes for 3.0 introduce breaking changes (B) 4.0 is the latest anyhow

tpdownes (changes requested):
Before merging:

- remove the cloud-storage module update
- consider wait until end of day and checking for a 4.70.0 release.

tpdownes (approved):

community/front-end/ofe/tf/main.tf
rohitramu: This was the smallest change I found which would get me past the following error during the "terraform validate" step in pre-commits:

```
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/google: no available releases match the given constraints >=
│ 3.53.0, ~> 3.53, ~> 4.0, < 5.0.0
```

Is there something better I could do to avoid that error?

tpdownes: I see your point. The most correct changes to allow TPG 4.x compatibility would be either:

- `~> 3.1` (4.x wasn't allowed in 3.0)
- `~> 4.0`

Since we should be testing these changes by running the OFE anyhow, I would just bite the bullet and do 4.0.

tpdownes: Please create a task in this iteration to test the OFE prior to release.

rohitramu: Thanks for the tip, I've updated the "cloud-storage" module to `~> 4.0`, and created the task to test OFE.

tpdownes (approved):

PR #1472, opened 2023-06-17
Allow providing short names for image project

rohitramu:
Currently, values provided to the `instance_image.project` property must be in the format `projects/<project-name>/global/images/family`:

```txt
instance_image:
  family: slurm-gcp-5-7-hpc-centos-7
  project: projects/schedmd-slurm-public/global/images/family
```

However for the module in `modules/compute/vm-instance`, the value provided is much simpler:

```txt
instance_image:
  family: slurm-gcp-5-7-hpc-centos-7
  project: schedmd-slurm-public
```

This PR allows users to provide just the `<project-name>` value to the the `project` property.


modules/scheduler/batch-job-template/main.tf
nick-stroud, line 1: For whatever reason, this module currently works fine currently with just providing the project name. See the [default](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/d6a3ef45165c0c925b77b2e8eeddc9f69c35c22b/modules/scheduler/batch-job-template/variables.tf#L175-L178) for example.

What is the rationale for adding the locals to this module?

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
nick-stroud, line 1: There are many instances where we use the long form project id in our examples. Please update examples and documentation to use the regular project id. 

You can search for `projects/schedmd-slurm-public` in our repo.

nick-stroud (changes requested):

modules/scheduler/batch-job-template/main.tf
rohitramu: Good catch.  Reverted.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
rohitramu: Good point.  Done.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
```diff
@@ -36,12 +36,19 @@ locals {
...
+
   # Handle VM image format from 2 sources, prioritize source_image* variables
   # over instance_image
   source_image_input_used = var.source_image != "" || var.source_image_family != "" || var.source_image_project != ""
-  source_image            = local.source_image_input_used ? var.source_image : lookup(var.instance_image, "name", "")
-  source_image_family     = local.source_image_input_used ? var.source_image_family : lookup(var.instance_image, "family", "")
-  source_image_project    = local.source_image_input_used ? var.source_image_project : lookup(var.instance_image, "project", "")
+  source_image            = local.source_image_input_used ? var.source_image : local.instance_image.name
+  source_image_family     = local.source_image_input_used ? var.source_image_family : local.instance_image.family
+  source_image_project    = local.source_image_input_used ? var.source_image_project : local.instance_image.project
```
nick-stroud, line 51: Is there any reason to not provide the same treatment to `source_image_project`?

You could leave all the code as it was and then just perform the conversion on `local. source_image_project`.
For example:
```
normalized_source_image_project = strcontains(local. source_image_project, "/") ? local.instance_image_project_raw : "projects/${local.normalized_source_image_project}/global/images/family"
```

Then just reference `normalized_source_image_project` below.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
```diff
@@ -36,12 +36,19 @@ locals {
   enable_public_ip_access_config = var.disable_controller_public_ips ? [] : [{ nat_ip = null, network_tier = null }]
   access_config                  = length(var.access_config) == 0 ? local.enable_public_ip_access_config : var.access_config
 
+  instance_image_project_raw = lookup(var.instance_image, "project", "")
+  instance_image = {
+    name    = lookup(var.instance_image, "name", "")
+    family  = lookup(var.instance_image, "family", "")
+    project = strcontains(local.instance_image_project_raw, "/") ? local.instance_image_project_raw : "projects/${local.instance_image_project_raw}/global/images/family"
```
nick-stroud, line 43: I think the long form is only needed when family is used (not name). Have you tested with name? I think you will want to also hide the transformation behind the condition that family is set.

nick-stroud (changes requested):

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
rohitramu: Good point.  I think I've done what you're asking, but let me know if not.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
rohitramu: I've removed this block altogether after my latest update.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
```diff
@@ -23,10 +23,11 @@ locals {
 
   # Handle VM image format from 2 sources, prioritize source_image* variables
   # over instance_image
-  source_image_input_used = var.source_image != "" || var.source_image_family != "" || var.source_image_project != ""
-  source_image            = local.source_image_input_used ? var.source_image : lookup(var.instance_image, "name", "")
-  source_image_family     = local.source_image_input_used ? var.source_image_family : lookup(var.instance_image, "family", "")
-  source_image_project    = local.source_image_input_used ? var.source_image_project : lookup(var.instance_image, "project", "")
+  source_image_input_used         = var.source_image != "" || var.source_image_family != "" || var.source_image_project != ""
+  source_image                    = local.source_image_input_used ? var.source_image : var.instance_image.name
+  source_image_family             = local.source_image_input_used ? var.source_image_family : var.instance_image.family
```
nick-stroud, line 28: The original code was performing lookups on `instance_image` for name and family (and project). What happens in the case of `var.instance_image.name` if name is not provided (it is optional). I suspect it will fail here or null will be returned and fail lower in the stack.

My gut is that the lookups were there for a reason and we will be safer to keep them.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
```diff
@@ -23,10 +23,11 @@ locals {
...
+  source_image_project            = local.source_image_input_used ? var.source_image_project : var.instance_image.project
+  source_image_project_normalized = strcontains(local.source_image_project, "/") ? local.source_image_project : "projects/${local.source_image_project}/global/images/family"
```
nick-stroud, line 30: I suspect that this conversion is only needed when we are selecting the image based on family (not name). To go further, I think you may see an error when providing this format with image name. If this is the case then we would only want to perform this conversion if 1) there are no slashes in the var and 2) family != "".

nick-stroud (changes requested):

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
rohitramu: Added lookups.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
rohitramu: Added the condition `local.source_image != ""`, so that if the image name is used, the project won't be modified.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
```diff
@@ -36,12 +36,14 @@ locals {
...
-  source_image            = local.source_image_input_used ? var.source_image : lookup(var.instance_image, "name", "")
-  source_image_family     = local.source_image_input_used ? var.source_image_family : lookup(var.instance_image, "family", "")
-  source_image_project    = local.source_image_input_used ? var.source_image_project : lookup(var.instance_image, "project", "")
+  source_image_input_used         = var.source_image != "" || var.source_image_family != "" || var.source_image_project != ""
+  source_image                    = local.source_image_input_used ? var.source_image : lookup(var.instance_image, "name", "")
+  source_image_family             = local.source_image_input_used ? var.source_image_family : lookup(var.instance_image, "family", "")
+  source_image_project            = local.source_image_input_used ? var.source_image_project : lookup(var.instance_image, "project", "")
+  source_image_project_normalized = local.source_image != "" || strcontains(local.source_image_project, "/") ? local.source_image_project : "projects/${local.source_image_project}/global/images/family"
```
nick-stroud, line 46: nit: This block is identical in 3 different modules. Sometimes we try to capture this logic in a separate file so we can prevent drift. 

Example: [threads_per_core_calc.tf](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/compute/vm-instance/threads_per_core_calc.tf) is found in two different modules but is forced to be maintained as identical via [duplicate-diff.py](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/d6a3ef45165c0c925b77b2e8eeddc9f69c35c22b/tools/duplicate-diff.py#L45-L46).

community/examples/hpc-slurm-ubuntu2004.yaml
nick-stroud, line 1: Once you are code complete I would recommend running all slurm tests. You can use the babysitter tool to execute: 

`tools/cloud-build/babysit_tests.py <commit> slurm5 --project <project>`

nick-stroud (approved):

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/main.tf
rohitramu: Done.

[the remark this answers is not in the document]
rohitramu: Done.

community/examples/hpc-slurm-ubuntu2004.yaml
rohitramu: Running.

nick-stroud (approved):

PR #1486, opened 2023-06-21
Remove `settingsToIgnore` from `useModule`

mr0re1:
Derive required information from value marks.


tpdownes (approved):
This is a good step forward integrating the improved parsing into our implementation.

PR #1496, opened 2023-06-23
Add functionality for Spack module to take data runners

nick-stroud:
`data_files` are a subset of `startup-script.runners`. They can only be of type data and must have a destination of an absolute path. While this same functionality could be achieved by using startup script directly, it was decided that it would be logically cleaner to define all of the spack configs in a single module.

It was decided to not duplicate the validation that startup script does. This leaves some minor naming mismatch in error messages as validation will report that "runner cannot do such and such" instead of "data_files".

Tested manually: Upload file via data runner and then cat the file in commands

### Submission Checklist


community/modules/scripts/spack-install/variables.tf
```diff
@@ -58,10 +58,24 @@ variable "log_file" {
   type        = string
 }
 
+variable "data_files" {
```
tpdownes, line 61: I think this would be much more clearly implemented as a `list(map(object(3 fields))` with the XOR validation of source/content. That said, this is also true of startup-script and it's a list(map) there, too.

At the very least, please cut/paste the logic for XOR. I realize that it is probably caught downstream, but I think it merits being done here.

tpdownes (approved):
Effectively, I believe sources must be absolute until we support a mechanism for staging data into the deployment directory. It's a "Toolkit limit" and not a bar I expect this PR to clear. I like that it also content, at least for text files.

Please make the change I request and merge.

douglasjacobsen (approved):

community/modules/scripts/spack-install/variables.tf
nick-stroud: As mentioned in description, my preference would be to rely on underlying validation. But for the sake of moving forward I will duplicate here as I don't feel strongly.

I suspect `list(map(object(3 fields))` is typo and you meant `list(object(3 fields))`. IIUC this would require user to define one field as null in blueprint (until optionals are available in TF 1.3)

```
data_files:
- source: null
  contents: echo 'hello world'
  destination: /this/that
```

To spare user from ugliness and maintain consistency with startup script I will leave as is.

PR #1497, opened 2023-06-23
DEPRECATE spack-install.environments in favor of using data_files and commands

nick-stroud:
This PR deprecates the `environments` variable and gives examples of how to accomplish the same with `commands`.

This change also fixes issue where commands did not have access to correct python environment where `google-cloud-storage` was installed.

### Submission Checklist


tools/validate_configs/test_configs/spack-environments.yaml
```diff
@@ -60,22 +60,36 @@ deployment_groups:
...
             specs:
               intel-mpi@20184.274 %gcc@10.3.0
-        packages:
-        - gromacs@2021.2 %gcc@10.3.0 ^intel-mpi@2018.4.274
-      - name: 'empty-test'
       gpg_keys:
       - type: 'file'
         path: '/tmp/spack_key.gpg'
       commands: |
+        if ! spack env list | grep -q test-file; then
```
douglasjacobsen, line 72: This is likely intentional, but I wanted to point out that with this pattern you're also ignoring changes in the data file if the environment has been set up once.

douglasjacobsen (approved):

tools/validate_configs/test_configs/spack-environments.yaml
nick-stroud: Yes. This is parity (or better) than what we were doing before except that now it is more visible to user.

I did not see any way to tell spack "make this already created file look like this yaml definition". Another option would be to always remove and create the env but I suspect that would be lots of needless work in most cases.

My personal recommendation would be to use this if to create a named env without a config. Then use spack add to build the environment as spack add is tolerant if the spec is already added. This way it would be idempotent.

nick-stroud:
Force pushed to fix typo in spack command:
`s/spack activate/spack env activate`

PR #1498, opened 2023-06-23
Add `pipefail` to Makefile to prevent swallowing failed tests

mr0re1:
```sh
go test -cover $(ENG) 2>&1 |  perl tools/enforce_coverage.pl
```

Ignored failed tests if `enforce_coverage.pl` was happy.


tpdownes (approved):
I understand that make already mimics "bash -e" so pipefail is enough.

NB: posix sh recently added pipefail but it's not in most Linux distributions yet.

PR #1499, opened 2023-06-23
Remove "failed tests" check from `enforce_coverage` since it doesn't …

mr0re1:
…work

Rely on `pipefail` instead #1498


cboneti (approved):
I assume you tested this extensively with tests failing. If so, approved.

PR #1504, opened 2023-06-25
Allow setting kubernetes labels on node group instances

issacg:
### Submission 

- [x]Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
- [ ] Confirm that "make tests" passes all tests
- [ ] Ensure that unit test coverage remains above 80%
- [x] Follow Cloud HPC Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

This pr allows for k8s labels to be pre-set on the node groups nodes.  Since the labels are (correctly!) ignored as part of the nodegroup lifecycle (so that manipulating node labels via kubectl doesn't cause flapping wrt terraform), this will only work on the initial deploy.  

I still found this to be valuable to keep my IaC up-to-date with changes made in production so it's that much easier to redeploy as needed, and also because (possibly out of ignorance) I don't know of any other way to cause an autoscaling node group to automatically get labels via kubectl.


community/modules/compute/gke-node-pool/variables.tf
```diff
@@ -176,6 +176,15 @@ variable "labels" {
   type        = map(string)
 }
 
+variable "node_labels" {
```
nick-stroud, line 179: My only thought on this is that there could be some user confusion between `labels` and `node_labels`. Due to HPC Toolkit mechanics around resource labels, the `labels` variable name cannot be changed.

Since we have already strayed from the names used by the terraform resource, I am thinking that `kubernetes_labels` or `kubernetes_node_labels` would be a better name as it would make the distinction clearer. I am leaning toward the prior and dropping the word `node` from the name as it possibility seems redundant since this is part of the `node-pool` module.

WDYT?

nick-stroud (changes requested):

community/modules/compute/gke-node-pool/variables.tf
issacg: My first instinct was to use `kubernetes_labels` but I second guessed myself, worrying it was too long as a variable name, so it was highly amusing for me to get this feedback. 😂

FWIW, for me, naming is hard because the idea is for the framework to be opinionated and therefore not just mirror the terraform modules, yet on the other hand when users need to customize, it's much easier for the engineers involved to have similar (or identical) names.

I'll make the change.

nick-stroud (approved):

nick-stroud:
Approved, waiting for tests.

PR #1506, opened 2023-06-26
Add support for Ephemeral Storage Local SSD API

issacg:
### Submission Checklist

If this is acceptable, I plan on also supporting the Local NVMe SSD Block API (for workloads where an emptyDir is not the desired/optimal construct) and adding a README section to briefly go over the differences.   Guidance from a maintainer on whether to do add to this PR or open a second PR is also welcome.


community/modules/compute/gke-node-pool/variables.tf
```diff
@@ -80,6 +80,20 @@ variable "image_type" {
   default     = "COS_CONTAINERD"
 }
 
+variable "local_ssd_count" {
```
nick-stroud, line 83: I would suggest the name `ephemeral_storage_local_ssd_count` to more closely tie it to the terraform API naming (`ephemeral_storage_local_ssd_config`) and to make the distinction clear between this and `local_nvme_ssd_block_config`.

nick-stroud (changes requested):

nick-stroud:
Guidance on Local NVMe SSD Block API:
I would approach exposing `local_nvme_ssd_block_config` in a separate PR. 

I have some follow up questions about your experience with Block SSDs and have opened #1509 if you care to discuss there.

community/modules/compute/gke-node-pool/variables.tf
issacg: Yes, that occurred to me too, as I was reviewing the final patch, but I decided to wait and see if the patch had favorable response before nitpicking on the naming.  I'll make the change.

issacg:
After making the above change, I'm second-guessing the naming.

`local_nvme_ssd_block_count` and `ephemeral_storage_local_ssd_count` will cause these two variable to be distanced from one another in the auto-generated docs, which might add confusion to users who are struggling to understand how to use these two variables.  Part of the problem is the upstream GCP API names, which are then reflected in Terraform, but here I'm suddenly thinking that end-user simplicity  [should trump mirroring the upstream names](https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1504#discussion_r1245025108).

Maybe if we called them `local_ssd_count_ephemeral_storage` and `local_ssd_count_nvme_block` here as a compromise?  WDYT @nick-stroud?

nick-stroud:
Looking at your response to #1509, it seems like a safe assumption that we will support both SSD options. In light of that I think keeping the docs together is a good idea as you suggest. I considered:

```
local_ssd_ephemeral_storage_count
local_ssd_nvme_block_count
```

but I think I prefer your suggestion of:

```
local_ssd_count_ephemeral_storage
local_ssd_count_nvme_block
```

Which reads to me like _Local SSD Count (type)_. Make the update and then I will merge.

nick-stroud (approved):

PR #1510, opened 2023-06-27
Make Windows VM building Toolkit-native

tpdownes:
This PR includes 2 changes:

1. The addition of a `windows-startup-script` modules (community, experimental) whose only function is to output a list of strings that each contain a PowerShell script to be used at boot or image-building time to customize Windows VMs.
  * NB: the `install_gpu_driver.ps1` is a verbatim copy of the (infrequently updated) [official script](https://github.com/GoogleCloudPlatform/compute-gpu-installation/blob/main/windows/install_gpu_driver.ps1). I believe it is best to maintain our own copy internal to the Toolkit and output a script from a local file. We can discuss this more if you object.
2. The modification of the Packer module to accept a list of strings. This modification is occurring before the current implementation of var.powershell_scripts is released, so I'd like to get it done now.
  * Current solution is subject to the problem that paths relative to the deployment directory are not the same as they were relative to the blueprint, so `var.powershell_scripts` is really only useful with absolute paths (generally undesirable). I believe this makes the whole solution more Toolkit native.

Note: I suspect that #1466 may have broken `use` and Packer list variable types. I will investigate and file an appropriate task. This module is written anticipating fixing that.

Example blueprint this enables:

```yaml
- group: primary
  modules:
  - id: network1
    source: modules/network/vpc
    settings:
      enable_iap_rdp_ingress: true
      enable_iap_winrm_ingress: true

  - id: windows_startup
    source: community/modules/scripts/windows-startup-script
    settings:
      install_nvidia_solutions: true

- group: packer
  modules:
  - id: image
    kind: packer
    use:
    - network1
    - windows_startup
    settings:
      source_image_family: windows-2016
      machine_type: n1-standard-8
      accelerator_count: 1
      accelerator_type: nvidia-tesla-t4
      disk_size: 75
      disk_type: pd-ssd
      omit_external_ip: false
      state_timeout: 15m
```

### Submission Checklist

### DIFF
--- a/community/modules/scripts/windows-startup-script/README.md
+++ b/community/modules/scripts/windows-startup-script/README.md
@@ -1,0 +1,104 @@
+## Description
+
+This module contains a set of scripts to be used in customizing Windows VMs at
+boot or during image building. Please note that the installation of NVIDIA GPU
+drivers takes, at minimum, 30-60 minutes. It is therefore recommended to build
+a custom image and reuse it as shown below, rather than install GPU drivers at
+boot time.
+
+> NOTE: the output `windows_startup_ps1` must be passed explicitly as shown
+> below when used with Packer modules. This is due to a limitation in the `use`
+> keyword and inputs of type `list` in Packer modules; this does not impact
+> Terraform modules
+
+### NVIDIA Drivers and CUDA Toolkit
+
+Many Google Cloud VM families include or can have NVIDIA GPUs attached to them.
+This module supports GPU applications by enabling you to easily install
+a compatible release of NVIDIA drivers and of the CUDA Toolkit. The script is
+the [solution recommended by our documentation][docs] and is [directly sourced
+from GitHub][script-src].
+
+[docs]: https://cloud.google.com/compute/docs/gpus/install-drivers-gpu#windows
+[script-src]: https://github.com/GoogleCloudPlatform/compute-gpu-installation/blob/24dac3004360e0696c49560f2da2cd60fcb80107/windows/install_gpu_driver.ps1
+
+```yaml
+- group: primary
+  modules:
+  - id: network1
+    source: modules/network/vpc
+    settings:
+      enable_iap_rdp_ingress: true
+      enable_iap_winrm_ingress: true
+
+  - id: windows_startup
+    source: community/modules/scripts/windows-startup-script
+    settings:
+      install_nvidia_driver: true
+
+- group: packer
+  modules:
+  - id: image
+    source: modules/packer/custom-image
+    kind: packer
+    use:
+    - network1
+    - windows_startup
+    settings:
+      source_image_family: windows-2016
+      machine_type: n1-standard-8
+      accelerator_count: 1
+      accelerator_type: nvidia-tesla-t4
+      disk_size: 75
+      disk_type: pd-ssd
+      omit_external_ip: false
+      state_timeout: 15m
+```
+
+## License
+
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.14.0 |
+
+## Providers
+
+No providers.
+
+## Modules
+
+No modules.
+
+## Resources
+
+No resources.
+
+## Inputs
+
+| Name | Description | Type | Default | Required |
+|------|-------------|------|---------|:--------:|
+| <a name="input_install_nvidia_driver"></a> [install\_nvidia\_driver](#input\_install\_nvidia\_driver) | Generate a PowerShell script that installs NVIDIA GPU drivers and the CUDA Toolkit | `bool` | `false` | no |
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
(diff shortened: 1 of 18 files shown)


community/modules/scripts/windows-startup-script/outputs.tf
nick-stroud, line 1: Description says:
> whose only function is to output a list of strings

This appears to be outputting a single string. Please rectify. If you plan to change interface in future include in description.

community/modules/scripts/windows-startup-script/README.md
```diff
@@ -0,0 +1,90 @@
...
+```yaml
+- group: primary
+  modules:
+  - id: network0
+    source: modules/network/pre-existing-vpc
+
+  - id: windows_startup
+    source: community/modules/scripts/windows-startup-script
+    settings:
+      install_nvidia_solutions: true
```
nick-stroud, line 23: Elsewhere we use `install_nvidia_driver`. The [script](https://github.com/GoogleCloudPlatform/compute-gpu-installation/blob/main/windows/install_gpu_driver.ps1) you reference is also called `install_gpu_driver`. Is there a reason to change terminology?

community/modules/scripts/windows-startup-script/README.md
nick-stroud, line 1: Please reference the source of the script in the documentation.

nick-stroud (changes requested):

community/modules/scripts/windows-startup-script/outputs.tf
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "windows_startup_ps1" {
+  description = "script to load and run all runners, as a string value."
+  value       = local.startup_ps1
```
nick-stroud, line 19: Would it be possible to output a list and then have the blueprint look like:
```
  - id: image
    kind: packer
    use: [network0]
    settings:
      windows_startup_ps1: $(windows_startup.windows_startup_ps1)
```

community/modules/scripts/windows-startup-script/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+> NOTE: the output `windows_startup_ps1` must be passed explicitly as shown
+> below when used with Packer modules. This is due to a limitation in the `use`
+> keyword and inputs of type `list` in Packer modules; this does not impact
+> Terraform modules
+
+```yaml
+- group: primary
+  modules:
+  - id: network0
```
nick-stroud, line 17: Every other example uses `network1`. Please use the same here.

nick-stroud (changes requested):

community/modules/scripts/windows-startup-script/README.md
tpdownes: 1. I do not want to assume that GPU = NVIDIA (even if ubiquitious)
2. In the other cases, the script only installs drivers and I am seeking to convey that this also installs the CUDA Toolkit (whether or not the script filename is entirely right)

community/modules/scripts/windows-startup-script/outputs.tf
tpdownes: With #1513 merged, I can have this do proper "use" syntax.

community/modules/scripts/windows-startup-script/outputs.tf
tpdownes: With #1513 merged, I can revert the change back to what you suggest and employ proper "use" syntax.

community/modules/scripts/windows-startup-script/README.md
tpdownes: addressed

tpdownes:
@mr0re1 with merge of #1513 this PR is improved and sent back to you. I have updated the example blueprint in README and PR description.

modules/packer/custom-image/image.pkr.hcl
```diff
@@ -134,9 +134,9 @@ build {
   # provisioner "powershell" blocks
   dynamic "provisioner" {
     labels   = ["powershell"]
-    for_each = var.powershell_scripts
+    for_each = var.windows_startup_ps1
     content {
-      script = provisioner.value
+      inline = split("\n", provisioner.value)
```
mr0re1, line 139: Should we handle `\n\r` as a "new line" since it's "friendly" module?

mr0re1 (approved):
LGTM. please comment on `\n` vs `\n\r` question.

modules/packer/custom-image/image.pkr.hcl
tpdownes: I think the DOS newline is `\r\n` (reverse). Strictly speaking, the current code ought to handle it.

WDYT of filtering out all `\r` with empty string?

mr0re1: Can splitting be avoided altogether? 
```hcl
inline = [provisioner.value]
```

mr0re1: I'am ok with keeping `\n`, because I'm not sure if there is indeed a problem.

community/modules/scripts/windows-startup-script/README.md
nick-stroud: I don't see relevance of point 1 to the final naming as you are still using nvidia in the name.
Point 2 seems like a reasonable tradeoff. My preference would still be for `install_nvidia_driver` for API consistency and document what you are actually doing. 

I think you are more likely to have a user get an error because they try `install_nvidia_driver` and it does not work. Than you are to have someone complain that cuda was installed. 

All that said, consider this a nit and use your judgement.

nick-stroud (approved):

community/modules/scripts/windows-startup-script/README.md
tpdownes: 1. the point is that I am purposefully using NVIDIA (not "GPU") in the name because it is not the only possible implementation of drivers and software libraries, even for NVIDIA accelerators.
2.  I'll consider `install_nvidia_driver` even though it is doing more than what it is doing elsewhere.

tpdownes: Changed to `install_nvidia_driver` as more in spirit of intent even if not precise in details.

PR #1516, opened 2023-06-27
Make `vm-instance.network_self_link` mandatory

mr0re1:
**Motivation**:
Prevent user from accidentally forgetting to specify network.
E.g.:
```yaml
modules:
  - id: network
    source: modules/network/vpc
  - id: vm
    # doesn't use `network`, uses default instead
```
Should be forced to either:
```yaml
  - id: vm
    use: [network]
```

Or 
```yaml
  - id: vm
    settings:
      network_self_link: "default" # explicitly use default

```


tpdownes:
Offline discussion concluded that we should try a `precondition` that checks `local.network_interfaces` for null network and subnetwork (which would catch combination of var.network_self_link and var.subnetwork_self_link)

PR #1519, opened 2023-06-28
Refactor HTCondor configuration file management

tpdownes:
- Fix HTCondor example job (comments must be on a separate line) and make it more useful for diagnosis (sleep 1000 vs echo "Hello, World!"
- NOP refactoring of existing configuration
  - simplify condor_config.d directory structure such that only 1 (or 2 for SchedD HA) files are created; rest of directory can be manually altered by user
  - render configuration files directly in Terraform templating language and upload as an object
  - refactor Ansible to download object only when necessary

The configuration refactoring anticipates a subsequent PR which will add support for a simple Windows startup script that downloads the object and triggers a reload. This is in lieu of adding full-fledged Ansible support on Windows.

I anticipate the bucket moving to a different module as the revised HTCondor solution is developed. Putting it in htcondor-configure is a step in the process and enables it to be done without modification to blueprint.

### Submission Checklist

### DIFF
--- a/community/examples/htc-htcondor.yaml
+++ b/community/examples/htc-htcondor.yaml
@@ -152,14 +152,15 @@ deployment_groups:
         destination: /var/tmp/helloworld.sub
         content: |
           universe       = vanilla
-          executable     = /bin/echo
-          arguments      = "Hello, World!"
+          executable     = /bin/sleep
+          arguments      = 1000
           output         = out.\$(ClusterId).\$(ProcId)
           error          = err.\$(ClusterId).\$(ProcId)
           log            = log.\$(ClusterId).\$(ProcId)
           request_cpus   = 1
           request_memory = 100MB
-          +RequireSpot   = true # if unset, defaults to false
+          # if unset, defaults to false
+          +RequireSpot   = true
           queue
 
   - id: htcondor_access
--- a/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
+++ b/community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
@@ -20,24 +20,17 @@
     job_queue_ha: false
     spool_dir: /var/lib/condor/spool
     condor_config_root: /etc/condor
-    role_file: 00-role
-    pool_file: 01-pool
-    cm_config_file: 02-central-manager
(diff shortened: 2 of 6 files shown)


community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
```diff
@@ -20,55 +20,53 @@
...
+    failed_when: config_update.rc != 0 and config_update.rc != 137
+    ansible.builtin.shell: |
+      set -e -o pipefail
+      REMOTE_HASH=$(gcloud --format="value(md5_hash)" storage hash {{ config_object }})
+
+      CONFIG_FILE="{{ condor_config_root }}/config.d/{{ ghpc_config_file }}"
+      if [ -f "${CONFIG_FILE}" ]; then
+          LOCAL_HASH=$(gcloud --format="value(md5_hash)" storage hash "${CONFIG_FILE}")
+      else
+          LOCAL_HASH="d41d8cd98f00b204e9800998ecf8427e"
```
nick-stroud, line 60: Why this? Add comment to explain magic hash.

nick-stroud (approved):

community/examples/htc-htcondor.yaml
```diff
@@ -152,14 +152,15 @@ deployment_groups:
         destination: /var/tmp/helloworld.sub
         content: |
           universe       = vanilla
-          executable     = /bin/echo
-          arguments      = "Hello, World!"
+          executable     = /bin/sleep
+          arguments      = 1000
```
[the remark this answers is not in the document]
tpdownes: It is not run during the test (which measures the number of static nodes requested). I agree we'd want something different for a test.

community/modules/scheduler/htcondor-configure/files/htcondor_configure.yml
tpdownes: It's perhaps a bit too clever, but one has to define the hash of an empty string (null terminated, 0 length) and this is it. In principle, if we somehow rendered an empty config file and had no file in place at all then no new (empty) file would be generated. Same in principle, different in detail so I'll modify it to read something entirely invalid (e.g. upper-case characters) and "self documenting."

PR #1531, opened 2023-06-30
Convert spack-install script to ansible

nick-stroud:
This change:
- Swaps out the remainder of the shell install spack script for ansible
- Refactors ramble install script for code reuse between spack and ramble

Note to reviewer:
- First review changes to `community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl`
- Then you can skip review of duplicate `community/modules/scripts/spack-install/templates/spack_setup.yml.tpl`
- After above conversation and ignoring lines from deleted file PR is 139 lines

Testing:
- I have manually tested both spack and ramble deployments
- I tested parallel spack build with 12 VMs all running install in parallel on filestore HDD. 
  - As designed only 1 machine grabbed lock and performed install

### Submission Checklist


community/modules/scripts/ramble-setup/main.tf
```diff
@@ -18,12 +18,15 @@ locals {
...
     "${path.module}/templates/ramble_setup.yml.tpl",
     {
-      install_dir = var.install_dir
-      ramble_url  = var.ramble_url
-      ramble_ref  = var.ramble_ref
-      chown_owner = var.chown_owner == null ? "" : var.chown_owner
-      chgrp_group = var.chgrp_group == null ? "" : var.chgrp_group
-      chmod_mode  = var.chmod_mode == null ? "" : var.chmod_mode
+      app_name              = "ramble"
+      profile_script        = ". {{ install_dir }}/share/ramble/setup-env.sh"
```
tpdownes, line 22: You should mimic the `if [ -f ` test in the Spack solution here. Is there a reason not to?

community/modules/scripts/spack-install/templates/spack_setup.yml.tpl
```diff
@@ -0,0 +1,77 @@
...
+      dest: /etc/profile.d/{{ app_name }}.sh
+      mode: '0644'
+      content: "{{ profile_script }}"
+
+  - name: Create parent of install directory
+    ansible.builtin.file:
+      path: "{{ install_dir | dirname }}"
+      state: directory
+
+  - name: Acquire lock
```
tpdownes, line 43: I think the more natural way of doing this would be something like

```
- name: Create a directory if it does not exist
  ansible.builtin.file:
    register: lockdir
    path: /etc/some_directory
    state: directory
    mode: '0755'
```

(this apporach might also make it easier for it to be a lockfile vs dir)

Then block other things on `when: lockdir.changed`. It avoids setting `failed_when: false` which is a bit of a sledgehammer that can cause you to not see other problems.

I understand you may not want to make progress on other aspects of this functionality and this appears to work. If you wish, create a task to evaluate this alternative approach.

community/modules/scripts/spack-install/templates/spack_setup.yml.tpl
```diff
@@ -0,0 +1,77 @@
...
+      mkdir "{{ install_dir | dirname }}/.install_{{ app_name }}_lock"
+    register: lock_out
+    changed_when: lock_out.rc == 0
+    failed_when: false
+
+  - name: Clones into installation directory
+    ansible.builtin.command: git clone --branch {{ git_ref }} {{ git_url }} {{ install_dir }}
+    when: lock_out.rc == 0
+
+  - name: chgrp on installation
```
tpdownes, line 54: Please avoid the clumsiness here by combining these into a single task using [filters](https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters.html#omitting-parameters); note setting the 2nd parameter to true so that it evaluates upon the Falsey empty string.

https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.default

```
    ansible.builtin.file:
      path: "{{ install_dir }}"
      owner: "{{ chown_owner | default(omit, true) }}"
      group: "{{ chgrp_group | default(omit, true)  }}"
      mode: "{{ chmod_mode | default(omit, true)  }}"
      recurse: true
    when:  lock_out.rc == 0
```

FYI: I believe comparison to empty string would trigger an opt-in linting error.

https://ansible.readthedocs.io/projects/lint/rules/empty-string-compare/

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
   hosts: localhost
   vars:
+    app_name: ${app_name}
+    profile_script: ${profile_script}
     install_dir: ${install_dir}
-    ramble_url: ${ramble_url}
-    ramble_ref: ${ramble_ref}
+    git_url: ${git_url}
+    git_ref: ${git_ref}
     chmod_mode: ${chmod_mode}
```
tpdownes, line 23: I suspect that the empty strings here render as "absolutely nothing". In which case, I'm not sure the comparisons to the empy string below are doing what you think they are doing.

The Ansible way is not to compare to empty string anyhow. See comment below.

tpdownes (changes requested):
- 2 actions requested, 3rd optional
- Additionally, please rename these files `.yml.tftpl` in line with [Terraform extension recommendation](https://developer.hashicorp.com/terraform/language/functions/templatefile)

I am not a fan of using Terraform templating to render Ansible since it requires you to define all variables at top of YAML even when you don't necessarily want to. And I don't trust ansible-lint to parse a Terraform template of a playbook which is, itself, a Jinja2 template. But, I don't think we have a good best practices guide for the team and this is fine until we do.

tpdownes:
> * 2 actions requested, 3rd optional

I realize I put some of the comments on the 2nd file. This is user/UI error. They are meant to apply to both playbooks.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Install Ramble
+- name: Install Application
```
douglasjacobsen, line 15: Could change application -> software here, to make this a bit more general

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
   hosts: localhost
   vars:
+    app_name: ${app_name}
```
douglasjacobsen, line 18: Could change `app_name` --> `sw_name`

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
-    ramble_url: ${ramble_url}
-    ramble_ref: ${ramble_ref}
+    git_url: ${git_url}
+    git_ref: ${git_ref}
     chmod_mode: ${chmod_mode}
     chown_owner: ${chown_owner}
     chgrp_group: ${chgrp_group}
+    finalize_setup_script: ${finalize_setup_script}
   tasks:
+  - name: Print Application Name
```
douglasjacobsen, line 28: Could change `Application` --> `Software`

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
+    ansible.builtin.debug:
+      msg: "Running installation for application: {{app_name}}"
+
+  - name: Add profile script for application
```
douglasjacobsen, line 32: Could change `application` -> `software`

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
+    ansible.builtin.copy:
```
douglasjacobsen, line 33: Could add something like `when: profile_script != ''` but I haven't worked out the right syntax.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
-    register: profile_check
-
-  - name: Add ramble to profile
-    ansible.builtin.copy:
-      dest: /etc/profile.d/ramble.sh
-      content: ". {{ install_dir }}/share/ramble/setup-env.sh"
-    when: not profile_check.stat.exists
+  - name: Finalize Setup
+    ansible.builtin.shell: "{{ finalize_setup_script }}"
+    when: lock_out.rc == 0
```
douglasjacobsen, line 77: Could change when to have `and finalize_setup_script != ''` or something.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
```diff
@@ -12,63 +12,66 @@
...
+
   - name: Create parent of install directory
     ansible.builtin.file:
       path: "{{ install_dir | dirname }}"
       state: directory
 
   - name: Acquire lock
     ansible.builtin.command:
-      mkdir "{{ install_dir | dirname }}/.ramble_lock"
+      mkdir "{{ install_dir | dirname }}/.install_{{ app_name }}_lock"
```
douglasjacobsen, line 45: Could change `app_name` -> `sw_name`

douglasjacobsen (approved):
Thanks for these changes. It's good to see ramble and spack using shared implementations where possible. :)

I made some suggestions, mostly just to make the shared stuff more general, but none of it is strictly needed.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
nick-stroud: Done.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
nick-stroud: Done.

community/modules/scripts/spack-install/templates/spack_setup.yml.tpl
nick-stroud: As discussed offline, I have opened a task and will circle back in separate pr where I can perform dedicated testing.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tpl
nick-stroud: You are more or less correct. It renders as follows:

```
    chmod_mode: a+rwxs
    chown_owner: 
    chgrp_group: 
```

I have tested this with the new comparison that you suggested and it is now working as expected.

nick-stroud:
> not a fan of using Terraform templating to render Ansible

I understand. I have made a point to not render anything into the body of the ansible and require all rendered variables are defined at the top. This allows for this file to be more or less reused in a non-rendered condition. It is written such that we could pass all the vars as args in startup-script but this looks very ugly. Without trying to solve the larger problem of how we pass args I am trying to make the script compatible with future changes.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
```diff
@@ -51,27 +52,15 @@
...
-      recurse: true
-    when: chown_owner != "" and lock_out.rc == 0
-
-  - name: chmod on installation
-    ansible.builtin.file:
-      path: "{{ install_dir }}"
-      mode: "{{ chmod_mode }}"
-      recurse: true
-    when: chmod_mode != "" and lock_out.rc == 0
+    when:  lock_out.rc == 0
```
tpdownes, line 62: there's a bit of extra whitespace here that I think was in my comment. Might as well remove it.

tpdownes (approved):
Minor whitespace nit/potential linting problem. Address and squash (or not) as you see fit, then merge.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
nick-stroud: Done.

PR #1553, opened 2023-07-06
Restrict VM-instance network variables validation

mr0re1:
* Remove default value for `var.network_self_link`;
* Enforce setting exactly one of `network_interfaces` or `network_self_link/subnetwork_self_link`;


modules/compute/vm-instance/main.tf
```diff
@@ -230,5 +230,16 @@ resource "google_compute_instance" "compute_vm" {
     ignore_changes = [
       metadata["ssh-keys"],
     ]
+
+    precondition {
+      condition     = (length(var.network_interfaces) == 0) != (var.network_self_link == null && var.subnetwork_self_link == null)
+      error_message = "Exactly one of network_interfaces or network_self_link/subnetwork_self_link must be specified."
+    }
+
+    precondition {
```
tpdownes, line 239: I do not believe that the 2nd precondition should be included. It breaks the following pattern:

```yaml
deployment_groups:
- group: first
  modules:
  - id: network1
  - id: simplevm
    use:
    - network1
    settings:
      machine_type: e2-medium
    outputs:
    - internal_ip
```

I don't think we should worry about network_self_link/subnetwork_self_link "disagreeing" if they are both supplied.

tpdownes (changes requested):

tpdownes (approved):

PR #1555, opened 2023-07-08
Remove "null by omitting value" from blueprints

mr0re1:
* Remove "null by omitting value" from blueprints;
* Improve error message for wrong YAML type of `Dict`;
* Add tests for YAML null-values.

### DIFF
--- a/community/examples/hpc-slurm-local-ssd.yaml
+++ b/community/examples/hpc-slurm-local-ssd.yaml
@@ -49,7 +49,7 @@ deployment_groups:
         auto_delete: true
         boot: false
       - device_name: test-disk-2
-        disk_name: {}
+        disk_name: null
         disk_size_gb: 375
         disk_type: local-ssd
         disk_labels: {}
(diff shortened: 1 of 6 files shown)


community/examples/hpc-slurm-local-ssd.yaml
```diff
@@ -42,17 +42,17 @@ deployment_groups:
...
+        disk_name: null
         disk_size_gb: 375
         disk_type: local-ssd
-        disk_labels:
+        disk_labels: {}
         auto_delete: true
         boot: false
       - device_name: test-disk-2
-        disk_name:
+        disk_name: {}
```
tpdownes, line 52: This should be `null`. It's causing a problem in PR validation.

tpdownes (approved):
My understanding is that we can't deny "omitted value" at parsing time due to our examples using `project_id: ####` so much. So, you are removing other omitted values from blueprints and adding better "plumbing" to test the 3 different ways a value can be `nil` in Go.

2 minor things to fix.

PR #1559, opened 2023-07-10
NVMe block storage for GKE

issacg:
### Submission Checklist


issacg:
One point I want to make in advance, is that this uses the variable naming requested in #1506, but then I can't pre-validate for mutual-exclusivity between the SSD types which leaves me personally feeling that this PR is somewhat wanting.  

We can go back to the alternate naming (a single variable block, as I mentioned in https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1506#issuecomment-1611253577) or else I'm open to other validation suggestions...

community/modules/compute/gke-node-pool/README.md
```diff
@@ -28,6 +28,33 @@ can be overridden using the `taints` setting. See
 [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/node-taints) for
 more info.
 
+### Local SSD Storage
+GKE offers two options for managing locally attached SSDs.  
+
+The first, and recommended, option is for GKE to manage this storage space
+ephemeral storage space on the node, which will then be automatically
```
nick-stroud, line 35: I think there is a typo here. You have `storage space ephemeral storage space`.

community/modules/compute/gke-node-pool/README.md
```diff
@@ -28,6 +28,33 @@ can be overridden using the `taints` setting. See
...
+
+The second, more complex, option is for GCP to attach these nodes as raw block
+storage. In this case, the cluster administrator is responible for software
+RAID settings, partitioning, formatting and mounting these disks on the host
+OS.  Still, this may be desired behavior in use cases which aren't supported
+by an `emptyDir` volume (for example, a `ReadOnlyMany` or `ReadWriteMany` PV).
+This can be accomplished using the [`local_ssd_count_nvme_block`] variable.
+
+The [`local_ssd_count_ephemeral_storage`] and [`local_ssd_count_nvme_block`]
+variables are mutually exclusive and cannot be mixed together.
```
nick-stroud, line 47: You could use a precondition block to cross validate.

Example precondition block:  
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8fb9539315da1b12a819d4bf7ddaf3c483c1c172/community/modules/compute/gke-node-pool/main.tf#L145-L148

Something like:
```
condition = var.local_ssd_count_ephemeral_storage == 0 || var.local_ssd_count_nvme_block == 0
```

You will also have to bump the required terraform version to 1.2, which is fine:  
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8fb9539315da1b12a819d4bf7ddaf3c483c1c172/community/modules/compute/gke-node-pool/versions.tf#L16

nick-stroud (changes requested):

community/modules/compute/gke-node-pool/README.md
issacg: Actually, you already use a validation block  at https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/8dfbecd552cff36a738f2966758e5af06f0c947d/community/modules/compute/gke-node-pool/variables.tf#L130

Doesn't that need 1.3.0? https://developer.hashicorp.com/terraform/language/v1.2.x/expressions/custom-conditions#input-variable-validation

If so, should I bump to 1.3.0 already? (or open another PR if you want it to be cleaner)?

nick-stroud: This is my understanding:
One can perform single variable validation by adding a validation block to a variable definition. This has been available for a while (since v0.13.0). I suspect that you might be confusing v1.3.0 with v0.13.0. 

[Reference](https://developer.hashicorp.com/terraform/language/v1.2.x/expressions/custom-conditions#input-variable-validation)

> Input variable validation is available in Terraform v0.13.0 and later.

Multi-variable validation can be performed using preconditions which are defined on a resource, data, or output block. This became available in v1.2.0.

[Reference](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions)

> Preconditions and postconditions are available in Terraform v1.2.0 and later.

---

Since this validation is of multiple independent variables, it must be done with a precondition.

v1.2.0 is the minimum requirement. v1.2.x also happens to be the maximum version that HPC Toolkit is currently supporting, so we can bump to v1.2.0 without issue.

---

Finally, It looks like gke-node-pool is already using a precondition so the current stated version of v1.0 is incorrect and should be bumped to v1.2 regardless of the outcome of this PR.

issacg: Oh, whoops you're right.  I misread 0.13.0 as 1.3.0.  Thanks for catching that as I was questioning my own memory of having used single variable validation for a while :)

nick-stroud (approved):

nick-stroud:
Waiting for integration test to pass then will attempt to resolve conflict.

nick-stroud:
PR-validation is consistently and erroneously failing. I have confirmed locally and in another cloned PR that tests pass.

PR #1563, opened 2023-07-11
Enable expressions in `Blueprint.Vars`

mr0re1:
Allow usage of expressions in `Blueprint.Vars` and references to other `vars`.

**NOTE**: Evaluation is happening during "expand", overwritten values (during apply) have no a effect on expressions that **used** them.

```yaml
vars:
  zone: (("${var.region}-c")) # order doesn't matter
  region: us-west4

# expanded
vars:
  region: us-west4
  zone: us-west4-c
```
### Errors:
```yaml
vars:
  region: $(vars.zone)
  zone: (("${var.region}-c"))
...
Error: cyclic dependency detected: "zone" -> "region"
on line 23, column 9:
23:   zone: (("${var.region}-c"))
```

```yaml
vars:
  region: $(foo.zone)
  zone: (("${var.region}-c"))
...
Error: non-global variable "zone" referenced in expression
on line 22, column 11:
22:   region: $(foo.zone)
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
-	"golang.org/x/exp/maps"
 	"gopkg.in/yaml.v3"
 
 	"hpc-toolkit/pkg/modulereader"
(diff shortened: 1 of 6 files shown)


tpdownes (approved):
Offline discussion suggested leaving this "dormant" as a feature for now and introduce it to users with advanced use cases.

PR #1567, opened 2023-07-11
Allow to specify `vm-instance.placement_policy.max_distance`

mr0re1:
* Switch to `google-beta` provider;
* Workaround lack of `optional` in TF 1.2 and desire to make `max_distance` optional,
   change type to `any`;
* Add validation of for "being a map/object", "only valid keys are usde".

### DIFF
--- a/modules/compute/vm-instance/README.md
+++ b/modules/compute/vm-instance/README.md
@@ -185,8 +185,8 @@ limitations under the License.
 | Name | Type |
 |------|------|
 | [google-beta_google_compute_instance.compute_vm](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_compute_instance) | resource |
+| [google-beta_google_compute_resource_policy.placement_policy](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_compute_resource_policy) | resource |
 | [google_compute_disk.boot_disk](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk) | resource |
-| [google_compute_resource_policy.placement_policy](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_resource_policy) | resource |
 | [null_resource.image](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [google_compute_image.compute_image](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_image) | data source |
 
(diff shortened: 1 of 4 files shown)


modules/compute/vm-instance/README.md
nick-stroud, line 1: I believe you need to update the required version of the beta provider in versions.tf. It is currently set to 4.12 which is very old and does not include the max_distance api.

nick-stroud (changes requested):

modules/compute/vm-instance/README.md
mr0re1: IIUC it has to be done as a separate PR.

nick-stroud: I don't think so, I am talking about this line: https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/a952bb7bec8ee89af38dab73d97556afd8765212/modules/compute/vm-instance/versions.tf#L26.

tpdownes (changes requested):
I believe the changes requested make sense:

- bump minimum provider for this module
- improve error message guidance
- improve validation of field with single appropriate value

I am not sufficiently aware of resource policies (or roadmap) to know whether we should extend the validation logic to address known bad combinations:

```
Error: Error creating ResourcePolicy: googleapi: Error 400: Invalid value for field 'resource.groupPlacementPolicy': '{  "vmCount": 2,  "availabilityDomainCount": 2,  "collocation": "COLLOCATED",  "maxDistance": 2}'. A Spread GroupPlacementPolicy should not have vm count set, invalid
```

For now, I would err on the side of letting the API errors bubble up rather than over-constraining the user. It is easier to add validation logic than to take away.

modules/compute/vm-instance/variables.tf
```diff
@@ -278,13 +278,31 @@ variable "bandwidth_tier" {
...
+  validation {
+    condition = alltrue([
+      for k in try(keys(var.placement_policy), []) : contains([
+      "vm_count", "availability_domain_count", "collocation", "max_distance"], k)
+    ])
+    error_message = <<-EOT
+    The supported fields for var.placement_policy are:
+    vm_count, availability_domain_count, collocation, max_distance.
+    EOT
+  }
```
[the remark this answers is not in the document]
nick-stroud: I would exclude validation for collocation. I can confirm it is validated by the terraform resource. Indicating type is still a good idea.

modules/compute/vm-instance/README.md
mr0re1: Thank you for clarification, completely missed it!

tpdownes (approved):

PR #1569, opened 2023-07-12
Output multiple errors instead of first one

mr0re1:
```yaml
// tst.yaml
...
  modules:
  - source: modules/network/vpc
    outputs: [gold, silver]

  - id: nw2
    source: modules/network/vpc
    settings:
      kill.switch: true
      kill@switch: true
      kill_switch: true
```

```shell
$ ./ghpc expand tst.yaml
Error: a module id cannot be empty
34:   - source: modules/network/vpc
        ^

Error: requested output was not found in the module, module:  output: gold
35:     outputs: [gold, silver]
                  ^

Error: requested output was not found in the module, module:  output: silver
35:     outputs: [gold, silver]
                        ^

42:       kill_switch: true
                       ^

Error: a setting name contains a period, which is not supported; variable subfields cannot be set independently in a blueprint.
40:       kill.switch: true
                       ^

Error: a setting name must begin with a non-numeric character and all characters must be either letters, numbers, dashes ('-') or underscores ('_').
41:       kill@switch: true
                       ^
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -127,10 +127,29 @@ func expandOrDie(path string) config.DeploymentConfig {
 	return dc
 }
 
+func findPos(path config.Path, ctx config.YamlCtx) (config.Pos, bool) {
+	pos, ok := ctx.Pos(path)
+	for !ok && path.Parent() != nil {
+		path = path.Parent()
+		pos, ok = ctx.Pos(path)
+	}
+	return pos, ok
+}
+
 func renderError(err error, ctx config.YamlCtx) string {
+	var me config.MultiError
+	if errors.As(err, &me) {
+		var sb strings.Builder
+		for _, e := range me.Errors {
+			sb.WriteString(renderError(e, ctx))
+			sb.WriteString("\n")
+		}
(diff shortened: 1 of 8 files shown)


tpdownes (commented):
This looks great. Would it be significant work to have this include errors such as

```
        nilErr := "deployment variable %s was not set"
```

e.g. for empty `vars.project_id` in our examples.

mr0re1:
This PR is preparatory steps + refactoring of module errors.

Grouping `validateVars` in isolation is rather trivial change.

Grouping all checks in `validateConfig` is tricky because there is a chance of getting a `panic`.
Simplie case: `checkModulesInfo` returned some errors, but proceed with `checkModuleSettings` that invokes `Module.InfoOrDie` -> panic.

To achieve fool grouping of `validateConfig` we need to:
* Get rid of `*OrDie` calls;
* Get rid of side-effects, e.g. `checkModulesAndGroups` sets `Group.Kind`, while it should be a part of `expand` stage.

I would address either in separate PR(s).

PR #1571, opened 2023-07-12
Add wrapper for HTCondor access points with management in Instance Groups

tpdownes:
Refactor existing blueprint using vm-instance and startup-script modules using a wrapper that adds the additional functionality of managing highly available HTCondor access points using a managed instance group.

### Submission Checklist

### DIFF
--- a/community/modules/scheduler/htcondor-access-point/main.tf
+++ b/community/modules/scheduler/htcondor-access-point/main.tf
@@ -1,0 +1,178 @@
+/**
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  # This label allows for billing report tracking based on module.
+  labels = merge(var.labels, { ghpc_module = "htcondor-access-point" })
+}
+
+locals {
+  network_storage_metadata = var.network_storage == null ? {} : { network_storage = jsonencode(var.network_storage) }
+  oslogin_api_values = {
+    "DISABLE" = "FALSE"
+    "ENABLE"  = "TRUE"
+  }
+  enable_oslogin_metadata = var.enable_oslogin == "INHERIT" ? {} : { enable-oslogin = lookup(local.oslogin_api_values, var.enable_oslogin, "") }
+  metadata                = merge(local.network_storage_metadata, local.enable_oslogin_metadata, var.metadata)
+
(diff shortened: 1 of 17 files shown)


community/modules/scheduler/htcondor-access-point/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  oslogin_api_values = {
+    "DISABLE" = "FALSE"
+    "ENABLE"  = "TRUE"
+  }
+  enable_oslogin_metadata = var.enable_oslogin == "INHERIT" ? {} : { enable-oslogin = lookup(local.oslogin_api_values, var.enable_oslogin, "") }
+  metadata                = merge(local.network_storage_metadata, local.enable_oslogin_metadata, var.metadata)
+
+  name_prefix = "${var.deployment_name}-ap"
+
+  example_runner = {
```
mr0re1, line 33: Why does this runner need to be added?

tpdownes: The most immediate explanation is that we cannot combine "use" matching of variables with explicit runners in the blueprint. Since our tutorials expect a helloworld.sub file, I want it to remain and this is a cheap/easy approach.

We could make it optional if it's an immediate issue. Obviously, it might be nice to have a method by which to "append_to_use" or "prepend_to_use" and it could be in the blueprint. But we don't have it now.

community/modules/scheduler/htcondor-access-point/main.tf
```diff
@@ -0,0 +1,178 @@
...
+    EOT
+  }
+
+  all_runners = concat(
+    var.access_point_runner,
+    [local.schedd_runner],
+    var.autoscaler_runner,
+    [local.example_runner]
+  )
+  disk_size_gb = max(var.disk_size_gb, data.google_compute_image.htcondor.disk_size_gb)
```
mr0re1, line 57: Should this be handled by some resource lifecycle condition?
Silently ignoring user settings  doesn't look good
```hcl
disk_size_gb  = coalesce(var.disk_size_gb, data.google_compute_image.htcondor.disk_size_gb)

...
condition = local.disk_size_gb >= data.google_compute_image.htcondor.disk_size_gb
error = "too small"
```

community/modules/scheduler/htcondor-access-point/main.tf
```diff
@@ -0,0 +1,178 @@
...
+
+  ap_config = templatefile("${path.module}/templates/condor_config.tftpl", {
+    htcondor_role       = "get_htcondor_submit",
+    central_manager_ips = var.central_manager_ips
+    spool_dir           = "${var.spool_parent_dir}/spool",
+  })
+
+  ap_object = "gs://${var.htcondor_bucket_name}/${google_storage_bucket_object.ap_config.output_name}"
+  schedd_runner = {
+    "type"        = "ansible-local"
```
mr0re1, line 67: Is there a reason for quoting obj keys? `"type" -> type`

mr0re1 (approved):

community/modules/scheduler/htcondor-access-point/main.tf
tpdownes: Developer error. Addressed in update.

community/modules/scheduler/htcondor-access-point/main.tf
tpdownes: Addressing this with a postcondition and verifying it has intended behavior.

PR #1576, opened 2023-07-12
Deprecate `DeploymentGroup.Kind`

mr0re1:
* Deprecate `DeploymentGroup.Kind`, parse but ignore the yaml field for backward compatibility;
* Add method to derive "kind";
* Add check "deployment group must have at least one module";

**Motivation:**
* The field doesn't improves usability, other than "extra check";
* Doesn't make sense if support is added for groups of mixed modules;
* Bloats the state;
* Set by `checkModulesAndGroups` that should not mutate the state;
* Is not mentioned in our docs/examples.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -84,7 +84,7 @@ func runDeployCmd(cmd *cobra.Command, args []string) {
 		cobra.CheckErr(shell.ImportInputs(groupDir, artifactsDir, expandedBlueprintFile))
 
 		var err error
-		switch group.Kind {
+		switch group.Kind() {
 		case config.PackerKind:
 			// Packer groups are enforced to have length 1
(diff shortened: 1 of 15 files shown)


mr0re1:
@cboneti  for "direction" review, code review can be done by other team members.

cboneti (commented):
If I recall correctly, the only reason we created group kinds was to ensure we could not mix module kinds in a group. It was never meant to be used directly.
As this is no longer necessary, I am generally in favor of deprecating it.  I quickly browsed through the key parts of the code and I like the direction this goes.

Please assign this PR to someone else to do the detailed code review.

tpdownes (approved):
LGTM. I think this leaves room for a `MixedKind` option if we ever elect to support that.

PR #1542, opened 2023-07-03
gvnic not supported by ubuntu_containerd

issacg:
### Submission Checklist

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cloud HPC Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

Removes gvnic when ubuntu_containerd is requested as a VM image

PR #1577, opened 2023-07-12
Insert a basic Ubuntu node pool into the GKE integration test

nick-stroud:
Prior to #1542 Ubuntu node pools did not work. This will prevent basic regressions for Ubuntu.

I expect there may be some controversy about using `echo` to edit the example in the test definition. It is a trade off between 1) making the public example more complex, 2) creating an alternative blueprint and not testing the public example, and 3) the proposed solution, modifying the blueprint on the fly.

### Submission Checklist


tpdownes (approved):
Its job is to catch failures, and it does that.

PR #1578, opened 2023-07-12
Use PR labels to auto-generate release notes

rohitramu:
This PR configures the auto-generation of release notes.  Config file: `.github/release.yml`.

It also adds a GitHub Action workflow to ensure PRs cannot be merged unless they have one of the accepted labels.  Workflow config file: `.github/workflows/pr-label-validation.yml`.  This is the list of accepted labels:

- chore
- key-new-features
- new-modules
- module-improvements
- improvements
- deprecations
- version-updates

I have tested the validation on this PR.

### DIFF
--- a/.github/workflows/pr-label-validation.yml
+++ b/.github/workflows/pr-label-validation.yml
@@ -1,0 +1,48 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-label-validation.yml
nick-stroud, line 1: Is there any way to test this before it gets to main? My concern is that we do release and there turns out there is some issue with this but we cannot just roll back because it is on main and "rollback" would require patch release.

Maybe test in a dummy repo. I will approve but please test before merging. If you feel that you cannot then let's discuss.

nick-stroud, line 1: Please make sure labels are created before merging this PR.

nick-stroud (approved):

.github/workflows/pr-label-validation.yml
rohitramu: Done.

rohitramu: I've tested the validation on this PR by adding/removing labels.  I've also tested the generation of release notes in a dummy repo.

PR #1585, opened 2023-07-14
Implement wrapper for HTCondor central manager

tpdownes:
Refactor existing blueprint using vm-instance and startup-script modules using a wrapper that adds the additional functionality of provisioning highly available HTCondor central manager using a managed instance group.

This PR makes the deliberate decision to rely upon MIGs for high availability rather than the "hot-hot" built-in high availability of HTCondor collector daemon and hot-cold HA of negotiator daemon. Reasoning:

1. The hot-hot solutions means you can query either collector for node informations, but their answers may sometimes differ slightly. They are "eventually consistent". This is fine for job scheduling, but a little less great for autoscaling decisions.
2. A MIG provides similar time recovery time for the negotiator daemon (the thing that matches jobs) as the hot-cold failover mechanism.
3. Using a MIG with stateful IP addresses means we do not know the IP addresses of the central managers until after the MIG is provisioned. This creates a cycle within the Terraform graph for the startup-script of the central managers themselves because you need to know the IP address of the other central manager to render it in the configuration file.
We can eliminate the cycle by configuring 1-and-only-1 central manager which is configured with CONDOR_HOST pointing to its own IP address (which is known by the machine and does not need to be rendered in the script).

### Submission Checklist

### DIFF
--- a/community/modules/scheduler/htcondor-base/README.md
+++ b/community/modules/scheduler/htcondor-base/README.md
@@ -1,12 +1,19 @@
 ## Description
 
-This module performs the following tasks:
+This module creates the basic security infrastructure of an HTCondor pool in
+Google Cloud.
+
+> **_NOTE:_** This module was previously named htcondor-configure. The interface
+> and responsibilities of this module have changed significantly. Please review
+> the [example](#example) and modify your blueprints accordingly.
+
+## Security setup
+
+This module will take the following actions:
 
 - store an HTCondor Pool password in Google Cloud Secret Manager
   - will generate a new password if one is not supplied
 - create service accounts for an HTCondor Access Point and Central Manager
-- create a Toolkit runner for an Access Point
-- create a Toolkit runner for a Central Manager
 
 It is expected to be used with the [htcondor-install] and
 [htcondor-execute-point] modules.
--- a/community/modules/scheduler/htcondor-base/README.md
+++ b/community/modules/scheduler/htcondor-base/README.md
@@ -213,7 +213,6 @@ limitations under the License.
 | Name | Source | Version |
 |------|--------|---------|
 | <a name="module_access_point_service_account"></a> [access\_point\_service\_account](#module\_access\_point\_service\_account) | terraform-google-modules/service-accounts/google | ~> 4.2 |
-| <a name="module_address"></a> [address](#module\_address) | terraform-google-modules/address/google | ~> 3.0 |
 | <a name="module_central_manager_service_account"></a> [central\_manager\_service\_account](#module\_central\_manager\_service\_account) | terraform-google-modules/service-accounts/google | ~> 4.2 |
 | <a name="module_execute_point_service_account"></a> [execute\_point\_service\_account](#module\_execute\_point\_service\_account) | terraform-google-modules/service-accounts/google | ~> 4.2 |
 | <a name="module_health_check_firewall_rule"></a> [health\_check\_firewall\_rule](#module\_health\_check\_firewall\_rule) | terraform-google-modules/network/google//modules/firewall-rules | ~> 6.0 |
@@ -223,7 +222,6 @@ limitations under the License.
 
 | Name | Type |
 |------|------|
-| [google_storage_bucket_object.cm_config](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_object) | resource |
 | [google_compute_subnetwork.htcondor](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_subnetwork) | data source |
 
 ## Inputs
@@ -231,7 +229,6 @@ limitations under the License.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_access_point_roles"></a> [access\_point\_roles](#input\_access\_point\_roles) | Project-wide roles for HTCondor Access Point service account | `list(string)` | <pre>[<br>  "roles/compute.instanceAdmin",<br>  "roles/monitoring.metricWriter",<br>  "roles/logging.logWriter",<br>  "roles/storage.objectViewer"<br>]</pre> | no |
-| <a name="input_central_manager_high_availability"></a> [central\_manager\_high\_availability](#input\_central\_manager\_high\_availability) | Provision HTCondor central manager in high availability mode | `bool` | `false` | no |
 | <a name="input_central_manager_roles"></a> [central\_manager\_roles](#input\_central\_manager\_roles) | Project-wide roles for HTCondor Central Manager service account | `list(string)` | <pre>[<br>  "roles/monitoring.metricWriter",<br>  "roles/logging.logWriter",<br>  "roles/storage.objectViewer"<br>]</pre> | no |
 | <a name="input_deployment_name"></a> [deployment\_name](#input\_deployment\_name) | HPC Toolkit deployment name. HTCondor cloud resource names will include this value. | `string` | n/a | yes |
 | <a name="input_execute_point_roles"></a> [execute\_point\_roles](#input\_execute\_point\_roles) | Project-wide roles for HTCondor Execute Point service account | `list(string)` | <pre>[<br>  "roles/monitoring.metricWriter",<br>  "roles/logging.logWriter",<br>  "roles/storage.objectViewer"<br>]</pre> | no |
@@ -245,8 +242,6 @@ limitations under the License.
 | Name | Description |
 |------|-------------|
 | <a name="output_access_point_service_account_email"></a> [access\_point\_service\_account\_email](#output\_access\_point\_service\_account\_email) | HTCondor Access Point Service Account (e-mail format) |
-| <a name="output_central_manager_ips"></a> [central\_manager\_ips](#output\_central\_manager\_ips) | Reserved internal IP address for use by Central Manager |
-| <a name="output_central_manager_runner"></a> [central\_manager\_runner](#output\_central\_manager\_runner) | Toolkit Runner to configure an HTCondor Central Manager |
 | <a name="output_central_manager_service_account_email"></a> [central\_manager\_service\_account\_email](#output\_central\_manager\_service\_account\_email) | HTCondor Central Manager Service Account (e-mail format) |
 | <a name="output_execute_point_service_account_email"></a> [execute\_point\_service\_account\_email](#output\_execute\_point\_service\_account\_email) | HTCondor Execute Point Service Account (e-mail format) |
 | <a name="output_htcondor_bucket_name"></a> [htcondor\_bucket\_name](#output\_htcondor\_bucket\_name) | Name of the HTCondor configuration bucket |
--- a/community/modules/scheduler/htcondor-base/variables.tf
+++ b/community/modules/scheduler/htcondor-base/variables.tf
@@ -69,9 +69,3 @@ variable "execute_point_roles" {
     "roles/storage.objectViewer",
   ]
 }
-
-variable "central_manager_high_availability" {
-  description = "Provision HTCondor central manager in high availability mode"
-  type        = bool
-  default     = false
-}
--- a/community/modules/scheduler/htcondor-central-manager/README.md
+++ b/community/modules/scheduler/htcondor-central-manager/README.md
@@ -1,0 +1,103 @@
+## Description
+
+This module provisions a highly available HTCondor central manager using a [Managed
+Instance Group (MIG)][mig] with auto-healing.
+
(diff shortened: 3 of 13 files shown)


community/modules/scheduler/htcondor-central-manager/versions.tf
```diff
@@ -0,0 +1,33 @@
...
+    time = {
+      source  = "hashicorp/time"
+      version = "~> 0.9"
+    }
+  }
+  provider_meta "google" {
+    module_name = "blueprints/terraform/hpc-toolkit:htcondor-central-manager/v1.20.0"
+  }
+
+  required_version = ">= 0.13.0"
```
nick-stroud, line 32: nit: I thought we decided on 1.0 as the minimum everywhere. I know we still have stragglers but for new modules I think 1.0 is more appropriate.

community/modules/scheduler/htcondor-central-manager/main.tf
```diff
@@ -0,0 +1,169 @@
...
+    "ENABLE"  = "TRUE"
+  }
+  enable_oslogin_metadata = var.enable_oslogin == "INHERIT" ? {} : { enable-oslogin = lookup(local.oslogin_api_values, var.enable_oslogin, "") }
+  metadata                = merge(local.network_storage_metadata, local.enable_oslogin_metadata, var.metadata)
+
+  name_prefix = "${var.deployment_name}-cm"
+
+  all_runners = flatten([var.central_manager_runner, local.schedd_runner])
+
+  cm_config = templatefile("${path.module}/templates/condor_config.tftpl", {})
```
nick-stroud, line 35: nit: I think the following names are clearer
local.cm_config -> local.cm_config_contents
local.cm_object -> local.cm_object_path

community/modules/scheduler/htcondor-central-manager/main.tf
```diff
@@ -0,0 +1,169 @@
...
+    postcondition {
+      condition     = self.disk_size_gb <= var.disk_size_gb
+      error_message = "var.disk_size_gb must be set to at least the size of the image (${self.disk_size_gb})"
+    }
+  }
+}
+
+data "google_compute_zones" "available" {
+  project = var.project_id
+  region  = var.region
```
nick-stroud, line 67: It seems like there is no option to pick a zone or turn off HA. Does this pose a potential issue for egress?

community/modules/scheduler/htcondor-central-manager/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  region          = var.region
+  labels          = local.labels
+  deployment_name = var.deployment_name
+
+  runners = local.all_runners
+}
+
+module "central_manager_instance_template" {
+  # tflint-ignore: terraform_module_pinned_source
+  source = "github.com/terraform-google-modules/terraform-google-vm//modules/instance_template?ref=84d7959"
```
nick-stroud, line 97: nit: I think I would prefer to use a version ref. Any reason to select a specific commit?

community/modules/scheduler/htcondor-central-manager/main.tf
```diff
@@ -0,0 +1,169 @@
...
+    response            = ""
+    proxy_header        = "NONE"
+    port                = 9618
+    request             = ""
+    request_path        = ""
+    host                = ""
+    enable_logging      = true
+  }
+
+  update_policy = [{
```
nick-stroud, line 144: I think it would be appropriate to document somewhere in layman's terms what the HA characteristics are.

community/modules/scheduler/htcondor-central-manager/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  oslogin_api_values = {
+    "DISABLE" = "FALSE"
+    "ENABLE"  = "TRUE"
+  }
+  enable_oslogin_metadata = var.enable_oslogin == "INHERIT" ? {} : { enable-oslogin = lookup(local.oslogin_api_values, var.enable_oslogin, "") }
+  metadata                = merge(local.network_storage_metadata, local.enable_oslogin_metadata, var.metadata)
+
+  name_prefix = "${var.deployment_name}-cm"
+
+  all_runners = flatten([var.central_manager_runner, local.schedd_runner])
```
nick-stroud, line 33: nit: move this below schedd_runner

community/modules/scheduler/htcondor-base/variables.tf
```diff
@@ -69,9 +69,3 @@ variable "execute_point_roles" {
     "roles/storage.objectViewer",
   ]
 }
-
-variable "central_manager_high_availability" {
```
nick-stroud, line 73: Deprecate instead of delete. Message should mention that using module gives HA by default.

community/modules/scheduler/htcondor-base/README.md
nick-stroud, line 1: nit: The name `htcondor-base` does not convey much to me and the documentation could use a one line that says what the module is (conceptually) as opposed to just what it does. `htcondor-setup` would make more sense to me I thin.

nick-stroud (changes requested):

community/modules/scheduler/htcondor-central-manager/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+}
+
+variable "subnetwork_self_link" {
+  description = "The self link of the subnetwork in which the HTCondor central manager will be created."
+  type        = string
+  default     = null
+}
+
+variable "instance_image" {
+  description = "Custom VM image with HTCondor and Toolkit support installed."
```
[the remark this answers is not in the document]
tpdownes: "and Toolkit support installed" means, more or less, the autoscaler virtual environment has been created. I'll change the text to read "Custom VM image with HTCondor installed using the htcondor-install module". We have the main example, but I'm wary about overloading variable descriptions when we already point to the example in the README.

We could do something like add a label to the image with Packer and enforce its existence using a terraform postcondition. There are organizations that enforce what labels are allowed to exist and this might break in those environments.

community/modules/scheduler/htcondor-base/README.md
tpdownes: I'll consider this as a separate PR before merging to develop.

community/modules/scheduler/htcondor-central-manager/main.tf
tpdownes: Done

community/modules/scheduler/htcondor-central-manager/README.md
```diff
@@ -0,0 +1,103 @@
...
+A reference implementation is included in the Toolkit module
+[htcondor-pool-secrets]. You may substitute implementations so long as they
+duplicate the functionality in the references. Usage is demonstrated in the
+[HTCondor example][htc-example].
+
+[htc-example]: ../../../../examples/README.md#htc-htcondoryaml--
+[htcondor-pool-secrets]: ../htcondor-pool-secrets/README.md
+[IDTOKEN]: https://htcondor.readthedocs.io/en/latest/admin-manual/security.html#introducing-idtokens
+
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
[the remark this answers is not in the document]
tpdownes: Done

community/modules/scheduler/htcondor-central-manager/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+    fs_type               = string,
+    mount_options         = string,
+    client_install_runner = map(string)
+    mount_runner          = map(string)
+  }))
+  default = []
+}
+
+variable "disk_size_gb" {
+  description = "Boot disk size in GB"
```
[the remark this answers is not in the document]
tpdownes: Probably bad. Good catch. I'll change to size of base HPC Rocky 8 image and prevent it from being null. With that, the `postcondition` block is helping us catch bad values.

community/modules/scheduler/htcondor-central-manager/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "project_id" {
+  description = "Project in which HTCondor pool will be created"
```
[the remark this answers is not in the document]
tpdownes: changing to central manager

community/modules/scheduler/htcondor-base/variables.tf
tpdownes: I'd prefer to avoid variable-by-variable deprecations, given that:

- the HTCondor modules are marked experimental
- the htcondor-base module only exists on this branch and is replacing a broader htcondor-configure module on main

I think the most appropriate deprecation-y process here would be to adopt the `checkMovedModules` mechanism with (potentially) expanded guidance beyond "it's the same module with a new name."

community/modules/scheduler/htcondor-central-manager/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+This module provisions a highly available HTCondor central manager using a [Managed
+Instance Group (MIG)][mig] with auto-healing.
+
+[mig]: https://cloud.google.com/compute/docs/instance-groups
+
+## Usage
+
+Although this provisions an HTCondor central manager with standard configuration,
+for a functioning node, you must supply Toolkit runners as described below:
```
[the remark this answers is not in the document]
tpdownes: Addressed

community/modules/scheduler/htcondor-central-manager/versions.tf
tpdownes: I've updated to 1.1 (which is accurate). My general feeling is that the child modules ought to be accurate, although 1.0 has been out 2 years now so it's not unrealistic to impose that full-stop.

community/modules/scheduler/htcondor-central-manager/main.tf
tpdownes: I'd rather keep the naming alignment across modules consistent than choose one here and another there. I'll add a note to my "debt tracker" as a potential mini-PR.

community/modules/scheduler/htcondor-central-manager/main.tf
tpdownes: I will also create a task to add var.zones to the other 2 infrastructure modules of the HTCondor solution.

community/modules/scheduler/htcondor-central-manager/main.tf
tpdownes: PTAL

community/modules/scheduler/htcondor-central-manager/variables.tf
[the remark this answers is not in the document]
nick-stroud: I think this is fine

I am just realizing that you have to build an image. Is this a new restriction. Before (as vm-instance) would we have been able to install via startup-script without building an image.

Either way I am fine with this now that we have deploy command, it just means that the simplest form of the blueprint will always have 3 groups.

community/modules/scheduler/htcondor-central-manager/main.tf
nick-stroud: LGTM

community/modules/scheduler/htcondor-base/variables.tf
nick-stroud: Yes to adding to checkMovedModules.

I will knock var deprecation down to nit but would still advocate for it. Your call.

User flow with old blueprint would be like (var deleted):
- create error: moved module
  - user updates blueprint
- create error: variable does not exist
  - user is confused on next steps??? maybe

User flow with old blueprint would be like (var deprecated):
- create error: moved module
  - user updates blueprint
- deploy error: fyi var has been deprecated, update your blueprint
  - user updates blueprint

nick-stroud (approved):

community/modules/scheduler/htcondor-central-manager/variables.tf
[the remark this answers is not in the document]
tpdownes: I believe one could provide the install runners to the inputs. It is a bad experience, however, due to the significant setup time and makes autoscaling unimpressive. I think we should steer away from this as much as possible.

I think the 3-group pattern is going to be more common with time.

community/modules/scheduler/htcondor-base/variables.tf
tpdownes: Added to checkMovedModules with guidance to read README.md in new location. README.md modified to tell user to read the example and modify blueprint.

PR #1586, opened 2023-07-14
HTCondor access point wrapper patches

tpdownes:
- Fix HTCondor access point playbook to ensure HTCondor is reloaded when configuration updates
- mimic `time_sleep` solution from #1585 to add IP/name outputs directly to module instead of a "here's a command to run"

### Submission Checklist

### DIFF
--- a/community/modules/scheduler/htcondor-access-point/outputs.tf
+++ b/community/modules/scheduler/htcondor-access-point/outputs.tf
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-output "list_instances_command" {
-  description = "Command to list Access Points provisioned by this module"
-  value       = local.list_instances_command
+output "access_point_ips" {
+  description = "IP addresses of the access points provisioned by this module"
+  value       = local.access_point_ips
+}
+
+output "access_point_name" {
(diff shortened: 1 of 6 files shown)


community/modules/scheduler/htcondor-access-point/outputs.tf
```diff
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-output "list_instances_command" {
```
rohitramu, line 17: Should we keep this output (in addition to the 2 new ones) to avoid breaking existing blueprints?

tpdownes: This is landing on a `htcondor-in-flight` branch I'm deliberately using to stage things before getting to `develop`. So it's not even on develop, much less a release.

rohitramu: Ok, makes sense!

rohitramu (approved):

tpdownes:
@rohitramu I observed a problem in the postcondition and fixed it with a new commit. I believe we should remove the list_instances_command output as it only existed on this feature branch for a few days.

PR #1617, opened 2023-07-24
Add MIG ID as output/input in HTCondor modules

tpdownes:
As part of transition to support N>2 instance configurations in HTCondor pools, jobs will need to identify their target by the ID of the managed instance group. This PR:

- adds the necessary output value and input variable
- removes the requirement that jobs specify `+RequireSpot = False` (or True); the "unset" value will now be allowed
  - logic: the MIG selection identifies pricing model and user can accidentally set these in conflict. Let's only require that one be set (MIG ID)
- renders the configuration logic necessary to guide the user toward setting `+RequireId = "MIG-ID"`

### Submission Checklist

### DIFF
--- a/community/modules/scheduler/htcondor-access-point/versions.tf
+++ b/community/modules/scheduler/htcondor-access-point/versions.tf
@@ -29,5 +29,5 @@ terraform {
     module_name = "blueprints/terraform/hpc-toolkit:htcondor-access-point/v1.20.0"
   }
 
-  required_version = ">= 0.13.0"
+  required_version = ">= 1.1"
 }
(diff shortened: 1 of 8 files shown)


community/modules/scheduler/htcondor-access-point/versions.tf
```diff
@@ -29,5 +29,5 @@ terraform {
     module_name = "blueprints/terraform/hpc-toolkit:htcondor-access-point/v1.20.0"
   }
 
-  required_version = ">= 0.13.0"
+  required_version = ">= 1.1"
```
rohitramu, line 32: Just curious what this upgrade was for - it seems like a significant jump.

rohitramu (approved):

community/modules/scheduler/htcondor-access-point/versions.tf
tpdownes: The use of `nullable = false` introduces a requirement for Terraform 1.1 or above.

PR #1606, opened 2023-07-20
Update DDN EXAscaler to 6.2

rohitramu:

PR #1623, opened 2023-07-27
Update DDN-EXAscaler module reference

rohitramu:
I updated the DDN-EXAscaler module reference to v6.2 in PR #1606, however there was a bug which caused validation to fail if `var.cls.node_count` was set to zero.  To get around this, I set the default value to 1.  See the following issue in the `DDNStorage/exascaler-cloud-terraform` repo for more information: https://github.com/DDNStorage/exascaler-cloud-terraform/issues/17

This PR:

- Updates the module reference to point to the commit for tag `v2.1.8`, where the fix was implemented.
- Reverts the change to the default value for `var.cls.node_count`, so that it is set to zero.

### DIFF
--- a/community/modules/file-system/DDN-EXAScaler/main.tf
+++ b/community/modules/file-system/DDN-EXAScaler/main.tf
@@ -42,7 +42,7 @@ locals {
 }
 
 module "ddn_exascaler" {
-  source          = "github.com/DDNStorage/exascaler-cloud-terraform//gcp?ref=e33f439a"
+  source          = "github.com/DDNStorage/exascaler-cloud-terraform//gcp?ref=a3355d50deebe45c0556b45bd599059b7c06988d"
   fsname          = var.fsname
   zone            = var.zone
   project         = var.project_id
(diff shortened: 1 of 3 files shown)


community/modules/file-system/DDN-EXAScaler/main.tf
```diff
@@ -42,7 +42,7 @@ locals {
 }
 
 module "ddn_exascaler" {
-  source          = "github.com/DDNStorage/exascaler-cloud-terraform//gcp?ref=e33f439a"
+  source          = "github.com/DDNStorage/exascaler-cloud-terraform//gcp?ref=a3355d50deebe45c0556b45bd599059b7c06988d"
```
tpdownes, line 45: nit: next time you can use `git rev-parse --short a3355d50deebe45c0556b45bd599059b7c06988d` to get something short but unique

tpdownes (approved):
Please resolve the `nit` without any action and merge.

community/modules/file-system/DDN-EXAScaler/main.tf
rohitramu: That's useful, thanks Tom!

PR #1625, opened 2023-07-27
Make the "Apply Changes" prompt clearer

rohitramu:


pkg/shell/common.go
```diff
@@ -99,7 +99,11 @@ func ApplyChangesChoice(c ProposedChanges) bool {
 	var userResponse string
 
 	for {
-		fmt.Print("Display full proposed changes, Apply proposed changes, Stop and exit, Continue without applying? [d,a,s,c]: ")
+		fmt.Printf(`Please select an option below [d,a,s,c]:
```
mr0re1, line 102: nit. No need for `Printf`, `Print` will do.

mr0re1 (approved):

PR #1626, opened 2023-07-28
Support N>2 groups of HTCondor execute points

tpdownes:
The solution currently only supports 2 groups of execute points because the naming convention supports a fixed name with "spot" optionally added. This commit modifies the approach to require user specification of a group name that is unique to each group of execute points; future work may avoid need for user to explicitly supply this value.

### Submission Checklist

### DIFF
--- a/community/modules/compute/htcondor-execute-point/variables.tf
+++ b/community/modules/compute/htcondor-execute-point/variables.tf
@@ -185,3 +185,13 @@ variable "guest_accelerator" {
     error_message = "The HTCondor module supports 0 or 1 models of accelerator card on each execute point"
   }
 }
+
+variable "name_prefix" {
+  description = "Name prefix given to hostnames in this group of execute points; must be unique across all instances of this module"
+  type        = string
+  nullable    = false
+  validation {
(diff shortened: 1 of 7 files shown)


community/modules/compute/htcondor-execute-point/variables.tf
```diff
@@ -185,3 +185,13 @@ variable "guest_accelerator" {
     error_message = "The HTCondor module supports 0 or 1 models of accelerator card on each execute point"
   }
 }
+
+variable "name_prefix" {
+  description = "Name prefix given to hostnames in this group of execute points; must be unique across all instances of this module"
```
rohitramu, line 190: Just a question, not requesting any action in this PR: is there a way to validate that the variable value is unique across all instances of this module?

rohitramu (approved):

community/modules/compute/htcondor-execute-point/variables.tf
tpdownes: Not in a good way, and also not without adding some functionality to the type of modules that we can write.

PR #1643, opened 2023-08-01
Fix: startup script could only be executed from filesystem root

nick-stroud:
`destpath` will always be a self contained path and does not need a preceding `/`. The `./` after the `&&` seems to have had no purpose and only worked since startup script is executed at the root directory. 

Manually tested:
- `destination: my_script.sh`
- `destination: home/my_script.sh`
- `destination: ./local/my_script.sh`

`vm-instance` runs startup script through a submodule. This was accounted for by swapping the source using an absolute path during testing for the [startup-script metamodule](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/61130586c901dd9467aefea0aad07629ff5e0531/modules/compute/vm-instance/startup_from_network_storage.tf#L58).

### Submission Checklist


jesuslopez2313 (changes requested):
modules/scripts/startup-script/templates/startup-script-custom.tpl

PR #1644, opened 2023-08-01
Star ccm update to include Slurm

jrossthomson:
Added a StarCCM on Slurm blueprint


jrossthomson:
@cboneti 

Nick is out of town. Who can I get to review this?

community/examples/tutorial-starccm-slurm.yaml
```diff
@@ -0,0 +1,124 @@
+# Copyright 2022 Google LLC
```
shakeebshams, line 1:
```suggestion
# Copyright 2023 Google LLC
```

community/examples/tutorial-starccm-slurm.yaml
```diff
@@ -0,0 +1,124 @@
...
+    use: [network1]
+    settings:
+      local_mount: /home
+
+  - id: login-script
+    source: modules/scripts/startup-script
+    kind: terraform
+    settings:
+      runners:
+      - type: shell
```
cboneti, line 48: isn't this similar to setting "configure_ssh_host_patterns" to "*", wouldn't it be better set to "star*"? [configure_ssh_host_patterns](https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/modules/scripts/startup-script#input_configure_ssh_host_patterns)

community/examples/tutorial-starccm-slurm.yaml
```diff
@@ -0,0 +1,124 @@
...
+        content: |
+          #!/bin/bash
+          google_mpi_tuning --hpcthroughput
+          google_mpi_tuning --nomitigation
+          yum -y install libExt libXext.x86_64 nmap
```
cboneti, line 68: Would you be able to add a comment describing why these libraries are needed?

cboneti (changes requested):
I am wondering if we should keep the old example (without slurm), thoughts? 
Also, PR validation is failing with a small error on the readme.md.

mr0re1:
PR-validation fails, you can find following in the step logs:
```
Step #2 - "pre-commit-run": examples/README.md:866:1: MD012: Multiple consecutive blank lines [Expected: 1, Actual: 2] (no-multiple-blanks)
```

community/examples/tutorial-starccm-slurm.yaml
jrossthomson: Is there a correct way to do this? I am getting prompted to accept keys when I run with `configure_ssh_host_patterns` set to [""]. 
```
 ssh starccmslu-debug-ghpc-1
The authenticity of host 'starccmslu-debug-ghpc-1 (10.0.0.5)' can't be established.
ECDSA key fingerprint is SHA256:ee/7Vve7r85pf6UyzXar41ToSkDviAZ50r5tef8XbBE.
ECDSA key fingerprint is MD5:04:0f:d7:19:67:7c:e6:e7:4c:f3:a1:57:2d:18:2b:e4.
Are you sure you want to continue connecting (yes/no)? 
```
@nick-stroud was the one who suggested the script as written based on a customer who was similarly stuck.

community/examples/tutorial-starccm-slurm.yaml
jrossthomson: It looks like the current version of the Slurm image works without the install. I will remove.

community/examples/tutorial-starccm-slurm.yaml
jrossthomson: That took multiple tries, but I get it now. Seems to work.

jrossthomson:
@cboneti I think I have all the corrections in place. Thanks!

community/examples/tutorial-starccm-slurm.yaml
cboneti: thanks, I should have been specific and said: configure_ssh_host_patterns: ["star*"], or something like that. Did you update the yaml?

cboneti:
Test fails with: 
failed to parse the blueprint in /tmp/tmp.dw7nFWZzGS/tutorial-starccm-slurm.yaml, check YAML syntax for errors, err=yaml: line 46: mapping values are not allowed in this context

community/examples/tutorial-starccm-slurm.yaml
jrossthomson: The key I was missing is that the script where your comment was is only used by the login node. I needed to create one to use with the compute nodes. A learning event for me!

PR #1646, opened 2023-08-02
Change description of instance image variable in CRD module to use ubuntu 2204 as default

harshthakkar01:


cboneti (approved):
Thanks for this. 
One suggestion is that in the future you add a little description that goes beyond the title, perhaps explaining why this was needed.

cboneti:
oh, and the test for labels failed, you probably need to add a label. :)

PR #1669, opened 2023-08-08
Add logic to retrieve current resource usage

mr0re1:


pkg/validators/quota.go
```diff
@@ -216,3 +218,56 @@ func serviceLimits(consumer string, service string) ([]*sub.ConsumerQuotaMetric,
...
+	s, err := cm.NewService(context.Background())
+	if err != nil {
+		return usageProvider{}, err
+	}
+
+	u := map[usageKey]int64{}
+	err = s.Projects.TimeSeries.List("projects/"+project_id).
+		Filter(`metric.type="serviceruntime.googleapis.com/quota/allocation/usage" resource.type="consumer_quota"`).
+		IntervalEndTime(time.Now().Format(time.RFC3339)).
+		// Quota usage metrics get duplicated once a day
```
rohitramu, line 255: Is this only true for metrics that haven't changed in 24 hours?  Or will we only be able to retrieve data that's up to 24 hours old?

rohitramu (approved):

pkg/validators/quota.go
mr0re1: If metric wasn't updated with 24 hours, it will be duplicated.
It is to prevent metrics that aren't update frequently to fall of the horizon.

PR #1675, opened 2023-08-10
Add check for invalid module id

harshthakkar01:
This adds a check in module validator to check and return error if the module id is set to invalid IDs such as `vars`.

### DIFF
--- a/pkg/config/validate.go
+++ b/pkg/config/validate.go
@@ -175,6 +175,9 @@ func validateModule(p modulePath, m Module, bp Blueprint) error {
 	if m.ID == "" {
 		errs.At(p.ID, fmt.Errorf(errorMessages["emptyID"]))
 	}
+	if m.ID == "vars" { // invalid module ID
+		errs.At(p.ID, errors.New("module id cannot be 'vars'"))
+	}
 	return errs.
 		Add(validateSettings(p, m, info)).
 		Add(validateOutputs(p, m, info)).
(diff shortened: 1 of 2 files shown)


pkg/config/validate.go
```diff
@@ -175,6 +179,9 @@ func validateModule(p modulePath, m Module, bp Blueprint) error {
 	if m.ID == "" {
 		errs.At(p.ID, fmt.Errorf(errorMessages["emptyID"]))
 	}
+	if invalidModuleID[m.ID] {
+		errs.At(p.ID, fmt.Errorf(errorMessages["invalidID"]))
```
mr0re1, line 183: * let's keep it simple `if m.ID == "vars"`, no need for set;
* there is no formatting happening , use `errors.New` instead of `fmt.Errorf`;
* `errorMessages` was a bad design decision made in the past, please don't use it, string literal will suffice;
* The error message can be more descriptive, e.g. `can not use "vars" as a module id`

harshthakkar01: Updated.

pkg/config/validate.go
```diff
@@ -175,6 +175,9 @@ func validateModule(p modulePath, m Module, bp Blueprint) error {
 	if m.ID == "" {
 		errs.At(p.ID, fmt.Errorf(errorMessages["emptyID"]))
 	}
+	if m.ID == "vars" { // invalid module ID
+		errs.At(p.ID, errors.New("can not use vars as module id"))
```
cboneti, line 179: perhaps phrase this error a little better?  For instance, what about:
"module id cannot be 'vars'".

harshthakkar01: Done.

mr0re1 (approved):

PR #1678, opened 2023-08-14
Add hint to error message `invalid module reference: var`

cdunbar13:
Made changes to include hints for the 'invalidMod' errors specifically in validateModuleSettingReference function.  The code will check the levenshtein distance of the reference and all possible modules it could legally reference.  If the distance is within 40% of the length of the module (for normalization), it will suggest that module.  If not, it will suggest vars for global variables.  

This also introduces a new error type HintError.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/agext/levenshtein"
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
 	"gopkg.in/yaml.v3"
@@ -129,7 +130,21 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
 		return nil
 	})
 	if mod == nil {
-		return nil, fmt.Errorf("%s: %s", errorMessages["invalidMod"], id)
+		err := InvalidModuleError{string(id)}
+		clMod := ""
+		minDist := -1.0
+		bp.WalkModules(func(m *Module) error {
+			dist := float64(levenshtein.Distance(string(m.ID), string(id), nil)) / float64(len(string(m.ID)))
+			if minDist == -1.0 || dist < minDist {
+				minDist = dist
+				clMod = string(m.ID)
+			}
+			return nil
+		})
+		if clMod != "" && minDist < 0.6 {
+			return nil, HintError{clMod, err}
+		}
+		return nil, err
 	}
 	return mod, nil
 }
@@ -143,7 +158,7 @@ func (bp Blueprint) ModuleGroup(mod ModuleID) (DeploymentGroup, error) {
 			}
 		}
 	}
-	return DeploymentGroup{}, fmt.Errorf("%s: %s", errorMessages["invalidMod"], mod)
+	return DeploymentGroup{}, InvalidModuleError{string(mod)}
 }
 
 // ModuleGroupOrDie returns the group containing the module; panics if unfound
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -50,7 +50,6 @@ var errorMessages = map[string]string{
 	"missingSetting":    "a required setting is missing from a module",
 	"settingsLabelType": "labels in module settings are not a map",
 	"invalidVar":        "invalid variable definition in",
-	"invalidMod":        "invalid module reference",
 	"varNotFound":       "Could not find source of variable",
 	"intergroupOrder":   "References to outputs from other groups must be to earlier groups",
 	"noOutput":          "Output not found for a variable",
@@ -130,7 +129,7 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
 		return nil
 	})
 	if mod == nil {
-		err := InvalidModuleError{string(id)}
+		err := InvalidModuleError{id}
 		clMod := ""
 		minDist := -1.0
 		bp.WalkModules(func(m *Module) error {
@@ -158,7 +157,7 @@ func (bp Blueprint) ModuleGroup(mod ModuleID) (DeploymentGroup, error) {
 			}
 		}
 	}
-	return DeploymentGroup{}, InvalidModuleError{string(mod)}
+	return DeploymentGroup{}, InvalidModuleError{mod}
 }
 
 // ModuleGroupOrDie returns the group containing the module; panics if unfound
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -137,7 +137,7 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
 
 // SuggestModuleIDHint return a correct spelling of given ModuleID id if one
 // is close enough (based on maxHintDist)
-func (bp *Blueprint) SuggestModuleIDHint(id ModuleID) (string, bool) {
+func (bp Blueprint) SuggestModuleIDHint(id ModuleID) (string, bool) {
 	clMod := ""
 	minDist := -1
 	bp.WalkModules(func(m *Module) error {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -36,6 +36,7 @@ const (
 	expectedVarFormat        string = "$(vars.var_name) or $(module_id.output_name)"
 	expectedModFormat        string = "$(module_id) or $(group_id.module_id)"
 	unexpectedConnectionKind string = "connectionKind must be useConnection or deploymentConnection"
+	maxHintDist              int    = 3 // Maximum levenshtein distance where we suggest a hint
 )
 
 var errorMessages = map[string]string{
@@ -129,23 +130,30 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
 		return nil
 	})
 	if mod == nil {
-		err := InvalidModuleError{id}
-		clMod := ""
-		minDist := -1.0
-		bp.WalkModules(func(m *Module) error {
-			dist := float64(levenshtein.Distance(string(m.ID), string(id), nil)) / float64(len(string(m.ID)))
-			if minDist == -1.0 || dist < minDist {
-				minDist = dist
-				clMod = string(m.ID)
-			}
-			return nil
-		})
-		if clMod != "" && minDist < 0.6 {
-			return nil, HintError{clMod, err}
-		}
-		return nil, err
+		return nil, UnknownModuleError{id}
 	}
 	return mod, nil
+}
+
+// SuggestModuleIDHint return a correct spelling of given ModuleID id if one
+// is close enough (based on maxHintDist)
+func (bp *Blueprint) SuggestModuleIDHint(id ModuleID) (string, bool) {
+	clMod := ""
+	minDist := -1
+	bp.WalkModules(func(m *Module) error {
+		dist := levenshtein.Distance(string(m.ID), string(id), nil)
+		if minDist == -1.0 || dist < minDist {
+			minDist = dist
+			clMod = string(m.ID)
+		}
+		return nil
+	})
+
+	if clMod != "" && minDist <= maxHintDist {
+		return clMod, true
+	}
+
+	return "", false
 }
 
 // ModuleGroup returns the group containing the module
@@ -157,7 +165,7 @@ func (bp Blueprint) ModuleGroup(mod ModuleID) (DeploymentGroup, error) {
 			}
 		}
 	}
-	return DeploymentGroup{}, InvalidModuleError{mod}
+	return DeploymentGroup{}, UnknownModuleError{mod}
 }
 
 // ModuleGroupOrDie returns the group containing the module; panics if unfound
--- a/pkg/config/errors.go
+++ b/pkg/config/errors.go
@@ -33,6 +33,23 @@ func (e BpError) Unwrap() error {
 	return e.Err
 }
 
+// HintError is an error wrapper to suggest other variable names
+type HintError struct {
+	varName string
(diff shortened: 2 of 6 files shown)


pkg/config/errors.go
```diff
@@ -33,6 +33,23 @@ func (e BpError) Unwrap() error {
 	return e.Err
 }
 
+// HintError is an error wrapper to suggest other variable names
+type HintError struct {
```
mr0re1, line 37: Let's make `HintError` a general hint wrapper, not just for this particular type of error:
```go
typ HintError struct {
  Hint string
  Err error
}
```

cdunbar13: Should we change the string created as well?  (i.e. Should the Hint passed in be the new error message to be concatenated to the end of the current error?) (e.g. Hint = "Do you mean vars.test?"). Alternatively we could call it VarHintError to make it more specific and keep the "Do you mean %s" in the error.

pkg/config/expand.go
```diff
@@ -545,3 +549,43 @@ func intersection(s1 []string, s2 []string) []string {
...
 	return is
 }
+
+// Find the closest module that the reference could be referring to (only ones that are defined before reference)
+// Returns a string with the closest module name (default vars)
+// Currently limited to finding strings that are at least 40% similar based on module string length and levenshtein distance
+func closestReference(bp *Blueprint, mod *Module, ref *Reference) string {
+	minDist := 0.0
+	var clMod Module
+	newRef := "vars"
```
mr0re1, line 559: `"vars"` is not a reference.

pkg/config/expand.go
```diff
@@ -545,3 +549,43 @@ func intersection(s1 []string, s2 []string) []string {
 	slices.Sort(is)
 	return is
 }
+
+// Find the closest module that the reference could be referring to (only ones that are defined before reference)
+// Returns a string with the closest module name (default vars)
+// Currently limited to finding strings that are at least 40% similar based on module string length and levenshtein distance
+func closestReference(bp *Blueprint, mod *Module, ref *Reference) string {
```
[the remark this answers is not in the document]
cdunbar13: If we don't need the references from preceding group, then this logic should all go in config.go:

```
// Module return the module with the given ID
func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
	var mod *Module
	bp.WalkModules(func(m *Module) error {
		if m.ID == id {
			mod = m
		}
		return nil
	})
	if mod == nil {
		return nil, InvalidModuleError{string(id)}
	}
	return mod, nil
}
```

and in ModuleGroup, where the invalidMod is used originally.

pkg/config/expand.go
cdunbar13: I'll change it to 'hint'

pkg/config/expand.go
```diff
@@ -545,3 +549,43 @@ func intersection(s1 []string, s2 []string) []string {
...
+	minDist := 0.0
+	var clMod Module
+	newRef := "vars"
+
+	if bp.Vars.Has(string(ref.Name)) {
```
mr0re1, line 561: `ref.Name` is already a string.

pkg/config/expand.go
mr0re1, line 556: This function doesn't look for reference but rather closes module_id(+ "vars"), let's name it appropriately, e.g. "hint module id fix"

pkg/config/expand.go
cdunbar13: Removed

pkg/config/expand.go
cdunbar13: Removed

pkg/config/expand.go
```diff
@@ -339,6 +340,9 @@ func validateModuleSettingReference(bp Blueprint, mod Module, r Reference) error
 	}
 
 	if err := validateModuleReference(bp, mod, r.Module); err != nil {
+		if strings.Contains(err.Error(), errorMessages["invalidMod"]) {
```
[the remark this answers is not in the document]
cdunbar13: This was addressed.

pkg/config/errors.go
cdunbar13: Addressed.

pkg/config/errors.go
```diff
@@ -43,6 +60,15 @@ func (err *InvalidSettingError) Error() string {
 	return fmt.Sprintf("invalid setting provided to a module, cause: %v", err.cause)
 }
 
+// InvalidModuleError signifies a problem with the supplied module name.
+type InvalidModuleError struct {
+	modID string
```
mr0re1, line 65: Make it exported, `ModuleID string`, or even `ModuleID ModuleID`

pkg/config/errors.go
```diff
@@ -43,6 +60,15 @@ func (err *InvalidSettingError) Error() string {
...
+}
+
+func (e InvalidModuleError) Error() string {
+	return fmt.Sprintf("%s: %s", errorMessages["invalidMod"], e.modID)
```
mr0re1, line 69: No, need for `errorMessages["invalidMod"]`, just use string literal here, and remove it from `errorMessages`.

pkg/config/expand.go
```diff
@@ -339,6 +339,9 @@ func validateModuleSettingReference(bp Blueprint, mod Module, r Reference) error
 	}
 
 	if err := validateModuleReference(bp, mod, r.Module); err != nil {
+		if _, ok := err.(InvalidModuleError); ok && bp.Vars.Has(r.Name) {
```
mr0re1, line 342: * IIUC. `err` will be of type `HintError` instead of `InvalidModuleError` if you implement it in `bp.Module()`, so you need to use `errors.Is` instead of casting.
*  The chance of having the same name for global vars and module outputs is hight (> 50%). It may not be reasonable to always suggest "vars" here. 
* Avoid double-wrapping.

I would advice to not do hinting in `Blueprint.Module`.

pkg/config/expand_test.go
```diff
@@ -230,7 +230,7 @@ func (s *MySuite) TestApplyUseModules(c *C) {
 
 		// Use ID doesn't exists (fail)
 		g.Modules[len(g.Modules)-1].ID = "wrongID"
-		c.Assert(dc.applyUseModules(), ErrorMatches, fmt.Sprintf("%s: %s", errorMessages["invalidMod"], used.ID))
+		c.Assert(dc.applyUseModules(), ErrorMatches, fmt.Sprintf("%s: %s - Did you mean '%s'\\?", errorMessages["invalidMod"], used.ID, using.ID))
```
mr0re1, line 233: Let's check.
```
err.Hint == "expected_hint"
errrs.Is(err, InvalidModuleError)
```

pkg/config/expand.go
cdunbar13: > * IIUC. `err` will be of type `HintError` instead of `InvalidModuleError` if you implement it in `bp.Module()`, so you need to use `errors.Is` instead of casting.

No, I don't think that's what it does.  If there isn't a close enough match, an InvalidModuleError is returned.  Either way I can switch out the type check for an error.Is function.

> * The chance of having the same name for global vars and module outputs is hight (> 50%). It may not be reasonable to always suggest "vars" here.

I don't think I always suggest it.  The logic is that if the error is an InvalidModuleError, that means no other module name was close enough to warrant a HintError and so nothing has been suggested yet.  As a result then we check if the reference name is within the pool of global variables, and if it is, then it is hinted.  Otherwise, nothing is suggested.

> * Avoid double-wrapping.

Is there no value in maintaining the error type from below?  We might want to check what kind of error is below the Hint, or just remove the HintError type altogether and just have hints as optional portions of the InvalidModuleError.  Alternatively, we could make HintError as part of the InvalidModuleError for inheritance, that way other things we want to give suggestions on could follow suit.

> 
> I would advice to not do hinting in `Blueprint.Module`.

Is there a reason why, given my explanations above?

pkg/config/errors.go
cdunbar13: Done

pkg/config/errors.go
cdunbar13: Done

pkg/config/expand_test.go
cdunbar13: I think I took care of this in a way that makes sense.  Please review.

pkg/config/config.go
```diff
@@ -129,7 +129,21 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
...
+		minDist := -1.0
+		bp.WalkModules(func(m *Module) error {
+			dist := float64(levenshtein.Distance(string(m.ID), string(id), nil)) / float64(len(string(m.ID)))
+			if minDist == -1.0 || dist < minDist {
+				minDist = dist
+				clMod = string(m.ID)
+			}
+			return nil
+		})
+		if clMod != "" && minDist < 0.6 {
```
mr0re1, line 143: Let's not use ratio, but absolute distance (say 3).
E.g. if `m.ID` the ration will be always `>= 1.0`, unless it's a match.

pkg/config/config.go
```diff
@@ -129,7 +129,21 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
 		return nil
 	})
 	if mod == nil {
-		return nil, fmt.Errorf("%s: %s", errorMessages["invalidMod"], id)
+		err := InvalidModuleError{id}
+		clMod := ""
```
mr0re1, line 133: Let's move it to separate function:
`suggestModuleIdFix(s string) (string, bool)`

pkg/config/errors.go
```diff
@@ -33,6 +33,23 @@ func (e BpError) Unwrap() error {
...
 }
 
+// HintError wraps another error to suggest other values
+type HintError struct {
+	Hint string
+	Err  error
+}
+
+func (e HintError) Error() string {
+	if len(e.Hint) > 0 {
```
mr0re1, line 43: Let's not do string substitution here. 
Instead of `HintError{"suggested_moduleid", errr}`, do `HintError{'Did you mean "suggested_moduleid"', errr}`

pkg/config/errors.go
```diff
@@ -43,6 +60,15 @@ func (err *InvalidSettingError) Error() string {
 	return fmt.Sprintf("invalid setting provided to a module, cause: %v", err.cause)
 }
 
+// InvalidModuleError signifies a problem with the supplied module name.
+type InvalidModuleError struct {
+	modID ModuleID
```
mr0re1, line 65: This field needs to be exported. Start with capital letter.
Let's keep it short
```go
type InvalidModuleError struct {
  ID ModuleID
}
```

pkg/config/errors.go
```diff
@@ -43,6 +60,15 @@ func (err *InvalidSettingError) Error() string {
...
+}
+
+func (e InvalidModuleError) Error() string {
+	return fmt.Sprintf("invalid module reference: \"%s\"", e.modID)
```
mr0re1, line 69: `s/reference/id/`

pkg/config/expand_test.go
```diff
@@ -230,7 +231,10 @@ func (s *MySuite) TestApplyUseModules(c *C) {
 
 		// Use ID doesn't exists (fail)
 		g.Modules[len(g.Modules)-1].ID = "wrongID"
-		c.Assert(dc.applyUseModules(), ErrorMatches, fmt.Sprintf("%s: %s", errorMessages["invalidMod"], used.ID))
+		err := dc.applyUseModules()
+		invModErr := InvalidModuleError{used.ID}
+		c.Assert(errors.Is(err, invModErr), Equals, true)
```
mr0re1, line 236: `s/Assert/Check/`

pkg/config/expand.go
```diff
@@ -339,6 +339,9 @@ func validateModuleSettingReference(bp Blueprint, mod Module, r Reference) error
...
+			err = HintError{fmt.Sprintf("vars.%s", r.Name), err}
+		}
```
mr0re1, line 344: Please add test case for `validateModuleSettingReference`

pkg/config/config.go
```diff
@@ -129,11 +130,32 @@ func (bp *Blueprint) Module(id ModuleID) (*Module, error) {
...
 	if mod == nil {
-		return nil, fmt.Errorf("%s: %s", errorMessages["invalidMod"], id)
+		return nil, UnknownModuleError{id}
 	}
 	return mod, nil
 }
 
+// SuggestModuleIDHint return a correct spelling of given ModuleID id if one
+// is close enough (based on maxHintDist)
+func (bp *Blueprint) SuggestModuleIDHint(id ModuleID) (string, bool) {
```
mr0re1, line 140: No need to take a pointer, take value.

cdunbar13: Done

pkg/config/errors.go
cdunbar13: Done

pkg/config/expand.go
cdunbar13: Done.

pkg/config/expand_test.go
cdunbar13: Done.

mr0re1 (approved):

pkg/config/config.go
cdunbar13: Done

pkg/config/expand.go
cdunbar13: Done.

pkg/config/config.go
cdunbar13: Done.Done.

PR #1682, opened 2023-08-15
Add Shielded VM support to Packer module

tpdownes:
Partially address #1674 

### Submission Checklist


modules/packer/custom-image/variables.pkr.hcl
```diff
@@ -242,3 +242,24 @@ EOD
   type        = list(string)
   default     = null
 }
+
+variable "enable_shielded_vm" {
```
mr0re1, line 246: nit. LGTM, but I would like to bring a few alternative options for you to consider:

1. Not sure if it's technically possible in Packer. Could we add validation that if user touches `shielded_instance_config` then  `enable_shielded_vm ` should be `true`
```py
not enable_shielded_vm  =>  shielded_instance_config.values == [true, true, true]
```

2. Have "tricky default value" for `shielded_instance_config` - if not set all false, otherwise default omitted keys to true.
```hcl
variable "shielded_instance_config" {
  type=map(bool)
  default=null 
  # + keys strings and value types validation
}
...
locals {
   enable_secure_boot  = shielded_instance_config == null ? 
     false  : 
     try(shielded_instance_config["enable_secure_boot"], true) 
  # same for enable_vtpm  and enable_integrity_monitoring 
}
```

**P.S.:** after writing it down, your option looks better than (2) )

mr0re1 (approved):

modules/packer/custom-image/variables.pkr.hcl
tpdownes: I'm choosing to replicating a pattern in active use at:

- https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/community/modules/scheduler/schedmd-slurm-gcp-v5-login/variables.tf#L231
- Also: Slurm v5 controller, node-group modules in Toolkit
- https://github.com/terraform-google-modules/terraform-google-vm/blob/master/modules/instance_template/variables.tf#L239

A nice property of this solution is `ghpc create --vars enable_shielded_vm=true blueprint.yaml` ought to globally enable all Shielded VM features. If I were writing my own modules, I would probably eliminate `enable_shielded_vm` in favor of `shielded_instance_config` object with defaults of false.

PR #1690, opened 2023-08-16
Refactor `cfg/validators` package

mr0re1:
* Change direction of dependency `config <-> validators`;
* Do not persist "default" validators in the expanded blueprint;
* Move control of validators execution form `config` to `cmd`;
* Remove stdout prints from `validators` and `config` packages;
* Remove validator inputs evaluation from specific validators, make it common;
* Utilize multi-error for validators;
* Simplify implementation of some validators;
* Remove `validatorName` type;
* Rename `validatorConfig -> Validator`;
* Put identifiers into quotes to improve error messages readability.

**Motivation:**
* The dependency `config -> validators` needed refactoring, the blueprint validator should be able to inspect the blueprint, instead of being a collection of GCP-specific functions + pretty renderer of logs;
* Finalize the effort of making core-packages stdout-free;
* Opens an opportunities for further refactoring;
* Simplifies addition of new validators, namely quota validation.

**Outputs comparison:**
**develop**:
```
arbuz is not an implemented validator
a required input project_id was not provided to test_region_exists!
warning: at least one required input was not provided to test_region_exists

module debug_node_group uses module network1, but matching setting and outputs were not found. This may be because the value is set explicitly or
One or more used modules could not have their settings and outputs linked.
warning: validator test_module_not_used failed

the deployment variable "useless" was not used in this blueprint
one or more deployment variables was not used by any modules
warning: validator test_deployment_variable_not_used failed

zone us-central1-a is not in region us-west1 in project ID io-playground or your credentials do not have permissions to access it
warning: validator test_zone_in_region failed

One or more blueprint validators has failed ... <same>
```

**PR:**
```
Error: unknown validator "arbuz"
27:   - validator: arbuz
                   ^
Error: error while trying to evaluate "region": :0,4-11: Unsupported attribute; This object does not have an attribute named "narnia".
29:     inputs: {region: "$(vars.narnia)"}
                ^
validator "test_module_not_used" failed:
module "debug_node_group" uses module "network1", but matching setting and outputs were not found. This may be because the value is set explicitly or set by a prior used module

validator "test_deployment_variable_not_used" failed:
Error: the variable "useless" was not used in this blueprint
24:   useless: 77
               ^

validator "test_zone_in_region" failed:
zone us-central1-a is not in region us-west1 in project ID io-playground or your credentials do not have permissions to access it

One or more blueprint validators has failed ... <same>
```


tpdownes (approved):
I think the re-factoring and ordering of dependencies makes a lot of sense. In my limited testing, it looks like explicitly defined validators are always written to the expanded blueprint.

If the claim about explicit validators is always true, please go ahead and merge. If there's some edge cases to that, please let me know first.

mr0re1:
> it looks like explicitly defined validators are always written to the expanded blueprint.

Correct, `original_bp(after CMD skipValidators applied).Validators == expanded_bp.Validators`

PR #1712, opened 2023-08-22
Improvements to hpc-enterprise blueprint

cboneti:
- print sinfo after Slurm partition test
- decrease the n2 node wait time to 300s
- increased the timeout for node cleanup to 15 minutes, instead of 10.


mr0re1 (approved):
preventive lgtm

PR #1718, opened 2023-08-24
Slurm image checks and Documentation Updates

cdunbar13:
Updates to check the instance_image family and project settings for slurm deployments.  Tied to v5.7 for compatibility, but can be expanded if needed.

Updates to documentation as well.


community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -24,3 +30,27 @@ locals {
     : "projects/${local.source_image_project}/global/images/family"
   )
 }
+
+data "google_compute_image" "slurm" {
+  family  = contains(keys(var.instance_image), "family") ? var.instance_image["family"] : null
+  name    = contains(keys(var.instance_image), "name") ? var.instance_image["name"] : null
+  project = var.instance_image["project"]
+
+  lifecycle {
```
tpdownes, line 39: I believe that we want to distinguish between 2 situations:

1. The blueprint is using the schedmd-slurm-public project but an incompatible image.
2. The blueprint is using another project entirely

I'm also concerned about throwing the words "not supported" around when we really mean "incompatible." Let's discuss specific wording offline.

I think you'll want two postconditions for this and it will probably look something like below. The error messages are not intended to be taken verbatim, but an idea of how to distinguish between the different scenarios.

```
  condition =  var.instance_image_override || self.project == "schedmd-slurm-public"
  error_message = "If using images published outside the schedmd-slurm-public project, you must set var.instance_image_override to true and build images according to LINK-TO-URL."
```

The OR condition could also be a `contains(keys...`. A second postcondition might look like

```
  condition =  self.project != "schedmd-slurm-public" || try(contains(local.supported_project_families[self.project], self.family), false)
  error_message = "If using images published in the schedmd-slurm-public project, you must use one of these compatible releases: X,Y,Z"
```

I don't think we want to get in the business of supporting alternative families within the schedmd project.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
```diff
@@ -550,6 +550,19 @@ variable "instance_image" {
   }
 }
 
+variable "instance_image_override" {
```
tpdownes, line 553: 1. Let's discuss offline "override" as the terminology to use. I like having the variable being lexically close to "instance_image."
2. I think the text should steer the user toward a statement like "the image must be built to be compatible with release according to LINK-TO-URL" rather than "custom image that is not supported".

tpdownes (commented):
Initial round of feedback.

community/modules/scheduler/schedmd-slurm-gcp-v5-login/variables.tf
```diff
@@ -308,6 +308,19 @@ variable "instance_image" {
   }
 }
 
+variable "instance_image_override" {
+  description = <<-EOD
+  A flag that designates that the user is aware that they are requesting
```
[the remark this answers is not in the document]
cdunbar13: > I think we should discuss this variable name and description a bit offline:
> 
> 1. I like that it is lexically next to instance_image so that it appears that way in documentation. I think I might choose a different word than override. Let's think a bit.

I believe I've gotten a bit closer to your intention with for this.  I changed the variable to `instance_image_compatible`, but we can discuss.

> 2. I think we should use different language than "not supported by" and more like "and must be produced with same release of Terraform module according to docs at https://X"

Understood, I believe the new language covers this in both the errors messages, variable descriptions, and documentation.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
cdunbar13: I've updated the language (but we should still discuss), and took a different approach for the two potential errors.  Instead of two postconditions, which may both trigger, I've created one postcondition with a conditional for the error message.  It should give a more accurate error message depending on if the user has specified an incorrect project or image family.

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
cdunbar13: Same as above.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
tpdownes: I do not see how my two postconditions can simultaneously trigger. Can you demonstrate a combination of settings that does?

My suggestion seems very direct to read and the error messages are a lot shorter. I want to get these error messages really down to "Here is your immediate problem. Visit URL for details." Preferably one line.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -110,6 +110,21 @@ variable "instance_image" {
   }
 }
 
+variable "instance_image_compatible" {
+  description = <<-EOD
+    A flag that designates that the user is aware that they are requesting
```
tpdownes, line 115: I suggest:

> that the user has built or copied a custom image that meets compatibility requirements defined at URL.
> This setting is ignored if using official images built by SchedMD

docs/vm-images.md
```diff
@@ -166,45 +206,53 @@ a description of our support for Windows images.
...
 support images not listed here, other public and custom images should work with
 the majority of modules with or without further customization, such as custom
 startup-scripts.
 
-## Slurm on GCP Custom Images
+### Slurm on GCP
+
+For Slurm images, please see [SchedMD's GitHub repository](https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md#public-image).
+
+#### Publically Published Images
```
tpdownes, line 221: Misspelling: "Publicly"

docs/vm-images.md
```diff
@@ -1,20 +1,60 @@
...
 For information on customizing VM images with extra software and configuration
 settings, see [Building Images](image-building.md).
 
 Please see the [blueprint catalog](https://cloud.google.com/hpc-toolkit/docs/setup/hpc-blueprint-catalog) for examples.
 
-For Slurm images, please see [SchedMD's GitHub repository](https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#public-image).
+## General Information
 
-## HPC CentOS 7
+### Instance Image
```
tpdownes, line 19: I'm not convinced that this new section is helping this PR especially as we have not yet harmonized the dtype of instance_image into an object of defined fields.

docs/vm-images.md
```diff
@@ -166,45 +206,53 @@ a description of our support for Windows images.
...
 startup-scripts.
 
-## Slurm on GCP Custom Images
+### Slurm on GCP
+
+For Slurm images, please see [SchedMD's GitHub repository](https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md#public-image).
+
+#### Publically Published Images
+
+SchedMD publishes "Slurm on GCP" public images, which have the following [documentation][slurm-gcp-images]. From there, you can see which public images are available and which software is installed on them.
```
tpdownes, line 223: I think you need to draft text here that is crystal clear that the Terraform module version and Packer module version must be the same. That should be the first sentence. The second (and third?) sentences should be that SchedMD publishes these images and that our modules will enforce compatibility when using images published by SchedMD.

The next paragaph should cover custom images. We can review that text in next round.

tpdownes (changes requested):
Make sure that we are using the data source as inputs to module where we can. We should have discussion with teammates about the project normalized value as that is funny, but (as I recall) for a good reason.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -24,3 +30,38 @@ locals {
     : "projects/${local.source_image_project}/global/images/family"
   )
 }
+
+data "google_compute_image" "slurm" {
+  family  = contains(keys(var.instance_image), "family") ? var.instance_image["family"] : null
```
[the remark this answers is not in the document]
cdunbar13: Done.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -15,6 +15,12 @@
...
 locals {
+  # Currently supported images and projects
+  supported_project_families = {
+    "schedmd-slurm-public" = ["slurm-gcp-5-7-debian-11", "slurm-gcp-5-7-hpc-rocky-linux-8", "slurm-gcp-5-7-ubuntu-2004-lts",
+    "slurm-gcp-5-7-ubuntu-2204-lts-arm64", "slurm-gcp-5-7-hpc-centos-7-k80", "slurm-gcp-5-7-hpc-centos-7"]
+  }
+
   source_image         = lookup(var.instance_image, "name", "")
   source_image_family  = lookup(var.instance_image, "family", "")
   source_image_project = lookup(var.instance_image, "project", "")
```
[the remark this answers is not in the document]
cdunbar13: Done.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/main.tf
```diff
@@ -62,9 +62,9 @@ locals {
     on_host_maintenance      = var.on_host_maintenance
     preemptible              = var.preemptible
     shielded_instance_config = var.shielded_instance_config
-    source_image_family      = local.source_image_family             # requires source_image_logic.tf
-    source_image_project     = local.source_image_project_normalized # requires source_image_logic.tf
-    source_image             = local.source_image                    # requires source_image_logic.tf
+    source_image_family      = local.source_image_family == "" ? "" : data.google_compute_image.slurm.family # requires source_image_logic.tf
```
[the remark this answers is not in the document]
cdunbar13: Resolved

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -15,12 +15,47 @@
...
-  source_image         = lookup(var.instance_image, "name", "")
-  source_image_family  = lookup(var.instance_image, "family", "")
-  source_image_project = lookup(var.instance_image, "project", "")
-  source_image_project_normalized = (
-    local.source_image != "" || length(regexall("/", local.source_image_project)) > 0
-    ? local.source_image_project
-    : "projects/${local.source_image_project}/global/images/family"
+  # Currently supported images and projects
+  known_project_families = {
+    "schedmd-slurm-public" = ["slurm-gcp-5-7-debian-11", "slurm-gcp-5-7-hpc-rocky-linux-8", "slurm-gcp-5-7-ubuntu-2004-lts",
```
tpdownes, line 20: Let's have the formatting align with rest of Toolkit choices. Drop the quotation marks around the key (it can only be a string in Terraform) and separate each value by line. e.g.

```hcl
known_project_families = {
  schedmd-slurm-public = [
    "slurm-gcp-5-7-debian-11",
    ...
  ]
}
```

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -16,52 +16,46 @@
...
-      condition = !var.instance_image_compatible || try(contains(local.supported_project_families[self.project], self.family), false)
-      error_message = (contains(keys(local.supported_project_families), self.project) ? <<-EOD
-      The image specified in the blueprint was found, but it is not compatible with this HPC Toolkit for Slurm module. The compatible image families are:
-
-      ${format("%-30s%-30s", "PROJECT", "IMAGE FAMILY")}
-      ${join("\n", [for proj in keys(local.supported_project_families) : join("\n", [for family in local.supported_project_families[proj] : format("%-30s%-30s", proj, family)])])}
-
-      Images that are not within this family may not have Slurm installed by default or may have faulty configurations.  If you are attempting to deploy a custom Slurm image, or an image with an unexpected version, please set the "instance_image_compatible" variable to "false" in the blueprint.
-      See https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/vm-images.md#slurm-on-gcp for more information.
+      condition     = var.instance_image_custom || try(contains(keys(local.known_project_families), self.project), false)
```
tpdownes, line 45: you do not need to wrap this in a `try` function.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -16,52 +16,46 @@
...
-      The project specified in "instance_image" was found, but is not guaranteed to contain images that are compatible with this module.  The project/s with compatible images are:
-
-      PROJECT
-      ${join("\n", [for proj in keys(local.supported_project_families) : proj])}
-
-      Please update the project specification in "instance_image", or set the "instance_image_compatible" variable to "false".
-      See https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/vm-images.md#slurm-on-gcp for more information.
+    }
+    postcondition {
+      condition     = var.instance_image_custom || !try(contains(keys(local.known_project_families), self.project), false) || try(contains(local.known_project_families[self.project], self.family), false)
```
tpdownes, line 51: I don't think `var.instance_image_custom` should appear in this condition. The logic is that the first condition will fire if you aren't using the schedmd-slurm-public project (in which case you must declare your intent to use a custom image) and the 2nd condition will fire if you are using schedmd-slurm-public but are using an incompatible family. There is not a situation where we want people to be able to override that logic within schedmd-slurm-public.

tpdownes: I think the logic reduces to:

```hcl
!contains(keys(local.known_project_families), self.project) || try(contains(local.known_project_families[self.project], self.family)
```

If Terraform did short-circuiting on booleans, you wouldn't need the `try`. But it doesn't.

tpdownes (changes requested):
I think we are close. If you agree with these changes, make them and then squash some of the back/forth so that the git history isn't so cluttered. After that, I'll try to break it with different combinations of project/name/family.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
cdunbar13: I'll make the fix removing the first condition, but it also makes it impossible for users to utilize (at their own peril) other schedmd images with the wrapper modules for v5.  I'd argue that `instance_image_custom: true` already lets them use any image, so why not other schedmd images?

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
cdunbar13: Resolved.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
cdunbar13: Resolved.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
cdunbar13: Resolved.

community/modules/scheduler/schedmd-slurm-gcp-v5-login/variables.tf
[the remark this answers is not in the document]
cdunbar13: Resolved.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -15,12 +15,53 @@
...
+data "google_compute_image" "slurm" {
+  family  = try(var.instance_image.family, null)
+  name    = try(var.instance_image.name, null)
+  project = var.instance_image.project
+
+  lifecycle {
+    postcondition {
+      condition     = var.instance_image_custom || contains(keys(local.known_project_families), self.project)
+      error_message = <<-EOD
+      '${self.project}' is not a known project with compatible Slurm image families. Use the 'instance_image_custom' flag to deploy custom images. See: https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/vm-images.md#slurm-on-gcp.
```
tpdownes, line 53: In interest of closing loop more quickly, I suggest that you modify this message to read:

```
      Images in project ${self.project} are not published by SchedMD. Images must be
      built by compatible releases of the Terraform and Packer modules following the
      guidance at https://goo.gle/hpc-slurm-images. Set var.instance_image_custom to
      true to silence this error and acknowledge that you are using a compatible image.     
```

community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -100,16 +100,32 @@ variable "instance_image" {
...
+    to use a custom and potentially incompatible image for this Slurm on
+    GCP module.
+
+    If the field is set to false, only the compatible families and project
+    names will be accepted.  The deployment will fail with any other image
+    family or name.  If set to true, no checks will be done.
+
+    See: https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/vm-images.md#slurm-on-gcp
```
tpdownes, line 123: Modify to https://goo.gle/hpc-slurm-images

docs/vm-images.md
```diff
@@ -1,20 +1,60 @@
...
 
 Please see the [blueprint catalog](https://cloud.google.com/hpc-toolkit/docs/setup/hpc-blueprint-catalog) for examples.
 
-For Slurm images, please see [SchedMD's GitHub repository](https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md#public-image).
+## General Information
 
-## HPC CentOS 7
+### Instance Image
+
+> **_Note:_** This information is true for most source modules, but there are some sources that have their own image specification.  Please read the documentation for any source module utilized.
```
tpdownes, line 21: I don't think we should include this change in this PR because it's really unrelated. But I will defer to @cboneti who plans a follow-up review for this PR.

docs/vm-images.md
```diff
@@ -166,45 +206,63 @@ a description of our support for Windows images.
...
+
+For Slurm images, please see
+[SchedMD's GitHub repository](https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md#public-image).
+
+### Custom Slurm Images
+
+> **_Warning:_** When building custom images, the Terraform and Packer modules
+must share the same version.
 
 HPC Toolkit modules based on terraform modules in [Slurm on GCP][slurm-gcp]
```
tpdownes, line 238: I would reduce the rest of the text to provide 2 links to `examples/README.md` with the anchors for image-builder.yaml and ml-slurm.yaml.

docs/vm-images.md
```diff
@@ -166,45 +206,63 @@ a description of our support for Windows images.
...
+
+### Publicly Published Slurm Images
+
+SchedMD publishes "Slurm on GCP" public images, which are documented
+[here][slurm-gcp-images]. This documentation covers which images are available
+and what software is installed on them.
+
+> **_Note:_** When deploying a Slurm cluster, the HPC Toolkit Terraform wrapper
+modules enforce that specific image versions are used, to ensure compatibility.
+Currently, the **slurm-gcp-5-7** image families are enforced for the Slurm v5
```
tpdownes, line 227: I would rather not name a specific release in the text especially as this only matches the beginning of each families name. Suggest removing the line in favor of:

```
Slurm images are compatible by the minor version releases of the Terraform and Packer modules. For example, images built for version 5.8 are compatible with all 5.8 Terraform modules beginning with 5.8.0. The version of the Slurm modules used by your copy of the Toolkit in the local filesystem can be inspected by looking for the `source` line in community/modules/compute/schedmd-slurm-gcp-v5-partition/main.tf.
```

The latest GitHub release supports [these images](https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md#published-image-family).

@cboneti This is definitely open for some thought.

tpdownes (changes requested):
This is my last review. I think we're at a place where the logic is good and the final text needs some massaging. My suggestions, frankly, reverse a bit of the "keep it super short" requests, so accept my apologies but I think they are clearer.

In addition to the changes, I've requested, please:

1. Correct a bunch of trailing newlines here and there.
1. Accept feedback from @cboneti 
1. Wait for #1728 to be merged, rebase off develop and update any explicit references from 5.7.6 to 5.8.0.
1. Update examples/image-builder.yaml and examples/ml-slurm.yaml to include the necessary extra setting.

tpdownes (commented):
LGTM. @cboneti requested an overall review when I was done.

cboneti (changes requested):
I left a few comments

docs/vm-images.md
```diff
@@ -203,8 +214,10 @@ These instructions apply to the following modules:
 * [schedmd-slurm-gcp-v5-node-group]
 
 [slurm-gcp]: https://github.com/SchedMD/slurm-gcp
-[slurm-gcp-packer]: https://github.com/SchedMD/slurm-gcp/tree/master/packer
-[slurm-gcp-images]: https://github.com/SchedMD/slurm-gcp/blob/master/docs/images.md
+[slurm-gcp-packer]: https://github.com/SchedMD/slurm-gcp/tree/5.7.6/packer
+[slurm-gcp-images]: https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md
+[slurm-gcp-published-images]: https://github.com/SchedMD/slurm-gcp/blob/5.7.6/docs/images.md#published-image-family
```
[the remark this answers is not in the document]
cdunbar13: For now we could probably point to their version branch (v5) instead of the tag.  It would present the same problem as before though, where they go ahead a version and it takes us a few days/weeks/months to update to that version.  That's what I'm guessing happened with the person that was wondering why the 6.1 image didn't work (we were pointing to the Slurm-GCP master branch).  

This problem seems related but it's not immediately apparent to me how a solution would solve all of the versioning issues.  I will try and think on it.  Maybe this is another task.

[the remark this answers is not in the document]
cdunbar13: Switched to v5, but it will have the issues mentioned above.

community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
```diff
@@ -15,12 +15,54 @@
...
-  source_image_family  = lookup(var.instance_image, "family", "")
-  source_image_project = lookup(var.instance_image, "project", "")
-  source_image_project_normalized = (
-    local.source_image != "" || length(regexall("/", local.source_image_project)) > 0
-    ? local.source_image_project
-    : "projects/${local.source_image_project}/global/images/family"
+  # Currently supported images and projects
+  known_project_families = {
+    schedmd-slurm-public = [
+      "slurm-gcp-5-7-debian-11",
```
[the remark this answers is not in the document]
cdunbar13: Updated

docs/vm-images.md
cboneti: is this resolved?

cdunbar13: Yes, resolved.

docs/vm-images.md
cdunbar13: Resolved.

cboneti (approved):

tpdownes (approved):

PR #1721, opened 2023-08-25
Refactor quota validation logic

mr0re1:
* Change logic of validation: Instead of piling together all buckets for metric regardless of dimensions scope, Limit validation logic to at most one `QuotaBucket` in `ConsumerQuotaLimit` per requested resource. This change fixes problem when "default" buckets were always used regardless of finding a better matching bucket within `ConsumerQuotaLimit`.

* Remove option to use aggregation other than "SUM", can be added in future if needed;
* Change `QuotaError` rendering to output human readable name instead of metric, also output dimensions scope to differentiate between zonal/regional/and others quotas;
* Execute API calls in parallel to speed it up.

**Before:**
```sh
validator "test_resource_requirements" failed:
not enough quota for resource "compute.googleapis.com/disks_total_storage", limit=4096 < requested=10250
not enough quota for resource "compute.googleapis.com/c2_cpus", limit=24 < requested=12004
not enough quota for resource "compute.googleapis.com/c2_cpus" in map[region:us-central1], limit=4800 < requested=12004
not enough quota for resource "compute.googleapis.com/cpus_per_vm_family", limit=0 < requested=1144
not enough quota for resource "compute.googleapis.com/cpus_per_vm_family" in map[region:us-central1], limit=0 < requested=1144
not enough quota for resource "compute.googleapis.com/cpus_per_vm_family" in map[vm_family:H3], limit=176 < requested=1144
not enough quota for resource "compute.googleapis.com/cpus_per_vm_family" in map[region:us-central1 vm_family:H3], limit=440 < requested=1144
not enough quota for resource "compute.googleapis.com/n2_cpus", limit=24 < requested=700
not enough quota for resource "compute.googleapis.com/ssd_total_storage", limit=500 < requested=100500
not enough quota for resource "compute.googleapis.com/ssd_total_storage" in map[region:us-central1], limit=55000 < requested=100500 + usage=10500
```

**After:**
```sh
validator "test_resource_requirements" failed:
not enough quota "C2 CPUs" as "1/{project}/{region}" in [region:us-central1], limit=4800 < requested=12004
not enough quota "Total SSD disk reserved (GB)" as "1/{project}/{region}" in [region:us-central1], limit=55000 < requested=100500 + usage=10500
not enough quota "CPU count per VM family." as "1/{project}/{region}/{vm_family}" in [region:us-central1 vm_family:H3], limit=440 < requested=1144
```

### DIFF
--- a/pkg/validators/quota.go
+++ b/pkg/validators/quota.go
@@ -39,7 +39,7 @@ type ResourceRequirement struct {
 	Dimensions map[string]string `cty:"dimensions"` // e.g. {"region": "us-central1"}
 }
 
-// InBucket returns true if the quota is in the QuotaBucket.
+// InBucket returns true if all dimensions specified in the bucket match dimensions the requirement.
 func (q ResourceRequirement) InBucket(b *sub.QuotaBucket) bool {
 	for d, v := range b.Dimensions {
 		if q.Dimensions[d] != v {
@@ -110,7 +110,19 @@ func validateResourceRequirements(rs []ResourceRequirement, up *usageProvider) (
 	return qerrs, errs.OrNil()
 }
 
-// Validate requirements for a single consume & service pair.
+// Validate requirements for a single consumer & service pair.
+// The `ServiceUsage.ConsumerQuotaMetrics` API call returns following structure:
+//
+// list[ConsumerQuotaMetric] - one per metric/quota, e.g. compute.googleapis.com/n2_cpus for `N2 CPUs`.
+// ├Metric
+// └list[ConsumerQuotaLimit] - one per quota scope (e.g. regional or zonal)
+// ....|.......................for N2 CPUs there are two: `N2-CPUS-per-project-region` & `N2-CPUS-per-project-zone`
+// ....├Unit - e.g. '1/{project}/{region}' for regional "scope"
+// ....└list[QuotaBucket] - represents the "slice" of the "scope" with specified limit
+// ........|................e.g. for `N2-CPUS-per-project-region` there will be buckets with specific region: `{'region': 'asia-east1'}`,
+// ........|................and one bucket, named `N2 CPUs (default)` in the UI, without `dimensions={}` to act as a wildcard and provide default limit.
+// ........├EffectiveLimit
+// ........└Dimensions - e.g. {"region": "us-central1"}
 func validateServiceRequirements(consumer string, service string, rs []ResourceRequirement, up *usageProvider) ([]QuotaError, error) {
 	qms, err := queryMetrics(consumer, service)
 	if err != nil {
@@ -141,7 +153,7 @@ func findBucket(r ResourceRequirement, ql *sub.ConsumerQuotaLimit) (*sub.QuotaBu
 		fmt.Errorf("unexpected default-less ConsumerQuotaLimit: %q", ql.Name)
 }
 
-// Set of requirements to the single bucket with context attached
+// Set of requirements that fall into the specific bucket with context attached
 type bucketRequirements struct {
 	QuotaMetric  *sub.ConsumerQuotaMetric
 	QuotaLimit   *sub.ConsumerQuotaLimit
@@ -182,6 +194,8 @@ func gatherBucketsRequirements(rs []ResourceRequirement, qms map[string]*sub.Con
 	return maps.Values(res), errs.OrNil()
 }
 
+// validateBucket aggregates (sum) all requirements and usage for the given bucket
+// and returns QuotaError if the bucket quota limit is not sufficient.
 func validateBucket(br bucketRequirements, up *usageProvider) []QuotaError {
 	if len(br.Requirements) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/validators/quota.go
```diff
@@ -86,149 +88,150 @@ func validateResourceRequirements(rs []ResourceRequirement, up *usageProvider) (
...
 
-		for _, limit := range ml {
-			qle := validateLimit(g, limit, up, agg)
-			qe = append(qe, qle...)
+// Find a bucket in the ConsumerQuotaLimit that matches the ResourceRequirement.
+func findBucket(r ResourceRequirement, ql *sub.ConsumerQuotaLimit) (*sub.QuotaBucket, error) {
+	// Iterate buckets in order from most to less specific
+	for i := len(ql.QuotaBuckets) - 1; i >= 0; i-- {
+		if r.InBucket(ql.QuotaBuckets[i]) {
+			return ql.QuotaBuckets[i], nil
```
nick-stroud, line 135: It seems like this is only finding a single bucket for a given quota limit (greedy return). From our discussion I thought it would fine all possible buckets. Can you please clarify. If we are only taking a single bucket and not all buckets then what is the criteria for the single bucket? To me it looks like you are just running through the list backwards. Is there only guaranteed to be a single match? Is the list sorted from least to most specific? Consider renaming to `findBestMatchBucket`.

pkg/validators/quota.go
```diff
@@ -86,149 +88,150 @@ func validateResourceRequirements(rs []ResourceRequirement, up *usageProvider) (
...
-	qe := []QuotaError{}
-	for _, bucket := range limit.QuotaBuckets {
-		vs := []int64{}
-		for _, r := range rs {
-			if r.InBucket(bucket) {
-				vs = append(vs, r.Required)
-			}
-		}
-		if len(vs) == 0 {
+// Set of requirements to the single bucket with context attached
```
nick-stroud, line 144: I understand this as "The set of requirements that match a specific bucket", emphasis on "match". My assumption is that there has been processing to match the requirements to the bucket. The wording of "requirements to bucket" is confusing. If so please update the description.

pkg/validators/quota.go
```diff
@@ -86,149 +88,150 @@ func validateResourceRequirements(rs []ResourceRequirement, up *usageProvider) (
...
+
 	}
-	return qe
+	return maps.Values(res), errs.OrNil()
 }
 
-func satisfied(requested int64, limit int64) bool {
-	if limit == -1 {
-		return true
+func validateBucket(br bucketRequirements, up *usageProvider) []QuotaError {
```
nick-stroud, line 185: Add a doc string description.

pkg/validators/quota.go
nick-stroud, line 1: We are having to do in depth processing perform this verification. Are we sure that a simpler way. It would make the most sense for the owners of the quota system to own all this logic and provide a "dry run" api where you could say, is there capacity for me to request xyz. Obviously this may not be the world we live in, but I want to make sure you have exhausted any such options.

nick-stroud (changes requested):
In general, I found the code is hard to read. I suspect this is the combination of many foreign quota concepts mixed with non-descriptive variable naming. After a few times through, and discussion with you, I feel comfortable with what is being done and it seems correct, but please lean on the side of more explanation to help with future maintainability.

pkg/validators/quota.go
mr0re1: Assume there is such API, say `checkQuota(metric, dimension, required_amount)`.
We couldn't use it because we have to deal with multiple requirements, that have to be aggregated (now just sum) along some dimensions.
Given we don't now the nature of dimensions we would have to do brute force -> 2^num_dimensions possible aggregations.
Also it could lead to the false positives:
Say there is a zonal quota, but our requirement contains both `region` & `zone` dimensions.
In our exhaustive search we will try aggregation by `{region}`, that would force validation to fallback to the "default" bucket `{}`, with potentially low quota we don't care about.

pkg/validators/quota.go
mr0re1: > From our discussion I thought it would fine all possible buckets. 

We find multiple buckets, but exactly one per `ConsumerQuotaLimit`, the `ConsumerQuotaLimit` defines the "scope" of quota, e.g. regional or zonal.
E.g. for `N2_CPUs`  we will check two buckets, one from regional-scope and another from zonal-scope.

pkg/validators/quota.go
mr0re1: Changed to `Set of requirements that fall into the specific bucket `

pkg/validators/quota.go
nick-stroud: >  If we are only taking a single bucket and not all buckets then what is the criteria for the single bucket? To me it looks like you are just running through the list backwards. Is the list sorted from least to most specific?

There is nothing in the function that leads me to believe we are returning the most specific match, it just looks like we are returning any match. Maybe I am missing something simple. Is the list known to be sorted with some special properties?

nick-stroud (approved):

PR #1730, opened 2023-09-01
Hotfix: Update NVIDIA Grid drivers to 15.3

rohitramu:
The Ubuntu base image was updated on 08/29, and our integration test for Chrome Remote Desktop started failing.  The error was likely caused by an incompatibility between an updated kernel and the NVIDIA Grid driver installer.

Updating the NVIDIA Grid driver installer to version 15.3 resolves the issue.


nick-stroud (approved):
More testing is needed for Debian but this fixes the observed failure on Ubuntu.

PR #1743, opened 2023-09-06
Create a helper tool to automate some of the steps of cleaning up a failed build

nick-stroud:
A helper tool that relieves some of the pain of internal cleanup of a failed build. 

Usage:
```
tools/cleanup-build.py my-project gs://my-bucket/test-name/build.tgz
```

Not covered in this tool:
- Identifying the failed builds
- finding the tgz file that goes along with failed builds
- handling cases such as left behind nodes or lock file needs overwrite

### Submission Checklist

### DIFF
--- a/tools/cleanup-build.py
+++ b/tools/cleanup-build.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


mr0re1:
It can be handy to enable shortcut invocation `cleanup-build.py <build_id>`, to avoid inspecting logs manually and can help with automation, e.g. `gcloud build list <recent_failures> --format=%id | xargs cleanup-build.py`

tpdownes (commented):
Without changes to `ghpc destroy`, I don't think this is going to work for multi-group deployments because you need to run the correct order of `ghpc export-outputs` and `ghpc import-inputs` commands on each group to regenerate the necessary TFVARS files.

I think the "most correct" thing to address would be to alter the behavior of `ghpc destroy` to do this for you (in a read-only fashion). Until we do that (not in this PR), we could (A) choose to accept this shortcoming (B) parse the YAML for the deployment group names and invoke the commands in order.

WDYT?

nick-stroud:
@tpdownes, I created a task for your suggestion B. Personally I would say we should put further energy into fixing with option A, which you pointed out offline already has a task. Option A will solve this problem for this tool but also for our users.

nick-stroud:
@mr0re1, I like your idea but will save for future PR. There is some minor complication as mapping from build id -> tgz file is not easily apparent.

tpdownes (approved):

PR #1744, opened 2023-09-07
Refine Slurm CRD blueprint and rename deployment_name in test

nick-stroud:
This change:
- use consistent deployment names
- use n2d instead of n2 to improve obtainability
- use larger machine for compute partition
- increase dynamic count of partitions since there is no cost for dynamic nodes

### Submission Checklist

### DIFF
--- a/community/examples/hpc-slurm-chromedesktop.yaml
+++ b/community/examples/hpc-slurm-chromedesktop.yaml
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,9 +18,15 @@ blueprint_name: slurm-crd
 
 vars:
   project_id:  ## Set GCP Project ID Here ##
-  deployment_name: slurm-chromedesktop
+  deployment_name: slurm-crd-01
   region: us-central1
   zone: us-central1-c
+  instance_image_crd:
+    family: slurm-gcp-5-8-debian-11
+    project: schedmd-slurm-public
+  instance_image:
+    family: slurm-gcp-5-8-hpc-centos-7
+    project: schedmd-slurm-public
 
 # Documentation for each of the modules used below can be found at
 # https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md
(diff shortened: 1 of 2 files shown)


community/examples/hpc-slurm-chromedesktop.yaml
```diff
@@ -74,8 +78,8 @@ deployment_groups:
   - id: compute_node_group
     source: community/modules/compute/schedmd-slurm-gcp-v5-node-group
     settings:
-      machine_type: n2-standard-4
-      node_count_dynamic_max: 1
+      machine_type: n2d-standard-16
+      node_count_dynamic_max: 20
```
harshthakkar01, line 82: Just for my understanding, why do we want to set this to large number ?

tools/cloud-build/daily-tests/tests/hpc-slurm-chromedesktop.yml
```diff
@@ -14,10 +14,10 @@
 ---
 
 test_name: hpc-slurm-chromedesktop
-deployment_name: "enter-{{ build }}"
+deployment_name: "slurm-crd-{{ build }}"
 # Manually adding the slurm_cluster_name for use in node names, which filters
 # non-alphanumeric chars and is capped at 10 chars.
-slurm_cluster_name: "enter{{ build[0:5] }}"
+slurm_cluster_name: "crd{{ build[0:5] }}"
```
harshthakkar01, line 20: nit: would be good to change this to `slurm-crd{{ build[0:5] }}` for consistency.

harshthakkar01:
Also,
1) Can you add label (release-chore) ? to make sure PR Label verification passes.
2) Run integration test locally to make sure, it is working fine.

Thanks.

community/examples/hpc-slurm-chromedesktop.yaml
nick-stroud: This is the max number of nodes that can be dynamically scaled to. If set to one then we can only create 1 node which is not much of a cluster. In a real world situation we would probably have a need to run multiple nodes. Since this is a max and nodes are dynamic, it will only deploy the nodes if there are jobs waiting for them.

tools/cloud-build/daily-tests/tests/hpc-slurm-chromedesktop.yml
nick-stroud: This was intentional, slurm will only use a few characters from the name (like 6 or maybe 8) when naming resources so if we use `slurm-crd-{build}` the build id will not make it into instance names. It will be clear that the instances are for slurm because the will append other identifiers like "controller".

nick-stroud:
Integration test is running, I will wait until it passes to merge.

harshthakkar01 (approved):

tools/cloud-build/daily-tests/tests/hpc-slurm-chromedesktop.yml
nick-stroud: Infact `slurm_cluster_name` must be a subset of the deployment name. Test caught it. I am rerunning with an update.

PR #1759, opened 2023-09-14
Add native support for ephemeral storage to GKE modules

nick-stroud:
This change adds native HPC Toolkit support for the following ephemeral storage options with GKE:
- In memory
- Local SSD
- SSD persistent disk
- Balanced persistent disk

These new features are documented and the storage-gke.yaml example has been updated to demonstrate the use of these new storage types.

Note to reviewer: This large change has been broken into several smaller commits which each stand on their own. **I would recommend reviewing the commits one at a time.**

Tested: I have manually tested all of the options individually and in concert. Through the use of `df -h` and `fio` I have satisfied myself that each option is backed by the expected storage type.

Not in this PR: Test automation for the storage-gke.yaml blueprint. This will be coming in a follow on PR.

### Submission Checklist


community/modules/compute/gke-job-template/variables.tf
```diff
@@ -123,6 +123,30 @@ variable "persistent_volume_claims" {
...
+    condition = alltrue([
+      for v in var.ephemeral_volumes :
+      contains(["pd-balanced", "pd-ssd", "memory", "local-ssd"], v.type)
+    ])
+    error_message = "Type must be one of 'pd-balanced', 'pd-ssd', 'memory', 'local-ssd'."
+  }
+  validation {
+    condition = alltrue([
+      for v in var.ephemeral_volumes :
+      length(regexall("^/.*", v.mount_path)) > 0
```
tpdownes, line 144: Please change this to `substr(v.mount_path, 0, 1) == "/"` for extreme explicitness since the `length(regexall` approach is going to go away with functions introduced in Terraform 1.3 and above.

tpdownes (changes requested):
This is not obvious to me, but is there any guarantee that the size_gb of the ephemeral_volumes in gke-job-template will be satisfiable by the local_ssd_count_ephemeral_storage for the node pool?

If not, how does this manifest to user? is there a low-hanging fruit pre/post-condition we can apply?

community/modules/compute/gke-job-template/variables.tf
nick-stroud: Done.

nick-stroud:
> This is not obvious to me, but is there any guarantee that the size_gb of the ephemeral_volumes in gke-job-template will be satisfiable by the local_ssd_count_ephemeral_storage for the node pool?
> 
> If not, how does this manifest to user? is there a low-hanging fruit pre/post-condition we can apply?

The short answer is no guarantee. If the user asks for too much the job will go pending and never start. Logs will state that job cannot be placed due to ephemeral storage request or if there are no nodes with SSDs then it will say that node selector requirements are not met. 

If I had to take a stab at a precondition, we could pass information about local ssd count from node pool to job. The challenge is that local SSDs each have 375 GiB capacity, but GKE  performs non-linear hold back so the actual available is less. We could do our best to model this holdback to set limits but the holdback could also be subject to change over time. My solution was to not delve into this and let the user hit the natural cap imposed by GKE.

community/examples/storage-gke.yaml
```diff
@@ -76,3 +79,57 @@ deployment_groups:
       - echo \$(date) >> /shared/timestamp.log; cat /shared/timestamp.log
       node_count: 3
     outputs: [instructions]
+
```
tpdownes, line 82: I agree that trying to enforce this can easily do more harm than good and should rely on K8S for user actions.

tpdownes (approved):

PR #1760, opened 2023-09-14
Guide user to solution when long family format

tpdownes:
The upstream Slurm modules were written to require a "long" project ID when using image familes:

```hcl
projects/${actual_project_id}/global/images/family
```

This format is not an actual project ID and is likely incompatible with other Toolkit modules, so it would break usage of `instance_image` as a deployment variable in the `vars` dictionary. It is effectively a workaround for constraints imposed by the [implementation of the upstream CFT instance_template module](https://github.com/terraform-google-modules/terraform-google-vm/blob/85fd51e618211a2333e3fb3b6a8d2addf74e2ad0/modules/instance_template/main.tf#L28). This module probably merits improvement, but its working to produces a valid value for the Compute Engine API's image parameter. It has many valid format and one of them is:

```hcl
projects/${actual_project_id}/global/images/family/${image_family_name}
```

Support for user specification of the project in this format was unintentionally dropped in bf729bdb. This commit adds validation to test when a user has provided this special ID format and guides them to use the simple project ID. The logic of constructing the special format is implemented entirely within the module.

In coding this solution, I came to the conclusion that the logic of validating the project's string contents is best done by the data source itself (via API failures) or via preconditions rather than validation blocks in the variable. Reasoning: you need lots of `try` functions in the validation blocks because the existence of the `project` field is not yet guaranteed. So the validation blocks now only service to guarantee the required fields of the `map(string)` variable.

After this change, the following error message will appear once for every Slurm module that uses the long format ID:

```
Error: Resource precondition failed

  on modules/embedded/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/source_image_logic.tf line 51, in data "google_compute_image" "slurm":
  51:       condition     = length(regexall("^projects/.+?/global/images/family$", var.instance_image.project)) == 0
    ├────────────────
    │ var.instance_image.project is "projects/schedmd-slurm-public/global/images/family"

The "project" field in var.instance_image no longer supports a long-form
ending in "family". Specify only the project ID.
```

### Submission Checklist

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
@@ -100,13 +100,13 @@ variable "instance_image" {
   }
 
   validation {
-    condition     = can(var.instance_image.project) && try(var.instance_image.project, "") != ""
-    error_message = "The \"project\" field is required for var.instance_image and cannot be the empty string."
+    condition     = can(coalesce(var.instance_image.project))
+    error_message = "In var.instance_image, the \"project\" field must be a string set to the Cloud project ID."
   }
 
   validation {
-    condition     = can(var.instance_image.name) != can(var.instance_image.family)
-    error_message = "Exactly one of \"family\" and \"name\" must be provided in var.instance_image."
+    condition     = can(coalesce(var.instance_image.name)) != can(coalesce(var.instance_image.family))
+    error_message = "In var.instance_image, exactly one of \"family\" or \"name\" fields must be set to desired image family or name."
   }
 }
 
--- a/community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v5-node-group/source_image_logic.tf
@@ -47,6 +47,11 @@ data "google_compute_image" "slurm" {
   project = var.instance_image.project
 
   lifecycle {
+    precondition {
+      condition     = length(regexall("^projects/.+?/global/images/family$", var.instance_image.project)) == 0
+      error_message = "The \"project\" field in var.instance_image no longer supports a long-form ending in \"family\". Specify only the project ID."
+    }
+
     postcondition {
       condition     = var.instance_image_custom || contains(keys(local.known_project_families), self.project)
       error_message = <<-EOD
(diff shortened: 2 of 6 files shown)


community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -100,8 +100,8 @@ variable "instance_image" {
   }
 
   validation {
-    condition     = can(var.instance_image.project) && try(var.instance_image.project, "") != ""
-    error_message = "The \"project\" field is required for var.instance_image and cannot be the empty string."
+    condition     = can(var.instance_image.project)
```
nick-stroud, line 103: We no longer check for empty string. Why? This is caught by data source but error is not intuitive. While I think it is unlikely that someone puts `project: ""` in the blueprint, the check was added for a reason. I will approve but expect either some rational for removing this check or to add it back.

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
tpdownes: I believe that I suggested that code while reviewing #1718 and can state with some authority that I added the check without having a clearly good reason. That said, examining the situation now, I think treating an empty string a identical to null in Terraform behavior is good for this value. Both are ways a user might reasonably think to express "I'm not setting this" and both are equally invalid from the perspective of the code.

So I updated code in line with your suggestion, albeit using `coalesce` as implementation and reducing the degree to which the message guides the user "away from empty string" and "toward the value it should represent." Like you, I don't think the user is likely to think empty string is a valid project ID.

nick-stroud: IIUC `coalesce` is handling the empty string case (returning false instead of true for `""`)? LGTM.

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
tpdownes: That is correct in spirit: `coalesce` errors when you supply it `null` or empty string. Thus, wrapping it in `can` (which returns `true` if the function does anything besides error) turns the errors into `false`.

PR #1766, opened 2023-09-15
Added bucket force_destroy variable to HCLS blueprint

cdunbar13:
The HCLS blueprint will fail during destruction if a user follows the instructions in the associated README, as there will be objects in the buckets.  This adds the option to force the destruction of the buckets as this is an example that does not produce results users would want to keep.

Also updated associated documentation to recommend that users enable this variable.

### DIFF
--- a/docs/videos/healthcare-and-life-sciences/README.md
+++ b/docs/videos/healthcare-and-life-sciences/README.md
@@ -115,8 +115,9 @@ storage intact and b) you can build software before you deploy your cluster.
    ```
 
    The `bucket_force_delete` variable makes it easier to tear down the
-   deployment. If it is set to `false`, buckets with objects (files) will not be
-   deleted and the `./ghpc destroy` command will fail partway through.
+   deployment. If it is set to the default value of `false`, buckets with
+   objects (files) will not be deleted and the `./ghpc destroy` command will
+   fail partway through.
 
    If the data stored in the buckets should be preseverved, remove the
    `--vars bucket_force_delete=true` portion of the command or set it to `false`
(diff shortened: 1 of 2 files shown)


docs/videos/healthcare-and-life-sciences/README.md
```diff
@@ -104,12 +104,23 @@ storage intact and b) you can build software before you deploy your cluster.
...
+   If you are running this as a test, and don't care about the files created in
+   the cloud buckets being destroyed, it is recommneded you run:
 
    ```bash
-   ./ghpc create docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml -w --vars project_id=<project>
+   ./ghpc create docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml -w --vars project_id=<project> --vars bucket_force_delete=true
    ```
 
+   The `bucket_force_delete` variable makes it easier to tear down the
+   deployment. If it is set to `false`, buckets with objects (files) will not be
```
nick-stroud, line 118: nit: I would call out that this is the default: 
`If it is set to the default value of `false`, buckets with objects...`

docs/videos/healthcare-and-life-sciences/hcls-blueprint.yaml
```diff
@@ -21,6 +21,7 @@ vars:
   deployment_name: hcls-01
   region: us-central1
   zone: us-central1-c
+  bucket_force_destroy: false
```
nick-stroud, line 24: nit: I have a slight preference for `force_destroy _buckets` as it reads more like a question. But what you have is also fine so feel free to ignore.

nick-stroud (approved):

PR #1767, opened 2023-09-15
Move task Ensure all nodes are powered down to slurm integration test from partition test

harshthakkar01:
### Submission Checklist


examples/hpc-slurm.yaml
```diff
@@ -21,6 +21,9 @@ vars:
   deployment_name: hpc-small
   region: us-central1
   zone: us-central1-a
+  # When set, active compute nodes will be cleaned up on destroy.
+  # Note that setting this option requires additional dependencies to be installed locally.
```
nick-stroud, line 25: I believe due to the `requires additional dependencies` this is intentionally left out of our examples.

I would propose the alternative action:
- do not change the blueprint
- In each slurm test (or at a minimum this slurm test) set `--vars enable_cleanup_compute=true`
- This will have the effect of turning it on in tests but leaving the example as is

You should be able to do this in the test definition: [example](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/6d3e6600e6e802f605a12336900da0294e40030e/tools/cloud-build/daily-tests/tests/hpc-enterprise-slurm.yml#L22-L24). Please test this out. Bools work on the actually command line but I don't know that his has been tested using this `cli_deployment_vars` in the test definition.

nick-stroud (changes requested):

examples/hpc-slurm.yaml
cboneti: +1

harshthakkar01: After offline discussion, we decided to move the Ensure all nodes are powered down to always block in slurm integration test. 
Updated accordingly.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
```diff
@@ -184,43 +203,59 @@
 
     ## Always cleanup, even on failure
     always:
+    - name: Ensure all nodes are powered down
+      ansible.builtin.command: sinfo -t 'POWERED_DOWN' --noheader --format "%n"
+      register: final_node_count
+      changed_when: False
+      until: final_node_count.stdout_lines | length == initial_node_count.stdout_lines | length
```
cboneti, line 210: This will break unless initial_node_count is acquired somehow before, which only happens with test partitions. If you want to move this into the slurm-integration-test, then you have to move the following *before* running the integration tests: 
```
- name: Count Slurm nodes
  ansible.builtin.shell:
    sinfo -t 'IDLE&POWERED_DOWN' --noheader --format "%n"
  args:
    executable: /bin/bash
  changed_when: False
  register: initial_node_count
  
  ```

File: examples/hpc-slurm.yaml (hunk 1 above)
Comment (nick-stroud) [reply examples/hpc-slurm.yaml]: I noted in offline discussion that there may be reasons to still go with my proposal here. I will remove my blocking request for change but would like to see the offline discussion resolve.

**Review by nick-stroud — COMMENTED** [review]
I still have a slight preference for using `enable_cleanup_compute=true` instead of custom logic in the test infra.

File: tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml (hunk 2 above)
Comment (harshthakkar01) [reply tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml:218]: Updated.

File: examples/hpc-slurm.yaml (hunk 1 above)
Comment (harshthakkar01) [reply examples/hpc-slurm.yaml]: Updated, after offline discussion we decided to keep both the solution. 

1) Add enable_cleanup_compute in the test file.
2) Move tasks from test partition to slurm integration test.

File: tools/cloud-build/daily-tests/tests/hpc-enterprise-slurm.yml [hunk 3]
```diff
@@ -22,6 +22,7 @@ zone: europe-west1-d
 cli_deployment_vars:
   region: europe-west1
   zone: "{{ zone }}"
+  enable_cleanup_compute: true
```
nick-stroud, line 25: This is a no-op since this option is already enabled for [this particular blueprint](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/0a301051fb4f9b704861eb2d3f8d5eedf1f7aa67/examples/hpc-enterprise-slurm.yaml#L35C13-L35C13).

I think we wanted to add similar functionality to the test for examples/hpc-slurm.yaml which does not have this setting.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/tests/hpc-enterprise-slurm.yml
harshthakkar01: Thanks for pointing this out. Updated slurm-v5-hpc-centos7 test variable.

tools/cloud-build/daily-tests/tests/slurm-v5-hpc-centos7.yml
```diff
@@ -20,6 +20,8 @@ deployment_name: "cent-v5-{{ build }}"
 # non-alphanumeric chars and is capped at 10 chars.
 slurm_cluster_name: "centv5{{ build[0:4] }}"
 zone: us-central1-a
+cli_deployment_vars:
+  enable_cleanup_compute: true
```
cboneti, line 24: are there no other slurm tests, why just adding to this one?

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
nick-stroud, line 1: Please run the integration tests that use this playbook.

nick-stroud (approved):

tools/cloud-build/daily-tests/tests/slurm-v5-hpc-centos7.yml
harshthakkar01: I don't see any other reference of hpc-slurm.yaml BP https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fhpc-toolkit+path%3A%2F%5Etools%5C%2Fcloud-build%5C%2Fdaily-tests%5C%2Ftests%5C%2F%2F++%22hpc-slurm.yaml%22&type=code in other tests.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
harshthakkar01: I will trigger all tests for this PR.

tools/cloud-build/daily-tests/tests/slurm-v5-hpc-centos7.yml
cboneti: I would look for all tests using [slurm-integration-test.yml](https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1767/files#diff-7cfee9141f93a04274d3ba1b990b0bf685d9481fdc364c8d90d96767491a376c)

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
harshthakkar01: All test passed except 1 PR-test-slurm-gcp-v5-hpc-centos7 (hpc-toolkit-dev) due to quota issue. But I have a successful run previously. 

Submitting this.

PR #1773, opened 2023-09-18
Add Pub/Sub subscriber role to Slurm compute node

tpdownes:
Adding this role will eliminate "ERROR: Cannot update and notify jobs with API failures" messages in /var/log/slurm/resume.log.

### Submission Checklist

### DIFF
--- a/examples/hpc-enterprise-slurm.yaml
+++ b/examples/hpc-enterprise-slurm.yaml
@@ -82,6 +82,7 @@ deployment_groups:
       project_roles:
       - logging.logWriter
       - monitoring.metricWriter
+      - pubsub.subscriber
       - storage.objectCreator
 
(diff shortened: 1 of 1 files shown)


tpdownes:
I was wrong about mechanism by which this error occurs. We should discuss with SchedMD team whther this should be logged with "ERROR" or "WARNING" as the [code] outwardly appears to suggest that this ERROR should go away as soon as the script is able to read the resume.log file. Since this script itself is creating the resume.log file, it may be a "first execution" error that is either safe to ignore or able to be worked around by flushing data in earlier writes.

[code]: https://github.com/SchedMD/slurm-gcp/blob/5ddcc635c35da7804521aa3c62358683ae0908d2/scripts/resume.py#L606

PR #1783, opened 2023-09-20
Add support to update mount options for file store instance

harshthakkar01:
### Submission Checklist

### DIFF
--- a/modules/file-system/filestore/README.md
+++ b/modules/file-system/filestore/README.md
@@ -26,6 +26,24 @@ creation will fail when you run `terraform apply`.
 
 [tiers]: https://cloud.google.com/filestore/docs/service-tiers
 
+### Filestore mount options
+After Filestore instance is created, you can mount this to the compute node
+using different mount options. Toolkit uses [default mount options](https://linux.die.net/man/8/mount)
+for all tier services. Filestore has recommended mount options for different
+service tiers which may overall improve performance. These can be found here:
+[recommended mount options.](https://cloud.google.com/filestore/docs/mounting-fileshares)
+While creating filestore module, you can overwrite these mount options as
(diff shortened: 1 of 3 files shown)


modules/file-system/filestore/README.md
```diff
@@ -26,6 +26,21 @@ creation will fail when you run `terraform apply`.
 
 [tiers]: https://cloud.google.com/filestore/docs/service-tiers
 
+### Filestore mount options
+Toolkit uses [default mount options](https://linux.die.net/man/8/mount) for all tiers
+services. Filestore recommends service specific mount options for better performance.
```
cboneti, line 31: s/for all tiers services/for all firestore tiers/

cboneti (changes requested):
Do you think this is enough documentation?

nick-stroud (approved):

modules/file-system/filestore/README.md
harshthakkar01: I reworded the documentation. I am not sure if we should put details like what are the recommended options for which tiers. We have a pointer to the recommended options from filestore. Here I only specified how customers can pass this parameter to update the default mount options. If they want to use different file system they can refer the filestore documentation and pass the tier specific recommended option in similar way.

LMKWYT.

cboneti: I would still replace "all tier services" for "all filestore tiers"

harshthakkar01: Done. Updated "all filestore tiers" to "all tier services"

harshthakkar01:
> Do you think this is enough documentation?

Had an offline discussion with Nick. Updated as per recommendation.

cboneti (approved):

PR #1788, opened 2023-09-22
Warn users that startup scripts are running.

cdunbar13:
This PR will add an warning to the startup-script module that alerts users when startup scripts are running when they login to a new VM.  It will also send an alert that the scripts are starting, or have an ended to any currently logged in users.


modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
+
+FINISH_LINE="startup-script exit status"
+ACTIVE=^activ
```
tpdownes, line 4: is the missing "e" deliberate?

modules/scripts/startup-script/main.tf
```diff
@@ -26,6 +26,19 @@ locals {
     destination = "install_cloud_ops_agent_automatic.sh"
   }] : []
 
+  warnings = [
+    {
+      type        = "shell"
+      content     = "chmod -R o+r /var/log/journal"
+      destination = "access_joural.sh"
```
tpdownes, line 33: typo: "journal"

modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -0,0 +1,59 @@
...
+WARNING=$(
+	cat <<EOF
+** WARNING **: The VM startup scripts for this have not been completed and
```
tpdownes, line 7: I think this text is a bit verbose and maybe difficult for non-native English speakers. We should meet offline to make it shorter and align with Google documentation principles. e.g.

"It is advised that you wait..." can become "Wait...".

modules/scripts/startup-script/main.tf
tpdownes: I'm not excited to change the permissions of an entire directory at boot, although I recognize this is a read-only change.

- it seems "wrong" because it impacts the domain of the system administrator
- a blind recursive chmod could take a long time to execute (would be better as a `find` command or determine very specific files we need to update)
- I haven't been convinced we've rejected alternatives

I'd like to see a very concise list of the situations we find this to be necessary. Probably some set of operating systems and OS Login configurations? If we're effectively solving a problem for 1 or 2 operating conditions, it may be simpler to consider -- for those situations -- a more generic warning (e.g. "startup scripts exist and you should be aware of them") that does not attempt to communicate the precise state of the system to the user.

modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -0,0 +1,59 @@
...
+The following command can also help display relevant information.
+
+Local shell:
+$ sudo journalctl -u google-startup-scripts.service
+EOF
+)
+
+# Check if end of startup scripts has been detected
+IS_ACTIVE=$(systemctl is-active google-startup-scripts.service | grep "${ACTIVE}")
+END_FOUND=$(journalctl -b 0 -u google-startup-scripts.service | grep "${FINISH_LINE}")
```
tpdownes, line 37: Nothing stops a user from starting google-startup-scripts multiple times after a boot. In fact, this is one way we might choose to help the user propagate Ansible updates. Is there a way to limit this command to the most recent activation of the google-startup-scripts.service unit.

(I did not see one in a quick scan)

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
```diff
@@ -30,6 +30,45 @@ readonly E_RUN_OR_DIE=5
 readonly E_MISSING_MANDATORY_ARG=9
 readonly E_UNKNOWN_ARG=10
 
+SUCCESS_MESSAGE=$(
+	cat <<-EOF
+		** NOTICE **: The VM startup scripts have finished running successfully.  
+		It is now safe to start using the system.
+	EOF
+)
+readonly SUCCESS_MESSAGE
```
tpdownes, line 39: is it possible to include this in the declaration? It looks funny coming so many lines later.

tpdownes (changes requested):
Broadly, the issue I see with a big "chmod" is that we are stepping on the toes of the sysadmin. We should discuss offline, but, I think you might consider a 2-stepped approach:

1. A simple PR with the "wall" messages and generic message(s) re: startup-scripts that is not dependent upon POSIX access to the SystemD journal or ability to read state of units.
2. A follow-up analysis / PR that extends the messages to provide more detailed information when it can.

I also have some concern about duplication of warning/error messages across two scripts. That can wait, but give it some thought.

modules/scripts/startup-script/files/running-script-warning.sh
cdunbar13: Yes, captures both `activating` and `active`

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
cdunbar13: I think the issue was that it didn't like the command `cat<<-EOF` with the assignment and readonly.  This was the solution I found after some research.

modules/scripts/startup-script/main.tf
cdunbar13: Will fix typo if we agree this is the way to move forward.

I understand the difficulty with the `chmod`, I don't like having to change permissions either.  That being said, there are a few different things at play:

1. I don't think its easy to determine which files to modify in the journal folder as they're created with what appear to be random suffixes (could get a list of the latest system.journal files or something)
2. The only alternative I can immediately think of, that maintains the error message, is to add users to the systemd-journal group, which is more difficult because users aren't created until login, or to use the `script-warning` file I created before.
   1. `script-warning` can be created by `passed-script.sh` file, which will get the exit code directly, and go around the need for users to need sudo, or to change permissions on any file
   2. The simplest alternative is to just ignore the error for users who log in after the scripts have finished.  The error will print for users who are logged in when the scripts finish.

We can discuss the rest offline.

cdunbar13: Additionally, we could also just add knobs to determine if the permissions for `/var/log/journal` will be given.  If it isn't the `profile.d` scripts won't be able to read anything and users who log in after the scripts will never see an error.

modules/scripts/startup-script/files/running-script-warning.sh
tpdownes: Please document this in a comment.

modules/scripts/startup-script/files/running-script-warning.sh
tpdownes: Well, I'm worried about defining a behavior. Do we target the boot-time activation of the service? Do we target all? Are we reporting success when the boot-time activation succeeds but subsequent activations fail?

And so-on. We should define the behavior and then confirm the code implements it.

PR #1795, opened 2023-09-26
Making CloudSQL to use internal IP address instead of external for Slurm Accounting DB.

ek-nag:
This PR modifies `community/modules/database/slurm-cloudsql-federation` module. It makes communication between the Slurm cluster controller node and SQL database over the private network instead of the external. This way SQL instance does not need to have external IP address. It allows for both Slurm cluster and SQL instance to be completely isolated from the external networks.

The resulting Slurmdbd configuration made by using this updated module on the test cluster:

```bash
[root@hpcsmall-controller scripts]# cat /etc/slurm/slurmdbd.conf 
# slurmdbd.conf
# https://slurm.schedmd.com/slurmdbd.conf.html

DebugLevel=info
PidFile=/var/run/slurm/slurmdbd.pid

################################################################################
#              vvvvv  WARNING: DO NOT MODIFY SECTION BELOW  vvvvv              #
################################################################################

AuthType=auth/munge
AuthAltTypes=auth/jwt
AuthAltParameters=jwt_key=/var/spool/slurm/jwt_hs256.key

DbdHost=hpcsmall-controller

LogFile=/var/log/slurm/slurmdbd.log

SlurmUser=slurm

StorageLoc=slurm_accounting

StorageType=accounting_storage/mysql
StorageHost=10.212.0.3
StoragePort=3306
StorageUser=slurm
StoragePass=mPgWEqy31A8f

################################################################################
#              ^^^^^  WARNING: DO NOT MODIFY SECTION ABOVE  ^^^^^              #
################################################################################
```

Snippet from the test blueprint to demonstrate use case:

```bash
  - id: slurm-sql
    source: community/modules/database/slurm-cloudsql-federation
    kind: terraform
    use: [network1]
    settings:
      sql_instance_name: slurm-sql8-ofe
      tier: "db-g1-small"

  - id: slurm_controller
    source: community/modules/scheduler/schedmd-slurm-gcp-v5-controller
    use:
    - network1
    - compute_partition
    - homefs
    - slurm-sql
    settings:
      instance_image:
        family: slurm-gcp-5-8-hpc-rocky-linux-8
        project: schedmd-slurm-public
```


cboneti:
Eimantas, the PR-validation test is failing.

community/modules/database/slurm-cloudsql-federation/main.tf
```diff
@@ -33,9 +33,27 @@ locals {
   sql_password      = var.sql_password == null ? random_password.password.result : var.sql_password
 }
 
+resource "google_compute_global_address" "private_ip_address" {
```
cboneti, line 36: I would like to have a little more explanation here, are you changing this so it always use the private ip address?  I guess this makes sense for slurm, but don't we want to leave the option to also have a public address?

ikatsardis: I was thinking of this also Carlos. I couldn't come up with a valid use case where one would need this sql database to have a public IP. If needed one would have a direct peering or already another way of accessing the slurm controller node already. In the spirit of reducing public IP addresses and the attack surface, I would prefer if we would remove this as an option. If there is indeed a use case I am missing and it wouldn't be possible to be solved by a direct vpc peering then I agree with you we should have it as an option.

tpdownes: My understanding is that the original use case was inter-regional database replication. Whether public IPs were actually required for that case, I can't say, but that's where I understood it to be coming from.

ikatsardis: That helps Tom thanks. Even the more reasons to actually force private vs public. Copying directly from our website I can see that for our replicas we also state that: "Creating a cross-region replica requires no networking setup. Global VPC uses private IP for replication traffic between regions—eliminating the need of complex VPN and VPC configuration, which would be otherwise needed to set up cross-region networking." and also: "Cross-region replication traffic remains private, without access to and from the public internet, when a private IP option is used." Based on these findings I would tend to agree with Eimantas and make private IP the default and only solution for now. If a use case comes up we can still revisit.

tpdownes: I can state with certainty that this module is the only known reason `nat_ips` exists as an output for the `modules/network/vpc` module.

I also agree that it seems unlikely to me that public IPs were strictly necessary in the first place. I'm also not sure this feature is actively used via the Toolkit. From purely technical perspective, I agree with the conclusion here.

ek-nag:
@cboneti I think I fixed the initial issue in pre-build, however it's still failing with `ERROR: build step 2 "us-central1-docker.pkg.dev/hpc-toolkit-dev/hpc-toolkit-repo/hpc-toolkit-builder" failed: step exited with non-zero status: 1
`, but I don't have permission to see the error. Would you mind checking what's wrong?

cboneti (approved):

PR #1797, opened 2023-09-28
Do not chmod spack installation unless user specifies mode

nick-stroud:
Use built in spack configuration to grant group write permissions. Do not chmod spack install unless user specifies a mode.

### Submission Checklist


douglasjacobsen (approved):

danielahlin (approved):
Looks good. I'm no ansible expert - I assume you've checked that "null" means nothing changes - right?

nick-stroud:
`I assume you've checked that "null" means nothing changes` > I did confirm this but it is really only true in the case that the script is being run as root. Although that is our case, I figured it is just safer to not run that task unless user has provided some input. I have pushed an update in a second commit.

tpdownes (approved):
I'm not convinced that the 2nd commit is really all that helpful, but I'll defer to your judgement as it was discussed.

nick-stroud:
The documentation for `ansible.builtin.file` says things like _ When left unspecified, it uses the current user unless you are root, in which case it can preserve the previous ownership._. My interpretation is that when run as root (as we do) this task will have no effect. If the script were to be run as a user with all inputs null, it would have an effect of changing permissions to that user/group. While this does not affect us in our normal use case, I didn't like having potential side effects when the user's intent was clearly to take no action.

PR #1800, opened 2023-09-30
Attach a (free) license to every built VM image

rohitramu:
This PR ensures that a (free) license is attached to each VM image built using the HPC Toolkit.  I have tested this by deploying the `image-builder.yaml` example blueprint: `<repo-root>/examples/image-builder.yaml`.


tpdownes (approved):
Could you please define the image_licenses parameter in the locals block and then use `local.image_licenses` in the actual build? I know it's equivalent, but as a matter of convention, I'd rather build-in the flexibility to later expose image_licenses to the user and merge in the local value.

Make sure to run `packer fmt` manually afterwards. Not yet automatic.

PR #1803, opened 2023-10-02
Add multigroup support for export and import variables to cleanup-build.py

cdunbar13:
Attempt to improve the cleanup-build.py by adding the ability to export and import variables to multigroup deployments (since it's not done automatically).  

May be worth adding in a flag to activate this function since it's not always necessary.

### DIFF
--- a/tools/cleanup-build.py
+++ b/tools/cleanup-build.py
@@ -50,8 +50,40 @@ def cp_from_gcs(gcs_source_uri: str, local_destination_path: str, project_id: st
     return destination
 
 def unpack_tgz(tar_file: str, destination_folder: str):
-  with tarfile.open(tar_file, "r:gz") as tar:
-    tar.extractall(destination_folder)
+    with tarfile.open(tar_file, "r:gz") as tar:
+        tar.extractall(destination_folder)
+
+# For multi-group deployments - attempt to export all group variables, and 
+# import them to downstream groups.  Dependent on blueprints needing groups to
+# be in dependent order top to bottom
+def export_import_vars(deployment_folder: str):
+    import re, mmap, subprocess, sys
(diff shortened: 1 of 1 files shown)


tools/cleanup-build.py
```diff
@@ -50,8 +50,40 @@ def cp_from_gcs(gcs_source_uri: str, local_destination_path: str, project_id: st
...
+    with tarfile.open(tar_file, "r:gz") as tar:
+        tar.extractall(destination_folder)
+
+# For multi-group deployments - attempt to export all group variables, and 
+# import them to downstream groups.  Dependent on blueprints needing groups to
+# be in dependent order top to bottom
+def export_import_vars(deployment_folder: str):
+    import re, mmap, subprocess, sys
+    # Simple check for multi-group (if all we have is primary and .ghpc we move on)
+    if len(os.listdir(deployment_folder)) > 2:
```
mr0re1, line 62: Terminate early, without increasing depth:
```py
if len(os.listdir(deployment_folder)) <= 2:
  return
...
```

tools/cleanup-build.py
```diff
@@ -50,8 +50,40 @@ def cp_from_gcs(gcs_source_uri: str, local_destination_path: str, project_id: st
...
+        ordered_dirs = []
+        # Get order of folder to simplify import and export order
+        with open(deployment_folder + '/.ghpc/artifacts/expanded_blueprint.yaml', mode="r") as f:
+            for l in f:
+                tmp = re.search(r'\s*- group:\s*([0-9a-z\-_]+)', l)
```
mr0re1, line 67: There is nothing preventing as from putting newline or using quotes, or use flow-style here. We shouldn't rely on `goyaml` implementation details.
I would recommend to use some yaml-parser (e.g. PyYaml) here.

PR #1806, opened 2023-10-03
Install specific version of shellcheck on cloud workstation first boot

nick-stroud:
`apt` will only install up to version v0.7.0 which gives spirous outputs. This is the method we use to install shellcheck on our builder.

### Submission Checklist

### DIFF
--- a/tools/cloud-workstations/configure-hpc-toolkit.sh
+++ b/tools/cloud-workstations/configure-hpc-toolkit.sh
@@ -36,5 +36,11 @@ if [[ ! -f $FLAG ]]; then
 	go install github.com/ramya-rao-a/go-outline@v0.0.0-20210608161538-9736a4bde949
 	go install github.com/rogpeppe/godef@v1.1.2
 	go install golang.org/x/tools/cmd/guru@latest
+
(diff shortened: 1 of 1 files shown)


cdunbar13 (commented):
Isn't this already handled in the Docker script?  I think that installs v0.9.0

nick-stroud:
> Isn't this already handled in the Docker script? I think that installs v0.9.0

Can you provide a link? I suspect you are thinking about the docker file that builds our cloud build runner. If so, I believe that is building an unrelated environment. This pr is trying to replicate that same behavior in the cloud workstation.

cdunbar13:
> > Isn't this already handled in the Docker script? I think that installs v0.9.0
> 
> Can you provide a link? I suspect you are thinking about the docker file that builds our cloud build runner. If so, I believe that is building an unrelated environment. This pr is trying to replicate that same behavior in the cloud workstation.

https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/tools/cloud-workstations/Dockerfile#L36

nick-stroud:
@cdunbar13, you are correct.

PR #1817, opened 2023-10-05
Fix: error on second run when mirror is already defined

nick-stroud:
Without conditional spack fails on second startup with error like "mirror already exists".

### Submission Checklist

### DIFF
--- a/tools/cloud-build/daily-tests/builds/batch-mpi.yaml
+++ b/tools/cloud-build/daily-tests/builds/batch-mpi.yaml
@@ -56,7 +56,7 @@ steps:
     BUILD_ID_SHORT=$${BUILD_ID_FULL:0:6}
     SG_EXAMPLE=examples/serverless-batch-mpi.yaml
 
-    sed -i "s/# spack mirror add .*/spack mirror add --scope site gcs_cache $${SPACK_CACHE_WRF//\//\\\/}/" $${SG_EXAMPLE}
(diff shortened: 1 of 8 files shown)


tools/cloud-build/daily-tests/builds/batch-mpi.yaml
```diff
@@ -56,7 +56,7 @@ steps:
     BUILD_ID_SHORT=$${BUILD_ID_FULL:0:6}
     SG_EXAMPLE=examples/serverless-batch-mpi.yaml
 
-    sed -i "s/# spack mirror add .*/spack mirror add --scope site gcs_cache $${SPACK_CACHE_WRF//\//\\\/}/" $${SG_EXAMPLE}
+    sed -i "s/#   spack mirror add .*/spack mirror add --scope site gcs_cache $${SPACK_CACHE_WRF//\//\\\/}/" $${SG_EXAMPLE}
```
rohitramu, line 59: Just curious, why do we need the extra space?

rohitramu (approved):

tools/cloud-build/daily-tests/builds/batch-mpi.yaml
nick-stroud: This sed is editing the blueprints that were updated, if you look at the change to the blueprint there is now extra space due to the indent from the if. 

The said does not bother with the if because the if is only there in the case of a reboot which will not happen in tests.

PR #1821, opened 2023-10-06
Update the install ramble deps runner based on the spack module

douglasjacobsen:
This merge updates the install_ramble_deps module, by changing the python package name, and using a venv for ramble (as is used in spack).

This fixes issues with the ramble module on rocky linux.

### Submission Checklist

### DIFF
--- a/community/modules/scripts/ramble-setup/templates/install_ramble_deps.yml.tpl
+++ b/community/modules/scripts/ramble-setup/templates/install_ramble_deps.yml.tpl
@@ -17,17 +17,24 @@
   hosts: localhost
   vars:
     ramble_ref: ${ramble_ref}
+    ramble_virtualenv_path: "/usr/local/ramble-python"
   tasks:
   - name: Install dependencies through system package manager
     ansible.builtin.package:
       name:
-      - python
+      - python3
       - python3-pip
       - git
+    register: package
+    changed_when: package.changed
+    retries: 5
+    delay: 10
+    until: package is success
 
   - name: Ensure a recent copy of pip
     ansible.builtin.pip:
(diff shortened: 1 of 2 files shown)


community/modules/scripts/ramble-setup/templates/install_ramble_deps.yml.tpl
```diff
@@ -17,17 +17,24 @@
...
+    register: package
+    changed_when: package.changed
+    retries: 5
+    delay: 10
+    until: package is success
 
   - name: Ensure a recent copy of pip
     ansible.builtin.pip:
       name: pip>=21.3.1
+      virtualenv: "{{ ramble_virtualenv_path }}"
```
nick-stroud, line 37: In Spack we do some magic to make sure that when a user lands on the machine they are using the correct python ([ref](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/0a301051fb4f9b704861eb2d3f8d5eedf1f7aa67/community/modules/scripts/spack-setup/main.tf#L24)). 

I am wondering if after your change the user will not be able to use ramble unless they activate the ramble virtual env. If so how will the user know to do this.

douglasjacobsen: Is there a better way to do this than to have a separate venv for every module? That seems really messy, and if the python deps are different in one app then you can't magically set the venv.

douglasjacobsen: Should be fixed now.

nick-stroud (approved):

PR #1823, opened 2023-10-06
Rename tests

alyssa-sm:
Renames the test names to match the blueprint names so the tgz file and the terraform backend state file can be uploaded to the same folder.


tools/cloud-build/daily-tests/tests/batch-mpi.yml
nick-stroud, line 1: Wait for the batch-mpi test to pass before merging.

nick-stroud (approved):

tools/cloud-build/daily-tests/tests/batch-mpi.yml
alyssa-sm: It passed!

PR #1836, opened 2023-10-13
Origin/caesolution

fschuerm:
This PR adds the cae/cae-slurm.yaml blueprint for the MVP Google HPC CAE solution as well as its accompanying documentation.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -18,6 +18,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [serverless-batch.yaml](#serverless-batchyaml-) ![core-badge]
   * [serverless-batch-mpi.yaml](#serverless-batch-mpiyaml-) ![core-badge]
   * [pfs-lustre.yaml](#pfs-lustreyaml-) ![core-badge]
+  * [cae/cae-slurm.yaml](#cae-slurm-) ![core-badge]
   * [hpc-slurm-ubuntu2004.yaml](#hpc-slurm-ubuntu2004yaml-) ![community-badge]
   * [hpc-intel-select-slurm.yaml](#hpc-intel-select-slurmyaml-) ![community-badge]
   * [pfs-daos.yaml](#pfs-daosyaml-) ![community-badge]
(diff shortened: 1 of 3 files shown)


cboneti (changes requested):

examples/README.md
```diff
@@ -18,6 +18,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [serverless-batch.yaml](#serverless-batchyaml-) ![core-badge]
   * [serverless-batch-mpi.yaml](#serverless-batch-mpiyaml-) ![core-badge]
   * [pfs-lustre.yaml](#pfs-lustreyaml-) ![core-badge]
+  * [cae-slurm.yaml](#cae-slurm-) ![core-badge]
```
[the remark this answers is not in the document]
fschuerm: good catch. I fixed it and pushed it to the PR

examples/cae/README.md
```diff
@@ -0,0 +1,245 @@
...
+> If you created a new project for testing of the CAE solution, the easiest way to
+> eliminate billing is to delete the project.
+
+When you would like to tear down the deployment, each stage must be destroyed.
+Since the `software_installation` and `cluster` depend on the network deployed
+in the `setup` stage, they must be destroyed first. You can use the following
+commands to destroy the deployment.
+
+```bash
+./ghpc destroy cae-slurm --auto-approve
```
cboneti, line 180: I would recommend not adding the --auto-approve here as it will not ask for confirmation or steps. It would be safer let the user interact. What do you think?

examples/cae/cae-slurm.yaml
```diff
@@ -0,0 +1,254 @@
...
+# It sets up the following infrastructure:
+#   * Google's H3 VMs, ideally suited for CAE workloads
+#   * Google's C3-highmem VM, suited for workloads with 16GB/core requirement
+#   * Google's Filestore NFS-based shared storage
+#   * Google's Chrome Remote Desktop
+#   * SLURM workload scheduler
+#
+blueprint_name: cae-slurm
+vars:
+  project_id:
```
cboneti, line 31: nit: usually we add ` ## Set GCP Project ID Here ##` to the project_id, to make it clear for the user that they have to add their project id.

cboneti (commented):
This is excellent work. Thank you for doing this!! 

I left two comments, none of which are blocking, so feel free to disagree. I can approve and merge as soon as you had the opportunity to review them.

examples/cae/README.md
fschuerm: Good point. I agree. I modified this part of the README accordingly

examples/cae/cae-slurm.yaml
fschuerm: Yes, I agree. This makes things clearer. Resolved.

cboneti (approved):

PR #1837, opened 2023-10-13
Update ramble modules to have data file management

douglasjacobsen:
This merge updates the ramble modules to have similar functionality to the spack modules, including data file creation / transfer.

### Submission Checklist


community/modules/scripts/ramble-execute/main.tf
```diff
@@ -15,34 +15,69 @@
...
+}
+
+locals {
+  commands_content = var.commands == null ? "echo 'no ramble commands provided'" : indent(4, yamlencode(var.commands))
 
   execute_contents = templatefile(
     "${path.module}/templates/ramble_execute.yml.tpl",
     {
-      pre_script = ". ${var.spack_path}/share/spack/setup-env.sh && . ${var.ramble_path}/share/ramble/setup-env.sh"
+      pre_script = ". /etc/profile.d/spack.sh && . /etc/profile.d/ramble.sh"
```
nick-stroud, line 28: IIUC you would like the option to use ramble-execute without spack-setup or even without ramble-setup. Previously the existence of the [spack|ramble]_path variables required that these were present. 

This pre script seems problematic since we now can't be sure that either of these files exist. Should we be testing for the files before calling them?

community/modules/scripts/ramble-execute/main.tf
```diff
@@ -15,34 +15,69 @@
...
+    content     = module.startup_script.startup_script
+    destination = "combined_install_spack_${local.combined_md5}.sh"
   }
+
+  bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 4)
+  bucket_name = "ramble-scripts-${local.bucket_md5}-${local.combined_md5}"
+}
+
+resource "google_storage_bucket" "bucket" {
+  count                       = var.gcs_bucket_path != null ? 0 : 1
```
nick-stroud, line 60: I thought this would be problematic when chaining execute modules. When a resource is conditionally created (such as in this case), it can sometimes cause problems for downstream dependent resources because Terraform cannot build out the resource graph without knowing the outcome of applying the upstream conditional. 

I tried to break this by deploying chained ramble-execute modules, but It deployed without problem. I think all of the work we are doing with hashing instead of using a random num is helping. Feel free to close this comment without action. I just wanted to provide some context in case you see an error along these lines.

community/modules/scripts/ramble-execute/outputs.tf
```diff
@@ -14,10 +14,25 @@
  * limitations under the License.
  */
 
+output "startup_script" {
+  description = "Path to the Ramble installation script."
+  value       = module.startup_script.startup_script
+}
+
+output "controller_startup_script" {
+  description = "Path to the Ramble installation script, duplicate for SLURM controller."
```
nick-stroud, line 23: See other comment about similar description. ~~Path to the~~

nick-stroud (changes requested):

community/modules/scripts/ramble-execute/main.tf
douglasjacobsen: Not exactly.

I would like to be able to call `ramble-setup` and `ramble-execute` as separate runners, rather than needing them to be chained.

So, this new format just requires that `*-setup` has run successfully on the machine you're trying to run `*-execute` on. Since both of them create the `/etc/profile.d/*.sh` scripts.

I updated it to be conditional though. Let me know if you dislike the change.

nick-stroud (changes requested):

community/modules/scripts/ramble-execute/variables.tf
```diff
@@ -85,10 +85,33 @@ variable "ramble_runner" {
...
+  type        = string
+  default     = null
+}
+
+variable "spack_profile_script_path" {
+  description = "Path to the Spack profile.d script. Created by an instance of spack-setup"
+  type        = string
+  default     = null
+  validation {
+    condition     = var.spack_profile_script_path != null
```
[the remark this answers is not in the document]
douglasjacobsen: It is technically possible, but the experiments you can run are very limited (almost all of Ramble's application definitions rely on spack right now)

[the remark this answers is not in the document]
douglasjacobsen: I wasn't aware that not defining a default would do the same thing, but yes I wanted an error message that said something like "This can be defined via chaining" but maybe that's better off in the description?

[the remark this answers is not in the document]
douglasjacobsen: I made this optional now. With a default set to the same default as the spack module. Given it's only conditionally loaded, it should be fine but let me know if you have any suggestions to improve this.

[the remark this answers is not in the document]
douglasjacobsen: I updated the description and used the pattern you mentioned.

community/modules/scripts/spack-execute/variables.tf
```diff
@@ -87,3 +88,13 @@ variable "gcs_bucket_path" {
   description = "The GCS path for storage bucket and the object, starting with `gs://`."
   type        = string
 }
+
+variable "spack_profile_script_path" {
+  description = <<-EOT
+    Path to the Spack profile.d script. Created by an instance of spack-setup.
+    Can be defined explicitly, or by chaining an instance of a spack-setup module
+    through a `use` setting.
+    Defaults to /etc/profile.d/spack.sh if not set.
```
nick-stroud, line 97: Remove this line. This module does not express any default.

community/modules/scripts/spack-execute/main.tf
```diff
@@ -40,7 +40,8 @@ locals {
     destination = "spack_execute_${local.execute_md5}.yml"
   }
 
-  runners = concat([var.spack_runner], local.data_runners, [local.execute_runner])
+  previous_runners = var.spack_runner != null ? [var.spack_runner] : []
```
nick-stroud, line 43: This null check will not be needed if you remove the default.

community/modules/scripts/ramble-execute/main.tf
```diff
@@ -15,34 +15,56 @@
...
+}
+
+locals {
+  commands_content = var.commands == null ? "echo 'no ramble commands provided'" : indent(4, yamlencode(var.commands))
 
   execute_contents = templatefile(
     "${path.module}/templates/ramble_execute.yml.tpl",
     {
-      pre_script = ". ${var.spack_path}/share/spack/setup-env.sh && . ${var.ramble_path}/share/ramble/setup-env.sh"
+      pre_script = "if [ -f ${var.spack_profile_script_path} ]; then . ${var.spack_profile_script_path}; fi; if [ -f ${var.ramble_profile_script_path} ]; then . ${var.ramble_profile_script_path}; fi;"
```
nick-stroud, line 28: If `ramble-execute: use: [ramble-setup] required` is true then I think you should drop the check for ramble profile and just call it.

community/modules/scripts/ramble-execute/variables.tf
```diff
@@ -37,22 +75,36 @@ variable "commands" {
 }
 
 variable "ramble_runner" {
-  description = <<-EOT
-    Ansible based startup-script runner from a previous Ramble step
-    Will automatically be prepended to the script generated by this module.
-  EOT
+  description = "Runner from previous ramble-setup or ramble-execute to be chained with scripts generated by this module."
   default     = null
```
nick-stroud, line 79: If `ramble-execute: use: [ramble-setup] required` is true then I think this should have no default.

nick-stroud (changes requested):

nick-stroud:
My comments are under the assumption that the following usage is permitted:

```
spack-execute: use: [spack-setup] required
ramble-execute: use: [ramble-setup] required
ramble-execute: use: [spack-setup] optional
```

If we get a little down the road and we want to change these assumptions, and the expression of theses assumptions in the code, I think that is fine. But if the above are our expectations then the code should reflect that.

community/modules/scripts/ramble-execute/variables.tf
```diff
@@ -37,22 +75,37 @@ variable "commands" {
...
-    condition     = var.ramble_runner == null || try(var.ramble_runner["type"] == "ansible-local", false)
-    error_message = "Ramble runner should be of type 'ansible'."
-  }
-  validation {
-    condition     = var.ramble_runner == null || contains(keys(var.ramble_runner), "content")
-    error_message = "Ramble runner should contain a 'content' key."
-  }
+}
+
+variable "gcs_bucket_path" {
```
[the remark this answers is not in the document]
douglasjacobsen: Yeah, after going through this I think startup scripts probably could use some cleanup regarding their bucket story.

community/modules/scripts/ramble-execute/main.tf
nick-stroud: discussed and agreed to offline.

nick-stroud (approved):

nick-stroud:
Running tests.

PR #1840, opened 2023-10-13
Organize test files

alyssa-sm:
Reorganized the test files for Ansible integration. Separated the files into two folders named test-setup and test-validation to better distinguish the difference.


rohitramu (approved):
Looks good!  Please run all the tests before merging.

PR #1841, opened 2023-10-14
Add support for reading `metadata.yaml` in CFT-format, fallack to har…

mr0re1:
…dcoded logic.

* Remove `RequiredApis` from `ModuleInfo`, replace with `Metadata`;
* Refactor `resreader_test.go` to re-use embedded files for localFS test;
* Use `LastIndex` instead `Index` on module source trimming to avoid potential errors;
* Move `defaultAPIList` to `metadata_legacy.go`.

Created BP:

```yaml
...
- id: gg
  source: github.com/terraform-google-modules/terraform-google-vm//modules/mig
```

Modified `metadata.go:GetMetadataSafe`:

```go
panic(fmt.Sprintf("%#v", mtd.Spec.Requirements.Services))
```

Got:

```sh
> make && ./ghpc create gg.yaml
panic: []string{"cloudresourcemanager.googleapis.com", "storage-api.googleapis.com", "serviceusage.googleapis.com", "compute.googleapis.com", "iam.googleapis.com"}
```

Matches: https://github.com/terraform-google-modules/terraform-google-vm/blob/master/modules/mig/metadata.yaml#L294

### DIFF
--- a/pkg/modulereader/metadata.go
+++ b/pkg/modulereader/metadata.go
@@ -1,0 +1,78 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package modulereader
+
+import (
+	"errors"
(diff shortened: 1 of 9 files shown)


tpdownes (changes requested):
The function `GetMetadata` does not appear to error if I give an invalid format. For example, if I add `community/modules/scheduler/schedmd-slurm-gcp-v5-controller/metadata.yaml` with this content:

```yaml
spec:
  requirements:
    services: yadayadayada
```

it returns a `nil` error and the empty list for `mtd.Spec.Requirements.Services`.

Other than that, I believe this is a very valuable addition to the Toolkit.

Please also add a label to the PR that will include it in the release notes.

mr0re1:
@tpdownes  thank you for being so thorough with review! There was a typo, fixed.

tpdownes (approved):
That change will certainly do it. LGTM.

PR #1846, opened 2023-10-17
Update image on chrome remote desktop from ubuntu 20.04 to Debian 12

cdunbar13:
Changed images for stability.  Briefly tested by checking for nvidia drivers and checking that the remote desktop booted correctly.  Running integration tests to make sure it's working on those.

The ramble updates are just snuck in as something I missed with the updated install_ansible PR I just merged.  Just making the link to the startup-script commit uniform.


community/modules/remote-desktop/chrome-remote-desktop/scripts/configure-grid-drivers.yml
```diff
@@ -88,8 +102,10 @@
     - name: Install GPU driver
       ansible.builtin.shell: |
         #jinja2: trim_blocks: "True"
-        {% if ansible_distribution_release == "jammy" %}
+        {% if ansible_distribution_release == "jammy"  %}
```
nick-stroud, line 105: nit: remove space added.

community/modules/remote-desktop/chrome-remote-desktop/scripts/configure-grid-drivers.yml
```diff
@@ -88,8 +102,10 @@
...
         CC=gcc-12 /tmp/NVIDIA-Linux-x86_64-525.125.06-grid.run --silent
+        {% elif ansible_distribution_release == "bookworm" %}
+        /tmp/NVIDIA-Linux-x86_64-525.105.17-grid.run --silent
```
nick-stroud, line 108: Can we find a way to not have these hard coded twice. I would suggest creating a dictionary with the part of the variable path (`vGPU14.2/NVIDIA-Linux-x86_64-510.85.02-grid.run`) and then use that in get and install. See how packages is done. To include `vGPU14.2` part you will have to use dest: /tmp/{path name}.

Another option would be to specify the same dest for each driver but I think I like leaving a breadcrumb of what driver was installed (and not renaming the file).

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -39,7 +39,8 @@ apt_wait() {
 install_python_deps() {
 	if [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release 2>/dev/null ||
 		grep -qi ubuntu /etc/os-release 2>/dev/null; then
-		apt-get install -y python3-distutils
+		apt-get update
```
nick-stroud, line 42: I cc'ed you on a task named "address release info changes for apt-get update". Please take a look and address here if needed.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
```diff
@@ -267,6 +267,15 @@
       ansible.builtin.debug:
         var: suspend_output.stdout_lines
 
+    - name: Check for stockout errors
```
nick-stroud, line 270: Please move the check for stockout errors to a separate PR.

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
```diff
@@ -62,8 +62,8 @@ variable "instance_image" {
     project = string
```
nick-stroud, line 62: Since we are no longer going to test Debian 11, WDYT of removing it from our supported images?

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
```diff
@@ -62,8 +62,8 @@ variable "instance_image" {
...
   })
   default = {
-    family  = "ubuntu-2204-lts"
-    project = "ubuntu-os-cloud"
+    family  = "debian-12"
```
nick-stroud, line 65: Please include an update to these docs to reflect the new state after this change: https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/docs/vm-images.md#supported-features

nick-stroud (changes requested):

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -39,7 +39,8 @@ apt_wait() {
...
+		apt-get install -y python3-distutils python3-venv
```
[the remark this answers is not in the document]
nick-stroud: Ignore. I see this is proposed in separate pr.

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -170,17 +172,18 @@ main() {
...
-	${python_path} -m virtualenv "${venv_path}"
+	${python_path} -m venv "${venv_path}"
 	venv_python_path=${venv_path}/bin/python3
 
+	# Upgrade pip if necesary
+	pip_version=$(${venv_python_path} -m pip --version | sed -nr 's/^pip ([0-9]+\.[0-9]+).*$/\1/p')
+	pip_major_version=$(echo "${pip_version}" | cut -d '.' -f 1)
+	if [ "${pip_major_version}" -lt "${REQ_PIP_MAJOR_VERSION}" ]; then
+		${venv_python_path} -m pip install --upgrade pip
+	fi
```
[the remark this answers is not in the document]
nick-stroud: Ignore. I see this is proposed in separate pr.

modules/scripts/startup-script/files/install_ansible.sh
nick-stroud: Ignore. I see this is proposed in separate pr.

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
cdunbar13: Done

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
cdunbar13: Do you mean from a larger HPC-Toolkit perspective?  

It might be worth bringing in Debian 12 as supported, setting Debian 11 as deprecated for a few months, then removing it.  I couldn't say that Debian 12 won't break anything.

community/modules/remote-desktop/chrome-remote-desktop/scripts/configure-grid-drivers.yml
cdunbar13: Done

community/modules/remote-desktop/chrome-remote-desktop/scripts/configure-grid-drivers.yml
cdunbar13: Done

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
nick-stroud: No, just from this module. 

The description above still says deb 11 is supported. It does not mention deb 12. That should probably be updated. You could also update the ansible to only support deb 12. I am not sure if it is validated anywhere else.

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
nick-stroud: SG. Nit: I might put a `*` in the untested boxes and say a note below like. "Not yet tested"

nick-stroud (changes requested):

community/modules/remote-desktop/chrome-remote-desktop/variables.tf
```diff
@@ -65,6 +69,16 @@ variable "instance_image" {
...
   }
+
+  validation {
+    condition     = contains(["ubuntu-2204-lts", "debian-12"], var.instance_image.family)
+    error_message = "In var.instance_image, the 'family' field must be 'ubuntu-2204-lts' or 'debian-12'."
+  }
+
+  validation {
+    condition     = contains(["ubuntu-os-cloud", "debian-cloud"], var.instance_image.project)
+    error_message = "In var.instance_image, the 'project' field must be 'ubuntu-os-cloud' or 'debian-cloud'."
```
nick-stroud, line 80: These validations are overly restrictive. What if I want to build a custom image based off these images. Also, I believe this will break the slurm use case which uses a slurm image.

I think the best we can say is that "if you use project x, then you must use family y".

docs/vm-images.md
nick-stroud, line 1: We now say Slurm Deb 12 is supported. Make sure that slurm crd test passes and blueprint is updated.

nick-stroud (changes requested):

nick-stroud:
I realize there is some scope creep on this task. My view on it is that the original goal was to "switch the default to Debian 11", but this PR became "upgrade CRD to Debian 12". 

 In the end, it feels like the right thing long term but I am just trying to put some narrative on how we got here.

I am happy with this as a single PR as it is not overly large, but for reference I think it could be broken up into several PRs:
- Enable Deb 12 and switch default
- Migrate Slurm test and support to Deb 12
- Deprecate Deb 11

PR #1849, opened 2023-10-18
Forces VMs to be replaced when placement policy is destroyed

nick-stroud:
Tested: 
- Deploy VM with no placement
- Add placement with max_distance: 1; re-deploy
- Edit `max_distance` from 1 -> 2; re-deploy
- remove placement; re-deploy

Each time it recognizes that VM must be destroyed.

Note: If one does not turn `auto_delete_boot_disk: false` you will still get an error on second apply as it will not find disk. This can be worked around by doing another apply. This disk issue is a known pain point that is unrelated to the placement policy error.

### Submission Checklist

### DIFF
--- a/modules/compute/vm-instance/README.md
+++ b/modules/compute/vm-instance/README.md
@@ -195,6 +195,7 @@ limitations under the License.
 | [google-beta_google_compute_resource_policy.placement_policy](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_compute_resource_policy) | resource |
 | [google_compute_disk.boot_disk](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk) | resource |
 | [null_resource.image](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| [null_resource.replace_vm_trigger_from_placement](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [google_compute_image.compute_image](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_image) | data source |
 
 ## Inputs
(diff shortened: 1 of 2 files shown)


tpdownes (changes requested):
I trust that this works but I think `google_compute_resource_policy.placement_policy` might be a list/array all by itself and replace_triggered_by takes a list. So I think the solution might be a bit fragile. Can you wrap the array with `flatten` and repeat test?

nick-stroud:
> Can you wrap the array with `flatten` and repeat test?

I tried but I get `Unexpected expression found in replace_triggered_by.` or `A static list expression is required.`.

From poking around online it looks like `replace_triggered_by` requires "static references" and cannot accept an expression. 

Additionally the documentation for `replace_triggered_by` states:
> If the reference is to a resource with multiple instances, a plan to update or replace any instance will trigger replacement.

So I think they have accounted for it to take a reference to a array resource, as is in this case. 

I propose to leave it as is.

tpdownes (approved):
I was not entirely sure how to read

>If the reference is to a resource with multiple instances, a plan to update or replace any instance will trigger replacement.

but reading it again, it does feel like yours is the correct interpretation.

nick-stroud:
Spot check test failed in way that makes it seem that standard no placement policy case does not work, although this was tested manually. 

I am dismissing approval so this doesn't get merged until I have a chance to debug.

nick-stroud:
While it looks like `replace_triggered_by` can handle an array of instances, it does not seem to handle the case when count=0. I have updated to rely on a null_resource as a proxy for a trigger. I am dubious of this methodology but wanted to put something forward that is at least working as expected.

tpdownes (approved):
I see what you are doing here. I think the most direct route would be this:

1. Create a random suffix and have its lifecycle managed via the same null_resource you've created.
2. Apply that random suffix to the name of the resource policy
3. Remove the lifecycle from the VM

You can consider this at nit level but I think the basic problem is that the self_link of the resource policy isn't changing and you probably should change it if you're changing its properties.

nick-stroud:
I cannot get rid of the lifecycle on the VM. I get the following error:
```
Error: Error adding resource policies: googleapi: Error 400: Invalid resource usage: 'Instance must be in stopped state'., invalidResourceUsage
```

It looks like it still tries to modify the resource policy on the live VM instead of destroying the VM.

PR #1857, opened 2023-10-19
Update install_ansible.sh to install python3-venv and be compatible with Debian 12

cdunbar13:
This change is a precursor to the Debian 12 update to CRD, which will be updated to be rebased off this branch.

This change is necessary because Debian 12's python environment is externally managed and using pip to install virtualenv breaks during the startup scripts.  Instead, I've moved to installing python3-venv through apt and updating pip from within the virtual environment.  

Additionally, at least on the Ubuntu image, python3-venv isn't immediately available with the apt list, so an apt-update is required before installing any packages.


nick-stroud:
I am reassigning to @tpdownes who has much more context on the python environment we setup up for ansible. I will stay on as I am interested interested in the review.

PR #1860, opened 2023-10-19
 Update install_ansible.sh to install python3-venv and be compatible with Debian 12

cdunbar13:
This change is a precursor to the Debian 12 update to CRD, which will be updated to be rebased off this branch.

This change is necessary because Debian 12's python environment is externally managed and using pip to install virtualenv breaks during the startup scripts. Instead, I've moved to installing python3-venv through apt and updating pip from within the virtual environment.

Additionally, at least on the Ubuntu image, python3-venv isn't immediately available with the apt list, so an apt-update is required before installing any packages.

### DIFF
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -15,6 +15,8 @@
 
 REQ_ANSIBLE_VERSION=2.11
 REQ_ANSIBLE_PIP_VERSION=4.10.0
+REQ_PIP_WHEEL_VERSION=0.41.2
+REQ_PIP_SETUPTOOLS_VERSION=59.6.0
 REQ_PIP_MAJOR_VERSION=21
 REQ_PYTHON3_VERSION=6
 
@@ -39,7 +41,8 @@ apt_wait() {
 install_python_deps() {
 	if [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release 2>/dev/null ||
 		grep -qi ubuntu /etc/os-release 2>/dev/null; then
-		apt-get install -y python3-distutils
+		apt-get update --allow-releaseinfo-change-origin --allow-releaseinfo-change-label
+		apt-get install -y python3-distutils python3-venv
 	fi
 }
 
@@ -81,7 +84,7 @@ install_python3_yum() {
 		echo "Unsupported version of centos/RHEL/Rocky"
 		return 1
 	fi
-	yum install --disablerepo="*" --enablerepo="${enable_repo}" -y python3 python3-pip
+	yum install --disablerepo="*" --enablerepo="${enable_repo}" -y python3 python3-pip python3-venv
 	python_path=$(rpm -ql python3 | grep 'bin/python3$')
 }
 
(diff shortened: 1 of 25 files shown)


tpdownes (changes requested):
Made offline request to alter invocation of `apt-get update` with extra flags due to known issues.

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -170,16 +174,19 @@ main() {
...
+	# Create pip virtual environment for HPC Toolkit
+	${python_path} -m venv "${venv_path}"
+	venv_python_path=${venv_path}/bin/python3
+
+	# Upgrade pip if necesary
+	pip_version=$(${venv_python_path} -m pip --version | sed -nr 's/^pip ([0-9]+\.[0-9]+).*$/\1/p')
 	pip_major_version=$(echo "${pip_version}" | cut -d '.' -f 1)
 	if [ "${pip_major_version}" -lt "${REQ_PIP_MAJOR_VERSION}" ]; then
-		${python_path} -m pip install --upgrade pip
+		${venv_python_path} -m pip install --upgrade pip
```
tpdownes, line 186: Does Debian 12 not allow for the system `pip` to be upgraded anymore? I do think it is beneficial to upgrade the system pip for users when we can.

The version of `pip` includes with some of our OSes is so old it can't properly resolve modern Python packages dependencies.

cdunbar13: You can't easily install anything for the system using pip.  You get an error stating that the system is managed externally and that if you want to install anything via pip, you should do so in a virtual environment.

I believe I've moved the pip upgrade to within the virtual environment, which is effectively the same, unless we're depending on this script to install system-wide packages.  We could use apt to install a newer version, or figure out a workaround if it's important, but it's confusing if we need to make system wide changes for a script that is installing things to a virtual environment.

tpdownes: The concern I have is that upgrading the system `pip` was a thing that I liked because, for better or worse, users install packages system-wide all the time.

The version of `pip` that comes with CentOS 7 for example, is 9.0.3. Going by memory, if you attempt `pip install google-cloud-storage` it will install a version of `google-cloud-storage` that is not compatible with Python 3.6. I could not tell you the complex set of Python package dependency resolution, but the long and short of it is that `pip < 20.3` is awful at handling conflicts and there is some lower version that began to understand things like "this package only works with Python 3.x" and above.

I recommend trying installation of google-cloud-storage on CentOS 7 with 9.0.3. I believe it will install something that is clearly not compatible with 3.6.

- https://pypi.org/project/google-cloud-storage/2.0.0/ (last to support 3.6)
- https://pypi.org/project/google-cloud-storage/2.1.0/ (dropped support)
- https://github.com/googleapis/python-storage/releases/tag/v2.1.0

On Debian 12, this isn't a concern in the first place since it comes with [23.0.1](https://packages.debian.org/bookworm/python3-pip) and _should_ bypass the upgrade in the first place.

I'd prefer to see this revert to upgrading the system-wide pip along with a comment that this may become unnecessary at some point in the future.

I still also advise testing and upgrading the pip inside the virtual environment. It does not always "come along for the ride".

tpdownes: In a quick search I also noticed that there is a "--break-system-packages" (or similar setting). Please do not use that.

The Debian 12 Behavior is Good And I Like It, But We Also Support Python 3.6 / pip 9.0.3 on CentOS and Rocky 8.

cdunbar13:
Install ansible times per OS:

- Debian 12: 1'25''
- Debian 11: 1'51''
- Rocky 8: 1'22''
- Ubuntu 20.04: 1'45''
- Ubuntu 22.04: 1'28''

PR #1861, opened 2023-10-20
Differentiate between key and value part of maps item path

mr0re1:
Currently error path points to a "value" part of key value path `Root.Group.At(i).Modules.At(j).Settings.Dot("scheduling")`. Split this path on two:

```py
...Settings.Key("scheduling")
...Settings.Dot("scheduling")
```

Example of problem:

```yaml
  settings:
    scheduling:
      maintenanceInterval: "PERIODIC"
```

Error:
```sh
200:         maintenanceInterval: "PERIODIC"
```

After fix:
```sh
199:       scheduling:
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -715,7 +715,7 @@ func (bp *Blueprint) evalVars() error {
 			}
 			if used[ref.Name] == 1 {
 				return BpError{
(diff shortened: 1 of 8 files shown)


mr0re1:
Back to drawing board. we can do better.

PR #1863, opened 2023-10-20
Render YAML-parsing errors in "rich" way

mr0re1:
* Make `NewYamlCtx` to always return meaningfull context that can be used to render YAML parsing errors;

* Parse `yaml.v3` error messages to extract line#;

* Add special type of `Path` to be used to point places outside of Blueprint schema;

* Do not add "arrow ^" if `Column == 0`;

* UNRELATED: supply `deployment_name is not set`-error with path.

```yaml
...
eldorado{}
...
```

```sh
# BEFORE
failed to parse the blueprint in tst.yaml, check YAML syntax for errors, err=yaml: line 6: could not find expected ':'

# AFTER
Error: YAML parsing error: could not find expected ':'
6: eldorado{}
```

```yaml
...
deployment_groups:
- group: primary
  shmodule: 8
  modules:
  - id: homefs
    source:
      type: git
```

```sh
# BEFORE
failed to parse the blueprint in tst.yaml, check YAML syntax for errors, err=yaml: unmarshal errors:
  line 9: field shmodule not found in type config.DeploymentGroup
  line 13: cannot unmarshal !!map into string

# AFTER
Error: YAML parsing error: field shmodule not found in type config.DeploymentGroup
9:   shmodule: 8

Error: YAML parsing error: cannot unmarshal !!map into string
13:       type: git
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -100,7 +100,7 @@ func printAdvancedInstructionsMessage(deplDir string) {
 func expandOrDie(path string) config.DeploymentConfig {
 	dc, ctx, err := config.NewDeploymentConfig(path)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal(renderError(err, ctx))
 	}
 	// Set properties from CLI
 	if err := setCLIVariables(&dc.Config, cliVariables); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/create.go
```diff
@@ -193,11 +193,23 @@ func renderError(err error, ctx config.YamlCtx) string {
...
+	if line < 0 {
+		line = 0
+	}
+	if line >= len(ctx.Lines) {
+		line = len(ctx.Lines) - 1
+	}
+
 	pref := fmt.Sprintf("%d: ", pos.Line)
-	arrow := strings.Repeat(" ", len(pref)+pos.Column-1) + "^"
+	arrow := " "
```
cdunbar13, line 205: Why a single space?

cdunbar13 (approved):
LGTM, just one question.

cmd/create.go
mr0re1: There was a strange behavior I didn't dig into much - with  empty `arrow` either format or print did swallow the last "empty" line, not sure why, but ` ` did fix it.

PR #1870, opened 2023-10-23
Point path of mapping item to key instead of value position

mr0re1:
Currently error path points to a "value" part of key value path `Root.Group.At(i).Modules.At(j).Settings.Dot("scheduling")`.

Example of problem (`scheduling` is not supported):

```yaml
  settings:
    scheduling:
      maintenanceInterval: "PERIODIC"
```

Error:
```sh
200:         maintenanceInterval: "PERIODIC"
```

After fix:
```sh
199:       scheduling:
```


cdunbar13 (approved):
LGTM

PR #1871, opened 2023-10-24
Add Slurm V6 controller, partition & nodeset modules

mr0re1:
# Changes from V5 to V6

## REMOVED obsolete variables:
* `enable_slurm_gcp_plugins`
* `enable_reconfigure`
* `enable_cleanup_subscriptions`

## Added login startup script:
* added `login_startup_script` similar to compute and controller ones;

## New logging variables:
* `enable_debug_logging` & `extra_logging_flags`;

## Limit `static_ips` to just one element
SlurmV6 takes only one IP address in `static_ip`, so we need to limit `static_ips` to at most one element.

## Do not output anything from controller module.
* `cloud_logging_filter` & `pubsub_topic` were used for login module in V5 but dependency direction was changed in V6;
* `controller_instance_id` - doesn't seem to be used anywhere;

## **network** moved from partition to nodeset

# TODOs:
### Forward-port variables from V5 to V6:
* `additional_networks`
* `nic_type`
* `access_config`
* `total_egress_bandwidth_tier`
*

## SKIPPED (present in SchedMD code, but omitted in Toolkit):
* `login_network_storage` - use same logic as V5, use only common `network_storage`;
* `network_ip` - use same logic as V5, use `static_ips` instead;
* `spot` (+ `termination_action`) - not present in V5, omit in ToolkitV6, though SchedMD supports it. TODO(?); `controller` only, `nodeset` V6 will support it;
* `network(network_self_link)` - V6 cluster doesn't provide `network` variable. Fallback to just `subnetwork_self_link`. TODO;

## Allow many-to-many relation between nodeset and partition
Currently it's implicitly limited to many-nodesets to one-partition.
AI: after aggregation of all nodesets into a list, remove all duplicates.

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
@@ -1,0 +1,200 @@
+## Description
+
+This module creates a nodeset data structure intended to be input to the
+[schedmd-slurm-gcp-v6-partition](../schedmd-slurm-gcp-v6-partition/) module.
+
+Nodesets allow adding heterogeneous node types to a partition, and hence
+running jobs that mix multiple node characteristics. See the [heterogeneous jobs
+section][hetjobs] of the SchedMD documentation for more information.
+
+To specify nodes from a specific nodesets in a partition, the [`--nodelist`]
+(or `-w`) flag can be used, for example:
+
+```bash
+srun -N 3 -p compute --nodelist cluster-compute-group-[0-2] hostname
+```
+
+Where the 3 nodes will be selected from the nodes `cluster-compute-group-[0-2]`
+in the compute partition.
+
+Additionally, depending on how the nodes differ, a constraint can be added via
+the [`--constraint`] (or `-C`) flag or other flags such as `--mincpus` can be
+used to specify nodes with the desired characteristics.
+
+[`--nodelist`]: https://slurm.schedmd.com/srun.html#OPT_nodelist
+[`--constraint`]: https://slurm.schedmd.com/srun.html#OPT_constraint
+[hetjobs]: https://slurm.schedmd.com/heterogeneous_jobs.html
+
+### Example
+
+The following code snippet creates a partition module using the `nodeset`
+module as input with:
+
+* a max node count of 200
+* VM machine type of `c2-standard-30`
+* partition name of "compute"
+* default nodeset name of "ghpc"
+* connected to the `network` module via `use`
+* nodes mounted to homefs via `use`
+
+```yaml
+- id: nodeset
+  source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
+  use:
+  - network
+  settings:
+    node_count_dynamic_max: 200
+    machine_type: c2-standard-30
+
+- id: compute_partition
+  source: community/modules/compute/schedmd-slurm-gcp-v6-partition
+  use:
+  - homefs
+  - nodeset
+  settings:
+    partition_name: compute
+```
+
+## Custom Images
+
+For more information on creating valid custom images for the node group VM
+instances or for custom instance templates, see our [vm-images.md] documentation
+page.
+
+[vm-images.md]: ../../../../docs/vm-images.md#slurm-on-gcp-custom-images
+
+## GPU Support
+
+More information on GPU support in Slurm on GCP and other HPC Toolkit modules
+can be found at [docs/gpu-support.md](../../../../docs/gpu-support.md)
+
+### Compute VM Zone Policies
+
+The Slurm on GCP nodeset module allows you to specify additional zones in
+which to create VMs through [bulk creation][bulk]. This is valuable when
+configuring partitions with popular VM families and you desire access to
+more compute resources across zones.
+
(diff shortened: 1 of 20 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
```diff
@@ -0,0 +1,73 @@
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
cboneti, line 1: Please add relevant info on the readmes.

cboneti (commented):
Please fix the READMEs and set default image family to Rocky.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
```diff
@@ -0,0 +1,204 @@
...
+This creates a controller node with the following attributes:
+
+* connected to the primary subnetwork of `network`
+* the filesystem with the ID `homefs` (defined elsewhere in the blueprint)
+  mounted
+* One partition with the ID `compute_partition` (defined elsewhere in the
+  blueprint)
+* machine type upgraded from the default `c2-standard-4` to `c2-standard-8`
+
+### Live Cluster Reconfiguration (`enable_reconfigure`)
```
cboneti, line 45: isn't `enable_reconfigure` gone now?

cboneti (commented):
one minor change as you left the enable_reconfigure key there. Either remove it, or explain it is no longer needed.

nick-stroud:
`login_network_storage - use same logic as V5, use only common network_storage` > I think this is worth some discussion. Personally I find the way that login magically inherits network storage from the controller to be confusing to users (Batch does the same thing I think and I don't really like it there either). The blueprint seems much more explicit and clear when the login has to declare its own dependencies.

PS. I have not reviewed the code and am mostly speaking from a v5 perspective. Maybe things have changed significanly with the inversion of the login/controller dependency order.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
```diff
@@ -0,0 +1,200 @@
...
+- id: nodeset
+  source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
+  use:
+  - network
+  settings:
+    node_count_dynamic_max: 200
+    machine_type: c2-standard-30
+
+- id: compute_partition
+  source: community/modules/compute/schedmd-slurm-gcp-v5-partition
```
harshthakkar01, line 50: Correct me if I am wrong. Should this be `schedmd-slurm-gcp-v6-partition` ?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -0,0 +1,388 @@
...
+  default     = null
+}
+
+variable "enable_spot_vm" {
+  description = "Enable the partition to use spot VMs (https://cloud.google.com/spot-vms)."
+  type        = bool
+  default     = false
+}
+
+variable "spot_instance_config" {
```
harshthakkar01, line 289: should this only be used when `enable_spot_vm` is true ?

community/modules/compute/schedmd-slurm-gcp-v6-partition/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  use_placement = [for ns in var.partition_conf : ns.nodeset_name if ns.enable_placement]
+
+  partition = {
+    default              = var.is_default
+    enable_job_exclusive = var.exclusive
+    network_storage      = var.network_storage
+    partition_conf       = var.partition_conf
+    partition_name       = var.partition_name
+    partition_nodeset    = [for ns in var.nodeset : ns.nodeset_name]
+    # partition_nodeset_dyn
```
harshthakkar01, line 26: Is this like, we want to update this in the future ?

harshthakkar01:
Also, is this PR ready to go to develop ? Just curious since we have slurm_v6 branch as well.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
mr0re1: Correct. We would add a validation for this if need arises.

community/modules/compute/schedmd-slurm-gcp-v6-partition/main.tf
mr0re1: Yes, it's TODO, removed commented lines

mr0re1:
> Also, is this PR ready to go to develop ? Just curious since we have slurm_v6 branch as well.

Yes, the plan is to land it into `develop` and remove `slurm_v6` branch.

mr0re1:
> login_network_storage - use same logic as V5, use only common network_storage > I think this is worth some discussion. Personally I find the way that login magically inherits network storage from the controller to be confusing to users (Batch does the same thing I think and I don't really like it there either). The blueprint seems much more explicit and clear when the login has to declare its own dependencies.

Will create a task specifically for this.

cboneti (approved):

PR #1877, opened 2023-10-25
Update ansible's usage of virtualenv to venv

cdunbar13:
Because of a fix to install_ansible.sh, we need to switch to using venv in the ansible scripts (virtualenv is no longer installed in installed_ansible.sh).


tpdownes (approved):
LGTM. Please merge once previous failing tests have passed.

PR #1879, opened 2023-10-25
Move set module settings validation into separate function

mr0re1:
**Motivation:** `applyGlobalVariables` is a wrong place to do it.

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -61,8 +61,42 @@ func (dc *DeploymentConfig) expand() error {
 		return err
 	}
 
+	if err := validateInputsAllModules(dc.Config); err != nil {
+		return err
+	}
+
 	dc.Config.populateOutputs()
 	return nil
+}
+
+func validateInputsAllModules(bp Blueprint) error {
+	errs := Errors{}
+	for ig, g := range bp.DeploymentGroups {
+		for im, m := range g.Modules {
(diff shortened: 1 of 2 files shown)


pkg/config/expand.go
```diff
@@ -61,10 +61,44 @@ func (dc *DeploymentConfig) expand() error {
...
+	for ig, g := range bp.DeploymentGroups {
+		for im, m := range g.Modules {
+			p := Root.Groups.At(ig).Modules.At(im)
+			errs.Add(validateModuleInputs(p, m, bp))
+		}
+	}
+	return errs.OrNil()
+}
+
+func validateModuleInputs(mp modulePath, m Module, bp Blueprint) error {
```
rohitramu, line 83: nit: The names seem kind of confusing at first glance (it becomes more obvious after looking at the code though).  Maybe these names might be a little bit easier to understand?

```suggestion
func validateInputsAllModules(bp Blueprint) error {
	errs := Errors{}
	for ig, g := range bp.DeploymentGroups {
		for im, m := range g.Modules {
			p := Root.Groups.At(ig).Modules.At(im)
			errs.Add(validateModuleInputs(p, m, bp))
		}
	}
	return errs.OrNil()
}

func validateModuleInputs(mp modulePath, m Module, bp Blueprint) error {
```

rohitramu (approved):

PR #1881, opened 2023-10-25
Simplify `DeploymentName` logic, under assumption that blueprint is valid

mr0re1:


pkg/config/config_test.go
```diff
@@ -515,45 +515,31 @@ func (s *MySuite) TestDeploymentName(c *C) {
 
 	// Is deployment_name a valid string?
 	bp.Vars.Set("deployment_name", cty.StringVal("yellow"))
-	dn, err := bp.DeploymentName()
-	c.Assert(dn, Equals, "yellow")
-	c.Assert(err, IsNil)
+	c.Check(checkDeploymentName(bp), IsNil)
```
harshthakkar01, line 518: Here and below for all tests.
It seems we test if checkDeploymentName function returns correct error or not. It would be good to add a assert to check if the deployment name is correctly set or not. Like how we had before.

```
c.Assert(dn, Equals, "yellow")
```

pkg/config/expand.go
```diff
@@ -86,11 +86,7 @@ func (dc *DeploymentConfig) expandBackends() {
 				}
 			}
 			if be.Type == "gcs" && !be.Configuration.Has("prefix") {
-				prefix := blueprint.BlueprintName
-				if deployment, err := blueprint.DeploymentName(); err == nil {
-					prefix += "/" + deployment
-				}
-				prefix += "/" + string(grp.Name)
+				prefix := fmt.Sprintf("%s/%s/%s", bp.BlueprintName, bp.DeploymentName(), grp.Name)
```
harshthakkar01, line 89: Is this always going to be executed after we validate blueprint ? Previous behavior is we check if there is error or not while getting deployment name.

PR #1882, opened 2023-10-25
Improve Spack and Ramble Refs

cdunbar13:
This PR should allow users of the Spack and Ramble modules to use commit hashes as refs for the version they procure.  

This was accomplished by ignoring errors on a bad `git clone --branch`, collecting the return code, and if it's not equal to 0, then it attempting to clone the repo and reset the repo to the ref.

Tested by deploying a modified hpc-slurm-gromacs blueprint. The tests varied the spack_ref field with `spack_ref: v0.19.0`, `spack_ref: 45accfa`, and without a ref specified.

After creation I checked `/sw/spack` directory to make sure that the branch matched the ref. All three matched with the ref that was selected.

Additionally I recorded the time to run the clone:

`spack_ref: v0.19.0`: 1'46"
`spack_ref: 45accfa`: 1'49"
no `spack_ref` specified: 1'55"
Original templates (single command): 1'44"

After reverting back to two tasks (`git clone --branch ref` with fail-through to `git clone ... && git checkout ref`, testing was done the same as above and the run times were not significantly different.

An unknown ref was also tested to ensure that the script failed during the second clone, which was successful.

### DIFF
--- a/community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
+++ b/community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
@@ -53,9 +53,18 @@
     changed_when: lock_out.rc == 0
     failed_when: false
 
-  - name: Clones into installation directory
+  - name: Clone branch or tag into installation directory
     ansible.builtin.command: git clone --branch {{ git_ref }} {{ git_url }} {{ install_dir }}
+    failed_when: false
+    register: clone_res
     when: lock_out.rc == 0
+
+  - name: Clone commit hash into installation directory
+    ansible.builtin.command: "{{ item }}"
+    with_items:
+      - git clone {{ git_url }} {{ install_dir }}
+      - git --git-dir={{ install_dir }}/.git checkout {{ git_ref }}
+    when: lock_out.rc == 0 and clone_res.rc != 0
 
   - name: Set ownership and permissions
     ansible.builtin.file:
--- a/community/modules/scripts/spack-setup/templates/spack_setup.yml.tftpl
+++ b/community/modules/scripts/spack-setup/templates/spack_setup.yml.tftpl
@@ -53,9 +53,18 @@
     changed_when: lock_out.rc == 0
     failed_when: false
 
-  - name: Clones into installation directory
+  - name: Clone branch or tag into installation directory
     ansible.builtin.command: git clone --branch {{ git_ref }} {{ git_url }} {{ install_dir }}
+    failed_when: false
+    register: clone_res
     when: lock_out.rc == 0
+
+  - name: Clone commit hash into installation directory
+    ansible.builtin.command: "{{ item }}"
+    with_items:
+      - git clone {{ git_url }} {{ install_dir }}
+      - git --git-dir={{ install_dir }}/.git checkout {{ git_ref }}
+    when: lock_out.rc == 0 and clone_res.rc != 0
 
   - name: Set ownership and permissions
     ansible.builtin.file:


community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
```diff
@@ -55,8 +55,17 @@
...
     ansible.builtin.command: git clone --branch {{ git_ref }} {{ git_url }} {{ install_dir }}
+    register: clone_res
+    failed_when: false
     when: lock_out.rc == 0
 
+  - name: Clone repository and set to specific commit
+    ansible.builtin.command: "{{ item }}"
+    with_items:
+      - git clone {{ git_url }} {{ install_dir }}
+      - git --git-dir={{ install_dir }}/.git reset --hard {{ git_ref }}
```
nick-stroud, line 66: I don't think we want to use `reset` here. Reset will move the branch that is currently checked out. For example if they have main checked out then main will get moved to the specific commit. 

I think what we want is a `git checkout <hash>` which will enter headless mode.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
nick-stroud, line 1: This will not be exercised by our tests. Please include what manual testing has been done in the PR description.

nick-stroud (changes requested):

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
cdunbar13: Tested by deploying a modified hpc-slurm-gromacs blueprint.  The tests varied the `spack_ref` field with `spack_ref: v0.19.0`, `spack_ref: 45accfa`, and without a ref specified.  

After creation I checked `/sw/spack` directory to make sure that the branch matched the ref.  All three matched with the ref that was selected.  

Additionally I recorded the time to run the clone:
- `spack_ref: v0.19.0`: 1'46"
- `spack_ref: 45accfa`: 1'49"
- no `spack_ref` specified: 1'55"
- Original templates (single command): 1'44"

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
cdunbar13: Done, simplified since we can just do checkout during the original task and get rid of the new one.

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
```diff
@@ -54,7 +54,10 @@
     failed_when: false
 
   - name: Clones into installation directory
-    ansible.builtin.command: git clone --branch {{ git_ref }} {{ git_url }} {{ install_dir }}
+    ansible.builtin.command: "{{ item }}"
+    with_items:
+      - git clone {{ git_url }} {{ install_dir }}
+      - git --git-dir={{ install_dir }}/.git checkout {{ git_ref }}
```
nick-stroud, line 60: While this looks more elegant there is a performance cost for the happy path. I honestly know just enough to be dangerous, but performing the checkout operation can be slow, especially when it happens over nfs. When we clone the repo without the `--branch` arg it will checkout the default branch. So basically we now have to do 2x checkouts for a git clone. 

I would go back to the logic of trying to clone the branch and if that doesn't work then clone and checkout the branch.

Open to discussion especially if you have new data.

nick-stroud (changes requested):

community/modules/scripts/ramble-setup/templates/ramble_setup.yml.tftpl
cdunbar13: Done

nick-stroud (approved):

PR #1883, opened 2023-10-25
Revert "Integration task for bulk insert  stockout errors"

cdunbar13:
Reverts GoogleCloudPlatform/hpc-toolkit#1858

Context: This is being reverted as it caused several integration tests to break.  It was assumed that all slurm based integration tests had `slurm_cluster_name` defined, but that was not the case.  

This will be resubmitted with a different solution.


tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
nick-stroud, line 1: Please add some more context in the description such as: "caused integration test regression".

nick-stroud (approved):

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
cdunbar13: Done.

PR #1895, opened 2023-10-27
Front End Deployment Changes

max-nag:
Allows the Front End source code to be deployed on the service machine through a git clone rather than building a tarball. The deploy.sh script will ask for a forked repository owner and branch (or they can be added to a config file) so that development changes can be tested. It would be best to verify in future work if the branch exists once provided, instead of the possibility of failing once the service machine infrastructure has been created


cboneti (approved):

cboneti:
WARNING: DO NOT MERGE WITHOUT TALKING TO NAG

ek-nag:
@cboneti tested this in my dev environment, everything worked as expected. Can you merge this if it's still possible without rebasing and if the OFE test still fails I will look into that in the separate PR.

PR #1902, opened 2023-10-30
Re-raise error in rescue block of slurm integration test

cdunbar13:
Noticed during today's daily tests that if an integration test fails (specifically things like `test-mounts.yml`), the rescue block of the `slurm-integration-test.yml` blocks the error from propagating to cloud build so we don't notice it. 

This PR creates an error and propagates it upward similar to how it is done during the rescue block of the setup portion of the integration test.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
@@ -168,7 +168,6 @@
     - name: Trigger failure (rescue blocks otherwise revert failures)
       ansible.builtin.fail:
         msg: "Failed while setting up test infrastructure"
-      when: true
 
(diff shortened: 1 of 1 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
```diff
@@ -220,6 +220,11 @@
         deployment_name: "{{ deployment_name }}"
         project: "{{ project }}"
 
+    - name: Trigger failure (rescue blocks otherwise revert failures)
+      ansible.builtin.fail:
+        msg: "Failed during the integration tests (see above)"
+      when: true
```
nick-stroud, line 226: Is `when: true` really necessary here?

nick-stroud (approved):

PR #1904, opened 2023-10-30
Do not set role-label in expand, rely on module embedded labels

mr0re1:
**Release note:**

---

`ghpc` will no longer populate `ghpc_role` label based on module source path for third-party modules. Please take care of setting it yourself if you rely on this label.

---

* Do not set role-label in expand, rely on module embedded labels;
* Add ghpc labels to packer module.

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -20,8 +20,6 @@ import (
 	"regexp"
 	"strings"
 
-	"path/filepath"
-
 	"hpc-toolkit/pkg/modulereader"
 
 	"github.com/agext/levenshtein"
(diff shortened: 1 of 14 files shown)


pkg/config/expand.go
nick-stroud, line 1: What behavior before and after this change when handling alternative sources such as github (example below)?

I think I would have expected we would have had special handling for non-Toolkit native modules but I don't see it here. 

Am I correct that
Before: this module would be labeled with `ghpc_role: modules`
After: it would not receive any `ghpc_role` label. 

`source: github.com/daos-stack/google-cloud-daos.git//terraform/modules/daos_server?ref=v0.4.1&depth=1`

mr0re1:
@nick-stroud , updated label to "deprecation", added release note snippet. Please review.

pkg/config/expand.go
mr0re1: Discussed offline. TL;DR: it's OK

nick-stroud (approved):

PR #1908, opened 2023-10-31
Update stockout check to account for logging delay

cdunbar13:
There's a slight delay with logging that means that we won't necessarily see stockout messages immediately after an integration test fails.  To account for this, I've adjusted the stockout check to wait until we get some content back in stdout from the logging command.  This also required a `failed_when: false` as we're just checking for logs, nothing is actually failing if we don't find any stockout issues.

I've also updated the label to be the generated command, so we can debug if any issues come up with the stockout check, in the future.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
@@ -209,15 +209,6 @@
       loop_control:
         loop_var: test
 
-    rescue:
-    - name: Set stockout check variable
-      ansible.builtin.set_fact:
-        stockout_test: true
-
-    - name: Trigger failure (rescue blocks otherwise revert failures)
-      ansible.builtin.fail:
-        msg: "Failed during the integration tests (see above)"
-
     ## Always cleanup, even on failure
     always:
     - name: Ensure all nodes are powered down
@@ -292,4 +283,3 @@
       vars:
         deployment_name: "{{ deployment_name }}"
         project: "{{ project }}"
-      when: stockout_test is defined
--- a/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
@@ -210,14 +210,9 @@
         loop_var: test
 
     rescue:
-    - name: Check for stockout errors
-      ansible.builtin.include_tasks:
-        file: tasks/check_stockout.yml
-        apply:
-          delegate_to: localhost
-      vars:
-        deployment_name: "{{ deployment_name }}"
-        project: "{{ project }}"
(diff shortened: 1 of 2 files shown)


cdunbar13:
The failure on PR-test-hpc-slurm-chromeremotedesktop shows the fix (so not an actual issue with this PR).

tools/cloud-build/daily-tests/ansible_playbooks/tasks/check_stockout.yml
nick-stroud, line 1: Just a naive question about Ansible loops. What is the schema of the output (`register: stockout`)? I am not sure since you are using `with_items` and `until` together. Would `stockout.stdout` be a concatenation of the two commands? I would have thought it would have returned some sort of array structure.

nick-stroud, line 1: IIUC, In the case that there is a failure with no stockouts this will hang for 50s before moving on. I think we should be cautious about adding this bloat to the integration tests. 

One idea: This logging lookup is not dependent on the cluster being live. Move this rescue block later in the test execution, after terraform destroy has been executed. This would give more time for the logs to propagate. At that point if no logs are available, just make this lookup once and do not wait for logs to show up later.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/tasks/check_stockout.yml
cdunbar13: I hadn't really thought about it (so good catch), but looking at [Ansible Loops](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html), it appears that my usage is correct and that while still in the loop, the registered variable is just for that iteration so all of it's components are immediately accessible.  Once you leave the task, the registered variable becomes a list of the results of that task.

cdunbar13: Another idea, what about having the task run asynchronously? We can set the timeout limit to be 2 minutes, but that way it's just running in the background while we destroy the cluster.

cdunbar13: Ultimately, I moved the stockout check to the end of the always block, reduced the number of retries and retry interval, and just set a variable in the rescue block that triggers the stockout check.  The check should only occur if we get into the rescue block of integration tests.

nick-stroud (approved):

tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml
nick-stroud, line 1: Run at least 1 integrant test that exercises this, just to make sure there are no silly mistakes.

nick-stroud (approved):

PR #1912, opened 2023-10-31
Create bucket inside Slurm cluster wrapper to control default name

mr0re1:
Create bucket inside Slurm cluster wrapper to set a unique name for the bucket.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
@@ -40,6 +40,31 @@ locals {
 
 }
 
+locals {
+  synt_suffix       = substr(md5("${var.project_id}${var.deployment_name}"), 0, 5)
+  synth_bucket_name = "${local.slurm_cluster_name}${local.synt_suffix}"
+}
+
+module "bucket" {
+  source  = "terraform-google-modules/cloud-storage/google"
+  version = "~> 3.0"
+
+  count = var.create_bucket ? 1 : 0
+
+  location   = var.region
+  names      = [local.synth_bucket_name]
+  prefix     = "slurm"
+  project_id = var.project_id
+
+  force_destroy = {
+    (local.synth_bucket_name) = true
+  }
+
+  labels = {
+    slurm_cluster_name = local.slurm_cluster_name
+  }
+}
+
 data "google_compute_default_service_account" "default" {
   project = var.project_id
 }
@@ -111,8 +136,8 @@ module "slurm_cluster" {
   slurm_cluster_name = local.slurm_cluster_name
   region             = var.region
 
-  create_bucket = var.create_bucket
-  bucket_name   = var.bucket_name
+  create_bucket = false
+  bucket_name   = var.create_bucket ? module.bucket[0].name : var.bucket_name
   bucket_dir    = var.bucket_dir
 
   controller_instance_config = local.controller_instance_config
@@ -150,4 +175,6 @@ module "slurm_cluster" {
   prolog_scripts = var.prolog_scripts
   epilog_scripts = var.epilog_scripts
   cloudsql       = var.cloudsql
+
+  depends_on = [module.bucket]
 }
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
@@ -124,6 +124,7 @@ limitations under the License.
 
 | Name | Source | Version |
 |------|--------|---------|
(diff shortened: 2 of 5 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
```diff
@@ -40,6 +40,31 @@ locals {
...
 }
 
+locals {
+  synt_suffix       = substr(md5("${var.project_id}${var.deployment_name}"), 0, 5)
+  synth_bucket_name = "${local.slurm_cluster_name}${local.synt_suffix}"
+}
+
+module "bucket" {
+  source  = "terraform-google-modules/cloud-storage/google"
+  version = "~> 3.0"
```
nick-stroud, line 50: nit: elsewhere in toolkit we reference 3.4 and 4.0 as a min. Maybe pick similar for consistency.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
```diff
@@ -40,6 +40,31 @@ locals {
...
+  names      = [local.synth_bucket_name]
+  prefix     = "slurm"
+  project_id = var.project_id
+
+  force_destroy = {
+    (local.synth_bucket_name) = true
+  }
+
+  labels = {
+    slurm_cluster_name = local.slurm_cluster_name
```
nick-stroud, line 64: nit: Ideally we would add resource tracking labels here. I notice that controller does not take labels. I think it should.

I am fine to add in separate PR, but please follow up.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
```diff
@@ -40,6 +40,31 @@ locals {
 
 }
 
+locals {
+  synt_suffix       = substr(md5("${var.project_id}${var.deployment_name}"), 0, 5)
+  synth_bucket_name = "${local.slurm_cluster_name}${local.synt_suffix}"
+}
+
+module "bucket" {
+  source  = "terraform-google-modules/cloud-storage/google"
```
nick-stroud, line 49: nit: my personal preference is to use the resource directly (see startup-script as example) instead of cft. I don't really see cft adding value here and it is just one more layer. But we do both in the toolkit so I am fine to leave it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
```diff
@@ -111,17 +136,16 @@ module "slurm_cluster" {
...
+  bucket_name   = var.create_bucket ? module.bucket[0].name : var.bucket_name
   bucket_dir    = var.bucket_dir
 
   controller_instance_config = local.controller_instance_config
 
   enable_login = var.enable_login
   login_nodes  = var.login_nodes
 
-  nodeset     = var.nodeset
-  nodeset_tpu = var.nodeset_tpu
```
nick-stroud, line 124: This seems unrelated. Why is it deleted?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
nick-stroud, line 1: Add rational for this in description: Unique naming of bucket.

nick-stroud (approved):
Approving to not hold up submission. Please respond to/disposition comments.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
mr0re1: tracking as a separate issue.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
mr0re1: hmm, strange rebase, reverted, thank you for spotting.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
mr0re1: I'm sticking to the same version used in SchedMD module, would keep it this way until we rethink this whole situation (hopefully soon).

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
mr0re1: I agree, though this is quick fix without changes to the interfaces, I'm open to change it near future.

nick-stroud (approved):

nick-stroud (approved):

PR #1914, opened 2023-11-01
Unify image variable under `instance_image` across most modules

cdunbar13:
This PR attempts to replace all module variables referring to images with instance image, using a similar format to those in the current SchedMD modules:

For example:
```tf
  type        = map(string)
  default = {
    family  = "ubuntu-2204-lts"
    project = "ubuntu-os-cloud"
  }
```
Image variables that used different names are now deprecated.  If the variable is assigned, it should fail the validation and present the user with an error message stating that the variable is deprecated.

This also updates the link to the newest version of vm-instance.


community/modules/file-system/DDN-EXAScaler/variables.tf
```diff
@@ -203,20 +203,52 @@ variable "boot" {
...
 variable "image" {
-  description = "Source image properties"
+  description = "DEPRECATED: Source image properties"
   type        = any
   # Ommiting type checking so validation can provide more useful error message
   # type = object({
   #   project = string
   #   family  = string
   # })
+  default = {
```
mr0re1, line 213: nit. Can we simply `default = null` here ?

tpdownes: I agree, at above nit level. Please change to a simple `null`.

community/front-end/ofe/infrastructure_files/workbench_tf/google/variables.tf
```diff
@@ -56,16 +56,56 @@ variable "folder_id" {
...
+# Deprecated, replaced by instance_image
+# tflint-ignore: terraform_unused_declarations
 variable "image_family" {
   description = "Image of the AI notebook."
   type        = string
   default     = "tf-latest-cpu"
+
+  validation {
+    condition     = var.image_family == null
+    error_message = "The 'image_family' setting is deprecated, please use 'instance-image' with the sub-settings 'project' and 'family' or 'name'."
```
tpdownes, line 68: You should either say `var.instance_image` or `instance_image`, but either way use an underscore. I would also describe the "sub-settings" as "fields" instead.

community/modules/file-system/nfs-server/main.tf
```diff
@@ -70,7 +76,7 @@ resource "google_compute_instance" "compute_instance" {
   boot_disk {
     auto_delete = var.auto_delete_disk
     initialize_params {
-      image = var.image
+      image = data.google_compute_image.compute_image.self_link
```
tpdownes, line 79: I believe that this is going to change the behavior of the module so that the VM is replaced whenever the family introduces a new image. I think this is, on balance, not desirable. If you look at the vm-instance module there is an example of how to make this work so that it triggers only on changes to user inputs.

Rather than re-invent the wheel, this maybe an opportunity to use vm-instance internally.

Regardless, the behavior should be fixed.

community/modules/file-system/nfs-server/variables.tf
```diff
@@ -47,10 +47,43 @@ variable "type" {
...
+# tflint-ignore: terraform_unused_declarations
 variable "image" {
   description = "the VM image used by the nfs server"
   type        = string
-  default     = "cloud-hpc-image-public/hpc-centos-7"
+  default     = null
+
+  validation {
+    condition     = var.image == null
+    error_message = "The 'image' setting is deprecated, please use 'instance-image' with the sub-settings 'project' and 'family' or 'name'."
```
tpdownes, line 59: Make same text changes as advised above. Make sure `git grep instance-image` doesn't find any hits and same for "sub-settings"

tpdownes (changes requested):

community/modules/file-system/nfs-server/variables.tf
cdunbar13: Done

community/front-end/ofe/infrastructure_files/workbench_tf/google/variables.tf
cdunbar13: Done

community/modules/file-system/DDN-EXAScaler/variables.tf
cdunbar13: Done

community/modules/file-system/nfs-server/main.tf
cdunbar13: Addressed, although I did not switch out google_compute_instance for vm-instance for this.

tpdownes (approved):
There are a couple ways we could have addressed nfs-server. For example, manually constructing valid value for `image` that points to the family or image ID without being the full self_link. I prefer that we are using the same solution as in vm-instance since it has known behavior that is favorable to the user and, when we refactor that, we can refactor this, too.

Please merge after a broad range of integration tests have passed.

modules/scheduler/batch-job-template/variables.tf
```diff
@@ -166,15 +166,42 @@ variable "native_batch_mounting" {
...
+    name: The name of the image. Mutually exclusive with family.
+    family: The image family to use. Mutually exclusive with name.
+    project: The project where the image is hosted.
+    EOD
+  type        = map(string)
   default = {
-    family  = "batch-hpc-centos-7-official"
-    project = "batch-custom-image"
+    project = "ubuntu-os-cloud"
+    family  = "ubuntu-2204-lts"
```
nick-stroud, line 194: We should keep the default image the batch hpc image.

tpdownes: Thank you for catching that. I agree @cdunbar13 and it probably explains the failure.

PR #1920, opened 2023-11-01
Test for tty before echo in /etc/profile.d scripts

dodecatheon:
In spack-setup and ramble-setup modules, profile scripts have echo diag output to the terminal. This breaks some ssh and most scp connections.

Therefore, a test is added to ensure that tty output is available.

'test -t 1' is shell-agnostic and tests whether stdout is open.


nick-stroud (approved):

nick-stroud:
Spack tests passed before rebasing.

nick-stroud (approved):

nick-stroud:
I have been convinced that releasing as is constitutes a regression. I have proposed merging this to release candidate so the fix is in our next release.

PR #1927, opened 2023-11-03
Allow ramble module to point at forked repositories

douglasjacobsen:
This merge allows the ramble-setup module to properly download a requirements.txt from the repository it will clone. Previously, if a ref was provided that was in a fork but not in the main repo, this module would fail.

After making this change, some general cleanup was applied to both the spack and ramble modules to allow more code to be shared between them.

### Submission Checklist


community/modules/scripts/ramble-setup/main.tf
```diff
@@ -45,16 +45,31 @@ locals {
...
-      ramble_ref = var.ramble_ref
+      virtualenv_path = var.ramble_virtualenv_path
     }
   )
 
   install_ramble_deps_runner = {
     "type"        = "ansible-local"
     "content"     = local.deps_content
     "destination" = "install_ramble_deps.yml"
+    "args"        = "-e virtualenv_path=${var.ramble_virtualenv_path}"
```
nick-stroud, line 58: There is some duplication going on here between the args and the templating. 

In this case `virtualenv_path`is getting written directly into the file via the templating that happens on L47. Then you are also passing the same variable in via ansible args. I would suggest only doing one or the other.

There is a general preference to have scripts instead of templates when possible. My suggestion would be to change install_ramble_deps back to a script and then just pass the path as an arg.

community/modules/scripts/ramble-setup/main.tf
```diff
@@ -45,16 +45,31 @@ locals {
...
+      install_dir     = var.install_dir
+      virtualenv_path = var.ramble_virtualenv_path
+    }
+  )
+
+  python_reqs_runner = {
+    "type"        = "ansible-local"
+    "content"     = local.python_reqs_content
+    "destination" = "install_ramble_reqs.yml"
     "args"        = "-e ramble_virtualenv_path=${var.ramble_virtualenv_path}"
```
nick-stroud, line 73: I believe this arg has no effect as this value is not used in install_ramble_python_deps.

community/modules/scripts/spack-setup/main.tf
```diff
@@ -61,11 +61,19 @@ locals {
...
+    }
+  )
+
   install_spack_deps_runner = {
     "type"        = "ansible-local"
-    "source"      = "${path.module}/scripts/install_spack_deps.yml"
+    "content"     = local.deps_content
     "destination" = "install_spack_deps.yml"
-    "args"        = "-e spack_virtualenv_path=${var.spack_virtualenv_path}"
+    "args"        = "-e virtualenv_path=${var.spack_virtualenv_path}"
```
nick-stroud, line 76: Again, you are injecting this same value twice. I would suggest just making it a script.

nick-stroud (changes requested):

community/modules/scripts/ramble-setup/main.tf
douglasjacobsen: Thanks! I honestly wasn't sure what the `args` part was for here.

nick-stroud (approved):

PR #1930, opened 2023-11-03
Hotfix for vm-instance to allow image names to be used correctly

cdunbar13:
Allows name or family to be used for the trigger, which was breaking the tests.  Exclusivity of the term is validated in variables.tf.

### DIFF
--- a/modules/compute/vm-instance/main.tf
+++ b/modules/compute/vm-instance/main.tf
@@ -89,13 +89,14 @@ locals {
 data "google_compute_image" "compute_image" {
   family  = try(var.instance_image.family, null)
   name    = try(var.instance_image.name, null)
-  project = var.instance_image.project
+  project = try(var.instance_image.project, null)
 }
 
 resource "null_resource" "image" {
   triggers = {
-    image   = var.instance_image.family,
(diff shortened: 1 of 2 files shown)


modules/compute/vm-instance/main.tf
```diff
@@ -94,7 +94,8 @@ data "google_compute_image" "compute_image" {
 
 resource "null_resource" "image" {
   triggers = {
-    image   = var.instance_image.family,
+    name    = try(var.instance_image.name, null),
+    family  = try(var.instance_image.family, null),
     project = var.instance_image.project
```
mr0re1, line 99: Let's wrap project lookup into `try` as well, for a good measure.
Please consider adding a validation test, I'm OK if you decide to do it as a separate PR.

cboneti: I am not, please do add a validation test as part of this PR

mr0re1 (approved):

modules/compute/vm-instance/main.tf
cdunbar13: Done

PR #1932, opened 2023-11-05
Slurm6. Fix race condition between GCS config files and instances

mr0re1:
There was a race between config files stored in GCS and instance using those files. Fixed by stopping using SchedMD `slurm_cluster` module, instead replicate (unwrap) its content within Toolkit's `schedmd-slurm-gcp-v6-controller` module. This let us to establish proper `depends_on = [slurm_files]` for controller instance resource.

**NOTE:** This changes made us to temporary remove support for using CloudSQL as DB backend since it was a point of circular dependencies with updated `depends_on`. To be adressed later.

Added Filestore usage to the `community/examples/hpc-slurm6.yaml` as it should work after the fix.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -0,0 +1,129 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# BUCKET
+
+locals {
+  synt_suffix       = substr(md5("${var.project_id}${var.deployment_name}"), 0, 5)
+  synth_bucket_name = "${local.slurm_cluster_name}${local.synt_suffix}"
```
nick-stroud, line 20: nit: a dash would be nice. Since cluster name is already chopped up it can sometimes not be clear where cluster name ends and suffix begins.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -0,0 +1,129 @@
...
+resource "google_storage_bucket_iam_binding" "legacy_readers" {
+  bucket  = local.bucket_name
+  role    = "roles/storage.legacyBucketReader"
+  members = compact(local.viewers)
+}
+
+# SLURM FILES
+locals {
+  ghpc_startup_script_controller = [{
+    filename = "ghpc_startup.sh"
```
nick-stroud, line 76: I think assumptions break down if I have multiple clusters using the same bucket. Very plausible now that we allow user to provide bucket. Could we add cluster name into the file names?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -0,0 +1,109 @@
...
+  region             = var.region
+  zone               = var.zone
+  # TODO: add supoort for cloudsql
+  # cloudsql           = var.cloudsql
+
+  instance_template = local.have_template ? var.instance_template : module.slurm_controller_template[0].self_link
+
+  enable_public_ip = !var.disable_controller_public_ips
+  # TODO: add support for network_tier
+  # network_tier       = var.controller_instance_config.network_tier
```
nick-stroud, line 103: Can we clean up the commented out lines. I am fine with the comment and even the left side of the `=`. The right side of the equals is referencing a var that does not exist in our code, so without more context it is quite confusing. I would update the comments to something like:

`# network_tier = TODO: add support for network_tier`

nick-stroud (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
mr0re1: By default files will be placed into separate directory with `cluster_name` in it.
Unless user specified both `bucket_name` and `bucket_dir` no collision should happen, 
otherwise we should let them to shoot in their foot since they tried so hard.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
mr0re1: done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
mr0re1: I'm hesitant to convolve this architecture change behaviour one. May do later.

PR #1945, opened 2023-11-07
Slurm6. Enable support for CloudSQL

mr0re1:
* Unwrap `slurm_cluster/modules/slurm_controller_instance` into `schedmd-slurm-gcp-v6-controller/controller.tf`;
* Simplify unwrapped `slurm_controller_instance`;
* Set `depends_on=[module.slurm_files,...` for `slurm_controller_instance`;
* Allow setting `var.cloudsql`.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
@@ -26,9 +26,14 @@ locals {
   ]
 
   have_template = var.instance_template != null && var.instance_template != ""
+
+  service_account = coalesce(var.service_account, {
+    email  = data.google_compute_default_service_account.default.email
+    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
+  })
 }
 
-# TEMPLATE
+# INSTANCE TEMPLATE
 module "slurm_controller_template" {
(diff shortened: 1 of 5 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -83,23 +84,88 @@ module "slurm_controller_template" {
...
+  ]
+}
+
+# SECRETS: CLOUDSQL
+resource "google_secret_manager_secret" "cloudsql" {
+  count = var.cloudsql != null ? 1 : 0
+
+  secret_id = "${local.slurm_cluster_name}-slurm-secret-cloudsql"
+
+  replication {
```
tpdownes, line 127: This syntax is deprecated and will break in TPG 5.0. It likely generates warnings you may not have noticed. Please adopt the syntax from 1d3a2b09

tpdownes (changes requested):
- Please make the minimal changes for syntax.
- I'm a little concerned at the presence of the "conditional creation of resource" (anti-)pattern in several places. This is not the only example of it in the Toolkit, but it can make dependency tracking difficult. I'd like to take a moment (myself) to look at the design doc and understand.
- I believe this is capturing all the "cleanup" variables behavior in once place (going by memory, this is part of the design) and that's a good thing!

tpdownes (approved):
I think the choice to point directly to _slurm_instance is probably reasonable as having a chain of 2 wrappers is not likely to be helping us.

PR #1946, opened 2023-11-07
Remove error-prone `errorMessage` map, remove deprecated `go-lint`

mr0re1:
* Remove error-prone `errorMessage` map, replace with named constants and sentinel errors;
* Remove deprecated `go-lint` hook, rely on `go vet` instead.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,7 +62,6 @@ repos:
   hooks:
   - id: go-fmt
   - id: go-vet
-  - id: go-lint
   - id: go-imports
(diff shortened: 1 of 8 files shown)


tpdownes (changes requested):
Please split out the go-lint portion into a separate PR. I agree with the decision to remove (lots of internal and external discussion going back years), but at least some of functionality can be replaced with staticcheck and it's really a separate purpose from the rest of the PR.

tpdownes (approved):
Thanks for #1961 and #1962 which are related to the requests here

PR #1949, opened 2023-11-08
🌈 Add colorization to the output

mr0re1:
Add color to the small set of outputs of `create` command.

Screenshot 2023-11-07 at 4 50 38 PM

Screenshot 2023-11-07 at 4 49 49 PM

Feature is hidden behind flag `--no-color=false` (default is `true`), also it is ignored if output doesn't support colors, see:

```
$ ./ghpc create tst.yaml -w --no-color=false 2> /tmp/out
$ cat /tmp/out
Error: invalid module id: "var" - Did you mean "vars"?
45:       node_count_dynamic_max: $(var.wut)
```

**NOTE:** All added dependencies are under `MIT` license.

### DIFF
--- a/cmd/color.go
+++ b/cmd/color.go
@@ -1,0 +1,39 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"os"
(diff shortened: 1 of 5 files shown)


cmd/color.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/fatih/color"
+	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
+)
+
+var noColorFlag bool = true
+
+func addColorFlag(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&noColorFlag, "no-color", true, "Disable colorized output.")
```
cboneti, line 28: This seems nice, why disable it by default?

cboneti (commented):
A nit, which is the choice of disabling it by default. Do you foresee the chance of having it break?

cboneti (approved):
I will approve in case you have a good reason for not making this default. But consider changing it if you feel it is reliable....

cmd/color.go
mr0re1: Because it's:
1) experimental
2) not fully implemented in terms of covered opportunities to colorize

We can make sure that it works and add more colors, than switch it on be default, and go full peacock.

PR #1955, opened 2023-11-09
Slurm6. Makde `nodeset.name` required

mr0re1:
The default value (`ghpc`) leads to collisions and poor user experience.


cboneti (changes requested):
I am worried by this change. Does it mean that users have to specify this extra setting in all node groups? before, as long as you have one node group per partition, everything worked. I understand that this was not ideal. Now you will force the user to come up with something and realistically everyone will hard code it...

Can we improve that experience?  Perhaps pick from the machine family?

mr0re1:
> Can we improve that experience? 

Back to drawing board.

PR #1956, opened 2023-11-09
Script run warning stage 2

cdunbar13:
Second stage of startup-script warning

- Creates file with warning message in `/run/startup-script-msg` 
- profile.d checks for that file and prints a message if it exists - if a user logs in before startup scripts have started an alternate warning is presented.

### DIFF
--- a/modules/scripts/startup-script/files/running-script-warning.sh
+++ b/modules/scripts/startup-script/files/running-script-warning.sh
@@ -13,17 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-WARNING=$(
-	cat <<EOF
-** NOTICE **: System services may not be running until startup scripts complete.
-The output of the command below will end with "Finished Google Compute Engine 
-Startup Scripts." when they are complete. Please review the output for any 
-errors which may indicate that services are unhealthy. 
+SCRIPT_COMPLETE_FILE="/run/startup_script_msg"
+WARNING_MESSAGE="** WARNING **: HPC Toolkit startup scripts have not started yet."
 
-sudo journalctl -b 0 -u google-startup-scripts.service
-EOF
-)
-
-echo
-echo "${WARNING}"
-echo
+# Ensure we're in an interactive terminal and not root
+if [ -t 1 ] && [ "$(id -u)" -ne 0 ]; then
+	# Check if the file has contents otherwise skip
+	if [ -s "$SCRIPT_COMPLETE_FILE" ]; then
+		echo
+		cat "$SCRIPT_COMPLETE_FILE"
+		echo
+	elif [ ! -f "$SCRIPT_COMPLETE_FILE" ]; then
+		echo
+		echo "$WARNING_MESSAGE"
+		echo
+	fi
+fi
--- a/modules/scripts/startup-script/files/startup-script-stdlib-head.sh
+++ b/modules/scripts/startup-script/files/startup-script-stdlib-head.sh
@@ -30,26 +30,12 @@ readonly E_RUN_OR_DIE=5
 readonly E_MISSING_MANDATORY_ARG=9
 readonly E_UNKNOWN_ARG=10
 
-SUCCESS_MESSAGE=$(
-	cat <<-EOF
-		** NOTICE **: The VM startup scripts have finished running successfully.  
-		Systems are configured and running.
-	EOF
-)
+SCRIPT_COMPLETE_FILE="/run/startup_script_msg"
+SUCCESS_MESSAGE="* NOTICE **: The HPC Toolkit startup scripts have finished running successfully."
 readonly SUCCESS_MESSAGE
-ERROR_MESSAGE=$(
-	cat <<-EOF
-		** ERROR **: The VM startup scripts have finished running, but produced an error.
-		Systems may be in an unhealthy state.
-	EOF
-)
+ERROR_MESSAGE="** ERROR **: The HPC Toolkit startup scripts have finished running, but produced an error."
 readonly ERROR_MESSAGE
-WARNING_MESSAGE=$(
-	cat <<-EOF
-		** WARNING **: The VM startup scripts for this machine have started.
-		Systems may not be configured or running.
-	EOF
-)
+WARNING_MESSAGE="** WARNING **: The HPC Toolkit startup scripts are currently running."
 readonly WARNING_MESSAGE
 
 stdlib::debug() {
@@ -82,6 +68,7 @@ stdlib::error() {
 stdlib::announce_runners_start() {
 	if [ -z "$recursive_proc" ]; then
 		wall -n "$WARNING_MESSAGE"
+		echo "$WARNING_MESSAGE" >"$SCRIPT_COMPLETE_FILE"
 	fi
 	export recursive_proc=$((${recursive_proc:=0} + 1))
 }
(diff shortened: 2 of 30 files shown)


modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -13,17 +13,21 @@
...
-The output of the command below will end with "Finished Google Compute Engine 
-Startup Scripts." when they are complete. Please review the output for any 
-errors which may indicate that services are unhealthy. 
+SCRIPT_COMPLETE_FILE="/run/startup_script_msg"
 
-sudo journalctl -b 0 -u google-startup-scripts.service
-EOF
+WARNING_MESSAGE=$(
+	cat <<-EOF
+		** NOTICE **: HPC-Toolkit startup scripts should start soon.
```
tpdownes, line 20: - No dash in "HPC Toolkit".
- Do not say "should" or "soon" which conveys uncertainty. Say "HPC Toolkit startup scripts are executing". Fewer, direct (and true) words.

modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -13,17 +13,21 @@
...
+		Systems may not be configured or running.
```
tpdownes, line 21: I still do not like "Systems" vs "Services" and I would like you to discuss it with the group.

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
```diff
@@ -92,8 +94,10 @@ stdlib::announce_runners_end() {
 	if [ "$recursive_proc" -le "0" ]; then
 		if [ "$exit_code" -ne "0" ]; then
 			wall -n "$ERROR_MESSAGE"
+			echo "$ERROR_MESSAGE" >"$SCRIPT_COMPLETE_FILE"
 		else
 			wall -n "$SUCCESS_MESSAGE"
+			cat /dev/null >"$SCRIPT_COMPLETE_FILE"
```
tpdownes, line 100: `echo -n "" >...` will be a shell built-in. `cat` is external. I'm sure it's installed nearly everywhere, but better to reduce external dependencies.

`type echo` will tell you what a given shell will use (there is also an echo binary).

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
```diff
@@ -30,23 +30,24 @@ readonly E_RUN_OR_DIE=5
 readonly E_MISSING_MANDATORY_ARG=9
 readonly E_UNKNOWN_ARG=10
 
+SCRIPT_COMPLETE_FILE="/run/startup_script_msg"
```
tpdownes, line 33: My instinct here would be `/run/ghpc/startup_script_msg`. Do you have a reason not to? Don't make change yet, I intend this to be start of discussion.

tpdownes (changes requested):
I will need to do more testing, but wanted to get some wording requests back to you.

tpdownes:
Testing with this blueprint, I still observe the "should start soon" message at login well after successful completion of the script. I saw the wall message indicating successful completion and then logged in again.

```yaml
---

blueprint_name: pr1956

vars:
  deployment_name: ss
  region: us-central1
  zone: us-central1-c

deployment_groups:
- group: first
  modules:
  - id: network
  - id: script
    settings:
      runners:
      - type: shell
        destination: sleep.sh
        content: |
          #!/bin/bash
          sleep 100
  - id: repro
    use:
    - network
    - script
    settings:
      machine_type: c2-standard-4
```

modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -13,17 +13,21 @@
...
-echo
-echo "${WARNING}"
-echo
+if [ -f "${SCRIPT_COMPLETE_FILE}" ]; then
+	echo
+	cat "${SCRIPT_COMPLETE_FILE}"
+	echo
+else
+	echo
+	echo "$WARNING_MESSAGE"
```
tpdownes, line 31: Testing with this blueprint, I still observe the "should start soon" message at login well after successful completion of the script. I saw the wall message indicating successful completion and then logged in again. I suspect this is the problem.

I thought intent was for this file to be long-lived and always indicate the status of the previous run.

```yaml
---

blueprint_name: pr1956

vars:
  deployment_name: ss
  region: us-central1
  zone: us-central1-c

deployment_groups:
- group: first
  modules:
  - id: network
  - id: script
    settings:
      runners:
      - type: shell
        destination: sleep.sh
        content: |
          #!/bin/bash
          sleep 100
  - id: repro
    use:
    - network
    - script
    settings:
      machine_type: c2-standard-4
```

tpdownes (changes requested):

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
tpdownes: It would also be nice if this could be defined once and sourced somehow. Please investigate.

modules/scripts/startup-script/files/running-script-warning.sh
```diff
@@ -13,17 +13,20 @@
...
+		** NOTICE **: HPC Toolkit startup scripts have not started yet.
+		Systems may not be configured or running.
+	EOF
 )
 
-echo
-echo "${WARNING}"
-echo
+if [ -f "${SCRIPT_COMPLETE_FILE}" ]; then
+	echo
```
tpdownes, line 25: These two echos are creating blank lines when the startup message is empty so that the prompt does not appear at the top of a fresh terminal. Please make it appear at the top (i.e. "normal").

I suggest you do this by incorporating any desired whitespace into the actual message file and ensuring that you do not add any additional newlines.

tpdownes (changes requested):
- please address the whitespace issue
- the /etc/profile.d script should exit early when effective uid = 0

e.g. 

```bash
if [ "$EUID" -eq 0 ]; then
  exit 0
fi
```

modules/scripts/startup-script/files/running-script-warning.sh
cdunbar13: My mistake, this didn't work because I hadn't linked vm-instance and two other modules in the `modules` folder.  Fixed.

modules/scripts/startup-script/files/running-script-warning.sh
cdunbar13: I saw this and was fixing it.  My alternate solution was to check if the file was empty and if it is, just don't print anything.

tpdownes: I considered that too. Do anything simple and robust that resolves it.

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
tpdownes: If possible, define this once. If not (or not easy/robust), then resolve this discussion.

Let's not consider any changes to location for now. We have an open task to better adopt [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) best practices in Toolkit generally.

modules/scripts/startup-script/files/running-script-warning.sh
cdunbar13: Done

modules/scripts/startup-script/files/startup-script-stdlib-head.sh
cdunbar13: I can't think of any easy ways to do this other than having the startup script modify the profile.d script which does not seem robust.  I will continue to think about it.

cdunbar13:
In regards to the `$EUID` check, shellcheck complained about `$EUID` not being available with POSIX, so I opted for `id -u`

tpdownes (approved):
Big improvement to command line experience with respect to startup-script completion and failure. Thanks!

PR #1961, opened 2023-11-10
staticcheck. Fix errors

mr0re1:
Fix errors pointed out by `staticcheck`. The majority of changes fall in of those categories:
* `ST1005` Incorrectly formatted error string;
* `U1000` Unused;
* `SA1019` Using a deprecated `ioutil`.

```sh
$ staticcheck ./...
$ # no more errors
```

### DIFF
--- a/pkg/inspect/modules_test.go
+++ b/pkg/inspect/modules_test.go
@@ -89,12 +89,6 @@ func all(ps ...predicate) predicate {
 	}
 }
 
-func not(p predicate) predicate {
-	return func(mod modInfo) bool {
-		return !p(mod)
-	}
-}
-
 func hasInput(name string) predicate {
 	return func(mod modInfo) bool {
 		_, ok := mod.Input(name)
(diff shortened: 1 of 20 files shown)


pkg/inspect/modules_test.go
```diff
@@ -89,35 +89,16 @@ func all(ps ...predicate) predicate {
...
-
-func ofRole(role string) predicate {
-	return func(mod modInfo) bool {
-		return mod.Role() == role
-	}
-}
-
 // Fails test if slice is empty, returns not empty slice as is.
 func notEmpty[E any](l []E, t *testing.T) []E {
-	if l == nil || len(l) == 0 {
```
rohitramu, line 120: Do we no longer want to check for `nil` here?

rohitramu (approved):

pkg/inspect/modules_test.go
mr0re1:
```go
var l []int
fmt.Printf("len(%#v) == %d", l, len(l))
// len([]int(nil)) == 0
```

PR #1962, opened 2023-11-10
Add `staticcheck` in place of deprecated `go-lint`

mr0re1:
https://github.com/golang/lint
> Golint is [deprecated and frozen](https://github.com/golang/go/issues/38968). There's no drop-in replacement for it, but tools such as [Staticcheck](https://staticcheck.io/) and `go vet` should be used instead.

* `go-vet` is already part of our pre-commit;
* add `go-static-check` to pre-commit hooks, all `staticcheck` are addressed in #1961

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,6 +73,10 @@ repos:
   hooks:
   - id: go-critic
     args: [-disable, "#experimental,sloppyTypeAssert"]
+- repo: https://github.com/Bahjat/pre-commit-golang
+  rev: v1.0.2
+  hooks:
+  - id: go-static-check
 - repo: https://github.com/adrienverge/yamllint
   rev: v1.32.0
   hooks:


tools/cloud-build/Dockerfile
```diff
@@ -43,7 +43,8 @@ RUN go install github.com/terraform-docs/terraform-docs@latest      && \
     go install github.com/go-critic/go-critic/cmd/gocritic@latest   && \
     go install github.com/google/addlicense@latest                  && \
     go install mvdan.cc/sh/v3/cmd/shfmt@latest                      && \
-    go install golang.org/x/tools/cmd/goimports@latest
+    go install golang.org/x/tools/cmd/goimports@latest              && \
+    go install honnef.co/go/tools/cmd/staticcheck@latest
```
tpdownes, line 47: also remove `golint` a few lines above

tools/cloud-build/dependency-checks/Dockerfile.precommit
```diff
@@ -25,4 +25,5 @@ RUN go install github.com/terraform-docs/terraform-docs@latest      && \
     go install github.com/go-critic/go-critic/cmd/gocritic@latest   && \
```
tpdownes, line 25: remove golint a few lines above

tpdownes (changes requested):
We can remove golint from development dependencies entirely right?

tpdownes (approved):
LGTM

PR #1964, opened 2023-11-11
Selectevly suggest to run `ghpc export-outputs`

mr0re1:
```sh
$ ./ghpc deploy tst
Testing if deployment group tst/primary requires adding or changing cloud infrastructure
Error: terraform plan for deployment group tst/primary failed; run "ghpc export-outputs" on previous deployment groups to define inputs
exit status 1

Error: No value for required variable

  on variables.tf line 32:
  32: variable "region" {

The root module input variable "region" is not set, and has no default value.
Use a -var or -var-file command line argument to provide a value for this
variable.

...

$ ./ghpc deploy tst
Testing if deployment group tst/primary requires adding or changing cloud infrastructure
Error: terraform plan for deployment group tst/primary failed
exit status 1

  on main.tf line 20, in module "network":
  20:   network_address_range = "over the hills and far away"
    ├────────────────
    │ var.network_address_range is "over the hills and far away"

IP address range must be in CIDR format.

This was checked by the validation rule at
modules/embedded/modules/network/vpc/variables.tf:65,3-13.
```

### DIFF
--- a/pkg/shell/terraform.go
+++ b/pkg/shell/terraform.go
@@ -17,6 +17,7 @@
 package shell
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -28,6 +29,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/hashicorp/terraform-exec/tfexec"
 	"github.com/zclconf/go-cty/cty"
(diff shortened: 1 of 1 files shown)


pkg/shell/terraform.go
```diff
@@ -135,17 +137,63 @@ func outputModule(tf *tfexec.Terraform) (map[string]cty.Value, error) {
...
+		var msg JsonMessage
+		if err := json.Unmarshal([]byte(sc.Text()), &msg); err != nil {
+			continue // silently skip the message
+		}
+		res = append(res, msg)
+	}
+	return res
+}
+
+func helpOnPlanError(msgs []JsonMessage) string {
```
tpdownes, line 166: Add a comment here that the function is enforcing the logic that the only undefined variables can possibly come from IGC references. This may not always be true.

tpdownes (approved):
Please add the requested comment. I think this is a good MVC for improving plan failures. I think we want to re-think the ghpc deploy/destroy experience within a larger context.

tpdownes:
For posterity, I tested destroy on a multi-group deployment without any artifacts generated and it produces the existing "try export-outputs" recommendation. Then I tried deploy on a single group deployment and mangled the terraform.tfvars in a similar manner as above.

PR #1965, opened 2023-11-12
Inject Blueprint ModuleID as specified variable

mr0re1:
* Add section `ghpc` to `metadata.yaml` schema;
* Add `inject_module_id` metadata to inject Blueprint ModuleID into specified field


community/examples/hpc-slurm6.yaml
```diff
@@ -37,51 +37,45 @@ deployment_groups:
...
     source: community/modules/compute/schedmd-slurm-gcp-v6-partition
-    use: [debug_nodeset, homefs]
+    use: [debugnodeset, homefs]
     settings:
-      partition_name: debug
       exclusive: false  # allows nodes to stay up after jobs are done
       is_default: true
 
-  - id: compute_nodeset
+  - id: computenodeset
```
cboneti, line 55: this will be the name of the nodes, right? it would almost be nicer to have a shorter name.

community/modules/compute/schedmd-slurm-gcp-v6-partition/metadata.yaml
```diff
@@ -0,0 +1,2 @@
+ghpc:
+  inject_module_id: partition_name
```
cboneti, line 2: I think this deserves a design doc. 
I like what you have done here, but injecting a varialbe "module_id" would also be very simple and transparent, since you then do this logic in terraform. I would be curious to know what others think. 
for instance: 
you could define a variable "module_id" that when set, ghpc will push the module id. Then in the main.tf, if partition_name == null and module_id != null, then partition_name = module_id.

In my opinion it would be preferrable as it would still allow this module to work without special ghpc behavior. 

On the other hand, what is nice here is the ability to tell ghpc to rename variables, meaning you would perhaps no longer need simple wrappers: one could have more or less complex metadata files that would instruct hpc how to map various variables... which is also powerful.

cboneti (commented):
I added a few comments.
I like this functionality, but would love to see it discussed a little with the other team members. A one pager design doc with the considerations I added in my comments could be useful.

pkg/modulereader/metadata.go
nick-stroud, line 1: Please add documentation on metadata.yaml schema and functionality. `modules/README.md` seems like the right place.

nick-stroud (changes requested):

pkg/modulereader/metadata.go
mr0re1: Added section to docs/module-guidelines.md

nick-stroud (approved):

PR #1972, opened 2023-11-13
Add native support for Google Cloud Storage with GKE

nick-stroud:
You can now configure a cluster and job to use a GCS bucket as a shared file system with GKE.

Note to reviewer:
- Each commit is self contained and can be reviewed independently
- First 3 commits are adding loosely related independent features that will be needed downstream
- Updated storage example job 
  - to write to separate files as concurrent file operations was problematic at scale
  - to use both shared file systems

Tested: Deployed storage-gke.yaml blueprint and ran job.

### Submission Checklist


community/modules/compute/gke-job-template/variables.tf
```diff
@@ -26,6 +26,12 @@ variable "node_count" {
   default     = 1
 }
 
+variable "completion_mode" {
+  description = "Sets value of `completionMode` on the job. Default uses indexed jobs. See [documentation](https://kubernetes.io/blog/2021/04/19/introducing-indexed-jobs/) for more information"
+  type        = string
+  default     = "Indexed"
```
mr0re1, line 32: nit. Shall we add enum-validation `{Indexed, NonIndexed}` ?

community/modules/file-system/cloud-storage-bucket/outputs.tf
```diff
@@ -62,3 +62,14 @@ output "gcs_bucket_path" {
     google_storage_bucket.bucket
   ]
 }
+
+output "gcs_bucket_name" {
+  description = "Bucket name."
+  # cannot use resource attribute, will cause lookup failure in startup-script
+  value = local.name
```
mr0re1, line 69: nit. Does it cause lookup failure even with `depends_on` set ?

community/modules/file-system/gke-persistent-volume/README.md
```diff
@@ -40,6 +41,36 @@ The following example creates a Filestore and then uses the
...
+    source: community/modules/scheduler/gke-cluster
+    use: [network1]
+    settings:
+      master_authorized_networks:
+      - display_name: deployment-machine
+        cidr_block: <your-ip-address>/32
+
+  - id: data-bucket
+    source: community/modules/file-system/cloud-storage-bucket
+    settings: { local_mount: /data }
```
mr0re1, line 59: nit. Is there a particular reason to use inline syntax. IIUC, unlike sequences, mapping is very picky for correct syntax (spaces) and can be a point of user frustration.

community/modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -20,12 +20,16 @@ locals {
 }
 
 locals {
-  location             = split("/", var.filestore_id)[3]
-  filestore_name       = split("/", var.filestore_id)[5]
+  is_gcs = (var.gcs_bucket_name != null)
+
+  filestore_id         = !local.is_gcs ? var.filestore_id : "projects/empty/locations/empty/instances/empty"
```
mr0re1, line 25: nit. Is there a meaning for `"projects/empty/locations/empty/instances/empty"` ?

mr0re1, line 25: Is there a special meaning to the `"empty"` values?

mr0re1 (approved):

community/modules/compute/gke-job-template/variables.tf
nick-stroud: In the referenced docs it sounds like they are planning on expanding the list of options. I thought safer to leave it open so update will not be needed as options grow:

>_we are considering an IndexedAndUnique mode_

community/modules/file-system/cloud-storage-bucket/outputs.tf
nick-stroud: Yes and No. IIUC there is some weird behavior because of this line:
https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/3abddcfbd245b0e6747917a4e55b30658414ffd7/modules/scripts/startup-script/main.tf#L130

If bucket name is not know at apply time then startup-script does not know if it is creating a bucket or not, and the bucket name gets written into vm and slurm startup scripts.

But this is a new variable that is not being used in such a convoluted way, so I agree to start out using it the right way.

community/modules/file-system/gke-persistent-volume/README.md
nick-stroud: Just trying to look cool. Good feedback I have updated.

community/modules/file-system/gke-persistent-volume/main.tf
nick-stroud: It is arbitrary. The logic of the module is that it authors both pv and pvc for GCS and Filestore. Then at the end it only applies the one that it is actually using. If this value is null then all subsequent calculations become ugly as we check for null on each one. I added a comment in the code to clarify that the value is arbitrary.

nick-stroud:
Will merge pending integration tests.

PR #1974, opened 2023-11-14
Rename misleading `login group_name -> name_prefix`

mr0re1:
The `group_name` is misleading name, as there is no notion of "group". Instead it's an "unique login node name".
The only reason to have it is to be able to use TF `for_each = { turn list of login_nodes into map node.name_prefix -> node}`

* Rename `group_name -> name_prefix`;
* Add validation for uniqueness.


community/modules/scheduler/schedmd-slurm-gcp-v6-login/variables.tf
nick-stroud, line 1: I suspect the origin has to do with `num_instances`. You might have two instances of the login module in your blueprint. One set of logins could be powerful machines for post processing, the other could be small machines for submission only. Each module could have multiple instances using the `num_instances` var. In this way you have two groups each with a `group_name`.

With all that said we do the same thin with vm-instance and we don't call it `group-prefix` so I am in favor of this change. Close after reading if you agree.

nick-stroud, line 1: Since you are renaming, would `name_prefix` be better than `name`. I think that would be my preference. Rational:
1. Naming would match slurm variable name
2. More accurate to how it is actually used
3. Commonality with vm-instance

mr0re1: `name_prefix` it is.

mr0re1: The only concern that it's not `prefix` but `suffix` ) But this is mistake made by SchedMD, I'm OK to propagate it.

nick-stroud: Do you mean that the provided string actually gets used as a suffix? If I provide `name_prefix: tower` what is the final name of the login node?

nick-stroud (approved):

PR #1975, opened 2023-11-14
Slurm6. Surface `login_network_storage`

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
nick-stroud, line 1: Ideally this would be a property of the login node node the controller. Given our current constraints do you know if that is possible?

mr0re1: Nope, all login nodes will recieve the same set of storages, this is done at python script level.

nick-stroud (approved):

PR #1976, opened 2023-11-14
Enable many-to-many nodeset partititon relation.

mr0re1:
Test blueprint:

```yaml
  - id: ns
    settings: { name: ns }

  - id: p1
    use: [ns]

  - id: p2
    use: [ns]

  - id: slurm_controller
    use: [p1, p2]
```


tpdownes (commented):
- This is "native" functionality of both Slurm-GCP 6.2.0 and Slurm itself, right?
- Can you provide a full blueprint to test?

My understanding is that the scaling behavior is defined at nodeset level but exclusive is set at partition level. Is there a potential conflict here that needs a validation?

This seems like functionality that ought to be tested regularly.

mr0re1:
> This is "native" functionality of both Slurm-GCP 6.2.0 and Slurm itself, right?

It's "native" for Slurm, but not for Slurm-GCP (I would say there is a bug).

> Can you provide a full blueprint to test?

TODO

> My understanding is that the scaling behavior is defined at nodeset level but exclusive is set at partition level. Is there a potential conflict here that needs a validation?

`exclusive` bit controls partition level settings `Oversubscribe` & `PowerDownOnIdle`, as well as label on instance, but it's expected to be unique per job (in case of exclusivity). 
So if there is any conflict, it's Slurm implementation to blame, not the wrapper.

PR #1977, opened 2023-11-14
Creation of script to check upcoming maintenance on nodes

cdunbar13:
Initial introduction of the check-upcoming-maintenance.py script to the tools folder. This script creates a class that holds onto project, zones, and regex search parameters for nodes that will have upcoming maintenance. Initializes by testing the parameters and doing an initial check for maintenance information.

Currently only stubs for performing maintenance and controls for doing additional maintenance checks.


tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+import re
+import subprocess
+from typing import List
+from optparse import OptionParser
+
+class NodeMaintenance:
+    '''
+    Class to keep track of project, zones, and search terms for nodes, as well as results from gcloud queries
+    '''
+    def __init__(self, project, zones, regex):
```
nick-stroud, line 11: I would provide default for `regex=None` to denote that it is optional.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+        if zones:
+            self.zones = zones.split(",")
+            for zone in self.zones:
+                res = subprocess.run("gcloud compute zones describe {}".format(zone), shell=True, capture_output=True, text=True)
+                if res.returncode != 0:
+                    raise Exception("{} does not exist".format(zone))
+        
+        self.regex = None
+        self.screen_reader = False
+        res = subprocess.run("gcloud config get accessibility/screen_reader",
```
nick-stroud, line 27: nit: Although not technically wrong, I would avoid using the same var for multiple calls. You could use more descriptive variable names that reflect the specific context of the result (ex. screen_reader_res). IMO this makes code more readable and avoids mistakes where one result bleeds into another.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+        self.zones = None
+        if zones:
+            self.zones = zones.split(",")
+            for zone in self.zones:
+                res = subprocess.run("gcloud compute zones describe {}".format(zone), shell=True, capture_output=True, text=True)
+                if res.returncode != 0:
+                    raise Exception("{} does not exist".format(zone))
+        
+        self.regex = None
+        self.screen_reader = False
```
nick-stroud, line 26: Most gcloud commands take [format](https://cloud.google.com/sdk/gcloud/reference#--format) argument. I would suspect this is to aid automation. I think picking a format like yaml or json would avoid the screen reader issue and would be more robust in the long term.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+    # Needs to be used after the command is fully created
+    def remove_accessibility(self, cmd: str) -> str:
+        return "gcloud config set accessibility/screen_reader False && " \
+               "{} && gcloud config set accessibility/screen_reader True".format(cmd)
+    
+    def get_maint_nodes(self) -> List[str]:
+        cmd = "gcloud alpha compute instances list --project={}".format(self.project)
+        if self.zones is not None:
+            cmd += " --zones=" + ",".join(self.zones)
+        cmd += " --filter=scheduling.maintenanceInterval:PERIODIC --format='table(name)'"
```
nick-stroud, line 49: Format `get` or `value` probably provide the cleanest output for this single field list. There is also `list`.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if self.screen_reader:
+            cmd = self.remove_accessibility(cmd)
+
+        res = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+        if res.returncode == 0:
+            self.maint_nodes = res.stdout.split("\n")[1:-1]
+            if self.regex:
+                self.maint_nodes = list(filter(self.regex.match, names))
```
nick-stroud, line 58: I think names not declared.

tools/check-upcoming-maintenance.py
nick-stroud, line 1: I think we will want some sort of unit testing for this tool. Please create a task for this.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+    parser.add_option("-p", "--project", dest="project",
+                      help="GCP project id")
+    parser.add_option("-z", "--zones", dest="zones",
+                      help="deployment zones (comma delimited, no spaces)")
+    parser.add_option("-n", "--node-regex", dest="node_regex",
+                      help="Regular expression search for nodes")
+    # parser.add_option("-m", "--check-maintenance", action="store_true", dest="check_maintenance",
+    #                   help="Get nodes that have periodic maintenance setup (default)")
+    # parser.add_option("-u", "--check-upcoming", action="store_true", dest="check_upcoming",
+    #                   help="Get nodes that have upcoming maintenance")
```
nick-stroud, line 120: What is the plan with these commented out lines?

tools/check-upcoming-maintenance.py
nick-stroud, line 1: Given the eventual scope of the tool  I would rename this file to `maintenance.py`. I might suggest putting it in its own folder so that we can eventually add a dedicated README and perhaps some tests.

nick-stroud (changes requested):

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import List
+from optparse import OptionParser
+
+class NodeMaintenance:
+    '''
+    Class to keep track of project, zones, and search terms for nodes, as well as results from gcloud queries
+    '''
+    def __init__(self, project, zones, regex):
+        self.project = project
+        res = subprocess.run("gcloud projects describe {}".format(self.project), shell=True, capture_output=True, text=True)
```
mr0re1, line 13: nit. you can use f-string instead of `format`:

```py
f"gcloud projects describe {self.project}"
```

mr0re1, line 13: There are multiple side-effects in this constructor. It's better to have `__init__` without side effects.
I would adivce to have a factory-method that takes you current `__init__` signature and build an instance, while 
change `__init__` contract to take already fetched&parsed state

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+        res = subprocess.run("gcloud config get accessibility/screen_reader", 
+                             shell=True, capture_output=True, text=True)
+        if res.returncode != 0:
+            raise Exception("Error getting accessibility/screen_reader information")
+        if res.stdout.split("\n")[0] == "True":
+            self.screen_reader = True
+        if regex:
+            self.regex = re.compile(regex)
+        self.maint_nodes = None
+        self.upc_maint = None
```
mr0re1, line 36: Is there any reason to make it a part of state ?

tools/check-upcoming-maintenance.py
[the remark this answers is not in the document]
cdunbar13: I believe this is done, but I'll resolve when the checks complete.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
+import os
+import re
+import subprocess
+from typing import List
+from optparse import OptionParser
```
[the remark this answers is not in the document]
cdunbar13: Done

tools/check-upcoming-maintenance.py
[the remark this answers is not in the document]
cdunbar13: Done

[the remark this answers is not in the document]
cdunbar13: Done, only thing that seems to be missing are the function comments.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+                      help="Regular expression search for nodes")
+    # parser.add_option("-m", "--check-maintenance", action="store_true", dest="check_maintenance",
+    #                   help="Get nodes that have periodic maintenance setup (default)")
+    # parser.add_option("-u", "--check-upcoming", action="store_true", dest="check_upcoming",
+    #                   help="Get nodes that have upcoming maintenance")
+
+    (options, args) = parser.parse_args()
+
+    if not options.project:
+        raise Exception("Project must be specified")
```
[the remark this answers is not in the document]
cdunbar13: Done.

tools/check-upcoming-maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+        for row in self.upc_maint:
+            print(row_format.format(*row))
+        print()
+
+if __name__ == "__main__":
+    parser = OptionParser()
+    parser.add_option("-p", "--project", dest="project",
+                      help="GCP project id")
+    parser.add_option("-z", "--zones", dest="zones",
+                      help="deployment zones (comma delimited, no spaces)")
```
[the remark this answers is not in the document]
cdunbar13: Done, but I would like to keep the previous commits there so I can find what I did if we do need it in the future.

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: OBE

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: Uncommented and in use.

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: Done.

tools/check-upcoming-maintenance.py
cdunbar13: Done.  Let me know if you had something else in mind.

tools/check-upcoming-maintenance.py
cdunbar13: Done. Let me know if you had something else in mind.

tools/check-upcoming-maintenance.py
mr0re1: why not just pass `maint_nodes` and `upc_maint` to `__init__` , while calling ` get_upcoming_maintenance` and `get_maintenance_nodes` from `node_maintenace_factory`?

mr0re1 (approved):

nick-stroud (approved):

PR #1978, opened 2023-11-17
Front End Deployment Unique Bucket Name

max-nag:
A UUID has been added to the storage bucket name so that it's unique and the deployment doesn't fail if the name has already been taken

### DIFF
--- a/community/front-end/ofe/tf/main.tf
+++ b/community/front-end/ofe/tf/main.tf
@@ -47,8 +47,7 @@ EOT
   default_labels = {
     ghpcfe_id = var.deployment_name,
   }
-  labels         = merge(var.extra_labels, local.default_labels)
-  generated_uuid = uuid()
+  labels = merge(var.extra_labels, local.default_labels)
 }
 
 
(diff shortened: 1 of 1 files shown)


tpdownes:
The integration test failed with the message below. This seems like an incredibly long bucket name. `uuid` is also the wrong function to use. From the docs:

> This function produces a new value each time it is called... In most cases we recommend using [the random provider](https://registry.terraform.io/providers/hashicorp/random/latest/docs) instead

```
╷
│ Error: error: bucket name validation failed ofe-deployment-10119f-cdb8dc1e-24c0-e9db-83d3-7f05d1536360-storage
│ 
│   with module.control_bucket.google_storage_bucket.buckets["storage"],
│   on .terraform/modules/control_bucket/main.tf line 40, in resource "google_storage_bucket" "buckets":
│   40: resource "google_storage_bucket" "buckets" {
│ 
╵

Error:  Terraform failed.
        Please check parameters and/or seek assistance.
```

What you want is something more like:

```
resource "random_id" "default" {
  byte_length = 2
}

...

  prefix     = "${var.deployment_name}-${random_id.default.hex}"
```

tpdownes (changes requested):
Make changes as noted in https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/1978#issuecomment-1819377485

max-nag:
Yeah that's fair enough. I can change it to use the random provider later in the week

tpdownes (approved):

PR #1979, opened 2023-11-17
Implement Custom Logging

alyssa-sm:
This PR implements a custom logging system to reduce the following print statements.

- fmt.Print
- fmt.Printf
- fmt.Println
- log.Fatal (Note: cases in which the error object was passed as the parameter were not replaced)
- log.Fatalf
- log.Printf
- log.Println

The custom logging system implemented has 3 levels.
1. Info - writes to stdout
2. Warn - writes to stderr
3. Fatal - writes to stderr and exits program

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"hpc-toolkit/pkg/config"
+	"hpc-toolkit/pkg/logging"
 	"hpc-toolkit/pkg/modulewriter"
 	"hpc-toolkit/pkg/validators"
 	"log"
@@ -83,31 +84,31 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 	deplDir := filepath.Join(outputDir, deplName)
 	checkErr(modulewriter.WriteDeployment(dc, deplDir, overwriteDeployment))
 
-	fmt.Println("To deploy your infrastructure please run:")
-	fmt.Println()
-	fmt.Printf(boldGreen("%s deploy %s\n"), execPath(), deplDir)
-	fmt.Println()
+	logging.Info("To deploy your infrastructure please run:")
+	logging.Info("")
+	logging.Info(boldGreen("%s deploy %s"), execPath(), deplDir)
+	logging.Info("")
 	printAdvancedInstructionsMessage(deplDir)
 }
 
 func printAdvancedInstructionsMessage(deplDir string) {
-	fmt.Println("Find instructions for cleanly destroying infrastructure and advanced manual")
-	fmt.Println("deployment instructions at:")
-	fmt.Println()
-	fmt.Printf("%s\n", modulewriter.InstructionsPath(deplDir))
+	logging.Info("Find instructions for cleanly destroying infrastructure and advanced manual")
+	logging.Info("deployment instructions at:")
+	logging.Info("")
+	logging.Info(modulewriter.InstructionsPath(deplDir))
 }
 
 func expandOrDie(path string) config.DeploymentConfig {
 	dc, ctx, err := config.NewDeploymentConfig(path)
 	if err != nil {
-		log.Fatal(renderError(err, ctx))
+		logging.Fatal(renderError(err, ctx))
 	}
 	// Set properties from CLI
 	if err := setCLIVariables(&dc.Config, cliVariables); err != nil {
-		log.Fatalf("Failed to set the variables at CLI: %v", err)
+		logging.Fatal("Failed to set the variables at CLI: %v", err)
 	}
 	if err := setBackendConfig(&dc.Config, cliBEConfigVars); err != nil {
-		log.Fatalf("Failed to set the backend config at CLI: %v", err)
+		logging.Fatal("Failed to set the backend config at CLI: %v", err)
 	}
 	if err := setValidationLevel(&dc.Config, validationLevel); err != nil {
 		log.Fatal(err)
@@ -116,13 +117,13 @@ func expandOrDie(path string) config.DeploymentConfig {
 		log.Fatal(err)
 	}
 	if dc.Config.GhpcVersion != "" {
-		fmt.Printf("ghpc_version setting is ignored.")
+		logging.Info("ghpc_version setting is ignored.")
 	}
 	dc.Config.GhpcVersion = GitCommitInfo
 
 	// Expand the blueprint
 	if err := dc.ExpandConfig(); err != nil {
-		log.Fatal(renderError(err, ctx))
+		logging.Fatal(renderError(err, ctx))
 	}
 
 	validateMaybeDie(dc.Config, ctx)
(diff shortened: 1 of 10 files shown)


cmd/root.go
```diff
@@ -71,7 +72,7 @@ func Execute() error {
 	log.SetFlags(0)
```
mr0re1, line 72: Let's remove it from here.
The `pkg/logging` should take care of 
> // Don't prefix messages with data & time to improve readability.

cmd/create.go
```diff
@@ -83,31 +84,31 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 	deplDir := filepath.Join(outputDir, deplName)
 	checkErr(modulewriter.WriteDeployment(dc, deplDir, overwriteDeployment))
 
-	fmt.Println("To deploy your infrastructure please run:")
-	fmt.Println()
-	fmt.Printf(boldGreen("%s deploy %s\n"), execPath(), deplDir)
-	fmt.Println()
+	logging.Info("To deploy your infrastructure please run:")
+	logging.Info("")
```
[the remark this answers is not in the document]
mr0re1: Other option is "let's not think about it now and address it later", 
At that case you will need to split you logger function into two, say:
`logging.Infof` & `logging.Ingoln`.

Pick either way )

pkg/logging/logging.go
```diff
@@ -0,0 +1,48 @@
...
+	fatallog = log.New(os.Stderr, "", 0)
+}
+
+// Info prints info to stdout
+func Info(fmt string, a ...any) {
+	infolog.Printf(fmt, a...)
+}
+
+// Warn prints info to stderr but does not end the program
+func Warn(fmt string, a ...any) {
```
mr0re1, line 40: It looks more like `Error` .

cmd/root.go
alyssa-sm: Removed.

cmd/create.go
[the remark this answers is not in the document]
alyssa-sm: I changed the logger to append new lines so `Println` behavior can be expected now. I agree that it seems better than `Print` behavior.

pkg/logging/logging.go
alyssa-sm: Renamed it to `Error` instead.

cmd/create.go
```diff
@@ -83,31 +84,31 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
...
+	logging.Info(boldGreen("%s deploy %s\n"), execPath(), deplDir)
```
mr0re1, line 89: Either remove `\n` from format string  (preferable) or `logging.Info("")`. Otherwise you are doing one more newline than original code.

cmd/create.go
```diff
@@ -83,31 +84,31 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
...
 
 func printAdvancedInstructionsMessage(deplDir string) {
-	fmt.Println("Find instructions for cleanly destroying infrastructure and advanced manual")
-	fmt.Println("deployment instructions at:")
-	fmt.Println()
-	fmt.Printf("%s\n", modulewriter.InstructionsPath(deplDir))
+	logging.Info("Find instructions for cleanly destroying infrastructure and advanced manual")
+	logging.Info("deployment instructions at:")
+	logging.Info("")
+	logging.Info("%s\n", modulewriter.InstructionsPath(deplDir))
```
mr0re1, line 98: remove `\n`.

cmd/expand.go
```diff
@@ -50,5 +50,5 @@ var (
 func runExpandCmd(cmd *cobra.Command, args []string) {
 	dc := expandOrDie(args[0])
 	checkErr(dc.ExportBlueprint(outputFilename))
-	fmt.Printf(boldGreen("Expanded Environment Definition created successfully, saved as %s.\n"), outputFilename)
+	logging.Info(boldGreen("Expanded Environment Definition created successfully, saved as %s.\n"), outputFilename)
```
mr0re1, line 53: remove `\n`

cmd/root.go
```diff
@@ -66,12 +67,9 @@ func init() {
 
 // Execute the root command
 func Execute() error {
-	// Don't prefix messages with data & time to improve readability.
-	// See https://pkg.go.dev/log#pkg-constants
-	log.SetFlags(0)
 	mismatch, branch, hash, dir := checkGitHashMismatch()
 	if mismatch {
-		fmt.Fprintf(os.Stderr, "WARNING: ghpc binary was built from a different commit (%s/%s) than the current git branch in %s (%s/%s). You can rebuild the binary by running 'make'\n",
+		logging.Error("WARNING: ghpc binary was built from a different commit (%s/%s) than the current git branch in %s (%s/%s). You can rebuild the binary by running 'make'\n",
```
mr0re1, line 72: remove `\n`

pkg/shell/common.go
```diff
@@ -95,11 +96,11 @@ func CheckWritableDir(path string) error {
 // skip making the proposed changes and continue execution (in deploy command)
 // only if the user responds with "y" or "yes" (case-insensitive)
 func ApplyChangesChoice(c ProposedChanges) bool {
-	log.Printf("Summary of proposed changes: %s", strings.TrimSpace(c.Summary))
+	logging.Info("Summary of proposed changes: %s", strings.TrimSpace(c.Summary))
 	var userResponse string
 
 	for {
-		fmt.Print(`(D)isplay full proposed changes,
+		logging.Info(`(D)isplay full proposed changes,
```
mr0re1, line 103: I think this one should stay `fmt.Printf` since it's user prompt it should always go to stdout.

cmd/create.go
alyssa-sm: Removed `\n`.

cmd/create.go
alyssa-sm: Removed.

cmd/root.go
alyssa-sm: Removed.

pkg/shell/common.go
alyssa-sm: I changed to back to `fmt.Print` since I assumed that's what you meant?

cmd/create.go
```diff
@@ -83,31 +84,31 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
...
 
 func printAdvancedInstructionsMessage(deplDir string) {
-	fmt.Println("Find instructions for cleanly destroying infrastructure and advanced manual")
-	fmt.Println("deployment instructions at:")
-	fmt.Println()
-	fmt.Printf("%s\n", modulewriter.InstructionsPath(deplDir))
+	logging.Info("Find instructions for cleanly destroying infrastructure and advanced manual")
+	logging.Info("deployment instructions at:")
+	logging.Info("")
+	logging.Info("%s", modulewriter.InstructionsPath(deplDir))
```
mr0re1, line 98: There is no point in `"%s"` format string. Just:

```go
logging.Info(modulewriter.InstructionsPath(deplDir))
```

cmd/create.go
```diff
@@ -134,29 +135,29 @@ func validateMaybeDie(bp config.Blueprint, ctx config.YamlCtx) {
...
-	log.Println("validators are shown below.")
-	log.Println("")
-	log.Println("- https://goo.gle/hpc-toolkit-troubleshooting")
-	log.Println("- https://goo.gle/hpc-toolkit-validation")
-	log.Println("")
-	log.Println("Validators can be silenced or treated as warnings or errors:")
-	log.Println("")
-	log.Println("- https://goo.gle/hpc-toolkit-validation-levels")
-	log.Println("")
+	logging.Error((renderError(err, ctx)))
```
mr0re1, line 138: Too many brackets

pkg/sourcereader/sourcereader_test.go
```diff
@@ -51,7 +51,7 @@ func setup() {
 	dirName := fmt.Sprintf("ghpc_sourcereader_test_%s", t.Format(time.RFC3339))
 	dir, err := os.MkdirTemp("", dirName)
 	if err != nil {
-		log.Fatalf("sourcereader_test: %v", err)
+		logging.Fatal("sourcereader_test: %v", err)
```
mr0re1, line 54: Giving it a second thought, there is no reason to use custom logging in `*_test.go` files.
If you agree, please reset all test files.

mr0re1: To clarify, I don't think that `log.Fatalf` is an appropriate way to terminate tests neither, it should be done by means of test library, but this is out of scope of this PR.

alyssa-sm: I will switch them. If we ever want custom logging for the test files, we can always include them back in.

mr0re1: #1985  for you to review

cmd/create.go
alyssa-sm: Done

pkg/sourcereader/sourcereader_test.go
alyssa-sm: I switched them back so custom logging isn't in any `*_test.go` files.

mr0re1 (approved):

PR #1993, opened 2023-11-27
Refactor `pkg/sourcereader` tests.

mr0re1:
* Don't hardcode creation of test dummy files, use embedded folder for this;
* Use `c.MkDir()` to one-line creation of test folders;
* Split `MySuite` into `local`, `embedded` and `goget` suites with different setup;

### DIFF
--- a/pkg/sourcereader/embedded.go
+++ b/pkg/sourcereader/embedded.go
@@ -28,7 +28,7 @@ import (
 var ModuleFS BaseFS
 
 // BaseFS is an join interface with the functionality needed
-// in copyDirFromModules. Works with embed.FS and afero.FS
+// in copyDir. Works with embed.FS and afero.FS
 type BaseFS interface {
 	fs.ReadDirFS
 	fs.ReadFileFS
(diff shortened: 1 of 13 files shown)


pkg/sourcereader/local_test.go
```diff
@@ -115,13 +58,29 @@ func (s *MySuite) TestGetModule_Local(c *C) {
...
 
 	// Invalid: Unsupported Module Source by LocalSourceReader
 	badSource := "gcs::https://www.googleapis.com/storage/v1/GoogleCloudPlatform/hpc-toolkit/modules"
-	err = reader.GetModule(badSource, dest)
+	err = s.r.GetModule(badSource, dest)
 	expectedErr = "source is not valid: .*"
 	c.Assert(err, ErrorMatches, expectedErr)
 }
+
+func (s *localSuite) TestCopyFromPath_Local(c *C) {
```
cdunbar13, line 72: Do you test if the path to copy from doesn't exist?

pkg/sourcereader/embedded.go
```diff
@@ -53,8 +53,8 @@ func copyFileOut(bfs BaseFS, src string, dst string) error {
 	return nil
 }
 
-// copyDirFromModules copies an FS directory to a local path
-func copyDirFromModules(bfs BaseFS, source string, dest string) error {
+// copyDir copies an FS directory to a local path
+func copyDir(bfs BaseFS, source string, dest string) error {
```
cdunbar13, line 57: Just a question about this function, how does isDIr handle links?  Is it possible that a link to a parent folder would cause the recursive function to infinite loop and OOM?

pkg/sourcereader/local_test.go
mr0re1: Good point. Added test for this

pkg/sourcereader/embedded.go
mr0re1: Good point, created a bug to look at it.

PR #1994, opened 2023-11-27
Add `addlicense` pre-commit hook

mr0re1:
* Add `addlicense` pre-commit hook;
* Remove explicit invokation of it.

_Motivation:_ to homogenize dev-side and github-side checks. It's annoying to see PR-validation fail with it.


tpdownes (approved):
This seems like an improvement and I don't believe it impacts the normal user experience unless they are making contributions to the Toolkit.

PR #1995, opened 2023-11-28
Overhaul golang pre-commit hooks

mr0re1:
* Remove flaky `go-build` as it can't be configures straight;
* Remove questionable `go-cyclo`;
* Remove skip-rules for `go-critic`;
* Remove `go-cyclo`;
* Remove sunsetted `github.com/tekwizely/pre-commit-golang`;
* Replace with analogues hooks.


cboneti:
I would like to make sure we don't increase cyclomatic complexity beyond a certain point, in fact, I would love to see it < 10 in all places. Would this still track that? I believe we enforced < 15 before.

cboneti (changes requested):
does this guarantee we don't inflace cyclomatic complexity?

mr0re1:
Restored `gocyclo` as "local" hook

cboneti (commented):
LGTM, but I would still like to see Tom's review.. :)

cboneti (approved):
LGTM, but I still would like to see Tom's review.

.pre-commit-config.yaml
```diff
@@ -64,27 +64,26 @@ repos:
     args: ['-c', '"Google LLC"', '-l', 'apache']
     exclude: docs/videos/healthcare-and-life-sciences/lysozyme-example/submit.sh
     pass_filenames: true
+  - id: gocyclo
+    name: gocyclo
+    language: system
+    entry: gocyclo
+    args: ['-over', '15']
+    files: '\.go$'
```
tpdownes, line 72: nit. instead of `files`, use:

```yaml
  types: [ go ]
```

See https://pre-commit.com/#filtering-files-with-types

.pre-commit-config.yaml
```diff
@@ -64,27 +64,26 @@ repos:
...
 - repo: https://github.com/tekwizely/pre-commit-golang
   rev: v1.0.0-rc.1
   hooks:
   - id: go-critic
-    args: [-disable, "#experimental,sloppyTypeAssert"]
+  - id: go-mod-tidy
 - repo: https://github.com/Bahjat/pre-commit-golang
   rev: v1.0.2
   hooks:
+  - id: go-fmt-import
```
tpdownes, line 83: Could the Bahjat hooks be replaced with:

- https://github.com/TekWizely/pre-commit-golang#go-fmt
- https://github.com/TekWizely/pre-commit-golang#go-imports
- https://github.com/TekWizely/pre-commit-golang#go-vet (and "sub"-hooks)

I don't see an alternative to go-unit-tests but... generally all of these repos seem equally dead. Are we getting anything out of this?

tpdownes (commented):
Let's discuss the value here. We appear to be migrating from a repo which affirmatively calls itself dead to another repo which is practically also dead.

tpdownes (approved):

PR #1997, opened 2023-11-28
Add modules metadata. Part 1 of 4

mr0re1:


harshthakkar01:
Would be good to add description as to why we are adding this metadata.

harshthakkar01 (approved):

PR #2002, opened 2023-11-28
Split `pkg/validators/validators.go` on multiple files.

mr0re1:
**NO CODE CHANGES** just moving around.


harshthakkar01:
Any reason why we want to move this ?

harshthakkar01 (approved):

mr0re1:
> Any reason why we want to move this ?

Too much for just one file. It's becoming a mess.

PR #2006, opened 2023-11-29
`ghpc import-inputs`. Fix panic; refactor to meet `gocyclo` requirements

mr0re1:
```sh ./ghpc deploy clues
$ 
./ghpc deploy clues

...

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

subnetwork_name_network = "default"
Collecting terraform outputs from clues/zero
Writing outputs artifact from deployment group zero to file clues/.ghpc/artifacts/zero_outputs.tfvars
collecting outputs for group one from group zero
Error: duplicate key subnetwork_name_network
```

### DIFF
--- a/pkg/shell/common_test.go
+++ b/pkg/shell/common_test.go
@@ -85,12 +85,14 @@ func (s *MySuite) TestCheckWritableDir(c *C) {
 }
 
 func (s *MySuite) TestMergeMapsWithoutLoss(c *C) {
-	m1 := map[string]int{"foo": 0}
-	m2 := map[string]int{"bar": 1}
+	t := map[string]int{"foo": 0}
+	f := map[string]int{"bar": 1}
 
-	mergeMapsWithoutLoss(m1, m2)
+	c.Check(mergeMapsWithoutLoss(t, f), IsNil)
+	c.Check(f, DeepEquals, map[string]int{"bar": 1})
+	c.Check(t, DeepEquals, map[string]int{"foo": 0, "bar": 1})
 
-	c.Assert(func() { mergeMapsWithoutLoss(m1, m1) }, PanicMatches, "unexpected key collision in maps")
+	c.Check(mergeMapsWithoutLoss(t, f), ErrorMatches, "duplicate key bar")
 }
 
 func (s *MySuite) TestValidateDeploymentDirectory(c *C) {
--- a/pkg/shell/terraform.go
+++ b/pkg/shell/terraform.go
@@ -332,6 +332,36 @@ func ExportOutputs(tf *tfexec.Terraform, artifactsDir string, applyBehavior Appl
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


pkg/shell/common_test.go
```diff
@@ -85,12 +85,14 @@ func (s *MySuite) TestCheckWritableDir(c *C) {
...
+	t := map[string]int{"foo": 0}
+	f := map[string]int{"bar": 1}
 
-	mergeMapsWithoutLoss(m1, m2)
+	c.Check(mergeMapsWithoutLoss(t, f), IsNil)
+	c.Check(f, DeepEquals, map[string]int{"bar": 1})
+	c.Check(t, DeepEquals, map[string]int{"foo": 0, "bar": 1})
 
-	c.Assert(func() { mergeMapsWithoutLoss(m1, m1) }, PanicMatches, "unexpected key collision in maps")
+	c.Check(mergeMapsWithoutLoss(t, f), ErrorMatches, "duplicate key (foo|bar)")
```
tpdownes, line 95: Why is this `(foo|bar)` and not `bar`?

pkg/shell/terraform.go
```diff
@@ -342,84 +372,67 @@ func ImportInputs(deploymentGroupDir string, artifactsDir string, expandedBluepr
...
 	}
 
-	var outfile string
+	var outFile string
+	var toImport map[string]cty.Value // input values to be imported
+
 	switch g.Kind() {
 	case config.TerraformKind:
-		outfile = filepath.Join(deploymentGroupDir, fmt.Sprintf("%s_inputs.auto.tfvars", g.Name))
+		outFile = fmt.Sprintf("%s_inputs.auto.tfvars", g.Name)
```
tpdownes, line 395: It seems to me that the different cases for outFile are not being treated the same anymore. This one doesn't have a directory path prepended to it. Is this intentional?

tpdownes (commented):
A couple questions to address.

pkg/shell/common_test.go
mr0re1: Good spot, thank you! The matcher was used for  `mergeMapsWithoutLoss(m1, m1)`, where error isn't deterministic. With new `mergeMapsWithoutLoss(t, f)`, it's always `bar`. Updated.

pkg/shell/terraform.go
mr0re1:
```go
// tf case
outFile = fmt.Sprintf("%s_inputs.auto.tfvars", g.Name)
// packer case
outFile = filepath.Join(modPath, fmt.Sprintf("%s_inputs.auto.pkrvars.hcl", mod.ID))

// after switch
outPath := filepath.Join(deploymentGroupDir, outFile)
```

Both cases end up in `deploymentGroupDir`.

tpdownes (approved):

tpdownes (approved):

PR #1193, opened 2023-04-19
Enable intergroup references in "advanced usage" mode

tpdownes:
- disable errors when references across groups are made
- add "advanced usage" warning to instructions for groups that make reference to previous groups
- modify metadata to eliminate "inputs" field and replace with the graph connections from the group to "deployment_inputs" and "intergroup_inputs"; these are the deployment variables and intergroup outputs necessary to resolve the group input settings (in Terraform these match group input variable names, for Packer they may not)
- enable deep resolution of deployment variables in Packer settings that are structural types
    - i.e. one can embed the value of `$(vars.zone)` in a object or tuple in `default.auto.pkrvars.hcl`
- eliminate "hidden" 2-field use keyword in favor of single field containing only the module

Packer groups are still subject to further development based upon active work building upon #1180. A multi-group Terraform deployment is shown below. If one executes the following command between invocations of `terraform apply` it will propagate the values to the second group. We anticipate automating this functionality in the near future.

```shell
terraform -chdir=igc-tf-test/zero output -json | jq 'with_entries(.value = .value.value)' > igc-tf-test/two/inputs.auto.tfvars.json
```

```yaml
---
blueprint_name: igc

vars:
  project_id: toolkit-demo-zero-e913
  deployment_name: igc-tf-test
  region: us-east4
  zone: us-east4-c

deployment_groups:
- group: zero
  modules:
  - id: network0
    source: modules/network/vpc
  - id: homefs
    source: modules/file-system/filestore
    use: [network0]
    settings:
      local_mount: /home
  - id: projectsfs
    source: modules/file-system/filestore
    use: [network0]
    settings:
      local_mount: /projects
- group: two
  modules:
  - id: network1
    source: modules/network/pre-existing-vpc
    use:
    - network0
  - id: vm
    source: modules/compute/vm-instance
    use:
    - network1
    - homefs
    - projectsfs
    settings:
      machine_type: n2-standard-4
```

A Packer blueprint is shown below. This has more limited functionality due to variable name matching and evaluation of settings that we have discussed offline. We anticipate automating both of these issues in near future.

```yaml
---
blueprint_name: igc

vars:
  project_id: toolkit-demo-zero-e913
  deployment_name: igc-pkr-test
  region: us-east4
  zone: us-east4-c

deployment_groups:
- group: zero
  modules:
  - id: network0
    source: modules/network/vpc
  - id: homefs
    source: modules/file-system/filestore
    use: [network0]
    settings:
      local_mount: /home
  - id: projectsfs
[...]

PR #2007, opened 2023-11-30
Update `checkOverwriteAllowed`

mr0re1:
**BREAKING CHANGE:**
Use `expanded_blueprint.yaml` to get list of groups instead of listing deployment dir. 

This will render old deployments (created by GHPC version 7 months old, before #1193 ) as "malformed" from POV of overwriting logic => user will not be able to overwrite
Other changes:
* Move it from `modulewriter` to `cmd`;
* Make it to return detailed error;
```sh
...
Error: you are attempting to remove a deployment group "one", which is not supported
```

```sh
...
Error: deployment folder "clues" already exists, use -w to overwrite
```

* Make `modulewriter` unaware of overwriting logic;
* Compare baked `ghpc_versions`, warn if mismatch.

```sh
...
WARNING: ghpc_version has changed from "v1.25.0-156-gd198216a" to "v1.25.0-155-g136bc46a"
To deploy your infrastructure please run:
...
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -303,7 +303,7 @@ func checkOverwriteAllowed(depDir string, bp config.Blueprint, overwriteFlag boo
 	// try to get previous deployment
 	expPath := filepath.Join(modulewriter.ArtifactsDir(depDir), modulewriter.ExpandedBlueprintName)
 	if _, err := os.Stat(expPath); os.IsNotExist(err) {
-		return fmt.Errorf("expanded blueprint file %q is missing, maybe your previous GHPC version is too old", expPath)
+		return fmt.Errorf("expanded blueprint file %q is missing, this could be a result of changing GHPC version between consecutive deployments", expPath)
 	}
 	prev, _, err := config.NewDeploymentConfig(expPath)
 	if err != nil {
@@ -311,7 +311,7 @@ func checkOverwriteAllowed(depDir string, bp config.Blueprint, overwriteFlag boo
 	}
 
 	if prev.Config.GhpcVersion != bp.GhpcVersion {
-		logging.Info("WARNING: ghpc_version has changed from %q to %q", prev.Config.GhpcVersion, bp.GhpcVersion)
+		logging.Info("WARNING: ghpc_version has changed from %q to %q, using different versions of GHPC to update a live deployment is not officially supported. Proceed at your own risk", prev.Config.GhpcVersion, bp.GhpcVersion)
 	}
 
 	if !overwriteFlag {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -24,6 +24,7 @@ import (
 	"hpc-toolkit/pkg/logging"
 	"hpc-toolkit/pkg/modulewriter"
 	"hpc-toolkit/pkg/validators"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


cmd/create.go
```diff
@@ -286,3 +288,41 @@ func filterYaml(cmd *cobra.Command, args []string, toComplete string) ([]string,
...
+
+	useWErr := fmt.Errorf("deployment folder %q already exists, use -w to overwrite", depDir)
+	// try to get previous deployment
+	expPath := filepath.Join(modulewriter.ArtifactsDir(depDir), modulewriter.ExpandedBlueprintName)
+	prev, _, err := config.NewDeploymentConfig(expPath)
+	if err != nil { // malformed previous deployment, rely on flag
+		if overwriteFlag {
+			return nil
+		}
+		return useWErr
```
nick-stroud, line 306: I think would prefer to just not allow overwrite in the malformed case. Can we tell the user to use an intermediate version to bridge the gap, or just stop them. In this case we know they are re-applying with a changing ghpc version which we specifically do not recommend. Maybe instead of version warning, in this case we do version error "you are trying to change version of ghpc and your previous version is too old"

cmd/create.go
```diff
@@ -286,3 +288,41 @@ func filterYaml(cmd *cobra.Command, args []string, toComplete string) ([]string,
...
+		if !newGroups[g.Name] {
+			return fmt.Errorf("you are attempting to remove a deployment group %q, which is not supported", g.Name)
+		}
+	}
+
+	if prev.Config.GhpcVersion != bp.GhpcVersion {
+		logging.Info("WARNING: ghpc_version has changed from %q to %q", prev.Config.GhpcVersion, bp.GhpcVersion)
+	}
+
+	if !overwriteFlag {
```
nick-stroud, line 324: Without thinking too hard about it I would put these checks in the following order:
1. warn on ghpc version - as just a warning no harm to always check
2. check for overwrite flag - if they have not approved overwrite then maybe we should not give warnings about deleting groups
3. error on group removal

Is there rational to use the current order?

cmd/create.go
nick-stroud, line 1: `created by GHPC version 7 months old` > Please update the description to have a specific reference to the last compatible version (or PR).

nick-stroud (changes requested):

cmd/create.go
```diff
@@ -295,15 +295,27 @@ func checkOverwriteAllowed(depDir string, bp config.Blueprint, overwriteFlag boo
...
-	useWErr := fmt.Errorf("deployment folder %q already exists, use -w to overwrite", depDir)
+	if _, err := os.Stat(modulewriter.HiddenGhpcDir(depDir)); os.IsNotExist(err) {
+		// hidden ghpc dir does not exist
+		return fmt.Errorf("folder %q already exists, and it is not a valid GHPC deployment folder", depDir)
+	}
+
 	// try to get previous deployment
 	expPath := filepath.Join(modulewriter.ArtifactsDir(depDir), modulewriter.ExpandedBlueprintName)
+	if _, err := os.Stat(expPath); os.IsNotExist(err) {
+		return fmt.Errorf("expanded blueprint file %q is missing, maybe your previous GHPC version is too old", expPath)
```
nick-stroud, line 306: nit: I know you are using the verbiage from my comment but the combination of "maybe" and "too old" look less professional than I would like.

Could we instead say "expanded blueprint file %q is missing, this could be a result of changing GHPC version between consecutive deployments."

cmd/create.go
```diff
@@ -295,15 +295,27 @@ func checkOverwriteAllowed(depDir string, bp config.Blueprint, overwriteFlag boo
...
-		if overwriteFlag {
-			return nil
-		}
-		return useWErr
+	if err != nil {
+		return err
+	}
+
+	if prev.Config.GhpcVersion != bp.GhpcVersion {
+		logging.Info("WARNING: ghpc_version has changed from %q to %q", prev.Config.GhpcVersion, bp.GhpcVersion)
```
nick-stroud, line 314: I think we should tell the user that this is not officially supported behavior. 

`"WARNING: ghpc_version has changed from %q to %q, using different versions of GHPC to update a live deployment is not officially supported. Proceed at your own risk"`

nick-stroud (changes requested):

nick-stroud (approved):

PR #2009, opened 2023-11-30
Update slurm modules to point to fork

mr0re1:


tpdownes (approved):
Please look at the Slurm build failure and merge if it is transient.

PR #2024, opened 2023-12-04
Don't test metadata services against legacy ones.

mr0re1:
Legacy metadata to be removed, added comment.


pkg/inspect/modules_test.go
nick-stroud, line 1: IIUC:
- check that metadata file exists
- check that services are defined, I am assuming that empty list is ok as long as it is not nil

If true then proceed.

nick-stroud (approved):

PR #2029, opened 2023-12-08
Disable `go-critic` hook

mr0re1:
Broken: https://github.com/go-critic/go-critic/issues/1388


cboneti:
what are you replacing it for?

mr0re1:
> what are you replacing it for?

I'm not replacing it with anythin, just temporarily disabling it, since it's broken.
Fallback to older binary version doesn't help, so I assume this is transitive dependency issue.

Disabling it to prevent PR-validation failures, that can (or not) start happening after next builder image roll out.

PR #2030, opened 2023-12-08
Allow the `wait-for-startup` module to take a list of instance names

rohitramu:
This PR adds a new setting to the `wait-for-startup` module called `instance_names` (plural), which extends the functionality of this module to wait for startup scripts to complete on multiple VM instances.

The `instance_names` setting takes a list of VM instance names, and allows the `name` output variable from the `compute/vm-instance` module to be passed in.  For example, here is a minimal snippet from a blueprint which was used to test this change:

```yaml
  - id: vm
    use:
    - network1
    - shared-fs
    - config-startup
    settings:
      instance_count: 8
  - id: wait-for-vm
    settings:
      instance_names: $(vm.name)
```

Before this change, the above snippet would've had to have been defined like this in order to wait for startup scripts to complete in all 8 instances (notice the plurality of `instance_name` vs. `instance_names`):

```yaml
  - id: vm
    use:
    - network1
    - shared-fs
    - config-startup
    settings:
      instance_count: 8
  - id: wait-for-vm-0
    settings:
      instance_name: $(vm.name[0])
  - id: wait-for-vm-1
    settings:
      instance_name: $(vm.name[1])
  - id: wait-for-vm-2
    settings:
      instance_name: $(vm.name[2])
  - id: wait-for-vm-3
    settings:
      instance_name: $(vm.name[3])
  - id: wait-for-vm-4
    settings:
      instance_name: $(vm.name[4])
  - id: wait-for-vm-5
    settings:
      instance_name: $(vm.name[5])
  - id: wait-for-vm-6
    settings:
      instance_name: $(vm.name[6])
  - id: wait-for-vm-7
    settings:
      instance_name: $(vm.name[7])
```

I have tested this change with both of the above snippets to ensure behavior stays the same whether the singular or plural version of the setting is used.  I have also tested with the following to ensure the list is combined with the singular setting correctly:

```yaml
  - id: vm
    use:
    - network1
    - shared-fs
    - config-startup
    settings:
      instance_count: 8
  - id: wait-for-vm-0
    settings:
      instance_name: $(vm.name[0])
      instance_names: $(vm.name)
```

Here's a full test blueprint (one set of VMs is made to fail to test the error case):

```yaml
blueprint_name: startup-vm-instance

vars:
  deployment_name: test-startup
  region: us-east4
  zone: us-east4-c

deployment_groups:
- group: first
  modules:
  - id: network1
    settings:
      # network_name: imb-test
  - id: script
    settings:
      runners:
      - type: shell
        destination: tmp.sh
        content: |
          #!/bin/bash
          sleep 60
  - id: script_fail
    settings:
      runners:
      - type: shell
        destination: tmp.sh
        content: |
          #!/bin/bash
          exit 1

  - id: vm0
    use:
    - network1
    - script
    settings:
      name_prefix: vm0
      machine_type: n1-standard-2
      instance_count: 2
  - id: wait-for-vm0
    settings:
      instance_name: $(vm0.name[0])
      instance_names: $(vm0.name)

  - id: vm1
    use:
    - network1
    - script_fail
    settings:
      name_prefix: vm1
      machine_type: n1-standard-2
      instance_count: 4
  - id: wait-for-vm1-0
    settings:
      instance_names: $(vm1.name)
```

### DIFF
--- a/community/modules/scripts/wait-for-startup/README.md
+++ b/community/modules/scripts/wait-for-startup/README.md
@@ -15,10 +15,19 @@ up a node.
 ### Example
 
 ```yaml
+- id: workstation
+  source: modules/compute/vm-instance
+  use:
+  - network1
+  - my-startup-script
+  settings:
+    instance_count: 4
+
+# Wait for all instances of the above VM to finish running startup scripts.
 - id: wait
   source: community/modules/scripts/wait-for-startup
   settings:
-    instance_name: $(workstation.name[0])
+    instance_names: $(workstation.name)
 ```
 
 ## License
@@ -40,18 +49,18 @@ limitations under the License.
 
 ## Requirements
 
-| Name | Version |
-|------|---------|
+| Name                                                                      | Version   |
+| ------------------------------------------------------------------------- | --------- |
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.14.0 |
-| <a name="requirement_google"></a> [google](#requirement\_google) | >= 3.83 |
-| <a name="requirement_null"></a> [null](#requirement\_null) | ~> 3.0 |
+| <a name="requirement_google"></a> [google](#requirement\_google)          | >= 3.83   |
+| <a name="requirement_null"></a> [null](#requirement\_null)                | ~> 3.0    |
 
 ## Providers
 
-| Name | Version |
-|------|---------|
+| Name                                                       | Version |
+| ---------------------------------------------------------- | ------- |
 | <a name="provider_google"></a> [google](#provider\_google) | >= 3.83 |
-| <a name="provider_null"></a> [null](#provider\_null) | ~> 3.0 |
+| <a name="provider_null"></a> [null](#provider\_null)       | ~> 3.0  |
 
 ## Modules
 
@@ -59,19 +68,21 @@ No modules.
 
 ## Resources
 
-| Name | Type |
-|------|------|
-| [null_resource.wait_for_startup](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| Name                                                                                                                                      | Type        |
+| ----------------------------------------------------------------------------------------------------------------------------------------- | ----------- |
+| [null_resource.validate_instance_names](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource)            | resource    |
+| [null_resource.wait_for_startup](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource)                   | resource    |
 | [google_compute_instance.vm_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_instance) | data source |
 
 ## Inputs
 
-| Name | Description | Type | Default | Required |
-|------|-------------|------|---------|:--------:|
-| <a name="input_instance_name"></a> [instance\_name](#input\_instance\_name) | Name of the instance we are waiting for | `string` | n/a | yes |
-| <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Project in which the HPC deployment will be created | `string` | n/a | yes |
-| <a name="input_timeout"></a> [timeout](#input\_timeout) | Timeout in seconds | `number` | `1200` | no |
-| <a name="input_zone"></a> [zone](#input\_zone) | The GCP zone where the instance is running | `string` | n/a | yes |
+| Name                                                                           | Description                                         | Type           | Default | Required |
+| ------------------------------------------------------------------------------ | --------------------------------------------------- | -------------- | ------- | :------: |
+| <a name="input_instance_name"></a> [instance\_name](#input\_instance\_name)    | Name of the instance we are waiting for             | `string`       | `null`  |    no    |
+| <a name="input_instance_names"></a> [instance\_names](#input\_instance\_names) | A list of names of the instances we are waiting for | `list(string)` | `[]`    |    no    |
+| <a name="input_project_id"></a> [project\_id](#input\_project\_id)             | Project in which the HPC deployment will be created | `string`       | n/a     |   yes    |
+| <a name="input_timeout"></a> [timeout](#input\_timeout)                        | Timeout in seconds                                  | `number`       | `1200`  |    no    |
+| <a name="input_zone"></a> [zone](#input\_zone)                                 | The GCP zone where the instance is running          | `string`       | n/a     |   yes    |
 
 ## Outputs
 
--- a/community/modules/scripts/wait-for-startup/README.md
+++ b/community/modules/scripts/wait-for-startup/README.md
@@ -49,18 +49,18 @@ limitations under the License.
 
 ## Requirements
 
-| Name                                                                      | Version   |
-| ------------------------------------------------------------------------- | --------- |
+| Name | Version |
+|------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.14.0 |
-| <a name="requirement_google"></a> [google](#requirement\_google)          | >= 3.83   |
-| <a name="requirement_null"></a> [null](#requirement\_null)                | ~> 3.0    |
+| <a name="requirement_google"></a> [google](#requirement\_google) | >= 3.83 |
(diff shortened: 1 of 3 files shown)


community/modules/scripts/wait-for-startup/README.md
```diff
@@ -15,10 +15,23 @@ up a node.
...
+# Wait for all instances of the above VM to finish running startup scripts.
+- id: wait
+  source: community/modules/scripts/wait-for-startup
+  settings:
+    instance_names: $(workstation.name)
 - id: wait
   source: community/modules/scripts/wait-for-startup
   settings:
-    instance_name: $(workstation.name[0])
+    instance_names: $(workstation.name)
```
nick-stroud, line 34: This duplication looks like a typo.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,42 @@
...
  * limitations under the License.
  */
+
+locals {
+  instance_names = toset(compact(concat(var.instance_names, [var.instance_name])))
+}
+
+resource "null_resource" "validate_instance_names" {
+  lifecycle {
+    precondition {
```
nick-stroud, line 23: I think you could put this precondition on the lifecycle of the data block and avoid having a separate null_resource

community/modules/scripts/wait-for-startup/main.tf
nick-stroud, line 1: What testing has been done: put in description. 

Some integration tests use this module, please trigger a few that use it. Have you observed success and failure in your manual testing? With the for_each + local_exec are we sure that it waits for all to succeed. The worst thing we could do would be to hide real failures.

community/modules/scripts/wait-for-startup/variables.tf
```diff
@@ -17,6 +17,13 @@
 variable "instance_name" {
   description = "Name of the instance we are waiting for"
   type        = string
+  default     = null
+}
+
+variable "instance_names" {
+  description = "A list of names of the instances we are waiting for"
```
nick-stroud, line 24: One of these descriptions should mention that we merge with the other.

nick-stroud (changes requested):

community/modules/scripts/wait-for-startup/main.tf
rohitramu: Done.

community/modules/scripts/wait-for-startup/main.tf
rohitramu: I've updated the description with relevant snippets from the tests I ran.

I have seen the output of success and failure cases (i.e. I put an `exit 1` in one of the scripts I tested) look identical between the current binary in develop and the binary built from this PR.

community/modules/scripts/wait-for-startup/README.md
rohitramu: Good catch, fixed.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,40 @@
...
  * limitations under the License.
  */
+
+locals {
+  instance_names = toset(compact(concat(var.instance_names, [var.instance_name])))
+}
+
 data "google_compute_instance" "vm_instance" {
-  name    = var.instance_name
+  for_each = local.instance_names
```
mr0re1, line 22: IIUC, `for_each` can't be used with values not known in advance.
If we one uses `random_pet` or `random_id` modules to come up with `instance_names`, this line will throw.

mr0re1: One way to fix it is to not use `toset(compact(concat(...))))` that has an unknown length,
but raw `var.instance_names` itself that has a better chances of having known length.

rohitramu: The `for_each` variable can only be assigned a `set()` value, otherwise it throws an error:

```
Error: Invalid for_each argument

  on modules/embedded/community/modules/scripts/wait-for-startup/main.tf line 22, in data "google_compute_instance" "vm_instance":
  22:   for_each = local.instance_names
    ├────────────────
    │ local.instance_names is tuple with 5 elements

The given "for_each" argument value is unsuitable: the "for_each" argument
must be a map, or set of strings, and you have provided a value of type
tuple.
```

Are there any other ways of preventing the possibility of an unknown length of instance names?  I can't think of any, but I'm certainly not a Terraform expert.

rohitramu: It looks like `count` doesn't have the same restriction as `for_each`: https://github.com/hashicorp/terraform/issues/30039.  I will use `count` instead.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,40 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+locals {
+  instance_names = compact(concat(var.instance_names, [var.instance_name]))
```
mr0re1, line 18: You can not use `compact` here for the same reasons.

rohitramu: Oops, sorry I missed that.  Fixed now.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,40 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+locals {
+  instance_names = concat(var.instance_names, [var.instance_name])
```
mr0re1, line 18: In case `var.instance_name` is not provided, how does `wait-for-startup-status.sh` handles empty(?) `INSTANCE_NAME`. 
I think better solution would be to deprecate `var.instance_name` in favor of `var.instance_names`

rohitramu: Good point.  I've changed it to have a separate resource for each variable.  This should cover all error cases.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,13 +13,27 @@
...
+  lifecycle {
+    precondition {
+      condition     = var.instance_name != null || length(var.instance_names) > 0
+      error_message = "At least one instance name must be provided"
+    }
+  }
+}
+
+data "google_compute_instance" "vm_instance_single" {
+  count = var.instance_name == null ? 0 : 1
```
mr0re1, line 27: It exhibits the same problem. I only see two options, either:
* let it fail;
* deprecate `var.instance_name`.

rohitramu: I'd like to avoid making a change that could break currently working blueprints.

I think this change would be big improvement to usability of the module.  Given that I haven't been able to repro the failure with any valid HPC Toolkit blueprint (but it is possible to make it fail with hand-written Terraform code) I would be in favor of just letting it fail.

@nick-stroud, any thoughts?

nick-stroud: I can repro with the following snippet:
```
  - id: new-bucket
    source: community/modules/file-system/cloud-storage-bucket
    settings:
      random_suffix: true

  - source: ./modules/compute/vm-instance
    id: workstation
    use:
    - network1
    settings:
      name_prefix: $(new-bucket.gcs_bucket_name)
      machine_type: n2-standard-2
      instance_count: 2
      startup_script: |
        sleep 10

  - id: wait0
    source: ./community/modules/scripts/wait-for-startup
    settings:
      instance_name: $(workstation.name[0])
```

There is no random naming behavior in vm-instance, but we name it after the bucket that has a random name.

Interestingly `instance_names: $(workstation.name)` has no issue, as the length of array is known even though the names are not. I assume it would be possible to randomize that as well although, much less likely to occur, I think.

One option would be to have the singular form always run (no count) and just make the script tolerant to null value being passed (exit immediately). In that case I think you could even go back to the original formulation where you combine the singular with the array, just don't perform the set function.

community/modules/scripts/wait-for-startup/variables.tf
nick-stroud: Assuming we get this to work. I would just say something along the lines of, "the module will wait for all VMs provided to either `instance_name` or `instance_names`." 

Basically make it clear we are checking both.

rohitramu:
Closing due to lack of bandwidth. Will reopen when I get time to work on it again.

PR #2032, opened 2023-12-08
Add `maintenance_interval` var to the SlurmV5 node group.

mr0re1:
DO_NOT_SUBMIT
Point SlurmV5 modules to tag `5.9.1_20231213`


community/modules/compute/schedmd-slurm-gcp-v5-node-group/variables.tf
```diff
@@ -304,6 +304,17 @@ variable "reservation_name" {
   nullable    = false
 }
 
+variable "maintenance_interval" {
+  description = "Specifies the frequency of planned maintenance events. The accepted values are: PERIODIC"
```
tpdownes, line 308: I would set `nullable = false` here. Among other reasons, this will generate an error (rather than validation failure):

```
> contains(["", "PERIODIC"], null)
╷
│
│
╵
```

tpdownes, line 308: In the description I would do somethimg more like:

```
The only accepted value is \"PERIODIC\". Leave omitted to avoid specifying an interval.
```

mr0re1:
> Practically speaking, we cannot merge this until we publish an image or we should use enable_devel in any blueprint that uses non-empty maintenance_interval?

Correct, I will put it on hold after all comments are resolved;

PR #2035, opened 2023-12-09
Use CFT-metadata struct directly instead of mimicing it.

mr0re1:


nick-stroud (approved):
I wondered if this would slow down build and ran a few tests with inconclusive results. Generally I think this is a good direction to head.

mr0re1:
> this would slow down build 

The whole `pkg/inspect` test takes `170ms` , I thinks it's toleratable. If `GetMetadata` is never called this package test run takes `160ms`. I think it's toleratable.

> tests with inconclusive results

You brought a good point.

PR #2045, opened 2023-12-13
Hotfix wait for startup update

cdunbar13:
This PR is to fix the wait-for-startup script in the wake of the guest agent being updated.  We needed to add a new matching string to account for the new error message, which was introduced about a month ago.

Testing was run by creating a vm-instance with a startup script with various changes:

1. Script that just runs "exit 0"
2. Script that runs "exit 1"
3. Script that slept for longer than the timeout
4. Script that slept and ended before the timeout

Tested on:
Latest hpc-centos-7 image
Latest rocky-8 image
`hpc-centos-7-v20220214` - for old agent
`rocky-linux-9-v20221102` - for old agent

This issue was introduced by a commit to [guest-agent](https://github.com/GoogleCloudPlatform/guest-agent/commit/5c85572e7f2b36d14668bef2c53e3acf7347d798)

### DIFF
--- a/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
+++ b/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
@@ -36,14 +36,19 @@ c1grep() { grep "$@" || test $? = 1; }
 now=$(date +%s)
 deadline=$((now + TIMEOUT))
 error_file=$(mktemp)
-fetch_cmd="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} --port 1 --zone ${ZONE} --project ${PROJECT_ID}"
+fetch_cmd="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} \
+           --port 1 --zone ${ZONE} --project ${PROJECT_ID}"
+# Match string for all finish types of the old guest agent and successful
+# finishes on the new guest agent
 FINISH_LINE="startup-script exit status"
+# Match string for failures on the new guest agent
 FINISH_LINE_ERR="Script.*failed with error:"
 
 until [[ now -gt deadline ]]; do
 	ser_log=$(
 		set -o pipefail
-		${fetch_cmd} 2>"${error_file}" | c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}"
+		${fetch_cmd} 2>"${error_file}" |
+			c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}"
 	) || {
 		cat "${error_file}"
 		exit 1
@@ -54,6 +59,8 @@ until [[ now -gt deadline ]]; do
 	now=$(date +%s)
 done
 
+# This line checks for an exit code - the assumption is that there is a number
+# at the end of the line and it is an exit code
 STATUS=$(sed -r 's/.*([0-9]+)\s*$/\1/' <<<"${ser_log}" | uniq)
 # This specific text is monitored for in tests, do not change.
 INSPECT_OUTPUT_TEXT="To inspect the startup script output, please run:"
--- a/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
+++ b/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
@@ -37,25 +37,26 @@ now=$(date +%s)
 deadline=$((now + TIMEOUT))
 error_file=$(mktemp)
 fetch_cmd="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} --port 1 --zone ${ZONE} --project ${PROJECT_ID}"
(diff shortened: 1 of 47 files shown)


community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
nick-stroud, line 1: Can you add what manual testing you have done to the PR description.

cdunbar13: Done.  Let me know if it needs more testing.

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -37,25 +37,26 @@ now=$(date +%s)
...
-	STATUS=$(sed -r 's/.*([0-9]+)\s*$/\1/' <<<"${ser_log}" | uniq)
-	if [[ -n "${STATUS}" ]]; then break; fi
-	echo "could not detect end of startup script. Sleeping."
+	if [[ -n "${ser_log}" ]]; then break; fi
+	echo "Could not detect end of startup script. Sleeping."
 	sleep 5
 	now=$(date +%s)
 done
 
+STATUS=$(sed -r 's/.*([0-9]+)\s*$/\1/' <<<"${ser_log}" | uniq)
```
nick-stroud, line 57: IIUC the sed does not chance as this just captures digits at the end of the line. 

Nit: I think this is sufficiently cryptic to add a comment with the intent. Something like `# Isolate any digits at the end of the captured line`

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -37,25 +37,26 @@ now=$(date +%s)
 deadline=$((now + TIMEOUT))
 error_file=$(mktemp)
 fetch_cmd="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} --port 1 --zone ${ZONE} --project ${PROJECT_ID}"
+FINISH_LINE="startup-script exit status"
+FINISH_LINE_ERR="Script.*failed with error:"
```
nick-stroud, line 41: nit: I would give some context to future readers as to why we have two options. If you can think of some descriptive naming that could be good but `OLD_FINISH_LINE` and `NEW_FINISH_LINE` don't seem great. Maybe just a comment that states that the output has changed over time and this script supports both old and new versions.

nick-stroud (dismissed):

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
cdunbar13: Done

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
cdunbar13: Done

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
nick-stroud, line 1: Please make sure you have rerun manual tests with the final version, just to make sure there is nothing silly we miss in the changes to make commands multiline. Alternatively leave commands in single line format.

Note: I prefer the multiline format but since we are releasing this straight away we should check the final version to be absolutely sure it is working.

nick-stroud (approved):

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
tpdownes: I'm a bit worried about the change to multiline as it isn't functioning like a HEREDOC string but literally executing a command as though you typed it over multiple lines. That said, if we run the checks (which I'll start now), I'll be satisfied.

tpdownes (changes requested):
This PR should not be merged until has been manually tested using older releases of the guest-agent.

tpdownes:
> This PR should not be merged until has been manually tested using older releases of the guest-agent.

The offending commit is present in these releaes:

- [20231212.00](https://github.com/GoogleCloudPlatform/guest-agent/releases/tag/20231212.00)
- [20231118.00](https://github.com/GoogleCloudPlatform/guest-agent/releases/tag/20231118.00)
- [20231115.00](https://github.com/GoogleCloudPlatform/guest-agent/releases/tag/20231115.00)

tpdownes (approved):
Further manual testing on old images was successful. I also believe the lustre PR check uses an image before the change in guest-agent.

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
cdunbar13: Done.

PR #2050, opened 2023-12-15
Allow usage of expression for `vars.deployment_name`

mr0re1:
* Perform deployment_name  validation after evaluation;
* Make `evalVars` to return new vars instead of mutating;
* Make `DeploymentName` to not return error.


cdunbar13 (approved):
LGTM

PR #2051, opened 2023-12-16
`vm-instance` is now able to `use` a service account

nick-stroud:
Update vm-instance service account interface to match #1427. This is an incremental move towards what I believe is our new standard interface for service accounts. This was in the way of some testing I was doing.

Now `vm-instance` can `use` a service account.

Tested:
- Manually deployed blueprint with vm-instance `use` a service account.

### Submission Checklist


modules/compute/vm-instance/main.tf
```diff
@@ -214,12 +214,9 @@ resource "google_compute_instance" "compute_vm" {
...
   }
 
-  dynamic "service_account" {
-    for_each = var.service_account == null ? [] : [var.service_account]
-    content {
-      email  = lookup(service_account.value, "email", null)
-      scopes = lookup(service_account.value, "scopes", null)
-    }
+  service_account {
+    email  = var.service_account_email
```
tpdownes, line 218: there is a behavioral change here in that the service account block is always being specified. did you consider using [this data source](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_default_service_account).

Going by memory, the code for this data source is just string interpolation of project # and the known pattern for default GCE service account. So it's lightweight.

tpdownes (commented):
A comment to begin short discussion.

modules/compute/vm-instance/main.tf
nick-stroud: On dynamic vs always specify: Technically, I do not think there is a behavioral change here. 

Omitting `service_account` has the same behavior as:
```
service_account {
    email  = null
    scope = "https://www.googleapis.com/auth/cloud-platform"
}
```

As in the VMs created are identical and switching from one to the other does not generate any plan in terraform.

I think now that `service_account_scopes` is a base variable with a non-null default, it makes more sense to always define the SA block instead of ignoring `service_account_scopes` when `service_account_email` is null.

---
On using `google_compute_default_service_account`:
`email` is listed as optional. When `null` is specified it uses the default service account. It seems more cumbersome to add conditional logic to lookup the default service account if that is what it is doing anyway.

I do think I could update the variable documentation to state that null results in default compute engine. 

---
My opinion on both counts would be to leave as is (with update to variable description), but this opinion is held loosely and I would fold on either to get this merged.

tpdownes: It wouldn't need be conditional, it could be something like

```hcl
# data block for default SA
...
  email  = coalesce(var. service_account_email, data.default_account...)
```

That said, I'll leave it at `nit` level. I strongly agree that an explicit, rather than dynamic service_account block is better.

tpdownes (approved):

modules/compute/vm-instance/main.tf
nick-stroud: Ack. Your proposal looks clean and I liked explicit breadcrumb trail that default will be used. After playing with it I am going to leave as is due to uncertainty of behavior in the case that default service account does not exist or has been disabled. I am not sure how terraform and/or gcp handle this case and do not want to insert myself in the middle.

PR #2053, opened 2023-12-17
Bring `$(...)` functionality on par with `((...))`

mr0re1:
* Use token-replacement instead of string-replacement for expresssion updates;
* Translate any BP-expression to TF-expressions by transforming used traversals;
* Remove notion of `((...))` from documentation.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -48,10 +48,9 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [Top Level Parameters](#top-level-parameters)
   * [Deployment Variables](#deployment-variables)
   * [Deployment Groups](#deployment-groups)
-* [Variables](#variables)
-  * [Blueprint Variables](#blueprint-variables)
-  * [Literal Variables](#literal-variables)
-  * [Escape Variables](#escape-variables)
+* [Variables and expressions](#variables-and-expressions)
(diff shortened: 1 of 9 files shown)


examples/README.md
```diff
@@ -1268,56 +1269,26 @@ deployment_groups:
          settings:
             key1: $(vars.zone)
             key2: $(resource1.name)
+            # access nested fields
+            key3: $(resource1.nodes[0].private_ip)
+            # arithmetic expression
+            key4: $(vars.num_nodes + 5)
+            # string interpolation expression
+            key5: $("${resource1.name}_${vars.zone}")
```
nick-stroud, line 1277: A common use case is people wanting to use a variable inside a script. Is there any way at present to do multiline string interpolation. The equivalent of:

```
contents: |
  #! /bin/bash
  sleep ${var.sleep_count}
```
It looks like the warning for string interpolation is gone. Do we now fully support string interpolation?

I think if any format of multi-line interpolation is possible it would be valuable to show as an example, and if not we should call it out and probably still warn on it.

nick-stroud (changes requested):

examples/README.md
mr0re1: Added support (+doc) for multiline expressions.

nick-stroud (approved):

PR #2057, opened 2023-12-18
Disable firewall rule logging by default

tpdownes:
At high packet rate, we observe performance impacts when Cloud VPC Firewall logging is enabled. This turns logging off by default but allows the user to enable it with metadata included or excluded.

Blueprint tested with all 3 allowed values for var.firewall_log_config. They can be modified in-place non-destructively:

```yaml
---
blueprint_name: startup-vm-instance

vars:
  deployment_name: testfix
  region: us-east4
  zone: us-east4-c

deployment_groups:
- group: first
  modules:
  - id: network1
    source: modules/network/vpc
    settings:
      firewall_log_config: EXCLUDE_ALL_METADATA
  - id: vm0
    use:
    - network1
    settings:
      name_prefix: vm0
      machine_type: n1-standard-2
```

### Submission Checklist


modules/network/vpc/variables.tf
```diff
@@ -227,3 +227,18 @@ variable "firewall_rules" {
   description = "List of firewall rules"
   default     = []
 }
+
+variable "firewall_log_config" {
+  type        = object({ metadata = string })
+  description = "Firewall log configuration for Toolkit firewall rules (var.enable_iap_ssh_ingress and others)"
+  default     = null
```
nick-stroud, line 234: What does null value represent? I assume it would be something along the lines of "platform default". 

If my assumption is correct, I would add the statement "null will result in using the platform default"

What is the platform default. We probably don't want to document as it may change. Would it be worth setting default explicitly to EXCLUDE since we know it can cause performance issues, it is a non-destructive change, and we are all about proposing sensible defaults for hpc?

nick-stroud (approved):

modules/network/vpc/variables.tf
tpdownes: The configuration is a bit weird and, perhaps, best to do something like OS Login settings.

- `null` = Logging OFF entirely
- `INCLUDE_ALL_METADATA` = Logging ON including "all" metadata
- `EXCLUDE_ALL_METADATA` = Logging ON excluding "all" metadata

tpdownes: The performance difference has been observed when I change from `INCLUDE_ALL_METADATA` to `null`. I have not tested `EXCLUDE_ALL_METADATA`.

tpdownes: Offline changes discussed, pick a "ternary logic" string with these allowed values:

- `DISABLE_LOGGING`
- `INCLUDE_ALL_METADATA`
- `EXCLUDE_ALL_METADATA`

nick-stroud (approved):

PR #2068, opened 2023-12-20
Slurm6. Automagicaly set `nodeset.name` from module id.

mr0re1:
```sh

  on main.tf line 41, in module "debug_nodeset":
  41:   name                   = "debug_nodeset"
    ├────────────────
    │ var.name is "debug_nodeset"

Nodeset name (var.name) must begin with a letter, be fully alphanumeric and
be 6 characters or less.
Got "debug_nodeset" (set to module id by default).
```

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
@@ -175,7 +175,7 @@ No modules.
 | <a name="input_machine_type"></a> [machine\_type](#input\_machine\_type) | Compute Platform machine type to use for this partition compute nodes. | `string` | `"c2-standard-60"` | no |
 | <a name="input_metadata"></a> [metadata](#input\_metadata) | Metadata, provided as a map. | `map(string)` | `{}` | no |
 | <a name="input_min_cpu_platform"></a> [min\_cpu\_platform](#input\_min\_cpu\_platform) | The name of the minimum CPU platform that you want the instance to use. | `string` | `null` | no |
-| <a name="input_name"></a> [name](#input\_name) | Name of the nodeset. | `string` | `null` | no |
+| <a name="input_name"></a> [name](#input\_name) | Name of the nodeset. | `string` | n/a | yes |
 | <a name="input_node_conf"></a> [node\_conf](#input\_node\_conf) | Map of Slurm node line configuration. | `map(any)` | `{}` | no |
 | <a name="input_node_count_dynamic_max"></a> [node\_count\_dynamic\_max](#input\_node\_count\_dynamic\_max) | Maximum number of dynamic nodes allowed in this partition. | `number` | `1` | no |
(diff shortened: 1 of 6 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/metadata.yaml
```diff
@@ -16,3 +16,5 @@
 spec:
   requirements:
     services: []
+ghpc:
+  inject_module_id: toolkit_module_id
```
nick-stroud, line 20: It seems like this should be set to `name`. Why do the metadata mapping just to create a new variable anyway? If it is because of restrictions in naming then we are going to see that on pretty much every module to some lesser extent. I think we just have to deal with it either through an error or through implicit manipulation of the string.

nick-stroud (changes requested):

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
```diff
@@ -175,7 +175,7 @@ No modules.
 | <a name="input_machine_type"></a> [machine\_type](#input\_machine\_type) | Compute Platform machine type to use for this partition compute nodes. | `string` | `"c2-standard-60"` | no |
 | <a name="input_metadata"></a> [metadata](#input\_metadata) | Metadata, provided as a map. | `map(string)` | `{}` | no |
 | <a name="input_min_cpu_platform"></a> [min\_cpu\_platform](#input\_min\_cpu\_platform) | The name of the minimum CPU platform that you want the instance to use. | `string` | `null` | no |
-| <a name="input_name"></a> [name](#input\_name) | Name of the nodeset. | `string` | `"ghpc"` | no |
+| <a name="input_name"></a> [name](#input\_name) | Name of the nodeset. | `string` | n/a | yes |
```
cboneti, line 178: should you not say this is automatically populated by the module id if not set?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -15,11 +15,13 @@
...
   description = "Name of the nodeset."
   type        = string
-  default     = "ghpc"
 
   validation {
     condition     = can(regex("^[a-z](?:[a-z0-9]{0,5})$", var.name))
-    error_message = "Nodeset name (var.name) must begin with a letter, be fully alphanumeric and be 6 characters or less. Regexp: '^[a-z](?:[a-z0-9]{0,5})$'."
+    error_message = <<EOT
+Nodeset name (var.name) must begin with a letter, be fully alphanumeric and be 6 characters or less.
+Got "${var.name}" (set to module id by default).
```
cboneti, line 23: here I am wondering if you should not do similarly to what we do in slurm cluster name, just pick a subset of the n first letters, remove spaces, dashes, etc....

PR #2074, opened 2023-12-27
Skip upgrade of wheel/setuptools if already installed

tpdownes:
The existing script attempts installation wheel and setuptools even if the current versions are already installed. The standard output appears harmless:

```
Requirement already satisfied: setuptools==59.6.0 in /usr/local/ghpc-venv/lib/python3.8/site-packages (59.6.0)
Requirement already satisfied: wheel==0.37.1 in /usr/local/ghpc-venv/lib/python3.8/site-packages (0.37.1)
```

But, this command actually generates traffic to the public internet. This causes failure when the script is run once in an image building environment (with public internet access) and then re-run in an envirionment without access to the internet. These changes skip package installation entirely when it's not necessary.

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -175,8 +175,8 @@ main() {
 		fi
 	fi
 
-	# Upgrade pip if necessary
-	# Only run if OS is not Debian 12 - Debian 12 does not allow for system level pip installs
+	# Upgrade system-wide pip
+	# Do not run on Debian 12 - system pip package modification is forbidden
 	if [ ! -f /etc/debian_version ] || [ "$(lsb_release -a 2>/dev/null | sed -n 's/Release:\s\+\([0-9]\+\).\?.*/\1/p')" -ne "12" ]; then
 		pip_version=$(${python_path} -m pip --version | sed -nr 's/^pip ([0-9]+\.[0-9]+).*$/\1/p')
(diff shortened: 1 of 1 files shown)


modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -196,7 +196,17 @@ main() {
 		${venv_python_path} -m pip install --upgrade pip
 	fi
 
-	${venv_python_path} -m pip install -U wheel==${REQ_PIP_WHEEL_VERSION} setuptools==${REQ_PIP_SETUPTOOLS_VERSION}
+	# upgrade wheel if necessary
+	wheel_pkg=$(${venv_python_path} -m pip list --format=freeze | grep "^wheel")
+	if [ "$wheel_pkg" != "wheel==${REQ_PIP_WHEEL_VERSION}" ]; then
```
cdunbar13, line 201: Will this trigger the issue it's trying to solve if the current version is greater than the required version?

tpdownes: Yes, it will force a downgrade. But I think that is intended behavior. Our goal (for better/worse) is to have a well defined set of versions in the ghpc-venv across all operating systems. So if a user upgrades the version of setuptools/wheel inside the ghpc-venv, they should expect it to revert upon the next execution of startup-script.

cdunbar13 (approved):

PR #2076, opened 2023-12-27
Add support for string interpolation

mr0re1:
Example:
```yaml
  - id: bahcha
    settings:
      runners:
      - type: data
        destination: /tmp/$(vars.path)/arbuz
        content: |
          #!/bin/bash
          echo "Hello $(vars.project_id) from $(vars.region)"
```

### Refactor escaping logic:
* Don't persist escape backslashes in `cty.Value` representation;
* Don't "escape-back" during HCL-write;
* Keep all escape logic & info to the YAML-unmarshal/marshall code.

### Treat every non-HCL-literal YAML string as an expression

* Split string on plain-string and BP-expression by tokenizing it (see `tokenizeBpString`)
* If it is one plain-string piece, return `cty.StringVal`;
* If it is one expression piece, return expression without string interpolation;
* return HCL string interpolation expression combining all parts.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1260,6 +1260,18 @@ Whenever possible, blueprint variables are preferred over literal variables.
 `ghpc` will perform basic validation making sure all blueprint variables are
 defined before creating a deployment, making debugging quicker and easier.
 
+### String Interpolation
+
+The `$(...)` expressions can be used within strings, see:
+
+```yaml
+settings:
+  title: Magnificent $(vars.name)
+  script: |
+    #!/bin/bash
+    echo "Hello $(vars.project_id) from $(vars.region)"
+```
+
 ### Escape Variables
 
 Under circumstances where the variable notation conflicts with the content of a setting or string, for instance when defining a startup-script runner that uses a subshell like in the example below, a non-quoted backslash (`\`) can be used as an escape character. It preserves the literal value of the next character that follows:
--- a/examples/README.md
+++ b/examples/README.md
@@ -1237,8 +1237,6 @@ The variable is referred to by the source, either vars for deploment variables
 or the module ID for module variables, followed by the name of the value being
 referenced. The entire variable is then wrapped in “$()”.
 
-Currently, string interpolation with variables is not supported.
-
 ### Literal Variables
 
 Literal variables should only be used by those familiar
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -449,17 +449,15 @@ func validateModuleUseReferences(p modulePath, mod Module, bp Blueprint) error {
...
 func checkBackend(b TerraformBackend) error {
-	const errMsg = "can not use variables in terraform_backend block, got '%s=%s'"
-	// TerraformBackend.Type is typed as string, "simple" variables and HCL literals stay "as is".
-	if hasVariable(b.Type) {
-		return fmt.Errorf(errMsg, "type", b.Type)
-	}
-	if _, is := IsYamlExpressionLiteral(cty.StringVal(b.Type)); is {
-		return fmt.Errorf(errMsg, "type", b.Type)
+	err := errors.New("can not use expressions in terraform_backend block")
+	val, perr := parseYamlString(b.Type)
```
harshthakkar01, line 453: Shoudn't we check here if perr is nil or not before calling IsExpressionValue ?

mr0re1: `IsExpressionValue` can safely take `cty.NilVal`, so there is no reason for separate check.

pkg/config/yaml.go
```diff
@@ -255,24 +256,33 @@ func (y *YamlValue) unmarshalScalar(n *yaml.Node) error {
 	}
 	y.Wrap(v)
 
-	if l, is := IsYamlExpressionLiteral(y.Unwrap()); is { // HCL literal
-		var e Expression
-		if e, err = ParseExpression(l); err != nil {
-			// TODO: point to exact location within expression, see Diagnostic.Subject
-			return nodeToPosErr(n, err)
+	if y.Unwrap().Type() == cty.String {
+		if y.v, err = parseYamlString(y.v.AsString()); err != nil {
```
nick-stroud, line 260: nit: My understanding is that you can access y.v directly since you know it is a string. Would it be more correct to use y.Unwrap().AsString()

examples/README.md
nick-stroud, line 1: Please add documentation of this feature and an example snippet. I would suggest a new section under `Variables` titled `String Interpolation`.

nick-stroud (changes requested):

nick-stroud (approved):

PR #2079, opened 2023-12-28
Support use of http/https proxy for pip/apt/yum package managers

tpdownes:
There is a customer request for package managers to function with http proxies. This PR configures pip and yum or apt (depending on OS) to use a configured proxy for both http and https traffic. It is outside of the scope of the Toolkit to provision the proxy. If no proxy is specified, then no proxy configuration is made. The configuration must occur before the installation of Ansible which depends upon pip/yum/apt.

Alternative of using HTTP_PROXY environment variables in Ansible installation was considered but rejected because it would also impact gcloud/gsutil and would be a temporary configuration. This configures all subsequent attempts to use pip/yum/apt.

I have tested with this blueprint (vm00 does not use proxy at all, and the other are 4 different OSes) and verified that each package manager attempts to use the configured proxy (and fails):

```yaml
---
blueprint_name: test-proxy

vars:
  deployment_name: proxy
  region: us-east4
  zone: us-east4-c

deployment_groups:
- group: first
  modules:
  - id: network1
    source: modules/network/vpc
  - id: script
    settings:
      http_proxy: http://my.proxy:80
      runners:
      - type: shell
        destination: test_proxy.sh
        content: |
          #!/bin/bash
          touch /etc/i_was_here
  - id: script00
    settings:
      runners:
      - type: shell
        destination: test_proxy2.sh
        content: |
          #!/bin/bash
          touch /etc/i_was_here
  - id: vm00
    use:
    - network1
    - script00
    settings:
      name_prefix: vm00
      machine_type: n1-standard-2
  - id: vm0
    use:
    - network1
    - script
    settings:
      name_prefix: vm0
      machine_type: n1-standard-2
  - id: vm1
    use:
    - network1
    - script
    settings:
      name_prefix: vm1
      machine_type: n1-standard-2
      instance_image:
        project: debian-cloud
        family: debian-12
  - id: vm2
```

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -236,6 +236,7 @@ No modules.
 | <a name="input_debug_file"></a> [debug\_file](#input\_debug\_file) | Path to an optional local to be written with 'startup\_script'. | `string` | `null` | no |
 | <a name="input_deployment_name"></a> [deployment\_name](#input\_deployment\_name) | Name of the HPC deployment, used to name GCS bucket for startup scripts. | `string` | n/a | yes |
 | <a name="input_gcs_bucket_path"></a> [gcs\_bucket\_path](#input\_gcs\_bucket\_path) | The GCS path for storage bucket and the object, starting with `gs://`. | `string` | `null` | no |
+| <a name="input_http_proxy"></a> [http\_proxy](#input\_http\_proxy) | Web (http and https) proxy configuration for pip, apt, and yum/dnf | `string` | `""` | no |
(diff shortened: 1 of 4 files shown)


harshthakkar01:
Just for my understanding, why would/ should attempt to use configured proxy fail by package manager in above case ?

harshthakkar01 (approved):

tpdownes:
> Just for my understanding, why would/ should attempt to use configured proxy fail by package manager in above case ?

Well, if you configure a proxy but the proxy doesn't exist then HTTP/HTTPS traffic generated by pip/apt/yum will simply fail. So, failure indicates "WAI" because it means pip/apt/yum are using the non-existent proxy.

PR #2083, opened 2024-01-03
VPC. Replace `s/_/-/` in `deployment_name` to avoid deploy-time error

mr0re1:


modules/network/vpc/main.tf
```diff
@@ -15,8 +15,9 @@
 */
 
 locals {
-  network_name    = var.network_name == null ? "${var.deployment_name}-net" : var.network_name
-  subnetwork_name = var.subnetwork_name == null ? "${var.deployment_name}-primary-subnet" : var.subnetwork_name
+  autoname        = replace(var.deployment_name, "_", "-")
```
rohitramu, line 18: Shouldn't deployment name never contain underscores?  If so, why don't we just throw an error instead during validation?

rohitramu (changes requested):

modules/network/vpc/main.tf
mr0re1: > Shouldn't deployment name never contain underscores?

No, it just needs to be a valid GCP label:
https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements

rohitramu (approved):

PR #2085, opened 2024-01-03
Extending GKE Scheduler module

ek-nag:
We have developed a custom Kubernetes Container Storage Interface (CSI) driver that utilizes Kernel modules. Currently, our custom packages, designed to work with these Kernel modules, are built specifically for the "UBUNTU_CONTAINERD" OS image. To improve flexibility and compatibility with Google Kubernetes Engine (GKE) system nodes, we propose a change to allow users the option in HPC Toolkit blueprint to select "UBUNTU_CONTAINERD" OS images for their system nodes pool, instead of the hardcoded default "UBUNTU_CONTAINERD" image. 

We are introducing an option to disable secure boot if required, catering to users with specific security or compatibility requirements, for example cases when using custom Kernel modules like us.

To further improve system organization and management, we are also adding the capability to specify Kubernetes labels for the system nodes. So, the features such as custom DeamonSets can be easily targeted towards specific type of nodes.


community/modules/scheduler/gke-cluster/variables.tf
```diff
@@ -270,3 +270,21 @@ variable "service_account" {
...
+  EOT
+  type        = map(string)
+  default     = null
+}
+variable "image_type" {
+  description = "The default image type used by NAP once a new node pool is being created. Use either COS_CONTAINERD or UBUNTU_CONTAINERD."
+  type        = string
+  default     = "COS_CONTAINERD"
+}
+variable "enable_secure_boot" {
```
nick-stroud, line 286: This module creates both the GKE cluster and the system node pool. All three of the new variables are specific to the system node pool. The other variables in this module that are specific to system node pool are prefixed with `system_node_pool_`. 

I would suggest using:
- `system_node_pool_kubernetes_labels`
- `system_node_pool_image_type`
- `system_node_pool_enable_secure_boot`

Also, move these variables to ~L173 so all the system node pool variables are listed together.

nick-stroud (changes requested):

nick-stroud:
Something went wrong in the rebase and now it is saying that 267 files have been edited.

PR #2088, opened 2024-01-03
Use `cty.Type` instead of `string` to represent type of vars.

mr0re1:
NOTE: after this change instead of `list` the `list(any)` will be written into `variables.tf`.

### DIFF
--- a/pkg/modulereader/hcl_utils.go
+++ b/pkg/modulereader/hcl_utils.go
@@ -61,9 +61,13 @@ func getHCLInfo(source string) (ModuleInfo, error) {
 	var vars []VarInfo
 	var outs []OutputInfo
 	for _, v := range module.Variables {
+		ty, err := GetCtyType(v.Type)
+		if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/modulereader/hcl_utils.go
```diff
@@ -82,15 +86,27 @@ func getHCLInfo(source string) (ModuleInfo, error) {
 	return ret, nil
 }
 
-// Transforms HCL type string into cty.Type
-func getCtyType(hclType string) (cty.Type, error) {
+// Transforms Terraform type string into cty.Type
```
nick-stroud, line 89: You s/HCL/Terraform/. How does this effect packer?

nick-stroud (approved):

pkg/modulereader/hcl_utils.go
mr0re1: Terraform type system is supperset of HCL one, by additional shorthand types `list` and `map`.
Packer is definetly a subset of TF one, no changes in behaviour.

PR #2091, opened 2024-01-04
Add test that check that modules don't output forbidden names

mr0re1:


pkg/inspect/modules_test.go
```diff
@@ -195,3 +195,23 @@ func TestMetadataInjectModuleId(t *testing.T) {
...
+		"labels":          nowhere,
+		"region":          nowhere,
+		"zone":            nowhere,
+		"deployment_name": nowhere,
+	}
+	for _, mod := range query(all()) {
+		t.Run(mod.Source, func(t *testing.T) {
+			for _, out := range mod.Outputs {
+				if where, ok := allowed[out.Name]; ok && !slices.Contains(where, mod.Source) {
+					t.Errorf("forbidden name for output %q", out.Name)
```
cboneti, line 212: what is a forbidden name? I would clarify what that means.

rohitramu: +1

Maybe something like "invalid output name '%s' - output names must be ..."?

rohitramu (changes requested):

pkg/inspect/modules_test.go
mr0re1: Added a comment
```go
// Global blueprint variables we don't want to get overwritten.
```

rohitramu: Oh I see, this is to prevent conflicts with blueprints.  Maybe the error message should say that explicitly, e.g. "the name `%s` is reserved for use by HPC Toolkit's built-in global variables".  It should help users to understand what caused the error.

rohitramu (changes requested):

pkg/inspect/modules_test.go
mr0re1: Users will never see this, it's unit test, for us and contributors.

cboneti: The assumption that users will never see this never works. :) Please fix the error message. :)

mr0re1: > The assumption that users will never see this never works. :) 

I don't see :
1) what does it have to do with users?
2) developers will go to the code and see what is happening, so what is wrong with error message?

rohitramu (approved):

PR #2092, opened 2024-01-04
Add `has_to_be_used` behaviour to some of modules

mr0re1:
```sh
Error: module "debug_nodeset" was not used - you need to add it to the `use`-block of downstream modules
40:   - id: debug_nodeset
        ^
```

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -56,6 +56,10 @@ func (dc *DeploymentConfig) expand() error {
 		return err
 	}
 
+	if err := validateModulesAreUsed(dc.Config); err != nil {
+		return err
+	}
+
 	dc.Config.populateOutputs()
 	return nil
 }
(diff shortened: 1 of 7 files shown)


pkg/config/expand.go
```diff
@@ -93,6 +97,35 @@ func validateModuleInputs(mp modulePath, m Module, bp Blueprint) error {
...
+	for ig, g := range bp.DeploymentGroups {
+		for im, m := range g.Modules {
+			if !m.InfoOrDie().Metadata.Ghpc.HasToBeUsed {
+				continue
+			}
+
+			if !used[m.ID] {
+				errs.At(Root.Groups.At(ig).Modules.At(im).ID,
+					HintError{
+						"you need to add it to the `use`-block of downstream modules",
```
cboneti, line 120: what if I assign an output of this module to an input of the downstream module? Wouldn't that work too?

mr0re1: Yes, that would work, module would be considered "used".  
 I intentionally keep the hint short to steer users toward `use`.

pkg/modulereader/metadata.go
nick-stroud, line 1: Add documentation to docs/module-guidelines.md

nick-stroud (changes requested):

nick-stroud:
My feeling is that this functionality is covering up an anti-pattern with the Slurm module architecture. The problem is that the Slurm modules using this feature (login/partitions) can only functionally exist in the context of a controller but as the modules do not create any infrastructure (data only) the terraform modules do not embody this constraint. 

To me the natural dependency chain should be that a partition uses a controller, a login node uses a controller. With this revised ordering you would naturally get this kind of error as a missing required variable from the login/partition module.

Given the current architecture, I approve of this quality of life improvement, but want to keep the door open to inverting the dependency chain in future versions of Slurm.

mr0re1:
>  this functionality is covering up an anti-pattern with the Slurm module architecture

Its usefulness goes beyond Slurm modules, though I agree that there is an anti-pattern indeed.
Other modules we may want to consider to use it with:
* `pre-existing-vpc`
* `startup-script`
* `community/modules/scripts/*`

nick-stroud (approved):

PR #2095, opened 2024-01-05
Unify usage and rendering of `HintError`

mr0re1:
Screenshot 2024-01-05 at 1 33 27 PM

Screenshot 2024-01-05 at 1 37 20 PM

### DIFF
--- a/cmd/render.go
+++ b/cmd/render.go
@@ -1,0 +1,87 @@
+// Copyright 2024 "Google LLC"
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


cmd/render.go
```diff
@@ -0,0 +1,94 @@
...
+
+	pref := fmt.Sprintf("%d: ", pos.Line)
+	arrow := " "
+	if pos.Column > 0 {
+		spaces := strings.Repeat(" ", len(pref)+pos.Column-1)
+		arrow = spaces + "^"
+	}
+
+	return fmt.Sprintf(`%s
+%s%s
```
cdunbar13, line 92: Same thing here, isn't it easier to use `\n` instead of actual newlines?

cdunbar13 (approved):

PR #2096, opened 2024-01-05
Simplify rendering of errors with Position instead of Path

mr0re1:
* Remove `internalPath`;
* Add `PosError` wrapper, render it specifically.

### DIFF
--- a/pkg/config/yaml.go
+++ b/pkg/config/yaml.go
@@ -68,15 +68,7 @@ func importBlueprint(f string) (Blueprint, YamlCtx, error) {
 
 	var bp Blueprint
 	if err = decoder.Decode(&bp); err != nil {
-		errs := Errors{}
-		for i, yep := range parseYamlV3Error(err) {
-			path := internalPath.Dot(fmt.Sprintf("bp_schema_error_%d", i))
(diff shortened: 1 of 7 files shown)


pkg/config/yaml.go
```diff
@@ -350,40 +338,37 @@ func (d Dict) MarshalYAML() (interface{}, error) {
...
+		errs.Add(parseYamlV3ErrorString(err.Error()))
 	}
-	return res
+	return errs
 }
 
 // parseYamlV3Error attempts to extract position and nice error message from yaml.v3 error message.
 // yaml.v3 errors are unstructured, use string parsing to extract information.
-// If no position can be extracted, returns (Pos{}, error.Error()).
+// If no position can be extracted, returns (Pos{}, error.Error()). !!!
```
cdunbar13, line 364: Why `!!!`?

cdunbar13 (approved):
Just one quick question about a comment.  LGTM

pkg/config/yaml.go
mr0re1: I use `!!!` as `"fix this before sending for review"` )

PR #2097, opened 2024-01-06
Slurm6. Move `examples/hpc-slurm` to V6

mr0re1:
* Move `examples/hpc-slurm` to V6;
* Updated `examples/README`;
* Remove `slurm-v5-hpc-centos7` test.


examples/hpc-slurm.yaml
```diff
@@ -28,53 +28,54 @@ vars:
...
-    source: community/modules/compute/schedmd-slurm-gcp-v5-partition
+    source: community/modules/compute/schedmd-slurm-gcp-v6-partition
     use:
-    - network1
     - homefs
-    - debug_node_group
+    - debug_nodeset
     settings:
       partition_name: debug
       exclusive: false # allows nodes to stay up after jobs are done
```
cboneti, line 58: I think it is important to keep placement disabled. In fact, the usual behavior is that if you enable placement, then exclusive is forced to true. We did have validation that warned the user about that.  So please make sure we have it again in V6.

cboneti (commented):
PTAL

examples/hpc-slurm.yaml
mr0re1: > I think it is important to keep placement disabled. 

We do, the setting moved from partition to nodeset ,  see L49

> We did have validation that warned the user about that. So please make sure we have it again in V6.

We do:

```hcl
  precondition {
    condition     = (length(local.use_placement) == 0) || var.exclusive
    error_message = "If any nodeset `enable_placement`, `var.exclusive` must be set true"
  }
```

cboneti (approved):

cboneti (approved):

PR #2120, opened 2024-01-10
Startup stackdriver option

nick-stroud:
Under some HPC workloads the Cloud Ops Agent has been observed to cause reduced performance. The old Stackdriver agent show the same performance regression. The recommendation is to use the legacy agent for HPC applications.

The integration test ensures that the neither agent causes a failure during installation. Since we are recommending Stackdriver I have switched the test to actually confirming that Stackdriver is running. 

Testing: Manually deployed both stackdriver and ops agent across the following images and confirmed the service was running:
- hpc-rocky-linux-8
- hpc-centos-7
- ubuntu-2004-lts
- debian-11

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/install_monitoring_agent.sh
+++ b/modules/scripts/startup-script/files/install_monitoring_agent.sh
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+set -e -o pipefail
 
 LEGACY_MONITORING_PACKAGE='stackdriver-agent'
 LEGACY_MONITORING_SCRIPT_URL='https://dl.google.com/cloudagents/add-monitoring-agent-repo.sh'
@@ -21,7 +22,7 @@ LEGACY_LOGGING_SCRIPT_URL='https://dl.google.com/cloudagents/add-logging-agent-r
 OPSAGENT_PACKAGE='google-cloud-ops-agent'
 OPSAGENT_SCRIPT_URL='https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh'
 
-install_legacy="${1:-true}"
+ops_or_legacy="${1:-legacy}"
 
 fail() {
 	echo >&2 "[$(date +'%Y-%m-%dT%H:%M:%S%z')] $*"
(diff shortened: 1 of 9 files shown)


nick-stroud:
Reminder: Add snippet of script to documentation.

modules/scripts/startup-script/main.tf
```diff
@@ -20,11 +20,16 @@ locals {
...
-    source      = "${path.module}/files/install_cloud_ops_agent.sh"
-    destination = "install_cloud_ops_agent_automatic.sh"
-  }] : []
+  monitoring_agent_installer = (
+    var.install_cloud_ops_agent || var.install_stackdriver_agent ?
+    [{
+      type        = "shell"
+      source      = "${path.module}/files/install_monitoring_agent.sh"
+      destination = "install_monitoring_agent_automatic.sh"
+      args        = var.install_cloud_ops_agent ? "false" : "true" # install legacy (stackdriver)
```
tpdownes, line 29: I think Terraform would render `!var.install_cloud_ops_agent` into a string, FYI, but I agree with explicitness here. +1.

That said, I think the script and this section would be better if instead of "true" you used "legacy".

modules/scripts/startup-script/files/install_monitoring_agent.sh
```diff
@@ -79,7 +94,13 @@ handle_redhat() {
 	}
 
 	install_opsagent() {
-		curl -s https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh | bash -s -- --also-install
+		curl -s "${OPSAGENT_SCRIPT_URL}" | bash -s -- --also-install
```
tpdownes, line 97: A couple broad comments on this script:

1. I think it would benefit from `set -e -o pipefail`
2. I'm dubious generally of piping from `curl` to `bash` without intermediate error checking. I realize that is a lot of changes in this script but a failed curl (e.g. no internet access) will pipe something to bash.
3. is_legacy_installed() seems like it has an `||` when it probably ought to have an `&&` relationship. I understand the power comes from having both legacy agents installed. That said, I see how it would complicate the script to enforce that level of detail. Thoughts? Maybe add a comment?

tpdownes (commented):
This is a very welcome and timely addition, thank you. I think the changes here can be minor. I don't think we need to solve  `curl | bash` unless testing indicates there is a real problem. The stdout might be empty in curl failure modes.

Rest is pretty staightforward.

modules/scripts/startup-script/main.tf
```diff
@@ -167,6 +172,13 @@ resource "google_storage_bucket_object" "scripts" {
     create = "10m"
     update = "10m"
   }
+
+  lifecycle {
+    precondition {
+      condition     = !var.install_cloud_ops_agent || !var.install_stackdriver_agent
```
mr0re1, line 178: nit. `!(A & B)` has better readability than `!A || !B`

tpdownes: +1

modules/scripts/startup-script/files/install_monitoring_agent.sh
nick-stroud: 1: done
2: discussed offline. Good to leave as is since error does not generate stdout and pipefail will catch error
3: I have updated logic to cover this case.

modules/scripts/startup-script/main.tf
nick-stroud: Done.

nick-stroud:
If you have something more elegant for an xor than the following then I am open to suggestions:

```
[[ -z "$(is_legacy_monitoring_installed)" && -n $(is_legacy_logging_installed) ]] ||
		[[ -n "$(is_legacy_monitoring_installed)" && -z $(is_legacy_logging_installed) ]]
```

`!=` does not work with the `-z` test and does not work with raw strings because the strings are different. I played with `grep -q` also but didn't see a great way to use in the context of the functions.

I am happy with what I have here but am open to suggestions.

tpdownes (approved):

PR #2137, opened 2024-01-15
Extending GKE Scheduler module

ek-nag:
We have developed a custom Kubernetes Container Storage Interface (CSI) driver that utilizes Kernel modules. Currently, our custom packages, designed to work with these Kernel modules, are built specifically for the "UBUNTU_CONTAINERD" OS image. To improve flexibility and compatibility with Google Kubernetes Engine (GKE) system nodes, we propose a change to allow users the option in HPC Toolkit blueprint to select "UBUNTU_CONTAINERD" OS images for their system nodes pool, instead of the hardcoded default "UBUNTU_CONTAINERD" image.

We are introducing an option to disable secure boot if required, catering to users with specific security or compatibility requirements, for example cases when using custom Kernel modules like us.

To further improve system organization and management, we are also adding the capability to specify Kubernetes labels for the system nodes. So, the features such as custom DeamonSets can be easily targeted towards specific type of nodes.


nick-stroud (approved):

nick-stroud:
Breadcrumb: This PR is a follow on from abandoned #2085.

PR #2138, opened 2024-01-15
Update spack gromacs example and references to Slurm V6

harshthakkar01:
### Submission Checklist

### DIFF
--- a/docs/tutorials/gromacs/spack-gromacs.md
+++ b/docs/tutorials/gromacs/spack-gromacs.md
@@ -5,7 +5,7 @@ easy for customers to deploy HPC environments on Google Cloud.
 
 In this tutorial you will use the HPC Toolkit to:
 
-* Deploy a [Slurm](https://github.com/SchedMD/slurm-gcp#readme) HPC cluster on
+* Deploy a [Slurm](https://github.com/GoogleCloudPlatform/slurm-gcp#readme) HPC cluster on
   Google Cloud
 * Use [Spack](https://spack.io/) to install the Gromacs application and all of
   its dependencies
@@ -13,10 +13,10 @@ In this tutorial you will use the HPC Toolkit to:
   cluster
 * Tear down the cluster
 
-Estimated time to complete:  
-The tutorial takes 2 hr. to complete,  
-of which 1.5 hr is for installing software  
-(without cache).  
+Estimated time to complete:
+The tutorial takes 2 hr. to complete,
+of which 1.5 hr is for installing software
+(without cache).
 
 > **_NOTE:_** With a complete Spack cache, the tutorial takes 30 min.
 
@@ -75,7 +75,7 @@ which should be open in the Cloud Shell Editor (on the left).
 
 This file describes the cluster you will deploy. It defines:
 
-* the existing default network from your project
+* a vpc network
 * a monitoring dashboard with metrics on your cluster
 * a definition of a custom Spack installation
 * a startup script that
@@ -84,7 +84,6 @@ This file describes the cluster you will deploy. It defines:
   * sets up a Spack environment including downloading an example input deck
   * places a submission script on a shared drive
 * a Slurm cluster
-  * a Slurm login node
   * a Slurm controller
   * An auto-scaling Slurm partition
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/gromacs/spack-gromacs.md
```diff
@@ -106,24 +105,18 @@ contains the terraform needed to deploy your cluster.
...
 
-The `terraform apply` command will generate a _plan_ that describes the Google
-Cloud resources that will be deployed.
-
-You can review the plan and then start the deployment by typing
-**`yes [enter]`**.
-
-The deployment will take about 30 seconds. There should be regular status updates
-in the terminal.
+You can also use below command to generate a plan that describes the Google Cloud resources that will be deployed.
```
nick-stroud, line 114: I would skip this. If they want to view the plan they can display the plan from the deploy command. Or you could point them to instructions file for more detailed terraform commands, but I think you could skip all together.

docs/tutorials/gromacs/spack-gromacs.yaml
```diff
@@ -88,12 +88,12 @@ deployment_groups:
...
 
-        spack config --scope defaults add config:build_stage:/apps/spack/spack-stage
+        spack config --scope defaults add config:build_stage:/opt/apps/spack/spack-stage
         spack config --scope defaults add -f /tmp/projections-config.yaml
         spack config --scope site add -f /tmp/slurm-external-config.yaml
 
-        spack install gcc@9.3.0 target=x86_64
-        spack load gcc@9.3.0 target=x86_64
+        spack install gcc@9.3.0 %gcc@8.5.0 target=x86_64
+        spack load gcc@9.3.0 %gcc@8.5.0 target=x86_64
```
nick-stroud, line 96: where is the 8.5 coming from. Is it needed?

docs/tutorials/gromacs/spack-gromacs.yaml
```diff
@@ -107,22 +107,27 @@ deployment_groups:
     source: modules/scripts/startup-script
     settings:
       runners:
+      - type: shell
+        destination: remove_lustre_client.sh
+        content: |
+          #!/bin/bash
+          rm /etc/yum.repos.d/lustre-client.repo
```
nick-stroud, line 114: I know this is probably not the only place but we need to document why we are doing this.

I would recommend we add a section to `docs/slurm-troubleshooting.md` and explain the issue. Then leave a comment on each of these startup scripts that points to the documentation.

docs/tutorials/gromacs/spack-gromacs.yaml
```diff
@@ -107,22 +107,27 @@ deployment_groups:
...
         content: |
           #!/bin/bash
-          source /apps/spack/share/spack/setup-env.sh
+          source /opt/apps/spack/share/spack/setup-env.sh
           spack env activate gromacs
-          chmod -R a+rwX /apps/spack/var/spack/environments/gromacs
-          mkdir -p /apps/gromacs
-          chmod a+rwx /apps/gromacs
-          cd /apps/gromacs
+          chmod -R a+rwX /opt/apps/spack/var/spack/environments/gromacs
```
nick-stroud, line 122: This should no longer be needed.

docs/tutorials/gromacs/spack-gromacs.yaml
```diff
@@ -107,22 +107,27 @@ deployment_groups:
...
+          mkdir -p /opt/apps/gromacs
+          chmod a+rwx /opt/apps/gromacs
```
nick-stroud, line 124: This should no longer be needed.

nick-stroud (changes requested):

docs/tutorials/gromacs/spack-gromacs.md
harshthakkar01: Removed.

docs/tutorials/gromacs/spack-gromacs.yaml
harshthakkar01: Removed.

docs/tutorials/gromacs/spack-gromacs.yaml
harshthakkar01: Removed.

docs/tutorials/gromacs/spack-gromacs.yaml
harshthakkar01: Removed.

docs/tutorials/gromacs/spack-gromacs.yaml
harshthakkar01: I am not sure if we should add this. Looks like slurm-troubleshooting doc is if there is any issue with deploying the cluster, how can customer navigate it. 
This script removes lustre client because of known issue we found out in slurm image. Should we put all the details over there  ?

nick-stroud: My concern is that a customer will look at this line in the blueprint and have no idea why it is needed. Can we give the user some more information that this is a workaround to a known issue. I am not sure the best form for that to take but I would probably like some reference to the problem in the blueprint since that is where the confusion is likely to occur.

harshthakkar01: Added a comment in the beginning. After we move to 6.3.1 I will remove this additional script from all BPs.

nick-stroud (approved):

PR #2142, opened 2024-01-17
Zone finding for TPU nodesets

rohitramu:


.gitignore
```diff
@@ -1,5 +1,8 @@
 #### GENERAL DEVELOPMENT
 
+# Top-level hidden files/folders
+.*
```
mr0re1, line 4: Could you please elaborate what you're hiding ?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
```diff
@@ -34,9 +34,8 @@ locals {
     docker_image = var.docker_image
 
     enable_public_ip = !var.disable_public_ips
-    # TODO: rename to subnetwork_self_link, requires changes to the scripts
```
mr0re1, line 37: pls don't remove TODO

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
```diff
@@ -89,6 +89,22 @@ variable "zone" {
...
+  description = <<-EOD
+    Additional nodes in which to allow creation of partition nodes. Google Cloud
+    will find zone based on availability, quota and reservations.
+    EOD
+  type        = set(string)
+  default     = []
+
+  validation {
+    condition = alltrue([
+      for x in var.zones : length(regexall("^[a-z]+-[a-z]+[0-9]-[a-z]$", x)) > 0
```
mr0re1, line 102: incorrect regex. There are zones with multiple digits, e.g. `europe-west10-a`.
Let's not make any assumptions, if there is any problems with zone, they will surface during plan (I hope).

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
```diff
@@ -34,9 +34,8 @@ locals {
...
-    subnetwork      = var.subnetwork_self_link
-    service_account = var.service_account
-    zone            = var.zone
+    subnetwork       = var.subnetwork_self_link
+    service_account  = var.service_account
+    zones            = toset(concat([var.zone], tolist(var.zones)))
```
mr0re1, line 39: This is breaking change, there is no support for `tpu_nodeset.zones` in neither controller wrapper nor SlurmGCP module.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
rohitramu: Good catch.  Fixed here and in the original code I copied from.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
rohitramu: Bad merge.  Thanks for catching this.  Added it back.

.gitignore
rohitramu: Good point.  I've changed it to hide only hidden folders.  I use hidden folders for my own temporary test files (e.g. blueprints, deployment files, etc.).

mr0re1: You can create your own .gitignore using
```sh
git config --global core.excludesfile $HOME/.gitignore
```

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
mr0re1: IMHO, We better to not have any assumptions about what is correct zone name. The most I would recommend is just "consist of letters, digits, _ and -"

rohitramu: I don't think underscores are allowed in zone names, so I left that out.

.gitignore
rohitramu: Cool, didn't know that.  Thanks!

Reverted the `.gitignore` changes.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
rohitramu: I've created a PR to support this in the SlurmGCP module: https://github.com/GoogleCloudPlatform/slurm-gcp/pull/16

PR #2148, opened 2024-01-19
Validation added to Slurm v5 login_startup_scripts_timeout

cdunbar13:
The login startup scripts timeout value is not being respected in `setup.py`.  This is because Slurm v5 adds a suffix to the login node, which causes the timeout look up to default to 300s.  This is currently being fixed in slurm-gcp, and once a new image is released, this validation can be removed.

This meant to be more of a warning for users who try to use this variable.  This value would go to the correct location, it just isn't used correctly.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
@@ -128,6 +128,11 @@ variable "login_startup_scripts_timeout" {
     EOD
   type        = number
   default     = 300
+
+  validation {
+    condition     = var.login_startup_scripts_timeout == 300
+    error_message = "Changes to login_startup_scripts_timeout (default: 300s) are not respected, this will be fixed in a later release"
+  }
 }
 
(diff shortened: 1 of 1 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
```diff
@@ -128,6 +128,11 @@ variable "login_startup_scripts_timeout" {
     EOD
   type        = number
   default     = 300
+
+  validation {
+    condition     = var.login_startup_scripts_timeout == 300
+    error_message = "Changes to login_startup_scripts_timeout (default: 300s) are not respected, this will be fixed in a later release"
```
tpdownes, line 134: Please change to "this is a known issue that will be..."

tpdownes (approved):

community/modules/scheduler/schedmd-slurm-gcp-v5-controller/variables.tf
cdunbar13: Done

PR #2150, opened 2024-01-22
Add commands to verify monitoring agents are active

nick-stroud:
Minor docs update for convenience.

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -161,6 +161,16 @@ sudo bash add-logging-agent-repo.sh --also-install
 sudo service stackdriver-agent start
 ```
 
+You can test if one of the agents is running using the following commands:
+
+```bash
+# For Cloud Ops Agent
+sudo systemctl status google-cloud-ops-agent
+
+# For Stackdriver Agent
+sudo service stackdriver-agent status
+```
+
 ### Example
 
 ```yaml


modules/scripts/startup-script/README.md
```diff
@@ -161,6 +161,16 @@ sudo bash add-logging-agent-repo.sh --also-install
 sudo service stackdriver-agent start
 ```
 
+You can test if one of the agents is running using the following commands:
```
tpdownes, line 164: I think you should include the output of these commands that indicates whether the service is active. Does `systemctl is-active X.service` include any stdout or just an exit code?

In particular, and I'm going by memory here `google-cloud-ops-agent.service` actually invokes other, different child systemd services based upon its configuration. Therefore, this parent service may not have clear output. The docs have a `*` in the status command?

https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/troubleshoot-install-startup

Might you just be better off linking to the troubleshooting section of docs?

Why use `service` for stackdriver-agent instead of systemctl? Any old SysV tabs are just converted to systemd units.

tpdownes (changes requested):
I don't think a novice user will know how to interpret the output of `

modules/scripts/startup-script/README.md
nick-stroud: Thanks for the feedback. 

1. I have updated cloud ops to use `is-active` as this is cleaner
2. I have updated to use `*` as used in the docs
3. I have added example output
4. I added command for legacy logging
5. I have realized that logging service was not being started (#2157)
6. Added links to official docs

`Why use service for stackdriver-agent instead of systemctl` >
There is a bit of mystery here to me. Systemctl reports status `inactive` (`sudo systemctl status stackdriver-agent.service`). `service` reports status `OK`.  Official docs say to look for `OK` from `service` and I have confirmed monitoring agent is working.

`Might you just be better off linking to the troubleshooting section of docs?`>
I have added links. I found it useful to have specific reference to commands close at hand for troubleshooting.

tpdownes: The fact that the agents are started as children of the "uber"-service is likely to blame for both of these considerations. Thanks for making the changes.

tpdownes (approved):

PR #2155, opened 2024-01-22
Add example of building Slurm on top of Rocky 8

nick-stroud:
Builds the latest Slurm V6 image on top of HPC Rocky 8 image and deploys a simple cluster using the image. This example is meant to be a demonstrator to show how it is possible to build Slurm. 

Notes:
- This calls ansible directly instead of using Slurm Packer module to avoid complexity of importing external Packer module.
- This does custom ansible install instead of using ansible runner as python upgrade breaks selinux bindings when done from within a virtual environment. 

Tested: 
- Manually run blueprint and test how cluster (`srun hostname`).
- Added integration test that does deploy/destroy

### Submission Checklist

### DIFF
--- a/community/examples/hpc-build-slurm-image.yaml
+++ b/community/examples/hpc-build-slurm-image.yaml
@@ -38,6 +38,8 @@ deployment_groups:
   - id: slurm-build-script
     source: modules/scripts/startup-script
     settings:
+      # Do not create Ansible virtual env; Install system wide Ansible below.
+      install_ansible: false
       runners:
       - type: shell
         destination: prep-for-slurm-build.sh
@@ -49,9 +51,10 @@ deployment_groups:
           dnf install -y python38
           alternatives --set python3 /usr/bin/python3.8
           python3 -m pip install pip --upgrade
-          python3 -m pip install ansible
+          python3 -m pip install ansible==4.10.0
           python3 -m pip install selinux
           export PATH=/usr/local/bin:$PATH
+          ansible --version
           ansible-galaxy role install googlecloudplatform.google_cloud_ops_agents
       - type: data
         destination: /var/tmp/slurm_vars.json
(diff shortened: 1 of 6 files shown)


nick-stroud:
To Reviewer: I will add documentation and integration test once you have given a first pass approval of the methodology of the blueprint.

community/examples/hpc-build-slurm-image.yaml
```diff
@@ -0,0 +1,116 @@
...
+deployment_groups:
+- group: setup
+  modules:
+  - id: network
+    source: modules/network/vpc
+
+  - id: slurm-build-script
+    source: modules/scripts/startup-script
+    settings:
+      runners:
```
tpdownes, line 41: I would set `install_ansible: false` explicitly and add a comment

- to avoid breaking upon nay future changes in behavior of startup-script
- allow user to add an ansible runner without breaking script (I bet it would work if it came after your installation)
- generally be explicit about a choice that matters

community/examples/hpc-build-slurm-image.yaml
```diff
@@ -0,0 +1,116 @@
...
+        destination: prep-for-slurm-build.sh
+        content: |
+          #!/bin/bash
+          set -e -o pipefail
+          # Slurm build on Rocky8 will upgrade to python38 as part of build
+          # This is not compatible with ansible-local runner
+          dnf install -y python38
+          alternatives --set python3 /usr/bin/python3.8
+          python3 -m pip install pip --upgrade
+          python3 -m pip install ansible
```
tpdownes, line 52: Is version of Ansible critical? Here you are installing a very recent copy of Ansible (good) but the rest of our Ansible solutions are tested with 2.11.

If the user were to add Ansible runners for the actual cluster, it would run on the copy of Ansible you are installing here. It might be wise to align with rest of Toolkit even if that mans an older version.

What do you think?

tpdownes (changes requested):

nick-stroud:
I submitted test manually (build 4c971653-eee2-45f3-8d9c-1cb7c2634145) and it is passing.  
I will update triggers once it is submitted.

tpdownes (approved):

PR #2156, opened 2024-01-22
Clarify that zone-finding isn't available for TPUs

rohitramu:
Correcting documentation to say that the TPU module doesn't support zone-finding.

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
@@ -85,7 +85,7 @@ variable "preserve_tpu" {
 }
 
 variable "zone" {
-  description = "Zone in which to create compute VMs. Additional zones in the same region can be specified in var.zones."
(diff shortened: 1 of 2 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
```diff
@@ -85,7 +85,7 @@ variable "preserve_tpu" {
 }
 
 variable "zone" {
-  description = "Zone in which to create compute VMs. Additional zones in the same region can be specified in var.zones."
+  description = "Zone in which to create compute VMs. Zone-finding is not currently available."
```
nick-stroud, line 88: nit: I would instead say: "Zone in which to create compute VMs. TPU partitions can only specify a single zone."

Zone-finding is not really an official term.

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/variables.tf
rohitramu: Makes sense, thanks for the tip.

PR #2166, opened 2024-01-25
DEMO. TF-native implementation of destroy-protection

mr0re1:
```yaml
# bp.yaml
blueprint_name: tst
vars:
  deployment_name: prot
deployment_groups:
- group: primary
  modules:
  - id: vpc
    source: modules/network/vpc
    protected: true
```

```hcl
// deployment::prot/primary/main.tf
module "vpc" {
  source          = "./modules/embedded/modules/network/vpc"
  deployment_name = var.deployment_name
  project_id      = var.project_id
  region          = var.region
}
resource "null_resource" "protection_vpc" {
  depends_on = ["module.vpc"]
  triggers = {
    deployment_name = var.deployment_name
    project_id      = var.project_id
    region          = var.region
  }
  lifecycle {
    prevent_destroy = true
  }
}
```

```sh
$ ./ghpc destroy prot $AA
Testing if deployment group prot/primary requires destroying cloud infrastructure
Error: terraform plan for deployment group prot/primary failed
exit status 1

Error: Instance cannot be destroyed

  on main.tf line 23:
  23: resource "null_resource" "protection_vpc" {

Resource null_resource.protection_vpc has lifecycle.prevent_destroy set, but
the plan calls for this resource to be destroyed. To avoid this error and
continue with the plan, either disable lifecycle.prevent_destroy or reduce
the scope of the plan using the -target flag.
```

Also changes to inputs  (e.g. `vars.region`) fill fail re-deployment as well:

```sh
$ ./ghpc deploy prot
Initializing deployment group prot/primary
Testing if deployment group prot/primary requires adding or changing cloud infrastructure
Error: terraform plan for deployment group prot/primary failed
exit status 1

Error: Instance cannot be destroyed

  on main.tf line 23:
  23: resource "null_resource" "protection_vpc" {
...
```

## Disabling protection
* Modify `bp.yaml`, remove `protected: true`
* `./ghpc reate bp.yaml -w`
* ` ./ghpc destroy prot $AA` - successfully destroys VPC

### DIFF
--- a/pkg/modulewriter/tfwriter.go
+++ b/pkg/modulewriter/tfwriter.go
@@ -24,7 +24,9 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/ext/typeexpr"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/zclconf/go-cty/cty"
 	"golang.org/x/exp/maps"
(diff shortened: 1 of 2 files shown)


pkg/modulewriter/tfwriter.go
```diff
@@ -166,27 +166,49 @@ func writeMain(
...
+
+	if mod.Protected {
+		rn := fmt.Sprintf("protection_%s", mod.ID)
+		rb := b.AppendNewBlock("resource", []string{"null_resource", rn}).Body()
+
+		mr := cty.StringVal(fmt.Sprintf("module.%s", mod.ID))
+		// TODO: Warning: Quoted references are deprecated
+		rb.SetAttributeValue("depends_on", cty.TupleVal([]cty.Value{mr}))
+
+		// tb := rb.AppendNewBlock("triggers", []string{}).Body()
```
cboneti, line 199: If we were to merge this, I would rather remove the commented lines.

pkg/modulewriter/tfwriter.go
```diff
@@ -166,27 +166,49 @@ func writeMain(
...
+		// for _, setting := range orderKeys(mod.Settings.Items()) {
+		// 	value := mod.Settings.Get(setting)
+		// 	tb.SetAttributeRaw(setting, config.TokensForValue(value))
+		// }
+		rb.SetAttributeRaw("triggers", config.TokensForValue(mod.Settings.AsObject()))
```
cboneti, line 204: In your example, this renders as "deployment_name", "project_id" and "region". So if any of these change, it would trigger a destruction, is that correct?

cboneti (commented):
I really like this, it seems both simple and powerful. 

I wonder if we could catch the terraform error and guide the user to the setting in the YAM, making this more friendly.

Finally, I would appreciate to have another pair of eyes. Will ask Tom.

pkg/modulewriter/tfwriter.go
mr0re1: Yes, we lock all input values on protected module

PR #2187, opened 2024-02-01
Updating spack and ramble buckets to use 8 digits of hex

cdunbar13:
Description in title.

Can't use b64 because of naming restrictions (no uppercase)

This was tested with the hpc-slurm-ramble-gromacs example with no errors.

### DIFF
--- a/community/modules/scripts/ramble-setup/main.tf
+++ b/community/modules/scripts/ramble-setup/main.tf
@@ -72,7 +72,7 @@ locals {
     "destination" = "install_ramble.yml"
   }
 
-  bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 4)
+  bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 8)
   bucket_name = "ramble-scripts-${local.bucket_md5}"
   runners     = [local.install_ramble_deps_runner, local.install_ramble_runner, local.python_reqs_runner]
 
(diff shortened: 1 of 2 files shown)


community/modules/scripts/ramble-setup/main.tf
```diff
@@ -72,7 +72,7 @@ locals {
     "destination" = "install_ramble.yml"
   }
 
-  bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 4)
+  bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 6)
```
nick-stroud, line 75: nit: I thought consensus was to go longer. I would say double it (8). @mr0re1 wanted even longer.

nick-stroud (approved):

community/modules/scripts/ramble-setup/main.tf
nick-stroud, line 1: I would recommend running a spack test. Not strictly mandatory.

nick-stroud (approved):

community/modules/scripts/ramble-setup/main.tf
cdunbar13: Done

community/modules/scripts/ramble-setup/main.tf
cdunbar13: I tested this, but just updated the initial comment with how I tested.  It was tested with 6, and I'm assuming there's no issue with 8 given that it doesn't exceed the length of the md5 or the max length of the GCS bucket name.

PR #2208, opened 2024-02-06
Add check that expanded blueprint is valid

mr0re1:


tools/validate_configs/validate_configs.sh
```diff
@@ -82,6 +82,24 @@ run_test() {
...
+		}
+	# TODO: re-enable this check once the trailing space in expressions is fixed, e.g.:
+	# -          labels: ((var.labels ))
+	# +          labels: ((var.labels))
+	#
+	# diff -u "${expanded}" "expanded_again.yaml" ||
+	# 	{
+	# 		echo "*** ERROR: re-expanded blueprint does not match the expanded one at ${exampleFile}"
+	# 		exit 1
+	# 	}
```
nick-stroud, line 102: nit: My personal preference would be to put this TODO comment in a task and keep the code clean since it is likely to be forgotten here.

nick-stroud (approved):

PR #2211, opened 2024-02-07
Show hint message if unsupported function is used

mr0re1:
Screenshot 2024-02-06 at 4 51 35 PM

**Context:**
Currently we have a few contexts that got evaluated by ghpc:
* `vars`;
* `module.settings` in packer groups;
* `validators.input` Our evaluation context only supports 2 functions: `merge` and `flatten`.

The rest of expressions (`module.settings` in TF groups) are not evaluated by ghpc => can use any valid HCL-syntax.


tpdownes (approved):

tpdownes:
@mr0re1 a thing to consider for future: exp/maps is part of stdlib in go 1.21.

PR #2221, opened 2024-02-08
Slurm6. more validations around `static/exclusive/placement/reservation`

mr0re1:
* Prevent usage of `static_nodes` and `exclusive` together
* Validate `enable_placement -> exclusive`
* Validate `reservation -> !enable_placement`

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
@@ -24,4 +24,12 @@ output "nodeset" {
     ], "${substr(var.machine_type, 0, 3)}:${var.disk_type}")
     error_message = "A disk_type=${var.disk_type} cannot be used with machine_type=${var.machine_type}."
   }
+
+  precondition {
+    condition     = (var.reservation_name == null) || !var.enable_placement
+    error_message = <<-EOD
+      If reservation is specified, `var.enable_placement` must be `false`.
+      If the specified reservation has a placement policy then it will be used automatically.
+    EOD
+  }
 }
--- a/community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
@@ -18,18 +18,16 @@ output "partitions" {
   value = [local.partition]
 
   precondition {
-    condition     = (length(local.use_placement) == 0) || var.exclusive
-    error_message = "If any nodeset `enable_placement`, `var.exclusive` must be set true"
+    condition     = (length(local.non_static_ns_with_placement) == 0) || var.exclusive
+    error_message = "If any non-static nodesets has `enable_placement`, `var.exclusive` must be set true"
   }
 
   precondition {
-    condition     = (length(local.use_placement) == 0) || contains(["NO", "Exclusive"], lookup(var.partition_conf, "Oversubscribe", "NO"))
-    error_message = "If any nodeset `enable_placement`, var.partition_conf[\"Oversubscribe\"] should be either undefined, \"NO\", or \"Exclusive\"."
-  }
-
-  precondition {
-    condition     = (length(local.use_placement) == 0) || (lookup(var.partition_conf, "SuspendTime", null) == null)
-    error_message = "If any nodeset `enable_placement`, var.partition_conf[\"SuspendTime\"] should be undefined."
+    condition     = (length(local.use_static) == 0) || !var.exclusive
+    error_message = <<-EOD
+    Can't use static nodes within partition with `var.exclusive` set to `true`.
+    NOTE: Partition's `var.exclusive` is set to `true` by default. Set it to `false` explicitly to use static nodes.
+    EOD
   }
 }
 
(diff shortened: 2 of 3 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -24,4 +24,12 @@ output "nodeset" {
     ], "${substr(var.machine_type, 0, 3)}:${var.disk_type}")
     error_message = "A disk_type=${var.disk_type} cannot be used with machine_type=${var.machine_type}."
   }
+
+  precondition {
+    condition     = (var.reservation_name == null) || !var.enable_placement
+    error_message = <<-EOD
+      If reservation is specified, `var.enable_placement` must be `false`.
+      Whatever (if any) placement policy is specified in the reservation will be used.
```
nick-stroud, line 32: nit s/`Whatever (if any) placement policy is specified in the reservation will be used.`/`If the specified reservation has a placement policy then it will be used automatically.`/

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
nick-stroud, line 1: I thought from out discussion we were going to allow the following case (which was disallowed in v5):
- static nodes > 0
- enable_placement: true
- exclusive: false

This config makes sense if the nodeset only has a single placement policy for all of the nodes. I think the current validation will block this case. From my initial testing this seemed possible but also the fact that it was working may not have been intentional (there were associated errors in logs). While I think it is sensible to support this case, I don't want to enable it if it is buggy/ not code complete.

nick-stroud: BTW Check out this commit: https://github.com/GoogleCloudPlatform/slurm-gcp/commit/30d9d90f5e7e7d762913bc52f6ea48401c97bade

It looks like static + placement is intentionally supported.

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
```diff
@@ -18,18 +18,16 @@ output "partitions" {
   value = [local.partition]
 
   precondition {
-    condition     = (length(local.use_placement) == 0) || var.exclusive
-    error_message = "If any nodeset `enable_placement`, `var.exclusive` must be set true"
+    condition     = (length(local.non_static_ns_with_placement) == 0) || var.exclusive
+    error_message = "If any non-static nodesets has `enable_placement`, `var.exclusive` must be set true"
```
nick-stroud, line 22: nit: I would use the terminology "autoscaling only" in place of "non-static", but just personal preference.

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
```diff
@@ -18,18 +18,16 @@ output "partitions" {
...
 
   precondition {
-    condition     = (length(local.use_placement) == 0) || contains(["NO", "Exclusive"], lookup(var.partition_conf, "Oversubscribe", "NO"))
-    error_message = "If any nodeset `enable_placement`, var.partition_conf[\"Oversubscribe\"] should be either undefined, \"NO\", or \"Exclusive\"."
-  }
-
-  precondition {
-    condition     = (length(local.use_placement) == 0) || (lookup(var.partition_conf, "SuspendTime", null) == null)
-    error_message = "If any nodeset `enable_placement`, var.partition_conf[\"SuspendTime\"] should be undefined."
+    condition     = (length(local.use_static) == 0) || !var.exclusive
```
nick-stroud, line 26: @tpdownes, exclusive + static behavior is sufficiently weird that @mr0re1 and I decided to prevent it with validation. I want to make sure this doesn't go against any of the testing you have done.

If we were to allow this setting, behavior is that even though nodes are static, they are deleted after jobs finish and slurm sync brings them back online. It is like an autoscaling cluster but where nodes are always warm.

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
tpdownes: I also believe this combination is rejected in v5, too, correct? There are cases in v5 where we have set:

```yaml
      partition_conf:
        OverSubscribe: EXCLUSIVE
```

Which is the Slurm meaning of "exclusive". I'm not aware of situations where we have set static and exclusive, but I think that's because it's not functionality in v5 at all.

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
mr0re1: ack

community/modules/compute/schedmd-slurm-gcp-v6-partition/outputs.tf
nick-stroud: I just deployed v5 with static nodes + exclusive: true. I don't think we explicitly blocked it.

But I agree, the oversubscribe feature + static is probably what people care about. That should still work as intended.

mr0re1: ack

PR #2222, opened 2024-02-09
Fix endless error render loop

mr0re1:
```sh
$ make && ./ghpc create null.yaml
Error: EOF
```


tpdownes (approved):
I think we will want to improve upon "EOF" as the result of reading an empty file, but this improves upon the status quo of crashing.

PR #2224, opened 2024-02-09
Initial implementation of deployment file flag

tpdownes:
From a secondary "deployment" file, implement

- merging deployment variables
- overriding Terraform backend defaults

Mark the flag hidden from the output of help menus. Does not yet implement validation level override or line/position error reporting.

To use, begin by defining a "deployment file":

```yaml
vars:
  project_id: a-project-that-exists

terraform_backend_defaults:
  type: gcs
  configuration:
    bucket: a-bucket-that-exists
```

And run

```shell
ghpc create -d deployment.yaml blueprint.yaml
```

### Submission Checklist

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -41,6 +41,11 @@ func init() {
 	cobra.CheckErr(createCmd.Flags().MarkDeprecated("config",
 		"please see the command usage for more details."))
 
+	deploymentFileFlag := "deployment-file"
+	createCmd.Flags().StringVarP(&deploymentFile, deploymentFileFlag, "d", "",
+		"Toolkit Deployment File.")
+	createCmd.Flags().MarkHidden(deploymentFileFlag)
+	createCmd.MarkFlagFilename(deploymentFileFlag, "yaml", "yml")
 	createCmd.Flags().StringVarP(&outputDir, "out", "o", "",
 		"Sets the output directory where the HPC deployment directory will be created.")
 	createCmd.Flags().StringSliceVar(&cliVariables, "vars", nil, msgCLIVars)
@@ -61,6 +66,7 @@ func init() {
 
 var (
 	bpFilenameDeprecated string
+	deploymentFile       string
 	outputDir            string
 	cliVariables         []string
 
@@ -83,7 +89,7 @@ var (
 )
 
 func runCreateCmd(cmd *cobra.Command, args []string) {
-	dc := expandOrDie(args[0])
+	dc := expandOrDie(args[0], deploymentFile)
 	deplDir := filepath.Join(outputDir, dc.Config.DeploymentName())
 	checkErr(checkOverwriteAllowed(deplDir, dc.Config, overwriteDeployment, forceOverwrite))
 	checkErr(modulewriter.WriteDeployment(dc, deplDir))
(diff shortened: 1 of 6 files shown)


pkg/config/yaml.go
```diff
@@ -74,6 +74,26 @@ func importBlueprint(f string) (Blueprint, YamlCtx, error) {
 	return bp, yamlCtx, nil
 }
 
+func importDeploymentFile(f string) (DeploymentSettings, YamlCtx, error) {
```
nick-stroud, line 77: Can we some code sharing with `importBlueprint` above. Either by pulling out reading the yaml into a separate shared function, or by making the function generic and passing in the type that is being decoded.

cmd/create.go
```diff
@@ -102,18 +108,29 @@ func printAdvancedInstructionsMessage(deplDir string) {
...
+
+	var ds config.DeploymentSettings
+	var dCtx config.YamlCtx
+	if dPath != "" {
+		ds, dCtx, err = config.NewDeploymentSettings(dPath)
+		if err != nil {
+			logging.Fatal(renderError(err, dCtx))
+		}
+	}
+	if err := setCLIVariables(&ds, cliVariables); err != nil {
```
nick-stroud, line 125: How I am reading this is that CLI vars supersedes deployment file, deployment file supersedes blueprint. Can you confirm? If so then I am happy.

cmd/create_test.go
```diff
@@ -90,33 +90,101 @@ func (s *MySuite) TestSetBackendConfig(c *C) {
...
+	// test priority-based merging of deployment variables
+	mergeDeploymentSettings(&bp, ds)
+	c.Check(bp.Vars.Items(), DeepEquals, map[string]cty.Value{
+		"project_id":      cty.StringVal("ds_test_project_id"),
+		"deployment_name": cty.StringVal("ds_deployment_name"),
+		"example_var":     cty.StringVal("bp_example_value"),
+	})
+
+	// check merging zero-value backends
+	ds = config.DeploymentSettings{
```
nick-stroud, line 128: Do not reuse the namespace `ds` `bp` for subsequent tests. Aka each time you define an object use the `:=` and come up with new naming.

if the same object works for multiple tests then fine to reuse, just not redefine. 

Reason: test is fragil to cases being inserted, harder to trace back to what objects are actually being tested.

nick-stroud (changes requested):

cmd/create.go
tpdownes: Yes. The conceptual intent is that command line variables are now the highest precedence deployment variables.

The technical intent is to implement a well defined behavior that aligns with the status quo: command line options always take precedence. It might be worth a reminder that `--vars` can be invoked multiple times, with later invocations at higher priority. e.g.

```
ghpc create -d test.yaml --vars "project_id=yads" --vars "project_id=bla"
```

and

```
ghpc create --vars "project_id=yada" --vars "project_id=bla" -d test.yaml 
```

both result in `vars.project_id="bla"` irrespective of the contents of the deployment file.

mr0re1 (approved):

pkg/config/yaml.go
tpdownes: I gave this some thought when writing it. I do not see an obvious answer in Go. Passing in a pointer would not help as it would need two different types of pointers.

I think the main opportunity might be to introduce custom errors before decoding, e.g. FileNotFound and YAMLParsingError. I'm not sure there's an obvious net simplification.

@mr0re1 ?

cmd/create_test.go
tpdownes: Please review https://github.com/GoogleCloudPlatform/hpc-toolkit/pull/2224/commits/ee4a2d8ad29a12ff660d7762ea37c0f60d9db13a

pkg/config/yaml.go
nick-stroud, line 77: `importBlueprint` has tests in `config_test.go`. `importDeploymentFile` should also have a test.

nick-stroud: How about the following (passes tests):

```
func readYaml(f string) (*yaml.Decoder, YamlCtx, error) {
	data, err := os.ReadFile(f)
	if err != nil {
		return &yaml.Decoder{}, YamlCtx{}, fmt.Errorf("%s, filename=%s: %v", errMsgFileLoadError, f, err)
	}
	decoder := yaml.NewDecoder(bytes.NewReader(data))
	decoder.KnownFields(true)

	yamlCtx, err := NewYamlCtx(data)
	if err != nil { // YAML parsing error
		return &yaml.Decoder{}, yamlCtx, err
	}
	return decoder, yamlCtx, nil
}

func importBlueprint(f string) (Blueprint, YamlCtx, error) {
	decoder, yamlCtx, err := readYaml(f)
	if err != nil {
		return Blueprint{}, YamlCtx{}, err
	}

	var bp Blueprint
	if err = decoder.Decode(&bp); err != nil {
		return Blueprint{}, yamlCtx, parseYamlV3Error(err)
	}
	return bp, yamlCtx, nil
}

func importDeploymentFile(f string) (DeploymentSettings, YamlCtx, error) {
	decoder, yamlCtx, err := readYaml(f)
	if err != nil {
		return DeploymentSettings{}, YamlCtx{}, err
	}

	var depl DeploymentSettings
	if err = decoder.Decode(&depl); err != nil {
		return DeploymentSettings{}, yamlCtx, parseYamlV3Error(err)
	}
	return depl, yamlCtx, nil
}
```

nick-stroud (changes requested):

pkg/config/yaml.go
tpdownes: I added a test for the public function that covers various failure methods of the internal private function (empty YAML, YAML with unknown fields).  I added these tests for importBlueprint while I was at it. Overall coverage up 0.1%.

nick-stroud (approved):

PR #2235, opened 2024-02-13
Use `n2-2` instead of `c3-8` in lustre tests

mr0re1:


tpdownes (approved):

tpdownes:
Given the current state of the Slurm v5.10.2 image, the failure of PR-test-lustre-slurm is expected. This should (and appears to) resolve the problems in PR-test-lustre-vm.

PR #2240, opened 2024-02-14
Remove stockout check as result was misleading and led to ignored tests

nick-stroud:
The output from this check has caused test misclassifications and humans to ignore failing tests. The data provided is generally already included in Slurm logs. We can consider adding this check back with different wording if it is found to be needed but at this point it is doing more harm than good.

### Submission Checklist


tools/cloud-build/daily-tests/ansible_playbooks/tasks/check_stockout.yml
```diff
@@ -1,44 +0,0 @@
...
-
-- name: Log compute stockout error
-  ansible.builtin.debug:
-    msg: |
-      "Abbreviated listing of nodes that could not be created:"
-      "{{ item.stdout }}"
-  when: item.stdout != ""
-  with_items: "{{ stockout.results }}"
-  loop_control:
-    label: "{{ item.cmd }}"
```
cdunbar13, line 44: Just as a point of reference, I believe this is the line that is causing the issue.  It prints out the lines under `with_items`.  We could change the label to something that wouldn't trigger the stockout result on the failure inspection.

cdunbar13 (approved):

PR #2250, opened 2024-02-16
Split service account creation from htcondor-setup

tpdownes:
Some users do not have IAM permissions to create service accounts with project-wide roles/permissions. This commit splits service account creation of the htcondor-setup module into an independent module. It also refactors each affected module to use internal Toolkit modules.

### Submission Checklist


mr0re1 (approved):

community/modules/scheduler/htcondor-service-accounts/outputs.tf
```diff
@@ -0,0 +1,39 @@
...
+  value       = module.central_manager_service_account.service_account_email
+  depends_on = [
+    module.central_manager_service_account
+  ]
+}
+
+output "execute_point_service_account_email" {
+  description = "HTCondor Execute Point Service Account (e-mail format)"
+  value       = module.execute_point_service_account.service_account_email
+  depends_on = [
```
mr0re1, line 36: nit. Please add some comment about `depends_on`? 
Looks like this problem ( I presume "email is available before all resources in the module are created" ) should be solved in `community/modules/project/service-account` , instead of boiling it up to the consumer, WDYT?

tpdownes: I agree it looks like it already is solved in community/modules/project/service-account. I'll remove the dependency and add a comment.

PR #2263, opened 2024-02-21
Split `chrome-remote-desktop` test on two `default` and `ubuntu`

mr0re1:
* Use fixed image `debian-12-bookworm-v20240110`
* Split chrome-remote-desktop test on two `default - debian` and `ubuntu`


tools/cloud-build/daily-tests/blueprints/crd-ubuntu.yaml
```diff
@@ -0,0 +1,37 @@
...
+  modules:
+  - id: network
+    source: modules/network/vpc
+
+  - id: crd
+    source: community/modules/remote-desktop/chrome-remote-desktop
+    use: [network]
+    settings:
+      install_nvidia_driver: true  # Must be set to true to accept Nvidia ToS for driver install
+      instance_image:
```
nick-stroud, line 30: nit: you could achieve the same default/ubuntu behavior with a single blueprint that does not define instance_image and then have one test pass in instance_image on the command line.

I just wanted to point it out incase you had not already thought of that. I am kind of indifferent as the way you have it is not that much duplication and is more explicit. So I would be fine either way.

nick-stroud (approved):

tools/cloud-build/daily-tests/blueprints/crd-ubuntu.yaml
mr0re1: acknowledged, frankly I was thinking about ways to do it but didn't come up with this solution.
Though it's good solution, I don't like the idea of using variables not declared in the blueprint.

PR #2268, opened 2024-02-21
Refactor `cmd`

mr0re1:
* Do not duplicate shared flags logic:
  * `artifactsDir` -> to reusable `addArtifactsDirFlag`;
  *  `addCreateFlags` calls `addExpandFlags` instead of copy-pasting flag setters from `expand.go`.
* Name all global flag variables explicitly to reduce namespace pollution, e.g. `forceOverwrite -> createFlags.forceOverwrite`;
* Use `Run` over `RunE` to make use of `checkErr` and unify rendering (not default to `cobra` rendering);
* Don't use `PreRun`, merge with `Run` to avoid usage of global vars for communication.

**Motivation:** preparation to upcoming changes (`deploy` does `create`) that would make fixed problems worse (copy all flags of `create` to `deploy`)

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -33,65 +33,46 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const msgCLIVars = "Comma-separated list of name=value variables to override YAML configuration. Can be used multiple times."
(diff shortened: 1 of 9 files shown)


tpdownes (approved):
This code is in need of some organization and I'm glad it's getting done. Thanks!

PR #2271, opened 2024-02-22
Use `bufio.Reader` instead of`fmt.Scanln` for user prompts

mr0re1:
```
Please select an option [d,a,s,c]: ggg
got "ggg"
(D)isplay ...
Please select an option [d,a,s,c]:
got ""
(D)isplay ...
Please select an option [d,a,s,c]: e f
got "e f"
(D)isplay ...
Please select an option [d,a,s,c]: s
got "s"
user chose to stop execution of ghpc rather than make proposed changes to infrastructure
```

### DIFF
--- a/pkg/shell/common.go
+++ b/pkg/shell/common.go
@@ -17,6 +17,7 @@
 package shell
 
 import (
+	"bufio"
 	"fmt"
 	"hpc-toolkit/pkg/config"
 	"hpc-toolkit/pkg/logging"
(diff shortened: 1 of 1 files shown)


tpdownes (changes requested):
One can break the loop this way as well:

```
Please select an option [d,a,s,c]: e f
expected newline
```

I think we should avoid focusing on the text of the error and consider what we want the user experience to be. I think it makes more sense to loop infinitely until valid input is provided.

WDYT of this?

- replace `logging.Fatal` with a lower level message to user (probably `logging.Error`)
- add a `continue` 

Would it help to capture the number of strings read? e.g.

```
 n, err := fmt.Scanln(&userResponse)
 if err != nil && n != 0 { ... }
```

mr0re1:
Switched to using `bufio.Reader` to avoid complexity of "expected number of results".

tpdownes (approved):

PR #2272, opened 2024-02-23
Make `config.Dict` immutable

mr0re1:
Replace mutator `Set` by builder `With`, see `pkg/config/dict.go`.

**Motivation:**
* Immutable structs are generally less  error-prone;
* Needed for easy `config.Blueprint.Clone()`;
* Stepping stone toward immutable `config.Blueprint`.

### DIFF
--- a/pkg/config/dict.go
+++ b/pkg/config/dict.go
@@ -27,11 +27,7 @@ type Dict struct {
 
 // NewDict constructor
 func NewDict(m map[string]cty.Value) Dict {
-	d := Dict{}
-	for k, v := range m {
-		d.Set(k, v)
-	}
-	return d
+	return Dict{m: maps.Clone(m)}
 }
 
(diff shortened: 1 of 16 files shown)


pkg/config/dict.go
```diff
@@ -51,26 +47,19 @@ func (d *Dict) Has(k string) bool {
...
-// Set adds/overrides value by key.
-// Returns reference to Dict-self.
-func (d *Dict) Set(k string, v cty.Value) *Dict {
-	if d.m == nil {
-		d.m = map[string]cty.Value{k: v}
-	} else {
-		d.m[k] = v
-	}
-	return d
+func (d Dict) With(k string, v cty.Value) Dict {
```
harshthakkar01, line 50: nit: comment

harshthakkar01 (approved):

PR #2275, opened 2024-02-23
Slurm6. Fix bug that prevents specifying `partition_conf`

mr0re1:
Referenced variable had moved from partition_conf to nodeset from v5 to v6.


nick-stroud:
I updated the PR description.

PR #2283, opened 2024-02-25
Fix for wait-for-startup.yml ansible task

cdunbar13:
The spack-gromacs test has failed the last few times it's been run.  This is due to an issue with the HPC-Rocky-8 image and the new Slurm-GCP-6.4 image.  In both images the hostname in /var/log/messages does not match the VM hostname, which causes the wait-for-startup.yml playbook to wait until the timeout.  At this point the test fails.

Two fixes have been added:
1. The `{vm_image}` has been removed from the task and the tasks that called it.  We will just rely on the `startup-script exit status` portion of the line, and now its taken from `journalctl` so that we know it's from this boot instead of potentially looking at the leftover data in `/var/log/messages`.  This also required switching to using an until loop instead of wait_for.  One side effect is that the number of failed attempts is printed out, which can be a number of lines for longer running startup-scripts.
2. The timeout has been reduced to 3 hours.  It was originally 6 hours which is longer than the timeout of the integration test that was calling the task.  This meant the integration test was timing out so the error was not as clear.

This has been tested by running the spack-gromacs and monitoring integration tests, which use this task.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
@@ -22,7 +22,7 @@
   become: true
   ansible.builtin.wait_for:
     path: /var/log/messages
-    search_regex: '.*{{ vm_name }}.*startup-script exit status ([0-9]+)'
+    search_regex: '.+startup-script exit status ([0-9]+)'
     timeout: "{{ timeout_seconds }}"
   register: startup_status
 - name: Fail if startup script exited with a non-zero return code
--- a/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
@@ -19,12 +19,13 @@
 
 - name: Wait for startup script to complete
   become: true
-  ansible.builtin.wait_for:
-    path: /var/log/messages
-    search_regex: '.+startup-script exit status ([0-9]+)'
-    timeout: "{{ timeout_seconds }}"
-  register: startup_status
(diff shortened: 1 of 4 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
```diff
@@ -15,17 +15,17 @@
...
   become: true
-  ansible.builtin.wait_for:
-    path: /var/log/messages
-    search_regex: '.*{{ vm_name }}.*startup-script exit status ([0-9]+)'
-    timeout: "{{ timeout_seconds }}"
-  register: startup_status
+  ansible.builtin.shell: journalctl -u google-startup-scripts.service -b | grep -oP "startup-script exit status \K\d+"
+  register: result
+  until: result.rc == 0
+  retries: "{{ (timeout_seconds/5)|int }}"
```
nick-stroud, line 25: 3 hr of this will generate over 2000 lines of logs.

nick-stroud: I don't love it but after looking for 30 minutes I don't see an alternative. I am going to approve and merge so we can see the test pass tonight. We can iterate if we don't like the logging.

nick-stroud (approved):

tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml
cdunbar13: There's a way to remove the log on a per-step basis, I was just worried that the entire step wouldn't be logged.  We could just increase the time step to a minute or more and that will cut the log lines significantly.  The extra minute of wait time shouldn't be too much of a hassle on other faster tests.

PR #2286, opened 2024-02-26
Allow expressions in backend specs

mr0re1:


tools/validate_configs/golden_copies/configs/igc_tf.yaml
```diff
@@ -21,6 +21,11 @@ vars:
   region: us-east4
   zone: $(vars.region)-c
 
+terraform_backend_defaults:
+ type: gcs
+ configuration:
+   enoz: $(vars.zone)
```
tpdownes, line 27: While our struct supports arbitrary key/value pairs, I'd rather we not test against them. It would fail "terraform init" (although I think the golden copies already fail due to use of "modules" subdir).

tpdownes (changes requested):
Please change the golden copies to use a valid configuration setting.

tpdownes (approved):

PR #2296, opened 2024-02-28
Add support for startup script per nodeset

mr0re1:
### testing done
```yaml
  - id: s1
    settings:
      runners:
      - type: shell
        destination: s1
        content: |
          #!/bin/bash
          echo "p1" > /tmp/secret

  - id: ns1
    source: ./community/modules/compute/schedmd-slurm-gcp-v6-nodeset
    use: [network, s1]
    settings:
      node_count_static: 1
      machine_type: n2-standard-2
      enable_placement: false # the default is: true
  
  - id: p1
    source: ./community/modules/compute/schedmd-slurm-gcp-v6-partition
    use: [ns1]
    settings:
      partition_name: p1
      
// Same for s2, ns2, p2
```

```sh
$ srun -N 1 -p p1 cat /tmp/secret
p1
$ srun -N 1 -p p2 cat /tmp/secret
p2
```


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
```diff
@@ -35,6 +35,11 @@ locals {
   public_access_config = var.disable_public_ips ? [] : [{ nat_ip = null, network_tier = null }]
   access_config        = length(var.access_config) == 0 ? local.public_access_config : var.access_config
 
+  ghpc_startup_script = [{
+    filename = "ghpc_nodeset_startup.sh"
```
nick-stroud, line 39: Is this problematic that each nodeset has the same filename. From your testing it seems not.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -84,10 +84,12 @@ locals {
     filename = "ghpc_startup.sh"
     content  = var.compute_startup_script
   }]
+  nodeset_startup_scripts = {
+  for ns in var.nodeset : ns.nodeset_name => ns.startup_script }
 }
 
 module "slurm_files" {
-  source = "github.com/GoogleCloudPlatform/slurm-gcp.git//terraform/slurm_cluster/modules/slurm_files?ref=6.4.2&depth=1"
+  source = "github.com/GoogleCloudPlatform/slurm-gcp.git//terraform/slurm_cluster/modules/slurm_files?ref=partition_script&depth=1"
```
nick-stroud, line 92: I affirm that this PR has the functionality we are looking for. 

Send it back to me when you have a tagged version in the Slurm repo.
PS. In final version, we will want to roll all references.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
nick-stroud, line 1: Testing we should also do:
- No startup script
- Central startup script on controller - propagates to all partitions
- Central startup script and custom startup script, what is the behavior? does custom override central or do they both run, in what order? We need to document this edge case behavior.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
mr0re1: ack

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
mr0re1: > No startup script

Integration tests will take care of it

> Central startup script on controller - propagates to all partitions
Central startup script and custom startup script, what is the behavior? does custom override central or do they both run, in what order?

```sh
$ srun -N 1 -p p1 cat /tmp/secret
script_common # both s1 & compute_startup_script
script_ns1
$ srun -N 1 -p p2 cat /tmp/secret
script_common # both s2 & compute_startup_script
script_ns2
$ srun -N 1 -p p3 cat /tmp/secret
script_common # only compute_startup_script
$
```

> We need to document this edge case behavior.

Done

nick-stroud (approved):

PR #2303, opened 2024-02-28
Add E2E integration test

mr0re1:
**Motivation:** to test invocations of `ghpc` binary in scenario close to actual usage.

test duration: `~7min`

### DIFF
--- a/tools/cloud-build/daily-tests/e2e.sh
+++ b/tools/cloud-build/daily-tests/e2e.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+depl_name="e2e-${BUILD_ID:0:6}"
+region="us-central1"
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/e2e.sh
```diff
@@ -0,0 +1,31 @@
...
+zone="us-central1-a"
+vars="project_id=$PROJECT_ID,deployment_name=$depl_name,region=$region,zone=$zone"
+
+# Already in a root of the repo
+make
+./ghpc create tools/cloud-build/daily-tests/blueprints/e2e.yaml --vars="$vars" -l ERROR
+./ghpc deploy "$depl_name" --auto-approve
+
+# check instance was created
+gcloud compute instances describe "${depl_name}-0" --project="$PROJECT_ID" --zone="$zone" >/dev/null
```
tpdownes, line 29: a filter on labels would be an interesting check

mr0re1: There are many check we could perform. But I would keep this test as simple as possible:
* It's not what this tests is for. It's to test that user can sucessfully use `ghpc` binary in simpiest of workflows.
* Increases chances of failure, not directly related to the subject of the test (`ghpc` binary) -> need to start worrying about clean up.

Frankly I'm on a fence about "check instance was created". I think it would be reasonable to remove it as well )
WDYT?

cboneti (approved):

tools/cloud-build/daily-tests/e2e.sh
tpdownes: I will approve, but if our goal is super simple, I might consider something that creates next to nothing, like pre-existing-vpc (although the default network is not guaranteed to exist)

tpdownes (approved):

tools/cloud-build/daily-tests/e2e.sh
mr0re1: Thank you, changed to `pre-existing-vpc`, runtime down to `~5 min`

PR #2311, opened 2024-02-29
Babysit. Allow to use PR number instead of SHA

mr0re1:
**BREAKING CHANGE**: to keep using SHA, one must specify it using `--sha`

```sh
$ ./babysit_tests.py --pr 2311  all -c=1
# same as
$ ./babysit_tests.py --sha 2e20882  all -c=1
```

### DIFF
--- a/tools/cloud-build/babysit_tests.py
+++ b/tools/cloud-build/babysit_tests.py
@@ -20,6 +20,7 @@ from google.cloud.devtools.cloudbuild_v1.types.cloudbuild import Build, ApproveB
 import time
 import argparse
 import subprocess
+import requests
 
 DESCRIPTION = """
 babysit_tests is a tool to approve & retry CloudBuild tests.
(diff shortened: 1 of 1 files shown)


tools/cloud-build/babysit_tests.py
```diff
@@ -280,12 +288,20 @@ def get_default_project():
     parser.add_argument("-r", type=int, default=1,
                         help="Number of retries, to disable retries set to 0, default is 1")
     args = parser.parse_args()
+
+    assert (args.sha is None) ^ (args.pr is None), "either --pr or --sha are required"
+    if args.pr is not None:
+        sha = get_pr(args.pr)["head"]["sha"][:7]
+    else:
+        sha = args.sha[:7]
```
nick-stroud, line 296: Technically the :7 should not be needed here or on L294 as it is handled in babysitter. I think best to only trim in one place. Makes sense to me to leave it long here and let it be trimmed by babysitter.

nick-stroud (approved):

PR #2313, opened 2024-03-01
Populate tags for all tests, based on babysitter selectors + new

mr0re1:
Last change before removing hardcoded selectors from babysit.


nick-stroud (approved):
Strongly in favor of #2305 for which this seems like a pre-req.

PR #2323, opened 2024-03-06
Allow `ghpc deploy blueprint.yaml`

mr0re1:
```sh
$ ./ghpc deploy examples/hpc-slurm-v6.yaml --vars=project_id=bullrun -l ERROR
Initializing deployment group hpc-slurm/primary
Testing if deployment group hpc-slurm/primary requires adding or changing cloud infrastructure
Deployment group hpc-slurm/primary requires adding or changing cloud infrastructure ...

$ ./ghpc deploy hpc-slurm
Testing if deployment group hpc-slurm/primary requires adding or changing cloud infrastructure
Deployment group hpc-slurm/primary requires adding or changing cloud infrastructure ...

$ ./ghpc deploy hpc-slurm --vars=zone=4
Error: cannot specify flag "vars" with DEPLOYMENT_DIRECTORY provided
```

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"hpc-toolkit/pkg/modulewriter"
 	"hpc-toolkit/pkg/shell"
+	"os"
 
 	"github.com/spf13/cobra"
 )
@@ -49,6 +50,14 @@ func getApplyBehavior() shell.ApplyBehavior {
 func addAutoApproveFlag(c *cobra.Command) *cobra.Command {
 	c.Flags().BoolVar(&flagAutoApprove, "auto-approve", false, "Automatically approve proposed changes")
 	return c
+}
+
+func checkExists(cmd *cobra.Command, args []string) error {
+	path := args[0]
+	if _, err := os.Lstat(path); err != nil {
+		return fmt.Errorf("%q does not exist", path)
+	}
+	return nil
 }
 
 func checkDir(cmd *cobra.Command, args []string) error {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -64,22 +64,26 @@ var (
 		Short:             "Create a new deployment.",
 		Long:              "Create a new deployment based on a provided blueprint.",
 		Run:               runCreateCmd,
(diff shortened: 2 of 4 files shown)


tpdownes:
@mr0re1 in the description above can you replace `$TOOLKIT_DEFAULT_VARS` with the actual string content?

cmd/utils.go
```diff
@@ -51,6 +52,17 @@ func addAutoApproveFlag(c *cobra.Command) *cobra.Command {
 	return c
 }
 
+func checkExists(cmd *cobra.Command, args []string) error {
+	path := args[0]
+	if path == "" {
```
tpdownes, line 57: This seems like undesirable behavior:

```
ghpc deploy ""
Error: .ghpc/artifacts must be a writable directory
```

There is also an internal bug tracking being able to run "ghpc import inputs DOT" (i.e. the period character) where the deployment name is not properly inferred from the basename of the DOT directory. These may be related.

tpdownes (changes requested):
Please address my comment either by fixing  bad behavior or by convincing me that it's not a problem.

Otherwise, the overall functionality LGTM and it reduces duplication.

cmd/utils.go
tpdownes: Similarly if you are in a deployment directory, you get:

```
~/b/testfix❯❯❯ ghpc deploy ""
Testing if deployment group first requires adding or changing cloud infrastructure
Deployment group first requires adding or changing cloud infrastructure
Summary of proposed changes: Plan: 10 to add, 0 to change, 0 to destroy.
(D)isplay full proposed changes,
(A)pply proposed changes,
(S)top and exit,
(C)ontinue without applying
Please select an option [d,a,s,c]:
```

mr0re1: Both of your examples are reproducible on `develop`. I would prefer to address it independently of this change. 

```sh
(develop %)$ ./ghpc deploy ""
Error: .ghpc/artifacts must be a writable directory
...
(develop %)$ ../ghpc deploy ""
Initializing deployment group primary
Testing if deployment group primary requires adding or changing cloud infrastructure
```

tpdownes (approved):

PR #2327, opened 2024-03-07
Remove slurm-lustre v5 example and references

harshthakkar01:
### Submission Checklist


docs/vm-images.md
nick-stroud, line 1: This is problematic. 
1) If you visualize the table this change breaks the formatting.
2) The way I would read this change to the table as an outsider is that we no longer support luster with Slurm-GCP. **This is not the case**. 

I believe Slurm GCP v6 should also support Lustre with CentOS and Rocky. Do we have an example in the repo of using lustre with each os? I am thinking enterprise slurm has lustre + centos and is tested, you could point to that. I don't know that we have migrated v6 enterprise slurm example but when we do that could be the other pointer.

nick-stroud (changes requested):

docs/vm-images.md
harshthakkar01: Updated.

We have v5 example `hpc-enterprise-slurm` using lustre and centos. So I updated it with that.
For v6 we dont currently have, I can hold this PR till we add any lustre + v6(rocky) example and then add that example here. LMKWYT

nick-stroud: I am fine to merge so long as we remember to link back to the v6 example when we create it. If you can create responsibility for that (I would put a task on a timer to remind you), then we can merge.

nick-stroud (approved):

docs/vm-images.md
harshthakkar01: I created a task and assigned it to myself. Will update it once we have example.

PR #2330, opened 2024-03-07
Remove slurm-gcp high io blueprint and integration test

harshthakkar01:
This PR adds SlurmGCP v6 version of high-io blueprint and adds integration test for it. 

### Submission Checklist


tools/cloud-build/daily-tests/blueprints/slurm-gcp-v6-high-io.yaml
nick-stroud, line 1: What is this blueprint covering that slurm-enterprise does not have? I wonder if the existence of high io is just from timing since it pre-dated enterprise. I realize it would have been better to ask this in planning.

harshthakkar01: You are right. Nothing special. What do you propose ? Should we remove this and v5 as well ?

nick-stroud: I agree with that course of action, so long as there is really no delta.

harshthakkar01: Updated to remove high-io-v5 example.

tools/cloud-build/daily-tests/blueprints/slurm-gcp-v5-high-io.yaml
nick-stroud, line 1: After reviewing high io and enterprise blueprints back to back, I believe the one thing we are loosing is that high io is the only blueprint that deploys a partition with multiple nodesets. 

Can we add multiple nodesets into another tested blueprint that we are keeping, one that is not in the examples folder, such as slurm-startup-script or lustre-slurm.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/blueprints/slurm-gcp-v5-high-io.yaml
harshthakkar01: We have blueprint with 1 partition and multiple nodeset. For v6, it is 
tools/validate_configs/test_configs/node-groups-v6.yaml. 
#2347 

Will that be sufficient ?

nick-stroud: Not if we don't deploy that blueprint. It provides some coverage as it will get run through ghpc create, but we will still be losing some coverage of deploying a blueprint with this configuration. I am hoping we can just tack this onto an existing test.

harshthakkar01: Gotcha. There is a blueprint htc-slurm having multiple nodesets in single partition. I added test for v6 here
#2348

nick-stroud (approved):

PR #2334, opened 2024-03-07
Added enable_devel for packer build to fix issue with bp

cdunbar13:
Tested using the hpc-build-slurm-image.yaml and a branch I had created in slurm-gcp.

### DIFF
--- a/community/examples/hpc-build-slurm-image.yaml
+++ b/community/examples/hpc-build-slurm-image.yaml
@@ -117,3 +117,4 @@ deployment_groups:
     settings:
       machine_type: n2d-standard-4
       instance_image: $(vars.built_instance_image)
+      enable_devel: false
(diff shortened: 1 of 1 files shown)


community/examples/hpc-build-slurm-image.yaml
```diff
@@ -117,3 +117,4 @@ deployment_groups:
     settings:
       machine_type: n2d-standard-4
       instance_image: $(vars.built_instance_image)
+      enable_devel: false
```
nick-stroud, line 120: Add a comment to this line, something like:
`# Will cause Slurm auto-scaling scripts to be sourced from built image

nick-stroud (approved):

community/examples/hpc-build-slurm-image.yaml
cdunbar13: Done

PR #2339, opened 2024-03-08
Implement `ghpc_stage` function to stage files into deployment

mr0re1:
Add function `ghpc_stage(path string) -> string`,
during `ghpc create` the staged files (directories) will be placed into dedicated directory within deployment folder and relative path (from any deployment group) will be the result.

### Usage:
Useful for referencing files by relative path and moving deployment folder.
```yaml
vars:
  ...
  # hello.sh is a local file at $(pwd)/hello.sh
  script_path: $(ghpc_stage("hello.sh"))
  ...

deployment_groups:
- group: primary
  modules:
  ...
  - id: script
    settings:
      runners:
      - type: shell
        destination: hello.sh
        source: $(vars.script_path)
  
  - id: workstation
    use: [network, script]
    settings:
      machine_type: n2-standard-2
```
### Limitations:

* Currently `ghpc_stage` function can only be used in `vars` block.
Attempt to use it directly in module settings would lead to create-time error:

Screenshot 2024-03-07 at 4 38 27 PM

* Errors (e.g. invalid path) can't not be attributed to the position in the blueprint yaml file yet.
Screenshot 2024-03-07 at 4 52 19 PM

* The value for `ghpc_stage` evaluated and actual staging are done during `create`. Changes to the `terraform.tfvars` will not affect it, unless `ghpc create` is called again. E.g.:
```yaml
vars:
  a: "path.txt"
  b: $(ghpc_stage(vars.a)) 
  # changes to vars.a after create will not change the value of vars.b
```

### Details:
```sh
$ tree ~/tmp/t2/
/usr/local/google/home/orlov/tmp/t2/
├── d
│   └── j.txt
├── f.txt
└── tst.yaml

$ cat ~/tmp/t2/f.txt
content-of-f-txt

$ cat ~/tmp/t2/d/j.txt
content-of-d-j-txt

$ cat ~/tmp/t2/tst.yaml
...
vars:
  fp: $(ghpc_stage("f.txt"))
  djp: $(ghpc_stage("d"))/j.txt

deployment_groups:
- group: primary
  modules:
  - id: f
    settings:
      network_name: $(file(vars.fp))
      region: us-central1

  - id: dj
    source: ./modules/network/pre-existing-vpc
    settings:
      network_name: $(file(vars.djp))
      region: us-central2

$ ./ghpc create ~/tmp/t2/tst.yaml --out=/tmp/

$ tree /tmp/tst/.ghpc/
/tmp/tst/.ghpc/
├── artifacts
│   ├── DO_NOT_MODIFY_THIS_DIRECTORY
│   └── expanded_blueprint.yaml
├── previous_deployment_groups
└── staged
    ├── d_8277e0910d
    │   └── j.txt
    └── f.txt_8a7cbf1fec

$ cat /tmp/tst/primary/terraform.tfvars
...
djp = "../.ghpc/staged/d_8277e0910d/j.txt"

fp = "../.ghpc/staged/f.txt_8a7cbf1fec"
...

$ ./ghpc deploy /tmp/tst
...
The subnetwork: content-of-d-j-txt could not be found in project: zzz and region: us-central2.
...
The subnetwork: content-of-f-txt could not be found in project: zzz and region: us-central1.
```


tpdownes:
This review does not focus on the implementation details, but the user experience. My first thought is that it is very good that we support HCL functions in the Toolkit, but they are not yet very well documented for users. Additionally, `ghpc_stage` doesn't manifest to the user any differently than an HCL function, so I'm not sure that it will be obvious where to look for the behavior of functions (esp. for users not familiar with Terraform). We will have to do a good job of documenting the file staging and how ghpc_stage values resolve in a blueprint.

An function-less alternative that I think you and the code reviewer might consider as more intuitive:

```yaml
vars:
...

local_files:
  file1: path/to/file
  dir1: path/to/dir

deployment_groups:
- group: grp0
  modules:
    id: mod0
    settings:
      script: $(local_files.file1.content)
      folder_path: $(local_files.dir1.path)
```

You could imagine other attributes like `abspath`, `basename`, `md5sum`, `zip` and so on. Even if these were implemented in our deployments as HCL functions, the behavior seems more self-explanatory and create-time validation could be performed. For example, `md5sum` attributes would be rejected when pointing to a directory. Other ideas:

- validator that notices when the path of the blueprint contains local_files not staged
- validator that warns when absolute paths are used in local_files (in my view is a bad practice).

pkg/config/staging.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/pkg/errors"
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+// Relative path from deployment group to the staging directory
+const StagingDir = "../.ghpc/staged"
+
+// Returns a map of absolute src to relative (to deployment folder) destination path
```
harshthakkar01, line 31: Can we simplify this comment ?

pkg/config/staging.go
```diff
@@ -0,0 +1,120 @@
...
+func (bp Blueprint) StagedFiles() map[string]string {
+	if len(bp.stagedFiles) == 0 {
+		return nil
+	}
+
+	if bp.path == "" {
+		panic("blueprint doesn't have known path, can't resolve staged files to absolute paths")
+	}
+	res := map[string]string{}
+	for src, dst := range bp.stagedFiles {
```
harshthakkar01, line 41: Should we log it somewhere how many stagedfiles found ? Or it is already being taken care of ?

pkg/config/staging.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (bp *Blueprint) makeGhpcStageImpl() func(src string) string {
+	// Move implementation instantiation to a separate function for easier testing
+	return func(src string) string {
+		// NOTE: we can't perform file validation here, because evaluation can be performed
+		// on expanded blueprints, and relative `src` will not be valid at that point.
+		// NOTE: this function needs to be deterministic, regardless of the invocation context.
+		hash := fmt.Sprintf("%x", md5.Sum([]byte(src)))[:10]
+		name := filepath.Base(src)
+		if name == "." || name == ".." || filepath.ToSlash(name) == "/" {
```
harshthakkar01, line 58: Some description as to what we do here and why replace `name` to `file` would be better.

pkg/config/staging.go
```diff
@@ -0,0 +1,120 @@
...
+		Type:        function.StaticReturnType(cty.String),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			src := args[0].AsString()
+			dst := impl(src)
+			return cty.StringVal(dst), nil
+		},
+	})
+}
+
+// Validate that there `ghpc_stage` is only used in `vars` declarations
```
harshthakkar01, line 86: nit: punctuation here and all function comments.

pkg/modulewriter/modulewriter.go
```diff
@@ -106,6 +112,63 @@ func WriteDeployment(bp config.Blueprint, deploymentDir string) error {
...
+	}
+	return true, nil
+}
+
+// stageFile copies a file from the src (absolute path) to the dst (relative to group directory in deployment directory)
+func stageFile(deplPath, absSrc, relDst string) error {
+	// relDst is relative to group folders ("../.ghpc/staged"),
+	// prepend any_group_dir to be "eaten" by ".."
+	absDst := filepath.Join(deplPath, "any_group_dir", relDst)
+	dstExists, err := doesExists(absDst)
```
harshthakkar01, line 150: nit: can be combined with below if check.

pkg/modulewriter/modulewriter.go
```diff
@@ -106,6 +112,63 @@ func WriteDeployment(bp config.Blueprint, deploymentDir string) error {
...
+	if err != nil {
+		return err
+	}
+
+	srcExists, err := doesExists(absSrc)
```
harshthakkar01, line 155: nit: can be combined with below if check.

pkg/config/config.go
```diff
@@ -242,6 +243,13 @@ type Blueprint struct {
 	Vars                     Dict
 	Groups                   []Group          `yaml:"deployment_groups"`
 	TerraformBackendDefaults TerraformBackend `yaml:"terraform_backend_defaults,omitempty"`
+
+	// internal & non-serializable fields
+
```
harshthakkar01, line 248: Is there something missing here ?

pkg/config/staging.go
mr0re1: Nah, we will only log if some problem occurs (e.g. missing file)

pkg/modulewriter/modulewriter.go
mr0re1: I don't see how.

pkg/modulewriter/modulewriter.go
mr0re1: I don't see how.

pkg/config/config.go
mr0re1: >  this entire section is for internal + non-serializable fields

this

pkg/config/staging_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		if diff := cmp.Diff(got, []StagedFile{
+			{"/root/abs.txt", "../.ghpc/staged/abs.txt_ffac5d1d6b"},
+			{"/zebra/one.txt", "../.ghpc/staged/one.txt_08bc3de154"},
+			{"/zebra/zero/one.txt", "../.ghpc/staged/one.txt_f8669c6c22"},
+		}); diff != "" {
+			c.Errorf("diff (-want +got):\n%s", diff)
+		}
+	}
+
```
harshthakkar01, line 72: nit: extra space

pkg/modulewriter/modulewriter_test.go
```diff
@@ -565,3 +565,78 @@ func (s *zeroSuite) TestWritePackerDestroyInstructions(c *C) {
...
+		if err := os.WriteFile(absDst, []byte("pulp"), 0644); err != nil {
+			c.Fatal(err)
+		}
+
+		c.Assert(stageFile(deplDir, f), IsNil)
+		dat, err := os.ReadFile(absDst)
+		c.Assert(err, IsNil)
+		c.Assert(string(dat), Equals, "pulp")
+	}
+
```
harshthakkar01, line 641: nit: extra space

harshthakkar01 (approved):

PR #2343, opened 2024-03-08
Babysit. Add `--name` and `--all` flags, remove `test_selectors` args

mr0re1:
**Breaking change:**
Remove args for `test_selectors`. Instead:
```sh
babysit_tests.py tag1 tag2 name1 name2 all ...
```
Use:
```sh
babysit_tests.py --tags tag1 tag2 --names name1 name2 --all ...
```


tools/cloud-build/babysit_tests.py
nick-stroud, line 1: Please describe in the description the ways in which you know this to be breaking.

nick-stroud (changes requested):

nick-stroud (approved):

PR #2344, opened 2024-03-08
Update terraform_backend_defaults documentation

alyssa-sm:
### Submission Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -71,6 +71,13 @@ under `vars` before using an example blueprint.
 
 ### (Optional) Setting up a remote terraform state
 
+There are two ways to specify [terraform backends] in HPC Toolkit: a default setting that propagates all groups and custom per-group configuration:
+
+* `terraform_backend_defaults` at top-level of YAML blueprint
+* `terraform_backend` within a deployment group definition
+
+Examples of each are shown below. If both settings are used, then the custom per-group value is used without modification.
+
 The following block will configure terraform to point to an existing GCS bucket
 to store and manage the terraform state. Add your own bucket name in place of
 `<<BUCKET_NAME>>` and (optionally) a service account in place of
@@ -87,6 +94,22 @@ terraform_backend_defaults:
     impersonate_service_account: <<SERVICE_ACCOUNT>>
 ```
 
+All Terraform-supported backends are supported by the Toolkit. Specify the backend using `type` and its [configuration block] using `configuration`.
+
+For the [gcs] backend, you must minimally supply the `bucket` configuration setting. The `prefix` setting is generated automatically as "blueprint_name/deployment_name/group_name" for each deployment group. This ensures uniqueness.
+
+If you wish to specify a custom prefix, use a unique value for each group following this example:
+
+```yaml
+deployment_groups:
+- group: example_group
+  terraform_backend:
+    type: gcs
+    configuration:
+      bucket: your-bucket
+      prefix: your/object/prefix
+```
+
 You can set the configuration using the CLI in the `create` and `expand`
 subcommands as well:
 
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -71,6 +71,10 @@ under `vars` before using an example blueprint.
 
 ### (Optional) Setting up a remote terraform state
 
+There are two ways to specify terraform backends in HPC Toolkit, defaults and per-group.
```
tpdownes, line 74: Make "terraform backends" a link to:

https://developer.hashicorp.com/terraform/language/settings/backends/configuration

examples/README.md
```diff
@@ -71,6 +71,10 @@ under `vars` before using an example blueprint.
...
+
+The first way is using terraform_backend_defaults at the top-level.
```
tpdownes, line 76: Make `terraform_backend_defaults` fixed width text.

examples/README.md
```diff
@@ -87,6 +91,12 @@ terraform_backend_defaults:
     impersonate_service_account: <<SERVICE_ACCOUNT>>
 ```
 
+It should be noted that "gcs" is used in the example above, but the configuration accepts anything that the backend accepts.
```
tpdownes, line 94: I'd replace this with something like "All Terraform-supported backends are supported by the Toolkit." Specify the backend using `type` and its [configuration block](https://developer.hashicorp.com/terraform/language/settings/backends/configuration#using-a-backend-block) using `configuration`. 

For the [gcs](https://developer.hashicorp.com/terraform/language/settings/backends/gcs) backend, you must minimally supply the `bucket` configuration setting. The `prefix` setting is generated automatically as "blueprint_name/deployment_name/group_name" for each deployment group. This ensures uniqueness.

If you wish to specify a custom prefix, use a unique value for each group following this example:

```yaml
deployment_groups:
- group: example_group
  terraform_backend:
    type: gcs
    configuration:
      bucket: your-bucket
      prefix: your/object/prefix
```

tpdownes (changes requested):
We need to link a bit more to Hashicorp docs and also give explicit example of a per-group setting.

examples/README.md
alyssa-sm: Done

examples/README.md
alyssa-sm: Done

examples/README.md
alyssa-sm: I think you phrased it nicely here so I replaced the section with your suggestion.

examples/README.md
```diff
@@ -71,6 +71,10 @@ under `vars` before using an example blueprint.
 
 ### (Optional) Setting up a remote terraform state
 
+There are two ways to specify [terraform backends] in HPC Toolkit, defaults and per-group.
```
tpdownes, line 74: > There are two ways to specify [terraform backends] in HPC Toolkit, defaults and per-group.

Let's say something more like

> There are two ways to specify [terraform backends] in HPC Toolkit: a default setting that propagates all groups and custom per-group configuration:
> 
> `terraform_backend_defaults` at top level of YAML blueprint
> `terraform_backend` within a deployment group definition
>
> Examples of each are shown below. If both settings are used, then the custom per-group value is used without modification.

tpdownes (changes requested):
We're close, but I notice that we don't have "terraform_backend" mentioned in the prose, only in the YAML example.

examples/README.md
alyssa-sm: Done

tpdownes (changes requested):

examples/README.md
```diff
@@ -71,6 +71,13 @@ under `vars` before using an example blueprint.
 
 ### (Optional) Setting up a remote terraform state
 
+There are two ways to specify [terraform backends] in HPC Toolkit: a default setting that propagates all groups and custom per-group configuration:
+
+`terraform_backend_defaults` at top-level of YAML blueprint
```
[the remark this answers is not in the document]
alyssa-sm: Good catch!

tpdownes (approved):

PR #2353, opened 2024-03-12
Update zone to us-central1-a for TPU blueprint and test and increase retries

harshthakkar01:
### Submission Checklist

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-slurm-v6-tpu.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-slurm-v6-tpu.yml
@@ -44,7 +44,7 @@
 - name: Run JAX verification
   register: jax_status
   failed_when: jax_status.rc != 0
-  retries: 3
+  retries: 10
   delay: 100
   ansible.builtin.command: |
     srun -N 1 -p tpu bash -c '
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/test-slurm-v6-tpu.yml
```diff
@@ -44,8 +44,8 @@
 - name: Run JAX verification
   register: jax_status
   failed_when: jax_status.rc != 0
-  retries: 3
-  delay: 100
+  retries: 5
+  delay: 200
```
nick-stroud, line 48: Why extend the delay also? Seems like you could do retries 10 delay 100 and probability would be better.

nick-stroud (changes requested):

tools/cloud-build/daily-tests/ansible_playbooks/test-slurm-v6-tpu.yml
harshthakkar01: I added 200 seconds to make sure if a TPU is not available we give enough time before we retry to avoid frequent retries in short time. WDYT ?

nick-stroud: The cost of requesting (and getting denied) is cheap, the cost of waiting an additional 100s to have the test move forward is expensive. It seems 100s was working ok before since we were already using that.

I would do retries 10, delay 100

harshthakkar01: Makes sense. Time is more expensive. :-) Updated to 10 retries with 100 sec interval.

nick-stroud (approved):

PR #2379, opened 2024-03-19
Deprecate `schedmd-slurm-gcp-v6-partition.network_storage`

mr0re1:
SlurmGCP V6 doesn't support specifying `network_storage` at partition level.


community/modules/compute/schedmd-slurm-gcp-v6-partition/variables.tf
nick-stroud, line 1: What follow on work do we have planned? Will we be exposing this on the nodeset?
Would it make more sense to first expose on the nodeset and then we can move all the partition usage to nodeset instead of just deleting. My fear is that all of the examples that you are removing will not get replaced when we restore this functionality.

mr0re1: > My fear is that all of the examples that you are removing will not get replaced when we restore this functionality.

I understand your concern, but examples remained functionally the same, because FS were duplicated on controller module -> I consider those to be bad examples.

Once nodeset functionality is enabled, we could add it to just a few(one) examples (e.g. enterprise) that explicitly showcase/document "fs per nodeset" functionality.

nick-stroud: SG, and agreed about bad examples. I will approve with the understanding that we are adding functionality back to nodeset in immediate future.

community/modules/compute/schedmd-slurm-gcp-v6-partition/variables.tf
```diff
@@ -199,3 +185,14 @@ variable "nodeset_dyn" {
...
 }
+
+# tflint-ignore: terraform_unused_declarations
+variable "network_storage" {
+  description = "DEPRECATED"
+  type        = any
+  default     = null
+  validation {
+    condition     = var.network_storage == null
+    error_message = "network_storage in partition module is deprecated and should be removed."
```
nick-stroud, line 196: Note for future (not this PR). When we add functionality to nodeset, please update this error to reference nodeset. You can close once you have read, so long as you agree.

nick-stroud (approved):

PR #2381, opened 2024-03-20
Allow nodeset to be used by multiple partitions

mr0re1:
### Side effects:
Removes validation for uniqueness of nodeset name. Given default value guaranteed to be unique this doesn't seem to be a big problem.
There are some avenues to restore this validation, but it's rather complex and should be addressed separately.

### Testing done:

```yaml
  - id: ns
    ...

  - id: p1
    use: [ns]
    ...

  - id: p2
    use: [ns]
    ...

  - id: slurm_controller
    use: [network, p1, p2]
    ...
```

```sh
$ srun -N 1 -p p1 hostname
orlovmulti-ns-0
$ srun -N 1 -p p2 hostname
orlovmulti-ns-0
```

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
@@ -13,9 +13,14 @@
 # limitations under the License.
 
 locals {
-  nodeset_map     = { for x in var.nodeset : x.nodeset_name => x }
-  nodeset_tpu_map = { for x in var.nodeset_tpu : x.nodeset_name => x }
-  nodeset_dyn_map = { for x in var.nodeset_dyn : x.nodeset_name => x }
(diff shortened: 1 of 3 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -13,9 +13,14 @@
 # limitations under the License.
 
 locals {
-  nodeset_map     = { for x in var.nodeset : x.nodeset_name => x }
-  nodeset_tpu_map = { for x in var.nodeset_tpu : x.nodeset_name => x }
-  nodeset_dyn_map = { for x in var.nodeset_dyn : x.nodeset_name => x }
+  nodeset_map_ell = { for x in var.nodeset : x.nodeset_name => x... }
```
cboneti, line 16: Perhaps validate that all elements in this list are equal.

cboneti (approved):

PR #2383, opened 2024-03-20
Set default validation level to `ERROR`

mr0re1:
**Motivation:** to safe new user from common deployment-time errors (e.g. APIs disabled)

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,10 @@ In the right side, expand the Filters view and then filter by label, specifying 
 
 ## Troubleshooting
 
+### Authentication
+
+Confirm that you have [properly setup Google Cloud credentials](#gcp-credentials)
+
 ### Slurm Clusters
 
 Please see the dedicated [troubleshooting guide for Slurm](docs/slurm-troubleshooting.md).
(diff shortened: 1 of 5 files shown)


tpdownes (changes requested):
I would like a section at the top of

https://goo.gle/hpc-toolkit-troubleshooting

that links to

https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main?tab=readme-ov-file#gcp-credentials

Something like:

```yaml
## Troubleshooting

### Authentication

Confirm that you have [properly setup Google Cloud credentials](#gcp-credentials)

### Slurm Clusters
```

tpdownes:
For record, here is example output:

```
~/p/h/hpc-toolkit (elError)❯❯❯ ghpc deploy example-with-bad-project.yaml
validator "test_project_exists" failed:
Error: project ID bad-project-id does not exist or your credentials do not have permission to access it

validators are shown below.

validation failed due to the issues listed above
```

mr0re1:
Thank you for suggestion. Done

tpdownes (approved):

PR #2388, opened 2024-03-21
Add `materialization` stage.

mr0re1:
* Add `materialization` (goes after `expanding`);
* Add `Clone` to the `Blueprint`;
* Improve error rendering by passing context.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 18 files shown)


pkg/config/config.go
```diff
@@ -73,6 +75,16 @@ type Group struct {
 	deprecatedKind interface{} `yaml:"kind,omitempty"` //lint:ignore U1000 keep in the struct for backwards compatibility
 }
 
+func (g *Group) Clone() Group {
```
harshthakkar01, line 78: nit: comment for exported function.

pkg/config/config.go
```diff
@@ -222,6 +234,14 @@ type Module struct {
 	WrapSettingsWith interface{} `yaml:"wrapsettingswith,omitempty"`
 }
 
+func (m *Module) Clone() Module {
```
harshthakkar01, line 237: nit: comment for exported function.

harshthakkar01:
Should we trigger couple of integration tests ?

harshthakkar01 (approved):

PR #2396, opened 2024-03-23
Fixing Makefile `terraform-format` command

cboneti:
`make terraform-format` can be used to call terraform-docs to clean up terraform format and generate README.md files.  A long time ago, we this called `terraform-docs json` which is no longer in use. 
This change makes removes the json part and improves the message.


mr0re1:
LGTM. But couldn't we remove `terraform-format` from `Makefile` all together? `pre-commit` will make changes in-place. What is use case to run `make terraform-format`?

mr0re1 (approved):

PR #2403, opened 2024-03-26
Fix "unused module validation" failures caused by crude deprecation

mr0re1:
In #2379 the datatype was change to `any`, that affected behavior of `use` that leads to validation failures: 
```
module "partition" uses module "fs2", but matching setting and outputs were not found. ...
286:     use: [nodeset, fs1, fs2]
```

Restore datatype back to list.


mr0re1:
> It's unclear to me what was done in https://github.com/GoogleCloudPlatform/hpc-toolkit/commit/2667200d3fae928e3b45bae4551ddd8e53067fea to migrate the feature elsewhere

Nothing. There is no definitive plan to migrate it anywhere.

community/modules/compute/schedmd-slurm-gcp-v6-partition/variables.tf
```diff
@@ -189,10 +189,18 @@ variable "nodeset_dyn" {
...
+    mount_options         = string,
+    client_install_runner = map(string)
+    mount_runner          = map(string)
+  }))
+  default = []
   validation {
-    condition     = var.network_storage == null
-    error_message = "network_storage in partition module is deprecated and should be removed."
+    condition     = length(var.network_storage) == 0
+    error_message = "network_storage in partition module is deprecated and should not be set."
```
cboneti, line 204: sorry, I don't really think this is the right thing to do.
I was very happy to see this comment here. I recently saw this error message and almost opened a bug.  
This message should say *where* it is valid to set network storage. Is this now in the node_set? is it in the controller?

tpdownes (approved):

PR #2406, opened 2024-03-27
Add support for `ghpc_stage` in module.settings

mr0re1:
During `materialization` stage, transform module settings by partially evaluating `ghpc_stage`.


tools/validate_configs/golden_copies/configs/igc_tf.yaml
```diff
@@ -41,6 +41,6 @@ deployment_groups:
     source: modules/file-system/filestore
     use: [network0] # wires network_id
```
harshthakkar01, line 42: nit: s/network0/network

pkg/config/expression.go
```diff
@@ -544,20 +559,19 @@ func parseBpLit(s string) (cty.Value, error) {
...
 		_, diag := hclsyntax.ParseExpression([]byte(s[:i]), "", hcl.Pos{})
 		if !diag.HasErrors() { // found an expression
-			exp, err := BlueprintExpressionLiteralToExpression(s[:i])
-			return exp, s[i+1:], err
+			return s[:i], s[i+1:], nil
 		}
 		err = diag // save error, try to find another closing bracket
 	}
-	return nil, s, err
+	return "", s, err
```
harshthakkar01, line 574: Why this is changed from nil to empty string ?

harshthakkar01:
nit: Description

tools/validate_configs/golden_copies/configs/igc_tf.yaml
mr0re1: Out of scope . Don't want to make this large even bigger for no good reason.

pkg/config/expression.go
mr0re1: Because signature have changed:
```go
- func greedyParseHcl(s string) (Expression, string, error) {
+ func greedyParseHcl(s string) (string, string, error) {
```

pkg/config/staging_test.go
```diff
@@ -97,3 +92,91 @@ func (s *zeroSuite) TestGhpcStageFunc(c *C) {
...
+	type test struct {
+		input string
+		want  string
+	}
+
+	// testing partial evaluation of `upper`
+	tests := []test{
+		{`upper("a")`, `"A"`},
+		{`file(upper("a"))`, `file("A")`},
+		{`lower("A") + upper("b")`, `lower("A")+"B"`},
```
harshthakkar01, line 106: Shoudn't `lower` function here update it to lowercase in the want string ?

mr0re1: > // testing partial evaluation of `upper` 

not `lower`

> ```{`lower("A") + upper("b")`, `lower("A")+"B"`},```

`lower` is not evaluated because it's not needed to evaluate `upper`

> ```{`upper(lower("A"))`, `"A"`},```

`lower` is evaluated because it's needed to evaluate `upper`

pkg/config/staging.go
```diff
@@ -88,38 +90,106 @@ func (bp *Blueprint) makeGhpcStageFunc() function.Function {
 	})
 }
 
-// Validate that the `ghpc_stage` is only used in `vars` declarations
-func (bp Blueprint) validateNoGhpcStageFuncs() error {
+// Update module settings in place, evaluating `ghpc_stage` expressions
+func (bp *Blueprint) evalGhpStageInModuleSettings() error {
```
harshthakkar01, line 94: Should be evalGhpcStageInModuleSettings ?

harshthakkar01 (approved):

PR #2408, opened 2024-03-27
Roll forward "Add example using Slurm static compute nodes"

nick-stroud:
Original PR #2393 caused test failures.  
Original PR was reverted in GoogleCloudPlatform/hpc-toolkit#2404.  

This PR is the roll-forward and contains the fix to the broken tests in a new commit. It makes the new task lists optional by providing a default of an empty list.


nick-stroud:
I have run one test that exercises each base-integration-test and slurm-integration-test tests to spot check.

tpdownes (approved):

PR #2410, opened 2024-03-27
Simplify & speed up integraion tests

mr0re1:
## Current test flow
```
golang:alpine -> fetch(30s) + make ghpc (3m) ↘
                                              ghpc:builder -> actual test (?s)
ghpc:builder -> just fetch (3m)              ↗

duration: max(3m, 3m30s) + ?s = 3m30s + ?s
```

## New test flow
```
ghpc:tester -> fetch (1m30s) +  make ghpc(3s) + actual test (?s)
duration:  1m33s + ?s
```

## Actions taken
* Create a new image `test-runner` that only has dependencies for tests (compare with builder):
  *  smaller (1GB vs 1.2GB);
  * less layers (9 vs 16);
  * fetches faster (1m30s vs 3m) 
* Preserve `go build` on the image cache to speed up `make ghpc ~3m -> ~3s`
* Update `monitoring` test to new flow, to showcase (9m40 vs 12m).

## Followups
Update rest of tests

### DIFF
--- a/tools/cloud-build/images/test-runner/Dockerfile
+++ b/tools/cloud-build/images/test-runner/Dockerfile
@@ -1,0 +1,48 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM golang:bullseye
+
+ENV GOCACHE=/ghpc_go_cache
+
+# copy the source code, to build the binary
+# use `/workspace` to match path in cache for future builds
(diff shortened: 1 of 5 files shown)


tools/cloud-build/images/test-runner/Dockerfile
nick-stroud, line 1: We now have:
```
tools/cloud-build/Dockerfile # builder
tools/cloud-build/ghpc-docker/Dockerfile # public container?
tools/cloud-build/images/test-runner # new test runner
```

Can we pick a common location, folder level for these. 

Maybe:
```
tools/cloud-build/images/test-runner
tools/cloud-build/images/builder
tools/cloud-build/images/ghpc-docker
```

nick-stroud, line 1: How hard would it be to build the builder off of the test-runner. This would make it much clearer what the delta is and stop drift.

nick-stroud (changes requested):

tools/cloud-build/images/test-runner/Dockerfile
mr0re1: That's easy, just use latest `test-runner` as a base for `builder`. Plus move it to the same build config, to guarantee  consistency (e.g. not latest was used).

mr0re1: > Maybe:

That's the plan, but goes outside of scope of this PR.

PR #2413, opened 2024-03-28
Remove "hard" dependency between login instance and controller instance

mr0re1:
To avoid unwanted distraction of login VM on controller VM changes (even **in-place**), 
remove the "hard" dependency.
Instead use "implicit" dependency on `slurm_files.checksum` to protect agains race conditions.
Instead use `lifecycle.replace_triggered_by ~ controller.self_link`

Other cahnges:
* Stop using `slurm-gcp.git//terraform/slurm_cluster/modules/slurm_login_instance`, use `_slurm_instance` instead to get access to necessary instance-level (not just through template) `labels`.
* Given opportunity use  `access_config` over `enable_public_ip`. 

**Testing**

```sh
# after changing `controller_startup_script` content
$ terraform plan 
...
# module.slurm_controller.module.slurm_login_instance["login"].google_compute_instance_from_template.slurm_instance[0] 
# will be updated in-place
  ~ resource "google_compute_instance_from_template" "slurm_instance" {
        id                        = "projects/io-playground/zones/us-central1-a/instances/orlovtst-login-login-001"
      ~ labels                    = {
          - "slurm_files_checksum" = "84613df9573297c805d8aa45abd3eb0b"
...

# after updating `controller.machine_type`:
module.slurm_controller.module.slurm_controller_instance.google_compute_instance_from_template.slurm_instance[0] must be replaced
module.slurm_controller.module.slurm_login_instance["login"].google_compute_instance_from_template.slurm_instance[0] will be replaced due to changes in replace_triggered_by
```

Current TF would require destruction (replacement) of login VM.


cboneti:
Don't you want to recreate the login node if the controller is recreated? or am I misunderstanding this?

community/modules/scheduler/schedmd-slurm-gcp-v6-login/main.tf
```diff
@@ -35,6 +35,7 @@ locals {
     }
   ]
 
+  public_access_config = [{ nat_ip = null, network_tier = "STANDARD" }]
```
nick-stroud, line 38: I think we should set network_tier=null. Default in terraform provider is PREMIUM. Public docs say "Use Premium Tier to optimize for performance". In vm-instance and previous slurm v5 login module we set network_tier to null which I assume is same as leaving to default of premium.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/login.tf
```diff
@@ -59,26 +59,27 @@ module "slurm_login_template" {
...
   instance_template = (
     each.value.instance_template != null && each.value.instance_template != ""
     ? each.value.instance_template
     : module.slurm_login_template[each.key].self_link
   )
-  network_tier  = each.value.network_tier
+  labels        = merge(each.value.labels, local.files_cs_labels)
   num_instances = each.value.num_instances
 
   region     = each.value.region
```
nick-stroud, line 81: There is a slight difference here. Before, due to using slurm login node module, the login node would use region of subnetwork and only fall back to var.region if region was not knowable from subnetwork. I am trying to think through all ramifications.

In our wrapper modules we always require subnetwork_self_link and region is optional. I am wondering if login node should take region at all. Would you ever deploy into a region different than the subnetwork.

nick-stroud: We had offline discussion here. I am going to leave it open for now so I remember to return to it.

nick-stroud:
This PR is pending some testing of login node behavior when controller is replaced.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/login.tf
mr0re1: While I agree with importance of resolving "explicit region/zone vs deduced from mandatory subnetwork".  I think this PR doesn't make it any worse, since login node also has `var.zone`(before and after the change) that is passed through.

I propose to handle this issue separately.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/login.tf
```diff
@@ -59,26 +59,30 @@ module "slurm_login_template" {
 
 # INSTANCE
 module "slurm_login_instance" {
-  source   = "github.com/GoogleCloudPlatform/slurm-gcp.git//terraform/slurm_cluster/modules/slurm_login_instance?ref=6.4.3&depth=1"
+  source   = "github.com/GoogleCloudPlatform/slurm-gcp.git//terraform/slurm_cluster/modules/_slurm_instance?ref=trigger_happy&depth=1"
```
nick-stroud, line 62: You still have your branch reference in here.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/login.tf
```diff
@@ -59,26 +59,30 @@ module "slurm_login_template" {
...
 
   region     = each.value.region
   static_ips = each.value.static_ips
   subnetwork = each.value.subnetwork
-  suffix     = each.key
   zone       = each.value.zone
 
-  depends_on = [module.slurm_controller_instance]
+  # trigger replacement of login nodes when the controller instance is replaced
+  replace_trigger = module.slurm_controller_instance.instances_self_links[0]
```
nick-stroud, line 87: IIUC, self link does not contain instance id. Does login replaced even when self link will remain the same before and after replaced. 

https://www.googleapis.com/compute/v1/projects/<project>/zones/us-central1-a/instances/<name>

My expectation on correct behavior is:
- when controller has in place update -> login should not be effected
- when controller is replaced -> ideally login also would not be affected, although it might be reasonable to replace login.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/login.tf
mr0re1: That's why it's labeled as "do-not-merge".

nick-stroud (approved):

PR #2418, opened 2024-03-28
Add hint to run `gcloud auth application-default login`

mr0re1:
Screenshot 2024-03-28 at 11 29 31 AM

### DIFF
--- a/pkg/validators/validators.go
+++ b/pkg/validators/validators.go
@@ -23,7 +23,14 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-const projectError = "project ID %s does not exist or your credentials do not have permission to access it"
+func projectError(p string) error {
+	return config.HintError{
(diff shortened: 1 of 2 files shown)


pkg/validators/validators.go
```diff
@@ -23,7 +23,14 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-const projectError = "project ID %s does not exist or your credentials do not have permission to access it"
+func projectError(p string) error {
+	return config.HintError{
+		Err: fmt.Errorf("project %q does not exist or your credentials do not have permission to access it", p),
+		Hint: "It may be the machine you are working on has not been authenticated.\n" +
```
nick-stroud, line 29: nit: I think it is fine as is. Possible improvement:

**It is possible** the machine you are working on has not been authenticated.

nick-stroud (approved):

PR #2421, opened 2024-03-28
Fixes for workstation creation - new extension added for yaml

cdunbar13:
Some users were having difficulty building workstations, so these are some fixes for that, as well as a new extension for parsing yaml.

This was tested by building out a new workstation and checking if the go and hpc-toolkit directories were present in the user's home directory.

Documentation was updated to alert users that the install and clone may take a few minutes after they log in, if they log in immediately after starting up the workstation.


tools/cloud-workstations/Dockerfile
```diff
@@ -50,14 +50,19 @@ RUN pip install --no-cache-dir --upgrade pip && \
...
 
-RUN chmod a+x /bin/$HPC_FILE && \
-    chmod a+x /etc/workstation-startup.d/$HPC_CONF_FILE
+COPY $CW_DIR/$HPC_WORK_FILE $START_DIR
+RUN chmod a+rx $START_DIR/$HPC_WORK_FILE
 
-RUN mkdir /etc/hpc-toolkit-config
-COPY $CW_DIR/code_oss_requirements.txt /etc/hpc-toolkit-config
+RUN mkdir $HPC_CONFIG_DIR
+COPY $CW_DIR/$CODE_REQS $HPC_CONFIG_DIR
```
mr0re1, line 67: AFAIK, no need for `RUN mkdir`, `COPY` should take care about creation of destination.

tools/cloud-workstations/Dockerfile
```diff
@@ -50,14 +50,19 @@ RUN pip install --no-cache-dir --upgrade pip && \
...
+ARG HPC_CONF_FILE=configure-hpc-toolkit.sh
+ARG CODE_REQS=code_oss_requirements.txt
+ARG HPC_CONFIG_DIR=/etc/hpc-toolkit-config
+ARG START_DIR=/etc/workstation-startup.d
+ARG INSTALL_DIR=/usr/local
 
-COPY $CW_DIR/$HPC_FILE /bin/
-COPY $CW_DIR/$HPC_CONF_FILE /etc/workstation-startup.d/
+COPY $CW_DIR/$HPC_CONF_FILE $INSTALL_DIR/bin
+RUN chmod a+rx $INSTALL_DIR/bin/$HPC_CONF_FILE
```
mr0re1, line 61: We could avoid `RUN chmod` by either:
* setting proper permission at the source (in the repo), since `COPY` preserves permissions;
* use `COPY --chmod=...`.

cdunbar13: Tested this out.  It's not native to the version of Docker that's running, so you need to set an environment variable `DOCKER_BUILDKIT=1` on the yaml file.  It seems to slow down the build, so I don't think we should change this.

tools/cloud-workstations/Dockerfile
cdunbar13: Addressed.

tools/cloud-workstations/200_configure-hpc-toolkit.sh
```diff
@@ -16,4 +16,4 @@
 # The docker script moves this to /etc/workstation-startup.d to run
 # the hpc toolkit configuration when the workstation is brought up for
 # the first time.
-sudo -i -b -u user bash /bin/configure-hpc-toolkit.sh
+runuser user /usr/local/bin/configure-hpc-toolkit.sh &
```
mr0re1, line 19: What is a point of going background here ?

cdunbar13: Pushing it to background is recommended on the google documentation.

The user will have control either way as code-oss should start earlier.  

No error will be presented to the user.  Their workstation would just not be configured correctly.  I'm not sure how we could present an error other than putting something in profile.d that says the system wasn't configured correctly when they start a terminal.  If they don't start a terminal, there's no easy way to alert them.

mr0re1 (approved):

PR #2448, opened 2024-04-04
Use "wait for startup" instead of "sleep 2min" in base tests

mr0re1:
Tried test (`cloud-batch`) runtime reduced by `>1min`


tpdownes (approved):
The only check that failed was not impacted by this change and failed for unrelated reasons.

PR #2449, opened 2024-04-04
Updating garther startup script and integration test

cdunbar13:
The rescue block was broken in the base integration test when the deploy phase went from terraform to ghpc.  This attempts to fix that, make the integration tests more consistent, and cleans up the serial port output.  It also makes it so if the serial log collection fails, that the infrastructure is still deleted.

Tested running the monitoring integration test with an wait-for-startup block that only waits 1 sec to create failures during deployments.

Direct issue:
```
TASK [Capture terraform stderr] ************************************************
fatal: [localhost]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'terraform_output' is undefined. 'terraform_output' is undefined

The error appears to be in '/workspace/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml': line 140, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    rescue:
    - name: Capture terraform stderr
      ^ here
```

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -137,17 +137,19 @@
 
     ## Cleanup and fail gracefully
     rescue:
-    - name: Capture terraform stderr
+    - name: Capture ghpc stderr
+      failed_when: false
       ansible.builtin.set_fact:
-        terraform_apply_stderr_one_line: "{{ terraform_output.results.1.stderr | replace('\n',' ') }}"
+        ghpc_stderr: "{{ deployment.stderr | replace('\n',' ') }}"
 
     - name: Gather logs
+      failed_when: false
       ansible.builtin.include_tasks:
         file: tasks/gather_startup_script_logs.yml
         apply:
           delegate_to: localhost
 
-    - name: Cleanup firewall and infrastructure
+    - name: Include rescue from ghpc failure
       ansible.builtin.include_tasks:
         file: tasks/rescue_ghpc_failure.yml
         apply:
@@ -159,7 +161,6 @@
     - name: Trigger failure (rescue blocks otherwise revert failures)
       ansible.builtin.fail:
         msg: "Failed while setting up test infrastructure"
-      when: true
 
 - name: Run Integration Tests
   hosts: remote_host
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -143,7 +143,6 @@
         ghpc_stderr: "{{ deployment.stderr | replace('\n',' ') }}"
 
     - name: Gather logs
-      failed_when: false
       ansible.builtin.include_tasks:
         file: tasks/gather_startup_script_logs.yml
         apply:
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/tasks/gather_startup_script_logs.yml
```diff
@@ -16,18 +16,28 @@
 - name: Assert variables are defined
   ansible.builtin.assert:
     that:
-    - terraform_apply_stderr_one_line is defined
+    - ghpc_stderr is defined
 
+- name: Get serial port command
+  failed_when: false
+  ansible.builtin.set_fact:
+    serial_port_cmd: '{{ ghpc_stderr | regex_findall("please run:\s+(.+?\s+--project\s+\S+)", "\\1") | first }}'
```
nick-stroud, line 24: What is the value of serial_port_cmd when regex is not found? What does `first` produce when list is empty ? Seems like it could cause error.

I want to be sure that 1) case that nothing is found does not throw error and 2) case that nothing is found results in `serial_port_cmd is defined` resolving to false.

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
nick-stroud, line 1: Can you run a few tests (at least one that uses each of base integration and slurm integration).

tools/cloud-build/daily-tests/ansible_playbooks/tasks/gather_startup_script_logs.yml
cdunbar13: I agree we should make sure all of this doesn't throw an error.  Alternatively, I'd propose that every task here simply uses the `failed_when: false` line to make sure we just keep working through since this is just logging, and we definitely want to get to the cleanup sections.  

The regex here isn't much different than the original line, so this shouldn't be a regression if I update the print out with `failed_when: false`

cdunbar13: I've updated this to include `| default("")` so if nothing is found, it will default to an empty string.  The next two tasks will check for an empty string and treat that as a failure.  This should improve the error handling above what was there before this PR.

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
cdunbar13: Yes, I've run this using lustre-slurm and monitoring integration tests.  On monitoring I forced errors by changing the wait timeout to 1 sec.  I tested it with a good and bad regex search to prove to myself that it functions properly in both situations.  The slurm tests will always fail here because their images don't produce serial port outputs, but it doesn't fail out, and the destruction of the deployment still proceeds correctly.

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
```diff
@@ -143,7 +143,6 @@
         ghpc_stderr: "{{ deployment.stderr | replace('\n',' ') }}"
 
     - name: Gather logs
-      failed_when: false
```
nick-stroud: The idea here is what? That the imported task cannot fail so we no longer need to protect it here? Technically the assert at the beginning could still trigger a failure. Maybe we just leave this in. Thoughts?

nick-stroud (approved):

tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
nick-stroud: I see, I had thought this was on develop, but now I see it was added in this pr.

nick-stroud (approved):

PR #2452, opened 2024-04-05
Add `pre-deploy` task that removes a module from blueprint

mr0re1:
* Add `pre-deploy` task that removes a module from blueprint;
* Move `Run Pre Deploy Tasks` **before** `Create Deployment Directory`;
* Add `Run Pre Deploy Tasks` to `base-integration-test`
* Remove `filestore homefs` from `hpc-slurm-chromedesktop`, saved `~6m` of run time.

### DIFF
--- a/tools/cloud-build/daily-tests/builds/hpc-slurm-chromedesktop.yaml
+++ b/tools/cloud-build/daily-tests/builds/hpc-slurm-chromedesktop.yaml
@@ -15,7 +15,7 @@
 ---
 tags:
 - m.chrome-remote-desktop
-- m.filestore
+- m.filestore # TODO: mark as removed during pre-deploy
 - m.schedmd-slurm-gcp-v5-controller
 - m.schedmd-slurm-gcp-v5-login
 - m.schedmd-slurm-gcp-v5-node-group
--- a/tools/cloud-build/daily-tests/remove_module.py
+++ b/tools/cloud-build/daily-tests/remove_module.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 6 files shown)


tools/cloud-build/daily-tests/remove_module.py
```diff
@@ -0,0 +1,52 @@
...
+  delete_module(bp, mod_id)
+   # remove module from "use"
+  for g in bp["deployment_groups"]:
+    for m in g["modules"]:
+      if mod_id in m.get("use", []):
+        m["use"].remove(mod_id)
+  # TODO:
+  # - move all settings of removed module into some "/dev/null" to prevent
+  #   "variable not used" validation errors
+  # - treat references to removed module (how ?)
```
nick-stroud, line 36: IIUC there is no handling of `$(<module_id>.<output>)` or the like. 

1. Add description and usage instructions. Include written warning in description that includes this gap in functionality.  
2. Could we do a quick RE search of the BP text to just confirm that there are no explicit references and error if there are. Could even look for `<module_id>.`. Maybe this is not needed because it will be caught by ghpc in following steps, but it would be clearer if the error came from the module removal step.

tools/cloud-build/daily-tests/builds/hpc-slurm-chromedesktop.yaml
```diff
@@ -15,7 +15,7 @@
 ---
 tags:
 - m.chrome-remote-desktop
-- m.filestore
+- m.filestore # TODO: mark as removed during pre-deploy
```
nick-stroud, line 18: Here is my concerns

1. Divergence between test and example: You could imagine some class of bug that we will miss due to this change.  CRD + Filestore + Slurm. While it seems unlikely, we should consider the value of any deviation between public example and test. What is the time savings for this test?
2. Loss of overall coverage: Maybe for 1) above we decide that in this case the test speedup is worth the divergence. Probably this would be ok in isolation because we have other Filestore + Slurm tests. Where do we holistically evaluate the matrix (A: nowhere)?

For 1), can you categorize the time savings? If it is meaningful then I would agree to try this out. 2) only becomes a concern if we start to apply this to more tests. Do you have any ideas or thoughts on 2).

PR #2454, opened 2024-04-08
A3 blueprint improvement

tpdownes:
We have seen customers mistakenly skip the step to update the Filestore IP address. Removing it entirely from the blueprint will ensure that the create/deploy commands will fail. We are also moving its setting to higher in the file so it will be more prominent to the user.

### Submission Checklist


tpdownes:
The alternatives here are:

- this PR
- removing the value from the blueprint but hacking `tools/validate_configs/validate_configs.sh` to supply a value for server IP during make tests
- enforcing a strict validation for var.server_ip in the pre-existing-network-storage module 

The 3rd option is appealing but it turns out var.server_ip is allowed to take on more complex values that are not just IP addresses (e.g. lustre) so adding it could easily do more harm than good. For now, we make the requirement to update IP address more obvious in the rendered markdown.

rohitramu (approved):

PR #2457, opened 2024-04-08
Allow wait-for-startup to accept a list of instance names

rohitramu:
This PR adds an input `instance_names` (note: plural) to allow users to specify a list of instances to wait on.  This is particularly useful when deploying a cluster of machines using the `vm-instance` module.

Note that this is not a breaking change, because the `instance_name` (note: singular) input has not been removed, and works the same way as before.

I tested this change using the following blueprint:
```
blueprint_name: wait-for-startup-test

vars:
  deployment_name: wait-for-startup-test
  region: us-central1
  zone: us-central1-a

deployment_groups:
- group: primary
  modules:
  - id: network1
    source: modules/network/vpc

  - id: startup-sleep
    settings:
      # timeout: 20
      configure_ssh_host_patterns:
      - '*'
      runners:
        - type: shell
          destination: /tmp/sleep_test.sh
          content: |
            #!/bin/bash
            sleep 30
            # exit 42
  - id: workstation
    use:
    - network1
    - startup-sleep
    settings:
      instance_count: 3
      name_prefix: roramu-test
  - id: wait
    settings:
      instance_name: $(workstation.name[0])
      instance_names: $(workstation.name)
```

### DIFF
--- a/community/modules/scripts/wait-for-startup/main.tf
+++ b/community/modules/scripts/wait-for-startup/main.tf
@@ -13,8 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+locals {
+  joined_instance_names = concat(var.instance_names, [var.instance_name])
+}
+
 data "google_compute_instance" "vm_instance" {
-  name    = var.instance_name
+  count = length(local.joined_instance_names)
+
+  name    = local.joined_instance_names[count.index]
   zone    = var.zone
   project = var.project_id
 }
@@ -23,12 +30,10 @@ resource "null_resource" "wait_for_startup" {
   provisioner "local-exec" {
     command = "/bin/bash ${path.module}/scripts/wait-for-startup-status.sh"
     environment = {
-      INSTANCE_NAME = var.instance_name
-      ZONE          = var.zone
-      PROJECT_ID    = var.project_id
-      TIMEOUT       = var.timeout
+      INSTANCE_NAMES = join(" ", distinct(compact(data.google_compute_instance.vm_instance[*].name)))
+      ZONE           = var.zone
+      PROJECT_ID     = var.project_id
+      TIMEOUT        = var.timeout
     }
   }
-
-  triggers = { instance_id_changes = data.google_compute_instance.vm_instance.instance_id }
 }
--- a/community/modules/scripts/wait-for-startup/variables.tf
+++ b/community/modules/scripts/wait-for-startup/variables.tf
@@ -19,6 +19,16 @@ variable "instance_name" {
   type        = string
 }
 
+variable "instance_names" {
+  description = "Names of the instances we are waiting for"
+  type        = list(string)
+  default     = []
+  validation {
+    condition     = var.instance_names != null
+    error_message = "The list of instances cannot be null (empty list is ok)"
+  }
+}
+
 variable "zone" {
   description = "The GCP zone where the instance is running"
   type        = string
--- a/community/modules/scripts/wait-for-startup/README.md
+++ b/community/modules/scripts/wait-for-startup/README.md
@@ -69,6 +69,7 @@ No modules.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_instance_name"></a> [instance\_name](#input\_instance\_name) | Name of the instance we are waiting for | `string` | n/a | yes |
+| <a name="input_instance_names"></a> [instance\_names](#input\_instance\_names) | Names of the instances we are waiting for | `list(string)` | `[]` | no |
 | <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Project in which the HPC deployment will be created | `string` | n/a | yes |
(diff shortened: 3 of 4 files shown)


community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,22 @@
...
 resource "null_resource" "wait_for_startup" {
   provisioner "local-exec" {
     command = "/bin/bash ${path.module}/scripts/wait-for-startup-status.sh"
     environment = {
-      INSTANCE_NAME = var.instance_name
-      ZONE          = var.zone
-      PROJECT_ID    = var.project_id
-      TIMEOUT       = var.timeout
+      INSTANCE_NAMES = join(" ", distinct(compact(concat(var.instance_names, [var.instance_name]))))
+      ZONE           = var.zone
```
mr0re1, line 27: While it's not directly related to this change, 
but using `zone` and `project` as separate inputs can be error-prone.E.g. was created 
Better option would be to use self-link. 

#2460 adds `self_link` to the output.

If you intend to use `data "google_compute_instance" "vm_instance"`  you better make use of returned `self_link`.

mr0re1, line 27: Do we assume instances are in the same zone/project?

rohitramu: Yup, that's correct, since that assumption was held before this change.  I don't want to make any breaking changes in this update, but it's something we can certainly think about as a separate change.

rohitramu: I don't think we can use a "data" block with triggers because of the issues we found previously with modifying lists in Terraform (this is why we agreed to move all the complexity into the script).

Maybe in the future, we can add a variable for `self_links` so that `wait-for-startup` can "use" `vm-instance`.  Unfortunately, it seems like `gcloud`'s  `get-serial-output` subcommand doesn't have an argument for self_link, so we'd probably need to parse the zone and project out of the self_link.

mr0re1: > because of the issues we found previously with modifying lists in Terraform

I don't see why we need any list modifications here.

> Unfortunately, it seems like gcloud's get-serial-output subcommand doesn't have an argument for self_link

`$ gcloud compute instances get-serial-port-output <self_link>` - works like a charm

mr0re1: Simply speaking that would be an implementation:

```hcl
data  "google_compute_instance" "instance_by_link" {
  # for each `var.self_links`
}

data  "google_compute_instance" "instance_by_name" {
  # for each `concat(var.instance_names, [var.instance_name])`
}

resource "null_resource" "wait_for_startup" {
   INSTANCES_NAMES = join(" ", distinct(concat(instance_by_link[*].self_link,  instance_by_name[*].self_link)))
}

```

Do you see any problems here ?

mr0re1: > We can't use for_each

Sure, since it's a list, just use `count` instead.

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-data "google_compute_instance" "vm_instance" {
-  name    = var.instance_name
-  zone    = var.zone
-  project = var.project_id
-}
+# data "google_compute_instance" "vm_instance" {
```
mr0re1, line 16: Pls remove

community/modules/scripts/wait-for-startup/main.tf
```diff
@@ -13,22 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-data "google_compute_instance" "vm_instance" {
```
mr0re1, line 16: It would still be useful to have this data source to keep `triggers` around.

community/modules/scripts/wait-for-startup/variables.tf
```diff
@@ -38,3 +48,13 @@ variable "timeout" {
     error_message = "The timeout should be non-negative"
   }
 }
+
+# variable "instance_discovery_timeout" {
```
mr0re1, line 52: Remove?

rohitramu: Removed.

community/modules/scripts/wait-for-startup/main.tf
rohitramu: I understand why it was useful when we wanted to trigger 1 script per machine, but now 1 instance of the script handles all machines.  How would we use triggers in this case?

community/modules/scripts/wait-for-startup/main.tf
rohitramu: Summary from offline conversation:
1. We will try to merge this change since it provides immediate value to other projects.
2. We need to come up with a plan to take a list of "self_link" values as an input to this module, then remove "name", "project" and "zone" inputs to improve the experience when using zone-finding or multiple zones.

rohitramu:
Closing in favor of #2515

PR #2460, opened 2024-04-09
Add `self_link` output to `vm-instance` module

mr0re1:
* Add `self_link` output to `vm-instance` module;
* Update `name` description;
* Update SSH instruction to use short-form.

**Motivation:** to remove any guess-work about what `zone` and `project` given instance belongs to.


tpdownes (approved):
One nit: I might consider "id" as the instructions change rather than the full self_link.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#id

Equal (I believe) uniqueness without quite such a large output.

mr0re1:
>  I might consider "id" as the instructions change rather than the full self_link.

`id` doesn't work nicely with `gcloud compute ssh` .
`gcloud compute ssh <id>` requires to additionally supply zone and project.

tpdownes:
> > I might consider "id" as the instructions change rather than the full self_link.
> 
> `id` doesn't work nicely with `gcloud compute ssh` . `gcloud compute ssh <id>` requires to additionally supply zone and project.

I think you may be confusing "id" and "name" (admittedly our documentation is not quite 100% on this). This appears to work just fine:

```
gcloud compute ssh projects/toolkit-demo-zero-e913/zones/us-east4-c/instances/vm2-0
```

PR #2465, opened 2024-04-10
Go binary downloads must include patch version

nick-stroud:
PR assumes go version of 0 of none is specified.

OFE tries to install go version from go.mod. Version bumped from 1.20 to 1.21 in ghpc v1.29.0 and beyond [commit](https://github.com/GoogleCloudPlatform/hpc-toolkit/commit/c32b3d114c291f96991713ebbd54a8ce3f8ccdbe).

Go provided binary installs for 1.20 but not for 1.21. We must use 1.21.0 instead.

Tested:
- Manually deployed and verified go was properly installed and `make` of ghpc succeeds.

### Submission Checklist

### DIFF
--- a/community/front-end/ofe/infrastructure_files/gcs_bucket/webserver/startup.sh
+++ b/community/front-end/ofe/infrastructure_files/gcs_bucket/webserver/startup.sh
@@ -159,6 +159,10 @@ EOF
 # Install go version specified in go.mod file
 #
 GO_VERSION=$(awk '/^go/ {print $2}' "/opt/gcluster/hpc-toolkit/go.mod")
+# Check if the version string has only two numbers separated by dots
+if [[ $GO_VERSION =~ ^([0-9]+\.[0-9]+)$ ]]; then
+	GO_VERSION="$GO_VERSION.0" # Append .0 if missing patch version
+fi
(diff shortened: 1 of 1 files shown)


community/front-end/ofe/infrastructure_files/gcs_bucket/webserver/startup.sh
```diff
@@ -159,6 +159,10 @@ EOF
 # Install go version specified in go.mod file
 #
 GO_VERSION=$(awk '/^go/ {print $2}' "/opt/gcluster/hpc-toolkit/go.mod")
+# Check if the version string has only two numbers separated by dots
```
rohitramu, line 162: Is there any case where it might only have the major version (single integer) instead of both major and minor versions?

rohitramu (approved):
Question was just for my understanding - no concerns.

community/front-end/ofe/infrastructure_files/gcs_bucket/webserver/startup.sh
nick-stroud: Even if technically yes (which may not even be the case), I think no. The version is coming from our go.mod file. We would never not specify the minor version.

PR #2470, opened 2024-04-12
Ensure that Filestore cleanup always re-enables the Filestore API

tpdownes:
Use a bash trap function to ensure that the Filestore API is always enabled. It is safe to run this command even when the API has already been enabled (i.e. it returns code 0 and leaves the service enabled)
 
A build where all commands succeeded:

https://console.cloud.google.com/cloud-build/builds/f42c571b-861b-4e10-be8c-d0c8fe72fb74?project=508417052821&e=13803378&mods=monitoring_api_prod

A build where I introduced a deliberate failure:

https://console.cloud.google.com/cloud-build/builds/c94d8e90-3152-4665-bfa6-ee535a0f0488?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod

### Submission Checklist

### DIFF
--- a/tools/clean-filestore-limit.sh
+++ b/tools/clean-filestore-limit.sh
@@ -13,7 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# shellcheck disable=SC2317
+# known issue that shellcheck wrongly detects trap function as unreachable
+
 set -e -o pipefail
+
+function enable_filestore_api() {
+	status=$?
+	echo "Re-enabling Filestore API..."
+	gcloud services enable file.googleapis.com --project "${PROJECT_ID}"
+	exit "$status"
+}
 
 BUILD_ID=${BUILD_ID:-non-existent-build}
 PROJECT_ID=${PROJECT_ID:-$(gcloud config get-value project)}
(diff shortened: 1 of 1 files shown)


tools/clean-filestore-limit.sh
```diff
@@ -35,6 +44,7 @@ fi
 
 # See https://cloud.google.com/filestore/docs/troubleshooting#system_limit_for_internal_resources_has_been_reached_error_when_creating_an_instance
 echo "Disabling Filestore API..."
+trap enable_filestore_api EXIT
```
rohitramu, line 47: Should this go after the next line?  I.e. shouldn't we put the trap after disabling the API?

tools/clean-filestore-limit.sh
```diff
@@ -13,8 +13,17 @@
...
 
+# shellcheck disable=SC2317
+# known issue that shellcheck wrongly detects trap function as unreachable
+
 set -e -o pipefail
 
+function enable_filestore_api() {
+	echo "Re-enabling Filestore API..."
+	gcloud services enable file.googleapis.com --project "${PROJECT_ID}"
+	exit 0
```
cboneti, line 24: won't this eat up any error that happens in the script? 
Say deleting a filestore instance failed, then it calls the trap, but then the trap has an exit 0... 
What do you think?

tools/clean-filestore-limit.sh
tpdownes: No, you want the trap in place first. It's also harmless.

tools/clean-filestore-limit.sh
tpdownes: Saw that too. Addressed.

rohitramu (approved):

PR #2471, opened 2024-04-12
Rewrite Filestore cleanup script in Python

rohitramu:
This PR rewrites the logic in `tools/clean-filestore-limit.sh` in Python.  There have been multiple cleanup failures recently which could be avoided with better error handling.  This is easier to do in a language like Python than in Bash.


tools/clean-filestore-limit.py
```diff
@@ -0,0 +1,127 @@
...
+
+    print("There are active Cloud Build jobs. Skip clean up, may require re-run.")
+    exit(1)
+
+
+def delete_filestore_instances():
+    """Deletes Filestore instances"""
+    print("Getting list of filestore instances...")
+    result = subprocess.run([
+        "gcloud", "filestore", "instances", "list", "--project", PROJECT_ID
```
mr0re1, line 58: consider using `format=json` to simplify logic and make it robust to "default format" changes

tpdownes:
Closing in favor of scheduling changes to cleanup implemented in #2484

PR #2472, opened 2024-04-13
Update Slurm troubleshooting docs to clarify behavior on stockout

rohitramu:
Updating documentation to clarify Slurm behavior when there is a stockout while Slurm is trying to bring up compute nodes.

The behavior is consistent for both v5 and v6, for both `srun` and `sbatch (I have tested all combinations).

### DIFF
--- a/docs/slurm-troubleshooting.md
+++ b/docs/slurm-troubleshooting.md
@@ -96,17 +96,21 @@ The solution here is to [request more of the specified quota](#gcp-quotas),
 It may be that the zone the partition is deployed in has no remaining capacity to create the
 compute nodes required to run your submitted job.
 
-You can confirm this by SSHing into the `controller` VM and checking the
-`resume.log` file:
+Check the `resume.log` file for possible errors by SSH-ing into the controller VM and running the following:
 
 ```shell
-$ cat /var/log/slurm/resume.log
-... bulkInsert operation errors: VM_MIN_COUNT_NOT_REACHED ...
+sudo cat /var/log/slurm/resume.log
+```
+
+One example of an error message which appears in `resume.log` due to insufficient capacity is:
+
+```text
+bulkInsert operation errors: VM_MIN_COUNT_NOT_REACHED
 ```
 
 When this happens, the the output of `sacct` will show the job's status as `NODE_FAIL`.
 
-Jobs submitted via `srun` will not be retried, however jobs submitted via `sbatch` will be retried.
+Jobs submitted via `srun` will not be requeued, however jobs submitted via `sbatch` will be requeued.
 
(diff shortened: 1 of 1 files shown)


docs/slurm-troubleshooting.md
```diff
@@ -90,6 +91,23 @@ The solution here is to [request more of the specified quota](#gcp-quotas),
...
+#### Insufficient Capacity
+
+It may be that the zone the partition is deployed in has no remaining capacity to create the
+compute nodes required to run your submitted job.
+
+You can confirm this by SSHing into the `controller` VM and checking the
+`resume.log` file:
+
+```shell
+$ cat /var/log/slurm/resume.log
```
nick-stroud, line 103: I believe sudo is required to read this log, correct? If so lets include it here so people know that is expected.

docs/slurm-troubleshooting.md
```diff
@@ -90,6 +91,23 @@ The solution here is to [request more of the specified quota](#gcp-quotas),
...
+... bulkInsert operation errors: VM_MIN_COUNT_NOT_REACHED ...
```
nick-stroud, line 104: As mentioned elsewhere, this is one possible error. Maybe say to check resume.log for possible errors, and then add as a NOTE that `VM_MIN_COUNT_NOT_REACHED` can be seen due to insufficient capacity. Basically make the statement less explicit. I don't want someone seeing this error and thinking it must be capacity. On the other hand, they might see a different error to this one. It is still good to call out `VM_MIN_COUNT_NOT_REACHED` as this particular message is not very intuitive.

docs/slurm-troubleshooting.md
```diff
@@ -90,6 +91,23 @@ The solution here is to [request more of the specified quota](#gcp-quotas),
...
+```
+
+When this happens, the the output of `sacct` will show the job's status as `NODE_FAIL`.
+
+Jobs submitted via `srun` will not be retried, however jobs submitted via `sbatch` will be retried.
```
nick-stroud, line 109: s/retried/requeued/

nick-stroud (changes requested):

docs/slurm-troubleshooting.md
rohitramu: Good point - updated.

docs/slurm-troubleshooting.md
rohitramu: Makes sense, updated.

docs/slurm-troubleshooting.md
rohitramu: Fixed.

nick-stroud (approved):

PR #2485, opened 2024-04-17
Add docs about `ghpc_stage` and other functions

mr0re1:


examples/README.md
```diff
@@ -1829,3 +1828,34 @@ deployment_groups:
...
+    settings:
+      runners:
+      - type: shell
+        destination: hi.sh
+        source: $(ghpc_stage("path/relative/to/blueprint/hi.sh"))
+        # or stage the whole directory
+        source: $(ghpc_stage("path"))/hi.sh
+        # or use it as input to another function
+        content: $(file(ghpc_stage("path/hi.sh")))
+````
```
nick-stroud, line 1861: Please add (or something similar):

> The `ghpc_stage` function will always look first in the path specified in the blueprint. If the file is not found at this path then `ghpc_stage` will look for the staged file in the deployment folder, if a deployment folder exists. This means that you can redeploy a blueprint (`ghpc deploy <blueprint> -w`) so long as you have the deployment folder from the original deployment, even if locally referenced files are not available.

nick-stroud (changes requested):

PR #2486, opened 2024-04-17
Slurm6. Replace vars `disable_X -> enable_X`

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -120,7 +120,7 @@ module "slurm_files" {
   prolog_scripts           = var.prolog_scripts
   enable_slurm_gcp_plugins = var.enable_slurm_gcp_plugins
 
-  disable_default_mounts = var.disable_default_mounts
+  disable_default_mounts = !var.enable_default_mounts
```
harshthakkar01, line 123: Are we planning to update this as well  ? All the way to slurm-gcp ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables_controller_instance.tf
```diff
@@ -91,10 +101,20 @@ variable "can_ip_forward" {
...
+  description = "If set to true. The controller will have a random public IP assigned to it. Ignored if access_config is set."
   type        = bool
-  default     = true
+  default     = false
+}
+
+variable "disable_controller_public_ips" { # tflint-ignore: terraform_unused_declarations
+  description = "DEPRECATED: Use `enable_controller_public_ips` instead."
+  type        = bool
+  default     = null
```
harshthakkar01, line 113: what would happen if we explicitly mention disable_controller_public_ips: false in the blueprint since condition only checks if it is null or not. Would that pass ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
mr0re1: This is a part of SlurmGCP modules , we will address them when/if we will merge them with wrappers.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables_controller_instance.tf
mr0re1: It will fail if it mentioned , regardless of the supplied value:
* `false | true` - condition `== null` will fire;
* `null` - type checker will complain "can't use null here, it's bool"

harshthakkar01 (approved):

harshthakkar01:
Looks like some files are needed to be updated.

harshthakkar01:
After tests pass, gtg from my side.

harshthakkar01 (approved):

PR #2501, opened 2024-04-19
Remove extra 'v' in versions generated by `create-release-candidate.sh`

rohitramu:


tpdownes (changes requested):
Please change OLD_VERSION to OLD_TAG. I think the script as-is uses VERSION in two different senses that introduced the error being resolved

tpdownes (approved):

Issue #2054, opened 2023-12-18
private_vpc_connection is not unique per cluster in slurm-sql module


ek-nag:
### Describe the bug

`module.slurm-sql.google_service_networking_connection.private_vpc_connection` is not unique per cluster. When two clusters are created in the same VPC and subnet both of them relies on the same `google_service_networking_connection`. When one cluster is destroyed the `google_service_networking_connection` gets removed breaking `slurm-sql` operation for the remaining cluster.

```
resource "google_service_networking_connection" "private_vpc_connection" {
  provider = google-beta

  network                 = var.network_id
  service                 = "servicenetworking.googleapis.com"
  reserved_peering_ranges = google_compute_global_address.private_ip_address[*].name
}
```
The easiest solution would be adding unique `name` argument to the `google_service_networking_connection` resource same as we doing for `google_compute_global_address.private_ip_address`.

```
resource "google_compute_global_address" "private_ip_address" {
  provider = google-beta

  name          = "slurm-cloudsql-private-ip-${random_id.resource_name_suffix.hex}"
  purpose       = "VPC_PEERING"
  address_type  = "INTERNAL"
  prefix_length = 16
  network       = var.network_id
}
```
However, [module in Terraform provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/service_networking_connection) does not have `name` argument.

### Steps to reproduce

1. Deploy two cluster in the same VPC and subnet including that use `community/modules/database/slurm-cloudsql-federation` module. For example:
```
  - source: community/modules/database/slurm-cloudsql-federation
    kind: terraform
    id: slurm-sql
    use: [hpc_network]
    settings:
      sql_instance_name: sql-cluster-b62b0a5b
      tier: "db-g1-small"
```

### Expected behavior

Make `module.slurm-sql.google_service_networking_connection.private_vpc_connection` unique. 
### Actual behavior

If two or more clusters reside in the same VPC and subnet they all use same `module.slurm-sql.google_service_networking_connection.private_vpc_connection` and when one of the clusters is deleted the ``module.slurm-sql.google_service_networking_connection.private_vpc_connection` is also removed breaking remaining clusters access to cloudsql slurm accounting database.

### Version (`ghpc --version`)

`ghpc version v1.26.1`

### Blueprint

If applicable, attach or paste the blueprint YAML used to produce the bug.

```yaml
blueprint_name: cluster-b62b0a5b

vars:
[...]

PR #2512, opened 2024-04-23
OFE: new features and fixes.

ek-nag:
This PR add the following features and fixes to the OFE:
* Image import functionality, allows user to import existing image from the project to use when creating a cluster. It's handy when deploying multiple OFEs or migrating images from different projects.
* Extending a list of nodes that allowed to have compact placement policy when creating partition.
* Fixing broken static path for the /admin page.
* Integrating #2054
* Extending a list of required APIs in the deployment script.


community/front-end/ofe/website/ghpcfe/cluster_manager/clusterinfo.py
```diff
@@ -221,9 +222,14 @@ def _prepare_ghpc_partitions(self, part_uses):
         for (count, part) in enumerate(self.cluster.partitions.all()):
             part_id = f"partition_{count}"
             if part.image is not None:
-                instance_image_yaml = f"""instance_image:
-        family: image-{part.image.family}
-        project: {self.cluster.project_id}"""
+                if part.image.source_image_family == "Imported":
+                    instance_image_yaml = f"""instance_image:
+            family: {part.image.family}
```
nick-stroud, line 227: nit: This might be more readable if you pull the conditional outside of the string and just define the string once:

```
                imported_image_fam = part.image.family
                regular_image_fam = f"image-{part.image.family}"
                image_family = imported_image_fam if part.image.source_image_family == "Imported" else regular_image_fam
                instance_image_yaml = f"""instance_image:
        family:{image_family}
        project: {self.cluster.project_id}"""
```

Totally optional, feel free to ignore and close.

community/front-end/ofe/website/ghpcfe/cluster_manager/image_import.py
```diff
@@ -0,0 +1,84 @@
...
+    from google.cloud import compute_v1
+    client = compute_v1.ImagesClient(credentials=credentials)
+    request = compute_v1.ListImagesRequest(project=project)
+    image_response = client.list(request)
+    existing_image_list = []
+    for image in image_response.items:
+        if cred_id == False:
+            existing_image_list.append([image.name, image.description, image.family])
+        else:
+            existing_image_list.append([cred_id,image.name, image.description, image.family])
```
nick-stroud, line 56: I am confused how we get away with having the appended lists being different lengths. It seems like this would cause issues downstream.

community/front-end/ofe/website/ghpcfe/forms.py
```diff
@@ -307,7 +307,7 @@ def clean(self):
         cleaned_data = super().clean()
         if cleaned_data["enable_placement"] and cleaned_data[
             "machine_type"
-        ].split("-")[0] not in ["c2", "c2d", "c3"]:
+        ].split("-")[0] not in ["a2", "a3", "c2", "c3", "c2d", "c3d", "g2", "h3", "n2", "n2d"]:
```
nick-stroud, line 310: In Slurm we switched to checking for types that we know will not work with placement instead of checking that type is allowed. This means that new types are allowed by default and do not require immediate change to make it work. 

Example: https://github.com/GoogleCloudPlatform/slurm-gcp/blob/03805b039bc6d878ee371c1513e164a80db362b1/scripts/resume.py#L627

Something to consider.

nick-stroud (approved):

PR #2513, opened 2024-04-23
Fix Race condition and add validation for mount_runner in community/modules/compute/notebook/main.tf 

jrossthomson:
@nick-stroud 

Fixes bugs:

- community/modules/compute/notebook/variables.tf: Add validation for mount_runner
- Fix Race condition in community/modules/compute/notebook/main.tf


community/modules/compute/notebook/main.tf
```diff
@@ -48,21 +48,22 @@ resource "random_id" "resource_name_suffix" {
...
+  content = local.content5
+  bucket  = local.bucket
+}
+
 resource "google_notebooks_instance" "instance" {
   name                = local.name
   location            = var.zone
   machine_type        = var.machine_type
   project             = var.project_id
   post_startup_script = "${var.gcs_bucket_path}/${local.post_script_filename}"
```
nick-stroud, line 62: I think another, arguably better, way to accomplish what you are doing with `depends_on` would be change this line to be:

```
# old
post_startup_script = "${var.gcs_bucket_path}/${local.post_script_filename}"
# new
post_startup_script = "${var.gcs_bucket_path}/${google_storage_bucket_object.mount_script.name}"
```

`depends_on` tends to be the nuclear option and can have some undesirable behavior such as destroying resources when really they could just be updated in place.

LMK if you are opposed to the suggestion. I am not set in stone on this.

community/modules/compute/notebook/variables.tf
```diff
@@ -68,4 +68,9 @@ variable "gcs_bucket_path" {
 variable "mount_runner" {
   description = "mount content from the google-cloud-storage module"
   type        = map(string)
+
+  validation {
+    condition     = (length(split(" ", var.mount_runner.args)) == 5)
+    error_message = "There must be 4 elements in the Mount Runner Arguments: ${var.mount_runner.args} \n "
```
nick-stroud, line 74: You have ==5 above but 4 in error message.

community/modules/compute/notebook/main.tf
jrossthomson: I'm not against it and happy to do it. I just don't know of any way to test whether it actually solves to proposed problem. I know `depends_on` is designed to ensure the resource is there. 

If you are comfortable with the proposed change, I have nothing to base an objection upon!

jrossthomson:
Wow, lots of checks waiting to run. Is this normal?

mr0re1:
> Wow, lots of checks waiting to run. Is this normal?

It's ok, we schedule all checks up, but don't run all of them.

mr0re1:
AFAIK, there is no test dedicated to this module. Assuming you tested it manually, LGTM.

mr0re1 (approved):

PR #2516, opened 2024-04-23
Make batch job runnables a list instead of single string

aaronegolden:
This way toolkit users can deploy batch jobs with multiple runnables. This can be convenient, for example, when the user wants to install something wih an initial "apt install" or "yum install" before running the "real" workload.

### Submission Checklist


modules/scheduler/batch-job-template/variables.tf
```diff
@@ -91,10 +91,12 @@ variable "log_policy" {
   }
 }
 
-variable "runnable" {
```
nick-stroud, line 94: As written, this is a breaking change. I would suggest one of the following:
1. Breaking change with warning: If we are going to make a breaking change, then generally we would leave this variable, but put in the description that it is 'DEPRECATED' and add a validator that checks that input value is always null. This would allow us to provide an error message if the variable is used saying something like "such and such is deprecated, use such and such instead"
2. We could take both the plural and singular forms and enforce that only one is set. This would take a bit of tf code which I could help write.

I think if we end up making plural version more complicated (as I suggest in separate comment) then I would lean toward accepting both so the simple form stays simple.

modules/scheduler/batch-job-template/templates/batch-job-base.yaml.tftpl
```diff
@@ -5,8 +5,10 @@ taskGroups:
         - barrier:
             name: "wait-for-node-startup"
         %{~ endif ~}
+        %{~ for runnable in runnables ~}
```
nick-stroud, line 8: If synchronized is turned on, I think we should add a barrier between every runnable. WDYT?

modules/scheduler/batch-job-template/templates/batch-job-base.yaml.tftpl
```diff
@@ -5,8 +5,10 @@ taskGroups:
...
         - script:
-            text: ${indent(12, chomp(yamlencode(runnable)))}
+            text: ${indent(20, chomp(yamlencode(runnable)))}
```
nick-stroud, line 10: Interesting that indent changed from 12 to 20. Why is that?

modules/scheduler/batch-job-template/variables.tf
```diff
@@ -91,10 +91,12 @@ variable "log_policy" {
...
-  description = "A string to be executed as the main workload of the Google Cloud Batch job. This will be used to populate the generated template."
-  type        = string
-  default     = "## Add your workload here"
+variable "runnables" {
+  description = "A list of strings to be executed as a sequence of shell scripts as the main workload of the Google Cloud Batch job. These will be used to populate the generated template."
+  type        = list(string)
```
nick-stroud, line 96: In previous discussion we had maybe using `list(object(...))`.

This would allow us to specify additional parameters. In our notes we had:
```
- id: step 1
  script: echo 'hello world'
  execution_mode: ?
```

We could leave off execution_mode for now since we don't have that behavior settled, but it would leave the door open to it in the future without breaking changes again. WDYT?

Whatever we agree on, I can help to hone the tf code if you need help.

modules/scheduler/batch-job-template/templates/batch-job-base.yaml.tftpl
aaronegolden: Ah, you are right. Will fix.

aaronegolden: I changed this so we get a "wait-for-script-to-finish" (instead of wait-for-workload) after each script when synchronized=true. This is to not imply that each script is the whole "workload".

modules/scheduler/batch-job-template/templates/batch-job-base.yaml.tftpl
aaronegolden: I don't remember what exactly happened but at one point during work on this I got it into my head that I needed the extra indentation to make the formatting come out properly. I probably had the wrong tildes ~ around some of my ifs or fors. I changed it back to 12 now and toolkit still generates well-formed job specs.

modules/scheduler/batch-job-template/variables.tf
aaronegolden: Sure, that will be more flexible for future enhancements. I have updated it to type(object) and updated the default value and example files as well.

modules/scheduler/batch-job-template/variables.tf
aaronegolden: I agree, leaning towards option 2 assuming we will add additional configurations for runnables later, but some users will want the shorter/simpler form.

nick-stroud: SG. It looks like this still needs to be implemented. Right now runnable is still being deleted.

modules/scheduler/batch-job-template/variables.tf
nick-stroud: Is there anything useful that we can/need to do with id right now? Like can we put it somewhere in the job such that it would show up in pantheon or in logging. If it will be unused then I would remove it, but still go with the dict style `script: <string>`, to give us room to grow.

nick-stroud (approved):

nick-stroud:
Approved, waiting for batch tests to run.

PR #2523, opened 2024-04-26
Fix `cleanup_compute.sh` caused by quoting.

mr0re1:
```sh
$ gcloud compute instances delete --quiet  "${nodes}"
ERROR: (gcloud.compute.instances.delete) Could not fetch resource:
 - Invalid value for field 'instance': 'nsfs-controller https://www.googleapis.com/compute/v1/projects/X/zones/us-central1-a/instances/nsfs-debugnodeset-0 https://www.googleapis.com/compute/v1/projects/X/zones/us-central1-a/instances/nsfs-debugnodeset-1'. 
  Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}'

$ gcloud compute instances delete --quiet  ${nodes}  
# OK
```


tpdownes (changes requested):
Before merging add another comment indicating that the lack of quotes is intentional and causes each new space-separated "word" to be treated as independent arguments.

tpdownes (approved):

PR #2535, opened 2024-04-30
Roll TGP to v5.28.0

alyssa-sm:
This PR rolls TGP from v4.84 to v5.28 by updating references to versions and sets the default for `deletion_protection` since that is new in 5.0.

### DIFF
--- a/pkg/modulewriter/tfwriter.go
+++ b/pkg/modulewriter/tfwriter.go
@@ -205,8 +205,8 @@ func getProviders(bp config.Blueprint) []provider {
 	}
 
 	return []provider{
-		{"google", "hashicorp/google", "~> 4.84.0", gglConf},
-		{"google-beta", "hashicorp/google-beta", "~> 4.84.0", gglConf},
+		{"google", "hashicorp/google", ">= 4.84.0, <= 5.28.0", gglConf},
+		{"google-beta", "hashicorp/google-beta", ">= 4.84.0, <= 5.28.0", gglConf},
 	}
 }
 
--- a/community/modules/compute/gke-node-pool/README.md
+++ b/community/modules/compute/gke-node-pool/README.md
@@ -180,15 +180,15 @@ limitations under the License.
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.2 |
-| <a name="requirement_google"></a> [google](#requirement\_google) | >= 4.75.1, < 5.0 |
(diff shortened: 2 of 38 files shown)


examples/machine-learning/ml-slurm-a3-1-image.yaml
```diff
@@ -90,7 +90,7 @@ deployment_groups:
           set -e -o pipefail
           ansible-galaxy role install googlecloudplatform.google_cloud_ops_agents
           ansible-pull \
-              -U https://github.com/GoogleCloudPlatform/slurm-gcp -C 5.10.6 \
+              -U https://github.com/GoogleCloudPlatform/slurm-gcp -C 92627dd \
```
tpdownes, line 93: This should not change in this PR.

pkg/modulewriter/tfwriter.go
```diff
@@ -205,8 +205,8 @@ func getProviders(bp config.Blueprint) []provider {
 	}
 
 	return []provider{
-		{"google", "hashicorp/google", "~> 4.84.0", gglConf},
-		{"google-beta", "hashicorp/google-beta", "~> 4.84.0", gglConf},
+		{"google", "hashicorp/google", ">= 4.84.0, <= 5.28.0", gglConf},
```
tpdownes, line 208: I thought the plan was `~> 4.84.0, ~> 5.28.0`

tpdownes (changes requested):
The change on the A3/machine-learning blueprint is necessary but should not require re-running tests.

pkg/modulewriter/tfwriter.go
tpdownes: If we change this, it will require re-running PR-validation, but if that passes, it does not require the whole suite of tests to be re-run. That said, I don't think the difference in the change is huge.

alyssa-sm: Nick and I discussed this offline. `~> 4.84.0, ~> 5.28.0` would not be able to find a version since both cannot be satisfied at once. So we decided to do the range.

tpdownes: Playing around with it, I see the desired behavior best matches the choice here.

examples/machine-learning/ml-slurm-a3-1-image.yaml
alyssa-sm: Changed this one back to 5.10.6.

tpdownes (approved):

PR #2537, opened 2024-04-30
Slurm6. Fix and improve clean up compute

mr0re1:
* Fix order of removal: 1. instances; 2. resource policies;
* Wait until controller is destroyed before doing cleanup;
* Don't trigger on changes `var.enable_cleanup_compute`.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
```diff
@@ -12,20 +12,32 @@
...
+    command = self.triggers.cmd
     when    = destroy
   }
 
+  lifecycle {
+    # IMPORTANT: we make use of terraform bug https://github.com/hashicorp/terraform/issues/13549
+    # Without this `create_before_destroy`, the provisioner would be executed on any change to 
+    # `var.enable_cleanup_compute`, even (true -> false).
+    # With this `create_before_destroy` the provisioner will be only executed during `destroy`.
+    create_before_destroy = true
```
nick-stroud, line 38: Can you clarify what is the behavior in the following conditions:

1. enable_cleanup_compute (true -> false) + TF bug open
1. enable_cleanup_compute (true -> false) + TF bug fixed
1. enable_cleanup_compute (false -> true) + TF bug open
1. enable_cleanup_compute (false -> true) + TF bug fixed

Ideally none of these would do anything, since we only want to destroy nodes on destroy.

nick-stroud: I am wondering if it would just be better to bite the bullet instead of expecting behavior to change later.

PR #2541, opened 2024-05-01
Slurm6. Fix order of clean up.

mr0re1:
No other changes, just swapping places of two blocks of code

Correct order:

1. Delete instances;
2. Delete resource policies.


tpdownes (approved):
I think there is a minor opportunity in a follow-up PR. I suspect we are calling delete agains the same set of nodes multiple times because they enter the [STOPPING state](https://cloud.google.com/compute/docs/instances/instance-life-cycle) before being deleted. Suggest:

1. A new first loop that runs delete against all nodes NOT in STOPPING state to avoid duplicated instance delete calls. Repeat until empty.
2. Retain the current loop to handle any edge cases in why the VM was STOPPING (maybe the user happened to execute `poweroff` from the shell)

As another PR, there are probably some tricks we could play to be 100% certain that the resource policies aren't being used before being deleted. But this is a significant change as it stands.

PR #2545, opened 2024-05-02
Perfrom import of inputs during `destroy`

mr0re1:


tpdownes (commented):
It seems to me that this PR should be combined with changes to walk the graph and export the outputs of each dependent group first.

In the real world, we tend to hit failures in this function when the deployment directory is freshly created. In this scenario, the likelihood of actually importing the inputs is low and we'll just progress to the same error we'd get otherwise.

Do you agree with this analysis?

mr0re1:
> It seems to me that this PR should be combined with changes to walk the graph and export the outputs of each dependent group first.
In the real world, we tend to hit failures in this function when the deployment directory is freshly created. In this scenario, the likelihood of actually importing the inputs is low and we'll just progress to the same error we'd get otherwise.
Do you agree with this analysis?

While this PR itself doesn't resolve any practical issues, it is a forerunner for another change that makes running `export/import` optional:

* fetch outputs from state if `exported` files is not present;
* don't perform `export` on groups that were skipped (`(C)ontinue without applying`).

I could bundle them all together but, I would prefer given them as small as possible. Any concerns ?

tpdownes:
> * don't perform `export` on groups that were skipped (`(C)ontinue without applying`).

I'm not sure about this bullet but I'd want to think about it. If this is a small step in a larger walk, I'm happy to merge with the understanding we're not yet solving most "real" problems.

tpdownes (approved):

PR #2546, opened 2024-05-02
Add utility `identity` module

mr0re1:
To use as:
* "poor man's"  `locals`;
* placeholder for testing.

### DIFF
--- a/community/modules/utils/identity/README.md
+++ b/community/modules/utils/identity/README.md
@@ -1,0 +1,36 @@
+## Description
+
+This module does not do anything, it just returns the value passed to it.
+It creates no resources and uses no providers.
+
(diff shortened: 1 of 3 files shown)


community/modules/utils/identity/README.md
```diff
@@ -0,0 +1,36 @@
+## Description
+
+This module does not do anything, it just returns the value passed to it.
+It creates no resources and uses no providers.
```
nick-stroud, line 4: Since we will not use this in example blueprints, please document what it is to be used for. What would keep me from deleting this as dead code as soon as it is submitted.

nick-stroud (changes requested):

PR #2552, opened 2024-05-03
Prevent multiple instances of these tests because will fail due to static network name & multi-group fails to cleanup

nick-stroud:
Until we can roll back #2530, these tests will fail and fail to cleanup if multiple instances are running at the same time.

Manually tested by submitting multiple runs.

### Submission Checklist

### DIFF
--- a/tools/cloud-build/daily-tests/builds/hcls-v6.yaml
+++ b/tools/cloud-build/daily-tests/builds/hcls-v6.yaml
@@ -34,6 +34,22 @@ tags:
 
 timeout: 14400s  # 4hr
 steps:
+# While using static network names we are gaurding against more than 1 instance running at a time (for multi-group tests)
+- id: check_for_running_build
+  name: gcr.io/cloud-builders/gcloud
+  entrypoint: /bin/bash
+  args:
+  - -c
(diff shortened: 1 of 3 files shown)


tools/cloud-build/daily-tests/builds/hcls-v6.yaml
```diff
@@ -34,6 +34,22 @@ tags:
 
 timeout: 14400s  # 4hr
 steps:
+# While using static network names we are gaurding against more than 1 instance running at a time (for multi-group tests)
+- id: check_for_running_build
+  name: gcr.io/cloud-builders/gcloud
+  entrypoint: /bin/bash
+  args:
```
mr0re1, line 41: nit. Let's move it into a shell script, do avoid copy pasting it.

mr0re1 (approved):

tools/cloud-build/daily-tests/builds/hcls-v6.yaml
nick-stroud: Good idea, will do in follow on PR just to get this in ASAP.

PR #2562, opened 2024-05-08
New output for private-service-access

cdunbar13:
When trying to use the public-service-access module with a filestore module, I ran into a dependency issue.  When a filestore module specifies a `connect_mode: PRIVATE_SERVICE_ACCESS` and that PSA is created in the same blueprint and group, it will likely throw an error, as the PSA won't necessarily be created before the filestore begins to be deployed.  Filestore seems to be able to figure out which PSA to use depending on the VPC passed to it, so it does not take any information from the Terraform module.

This is a perfect example of why the toolkit needs a `depends_on` field in the blueprints so we can manually specify upstream dependencies.

Barring that, a quick solution was to output the `network_id` so that the filestore module could take the psa module's outputs creating the dependency that is needed.

Update: I have moved away from using the output from psa.  Now filestore takes in the output from psa and checks if it exists and uses that to set the `connect_mode: PRIVATE_SERVICE_ACCESS`.  This creates the correct dependency.

This has been tested on hpc-enterprise-slurm.yaml by adding a psa, and passed it into the projectfs filestore.  The filestore uses the correct connection mode, and doesn't interrupt the normal behavior.

### DIFF
--- a/community/modules/network/private-service-access/README.md
+++ b/community/modules/network/private-service-access/README.md
@@ -81,6 +81,5 @@ No modules.
 
 | Name | Description |
 |------|-------------|
-| <a name="output_network_id"></a> [network\_id](#output\_network\_id) | The id of the network that the private-service-access has been set up on. |
 | <a name="output_private_vpc_connection_peering"></a> [private\_vpc\_connection\_peering](#output\_private\_vpc\_connection\_peering) | The name of the VPC Network peering connection that was created by the service provider. |
 <!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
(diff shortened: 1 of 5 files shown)


mr0re1:
This approach creates a problem with `uses` on filestore module. There is no explicit expectation that user needs to specify PSA module instead of VPC module in filestore `uses` list.
 
Let's consider an alternative:
Add an input `optional(string) private_vpc_connection_peering` to the filestore module.
If `connect_mode == PRIVATE_SERVICE_ACCESS` that input has to be set and will be used in `depends_on` with filestore resource.

mr0re1 (approved):

PR #2564, opened 2024-05-09
Explicitly set default emails for login/nodeset service account

alyssa-sm:
This PR explicitly sets a default email for login/nodeset service account.

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
@@ -20,8 +20,10 @@
 locals {
   name = substr(replace(var.name, "/[^a-z0-9]/", ""), 0, 14)
 
+  service_account_email = coalesce(var.service_account_email, data.google_compute_default_service_account.default.email)
+
   service_account = {
-    email  = var.service_account_email
+    email  = local.service_account_email
     scopes = var.service_account_scopes
   }
 
@@ -49,4 +51,8 @@ locals {
     reserved        = var.reserved
     network_storage = var.network_storage
   }
+}
+
+data "google_compute_default_service_account" "default" {
+  project = var.project_id
 }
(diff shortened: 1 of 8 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
```diff
@@ -50,3 +52,6 @@ locals {
     network_storage = var.network_storage
   }
 }
+
+data "google_compute_default_service_account" "default" {
```
mr0re1, line 56: We got to specify `project_id`, please add it as a variable to this module.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
```diff
@@ -96,3 +98,6 @@ locals {
     network_storage   = var.network_storage
   }
 }
+
+data "google_compute_default_service_account" "default" {
```
mr0re1, line 102: same

mr0re1:
That will do, I'm OK to LGTM it after other comments are addressed.
But, I would strongly consider an alternative solution: fix it in `slurm_instance_template/main.tf`, two reasons:
* there is clearly a bug in `slurm_instance_template`;
* avoid bloating terraform graph with more nodes.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu/main.tf
alyssa-sm: Added.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
alyssa-sm: Added as well.

mr0re1 (approved):

PR #2569, opened 2024-05-10
Update to show example of the private-service-access in hpc-enterprise-slurm

cdunbar13:
Adds the `private-service-access` module to the hpc-enterprise-slurm blueprints and updates the `projectsfs` module to use PSA as an example.

### DIFF
--- a/examples/hpc-enterprise-slurm.yaml
+++ b/examples/hpc-enterprise-slurm.yaml
@@ -57,6 +57,10 @@ deployment_groups:
   - id: network1
     source: modules/network/pre-existing-vpc
 
+  - id: private-service-access
+    source: community/modules/network/private-service-access
+    use: [network1]
+
   - id: controller_sa
     source: community/modules/project/service-account
     settings:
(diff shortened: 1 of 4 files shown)


tpdownes (approved):
The integration test configuration needs to be updated to include the tag for the new modules. Otherwise, LGTM. Feel free to squash that.

examples/hpc-enterprise-slurm.yaml
```diff
@@ -95,9 +99,10 @@ deployment_groups:
 
   - id: projectsfs
     source: modules/file-system/filestore
-    use: [network1]
+    use: [network1, private-service-access]
     settings:
       local_mount: /projects
+      connect_mode: PRIVATE_SERVICE_ACCESS
```
nick-stroud, line 105: I am late to the party, but was it considered that this variable could be passed via use?

cdunbar13: I don't think both values for that variable could be passed via `use`.  You are semi-correct because in the blueprint as it stands `connect_mode: PRIVATE_SERVICE_ACCESS` is redundant because by passing the psa module into the filestore it will automatically set the connect mode in the module.  I left it in so that it was apparent to users what was happening.

That being said, if a user has a PSA already set up in their account you still need `connect_mode` to allow both options if no psa module is in the blueprint with the filestore module.

PR #2583, opened 2024-05-13
Add input type test for `additional_networks`

mr0re1:


tpdownes (approved):

tpdownes:
I think there is some user experience here that we need to look at it for adoption `optional` but having this integration test will ensure we do that uniformly in the Toolkit.

There is not yet a way to detect the additional_networks field within the `node_groups` parameter to the partition module?

mr0re1:
> There is not yet a way to detect the additional_networks field within the node_groups parameter to the partition module?

Sure we can do that

> Is there a way to treat network_interfaces (vm-instance) as identical in structure?

Sure, we can do that as well

PR #2587, opened 2024-05-14
Don't destroy VPCs before `cleanup_compute` is done

mr0re1:
Expand networks that cleanup compute depends on so that if nodesets are using additional networks, they wait to be deleted until cleanup compute has run.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
@@ -12,6 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+locals {
+  cleanup_dependencies_agg = flatten([
+    [
+      for ns in var.nodeset : [
+        ns.subnetwork_self_link,
+        [for an in ns.additional_networks : an.subnetwork]
+      ]
+    ],
+    [for ns in var.nodeset_tpu : ns.subnetwork],
+  ])
+}
+
+resource "null_resource" "cleanup_compute_depenencies" {
+  count = length(local.cleanup_dependencies_agg)
+}
 
 resource "null_resource" "cleanup_compute" {
   count = var.enable_cleanup_compute ? 1 : 0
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
```diff
@@ -26,10 +41,5 @@ resource "null_resource" "cleanup_compute" {
     when    = destroy
   }
 
-  depends_on = [
-    # Depend on controller network, as a best effort to avoid
-    # subnetwork resourceInUseByAnotherResource error
-    # NOTE: Can not use nodeset subnetworks as "A static list expression is required"
-    var.subnetwork_self_link,
-  ]
+  depends_on = [null_resource.cleanup_compute_depenencies]
```
nick-stroud, line 44: Are we sure this does not instigate inopportune running of the cleanup script? For example, could adding a nodeset and redeploying cause cleanup to run now that cleanup is dependent on that new nodeset network.

nick-stroud, line 44: Terraform docs: _We recommend always including a comment that explains why using depends_on is necessary._.
Please add a comment.

nick-stroud (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
```diff
@@ -12,6 +12,21 @@
...
+        [for an in ns.additional_networks : an.subnetwork]
+      ]
+    ],
+    [for ns in var.nodeset_tpu : ns.subnetwork],
+  ])
+}
+
+resource "null_resource" "cleanup_compute_depenencies" {
+  count = length(local.cleanup_dependencies_agg)
+}
```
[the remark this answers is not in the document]
mr0re1: > Should it also include var.subnetwork_self_link? 

I would not:
* If this hack doesn't work , we're in trouble with or without `var.subnetwork_self_link` in it.
* Adds implication that it somehow related with controllers VPC.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/cleanup.tf
mr0re1: Tested adding & removing nodesets, the clean up  wasn't triggered

PR #2598, opened 2024-05-16
OFE: Add support for Shared VPCs.

LasseHjorth:
This PR is a continuation of [#2597 2597] - but specific for OFE and this PR is also dependent of this.

The following changes explained:

* Allow Importing existing VPC to discover shared vpc and subnets.
* Fixed so page doesn't error out if there is no networks/subnets to import.
* Import existing VPC with full self_link names as cloud_ids.
* Moved network config to own blueprint jinja template. 
* Removed private-service-access - this will break creating multiple clusters on the same VPC network, and belongs in community/front-end/ofe/infrastructure_files/vpc_tf/GCP.
* Modified create filestore filesystem to work with Shared VPCs. (connect mode).
* Fixed bug in create filestore filesystem, where ghpc was using machine identity and not the identity selected in frontend, to build TF code.
* Fixed bug in VPCDetailedView  where any network would not be displayed if an ongoing cluster config is missing subnet.


sarthakag:
Hi @LasseHjorth, checking in on this PR. 
Since it has been inactive for several months and has some unresolved conflicts/comments, we are planning to close it to keep our backlog tidy.
If you are still working on this, please let us know! Otherwise, we’ll close it for now, but we'd be happy to reopen it later if you return to it.

PR #2599, opened 2024-05-16
Make Netapp Volumes work with pre-existing nfs shares

LasseHjorth:
When using Netapp Volumes as existing mount point world wide execute permissions are missing.

### DIFF
--- a/community/modules/file-system/cloud-storage-bucket/scripts/mount.sh
+++ b/community/modules/file-system/cloud-storage-bucket/scripts/mount.sh
@@ -56,3 +56,6 @@ fi
 echo "Mounting --target ${LOCAL_MOUNT} from fstab"
 mkdir -p "${LOCAL_MOUNT}"
 mount --target "${LOCAL_MOUNT}"
+
+# Support for Netapp Volumes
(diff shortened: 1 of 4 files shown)


community/modules/file-system/cloud-storage-bucket/scripts/mount.sh
```diff
@@ -56,3 +56,6 @@ fi
 echo "Mounting --target ${LOCAL_MOUNT} from fstab"
 mkdir -p "${LOCAL_MOUNT}"
 mount --target "${LOCAL_MOUNT}"
+
+# Support for Netapp Volumes
+chmod a+x "${LOCAL_MOUNT}"
```
mr0re1, line 61: Could you please provide more details about necessity of such permissions?
Would `chmod ug` / `chmod u` suffice?

Thank you.

LasseHjorth: @mr0re1 for some reason, when using filestore this permission is already there. 

When mounting a netapp Volume it's not. When mounting /home i.e from netapp, it causes that users cannot get into their homedir. When i did the comparison, it was the only permission missing afair - hence the result.

PR #2606, opened 2024-05-18
feat: Update database version variable in slurm-cloudsql-federation module

tfhartmann:
Adding the ability to pick between database versions in the `slurm-cloudsql-federation` module, defaults to MYSQL_5_7 so as not to break backwards compatibility.

### Submission Checklist

### DIFF
--- a/community/modules/database/slurm-cloudsql-federation/README.md
+++ b/community/modules/database/slurm-cloudsql-federation/README.md
@@ -73,6 +73,7 @@ No modules.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
+| <a name="input_database_version"></a> [database\_version](#input\_database\_version) | The version of the database to be created. | `string` | `"MYSQL_5_7"` | no |
 | <a name="input_deletion_protection"></a> [deletion\_protection](#input\_deletion\_protection) | Whether or not to allow Terraform to destroy the instance. | `string` | `false` | no |
 | <a name="input_deployment_name"></a> [deployment\_name](#input\_deployment\_name) | The name of the current deployment | `string` | n/a | yes |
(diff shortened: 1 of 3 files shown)


nick-stroud (approved):

nick-stroud:
@tfhartmann I have gone ahead and merged the PR as it is strictly an improvement over the status quo and I don't want to hold up the change.

I did have one question: What is the basis for `MYSQL_5_7 or MYSQL_8_0`? There are many other supported values for [this field](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#database_version). Do we know that only these two versions work? Have we only tested these two versions but others may or may not work?

I could imaging instead saying in the description that we recommend MYSQL_5_7 or MYSQL_8_0 as they have been tested, then linking to the field in terraform docs and not validating the value.

tfhartmann:
> I did have one question: What is the basis for `MYSQL_5_7 or MYSQL_8_0`? There are many other supported values for [this field](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#database_version). Do we know that only these two versions work? Have we only tested these two versions but others may or may not work?

Good question! I limited it because it's my understanding the Slurm only officially [supports](https://slurm.schedmd.com/platforms.html#database) MySQL and MariaDB   normally i'd have added more options, but when I read that page, I decided to limit it.  If other folks are running Slurm happily using some of the other options, than by all means, lets add em in!

tfhartmann:
> I could imaging instead saying in the description that we recommend MYSQL_5_7 or MYSQL_8_0 as they have been tested, then linking to the field in terraform docs and not validating the value.

Yeah, absolutely, maybe bump the default to MYSQL_8_0 and allow folks the option!

PR #2615, opened 2024-05-21
NFS export options for Filestore

wiktorn:
Add NFS export options for Filestore that are necessary for clusters running in non-RFC1918 network ranges.


modules/file-system/filestore/variables.tf
```diff
@@ -114,6 +114,16 @@ variable "connect_mode" {
   }
 }
 
+variable "nfs_export_options" {
+  description = "Define NFS export options."
+  type = object({
```
mr0re1, line 119: `nfs_export_options` block can be repeated multiple times (up to 10), let's make it `list` of objects instead.

wiktorn: Made it a non-nullable list.

mr0re1 (approved):

mr0re1:
Thank you for contribution!

PR #2653, opened 2024-05-31
Update slurm-gcp module to use custom endpoints.

cdunbar13:
As part of the process of decoupling ghpc from the GCP prod environment, this PR adds two new variables to the slurm-gcp v6 set of modules.  The two values are intended to go into the config.yaml that slurm-gcp creates and allow users to access other API endpoints.

Tested by adding:

```
      universe_domain: test.com
      custom_endpoints:
        compute: test.blah
```

To a controller's settings and verified by checking `/slurm/scripts/config.yaml` to ensure that the contents were transferred successfully.

This PR is waiting for a new release of Slurm-GCP which is required for these changes.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
```diff
@@ -17,12 +17,17 @@ set -e -o pipefail
...
 cluster_name="$2"
+gcloud_bin="$3"
 
 if [[ -z "${project}" || -z "${cluster_name}" ]]; then
-	echo "Usage: $0 <project> <cluster_name>"
+	echo "Usage: $0 <project> <cluster_name> [<gcloud location>]"
 	exit 1
 fi
 
+if [[ -z "${gcloud_bin}" ]]; then
```
mr0re1, line 27: I wouldn't use "optional" positional arguments, it's error prone, let's just make all 3 arguments required

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
```diff
@@ -612,3 +612,31 @@ EOD
...
+    compute = null
+  }
+}
+
+variable "gcloud_executable" {
+  description = "Location of the gcloud executable to be used during cleanup"
+  type        = string
+  default     = "gcloud"
+
+  validation {
```
mr0re1, line 638: I wouldn't make an assumption that the machine that does deployment is the same that would do destroy.
Also `enable_cleanup` can be disable, so this check is excessive. Let's remove it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
```diff
@@ -612,3 +612,31 @@ EOD
...
+  description = "Domain address for alternate API universe"
+  type        = string
+  default     = "googleapis.com"
+  nullable    = false
+}
+
+variable "endpoint_versions" {
+  description = "Version of the API to use (The compute service is the only API currently supported)"
+  type = object({
+    compute = string
```
mr0re1, line 626: nit.
```hcl
compute = optional(string)
```

mr0re1 (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
cdunbar13: It's no longer optional, but it does allow for null inputs.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
cdunbar13: I agree with this, and I think my latest changes address this so that most users will never need to use the added `gcloud_override_path`, but this new change just adds to the path, so it should still search correctly.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
cdunbar13: I don't want this to be optional, and set it to default to `beta` which is what slurm-gcp was using by default for compute.

mr0re1 (approved):

PR #2659, opened 2024-06-03
OFE: new features

ek-nag:
This PR adds the following features to the OFE:
* Allows user to enable Tier 1 networking for nodes in the cluster partition.
* Allows user to specify nodes reservation name for nodes used in cluster partition.
* Adding support for C4 type nodes.


community/front-end/ofe/website/ghpcfe/cluster_manager/cloud_info.py
```diff
@@ -359,9 +360,11 @@ def get_cpu_price(num_cores, instance_type, skus):
             "n2d": "N2D AMD Instance Core",
             "h3": "Compute optimized Core",
             "c3": "Compute optimized Core",
+            "c4": "Compute optimized Core",
             "c2": "Compute optimized Core",
             "c2d": "C2D AMD Instance Core",
             "c3d": "C3D AMD Instance Core",
+            "c4": "C4 Intel Emerald Rapids",
```
nick-stroud, line 367: Should `c4` be in this list twice?

community/front-end/ofe/website/ghpcfe/models.py
```diff
@@ -937,9 +944,9 @@ class ClusterPartition(models.Model):
         default="pd-standard",
     )
     boot_disk_size = models.PositiveIntegerField(
-        validators=[MinValueValidator(49)],
+        validators=[MinValueValidator(120)],
         help_text="Boot disk size (in GB)",
-        default=50,
+        default=120,
```
nick-stroud, line 949: What is the reason for changing the default? What is the reason for raising min validator above 50. I believe 50 gb worked previously. Whatever min value is set here forces all compute nodes to adopt that as a minimum. In a large cluster this could be expensive, especially if the on disk storage is not needed.

community/front-end/ofe/website/ghpcfe/views/clusters.py
```diff
@@ -470,6 +470,27 @@ def form_valid(self, form):
...
+                        cpu_count_str = part.machine_type.split('-')[-1]
+                        lssd_cpu_count_str = part.machine_type.split('-')[-2]
+                        if cpu_count_str.isdigit():
+                            cpu_count = int(cpu_count_str)
+                            logger.info(f"CPU count: {cpu_count}, VM type: {part.machine_type}.")
+                        elif lssd_cpu_count_str.isdigit():
+                            cpu_count = int(lssd_cpu_count_str)
+                            logger.info(f"CPU count: {cpu_count}, VM type: {part.machine_type}.")
+                        else:
+                            raise ValidationError(f"Cannot extract vCPU count from machine type {part.machine_type}")
```
nick-stroud, line 483: I fear this might be overly restrictive. What about `e2-medium` or `a2-ultragpu-2g`. Is it possible to make this a warning instead of an error? Maybe only perform validation when extract is successful. 

PS, I believe the vcpu count is accessible  via API. aka `gcloud compute machine-types list`. I am not suggesting to use this gcloud command, just that I suspect there is some equivalent in the python client.

community/front-end/ofe/website/ghpcfe/cluster_manager/cloud_info.py
ek-nag: fixed

community/front-end/ofe/website/ghpcfe/models.py
ek-nag: Yep, that make sense. It was misunderstood to when developer was asked to increase disks size, mostly the ask was to increase disk size on the login nodes. Changed back to 50 gb on compute nodes

community/front-end/ofe/website/ghpcfe/views/clusters.py
ek-nag: Updated the logic, it was slightly over engineered. vcpu count extracting from API is on our tasks list, we are working on better way of dynamically pulling such information.

nick-stroud (dismissed):

nick-stroud:
I need to do some git history editing avoid including updates from develop.

nick-stroud:
Closed in favor of #2700 which has identical content minus the merge back from develop.

PR #2662, opened 2024-06-04
Update GKE ML blueprint to use native GKE driver

ankitkinra:
Updated blueprint to remove the custom installation of the GPU driver and instead use the GKE provided setting `gpu_driver_version` . Detailed list of changes

* Used the native config `gpu_driver_version`  present in `gpu_driver_installation_config` added in https://github.com/hashicorp/terraform-provider-google/issues/15151
* Removed `install-nvidia-drivers` which did the custom installation of drivers
* Removed dependency on `kubernetes-operations`


community/examples/ml-gke.yaml
```diff
@@ -48,18 +48,18 @@ deployment_groups:
         cidr_block: $(vars.authorized_cidr)
     outputs: [instructions]
 
-  # Docs at https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/community/modules/scripts/kubernetes-operations
```
nick-stroud, line 51: I believe our documentation in examples/README.md may refer to this module, if so we should update that documentation.

nick-stroud, line 51: Are there any other examples using the Kubernets-operations module. If not we should probably delete all of the module documentation in our Repo. This would be under community/modules/kubernetes-operations/ and also from modules/README.md

nick-stroud (changes requested):

examples/README.md
```diff
@@ -1382,10 +1382,10 @@ Toolkit. It includes:
 > work. See note below.
 
 * Creation of a regional GKE cluster.
-* Creation of an autoscaling GKE node pool with `a2` machines each with 8
-  attached A100 GPUs.
-* Configuration of the cluster using the [`kubernetes-operations`] module to
-  install nvidia drivers.
+* Creation of an autoscaling GKE node pool with `g2` machines each with 1
+  attached L4 GPUs. Note: We tested this blueprint with `a2` machines as well,
+  but as capacity is hard to find we are using `g2` for example at the moment.
```
nick-stroud, line 1387: nit: I would reword to remove the word "we". AKA, "this blueprint was tested..."

community/examples/ml-gke.yaml
```diff
@@ -19,6 +19,7 @@ vars:
   project_id:  ## Set GCP Project ID Here ##
   deployment_name: ml-01
   region: us-central1
+  zones: [us-central1-c]
```
nick-stroud, line 22: Did you mean for this to make it in the final version, or is it an artifact from testing?

nick-stroud (changes requested):

tpdownes (commented):
This comment was stuck pending. I will see if it still applies with the changes.

community/examples/ml-gke.yaml
ankitkinra: Updated

ankitkinra: Deleted references to `kubernetes-operations` module, there is only one reference remaining which is around an output variable which is required by the `kubernetes-operations`

Remaining lines which I plan to clean up in a separate CL unless we should do this together as well:
*  https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/2af3c420f05970edb11f755af6556fde500b77d9/community/modules/scheduler/gke-cluster/README.md?plain=1#L159
* https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/2af3c420f05970edb11f755af6556fde500b77d9/community/modules/scheduler/gke-cluster/outputs.tf#L23

community/examples/ml-gke.yaml
```diff
@@ -48,18 +48,18 @@ deployment_groups:
...
-    use: [gke_cluster]
-    settings:
-      install_nvidia_driver: true
-
   - id: a2-pool
     source: community/modules/compute/gke-node-pool
     use: [gke_cluster]
     settings:
       machine_type: a2-highgpu-8g
+      guest_accelerator:
```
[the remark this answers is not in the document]
ankitkinra: Should not apply anymore but I have added an example setting with a2 machines which has been tested once (but could not run job as I was not able to get capacity)

community/examples/ml-gke.yaml
ankitkinra: I can remove this as this was a change from a2 , when I was not able to get capacity , for g2 I probably don't need this .

ankitkinra: Turns out I was wrong, `nvidia-l4` is not available in all zones so hardcoded to `us-central1-c` where it is present. Will add a comment to blueprint to record this reason.

community/examples/ml-gke.yaml
```diff
@@ -19,6 +19,8 @@ vars:
   project_id:  ## Set GCP Project ID Here ##
   deployment_name: ml-01
   region: us-central1
```
tpdownes, line 21: I suggest considering asia-southeast1-b in asia-southeast1 as a region with more capacity.

examples/README.md
```diff
@@ -1382,10 +1382,28 @@ Toolkit. It includes:
 > work. See note below.
 
 * Creation of a regional GKE cluster.
-* Creation of an autoscaling GKE node pool with `a2` machines each with 8
-  attached A100 GPUs.
-* Configuration of the cluster using the [`kubernetes-operations`] module to
-  install nvidia drivers.
+* Creation of an autoscaling GKE node pool with `g2` machines each with 1
+  attached L4 GPUs. Note: This blueprint was tested this blueprint with `a2` machines
+  as well, but as capacity is hard to find we are using `g2` for example at the moment.
```
nick-stroud, line 1387: Minor grammar errors, I recommend the following corrected version:

 Note: This blueprint has also been tested with `a2` machines, but as capacity is hard to find the example uses `g2` machines which have better obtainability. If using with `a2` machines it is recommended to first obtain an automatic reservation.

nick-stroud (approved):

PR #2674, opened 2024-06-10
Documentation and example clound function for shared vpcs.

LasseHjorth:
Documentation and example clound function for shared vpcs.

### DIFF
--- a/community/other/cloud-function-for-shared-vpcs/main.tf
+++ b/community/other/cloud-function-for-shared-vpcs/main.tf
@@ -1,0 +1,174 @@
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


community/other/cloud-function-for-shared-vpcs/main.tf
nick-stroud, line 1: If I understand correctly this terraform would essentially be a pre-step that would set up the shared vpc and cloud function to give appropriate permissions before deploying a cluster in a service project using a blueprint.

This submission would be the first of its kind. The HPC Toolkit repo does not host terraform that is not Toolkit compatible. It seems to me the way this terraform is written is meant to be a self contained deployment (aka no inputs/outputs and not modular). While the example is informative, I believe it strays from the purpose of the Toolkit to build composable modules.

Several ideas on a path forward.
1. Modularize the code and create a blueprint that could be used with the Toolkit. This would be a non trival amount of work but would reuse existing modules and produce new modules that could be reused in future.

  - network/vpc module - exists
  - project/iam-custom-role module - needs development
  - project/service-account module - exists
  - pubsub/topic module - exists but expand to include log sink functionality 
  - file-system/cloud-storage-bucket module - exists
  - cloud-function/serviceaccount-audit-logs-watcher module - needs development

Possible alternative breakdown could be:

  - network/vpc module - exists
  - project/service-account module - exists
  - file-system/cloud-storage-bucket module - exists
  - cloud-function/serviceaccount-audit-logs-watcher - new module that takes in bucket, vpc, and service account and contains rest of functionality. 

2. Host this example in some other repo and refer to it from the shared vpc documentation in this PR.
3. Maybe there is a deeper level we could place this code that is more appropriate such as /community/examples/shared-vpc-setup-helper/<this code>

I will confer with the team and decide what options are reasonable.

PR #2678, opened 2024-06-10
Move `runners` from `map(string)` to `object`

mr0re1:
* Limit changes to `startup-script` module only.
* Improve testing around data types by using regex name matching.


nick-stroud:
reopen when ready

PR #2680, opened 2024-06-11
Add known gpu types and their accelerators to gke module

ankitkinra:
## Add known gpu types and their accelerators to gke module

Changes in this PR
* Added a gpu file similar to https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/627b43aa3f238242b6ca98a5575541009ac411df/modules/compute/vm-instance/gpu_definition.tf#L29 which allows GKE module to derive the accelerator type and gpu count from the machine type.
* Additionally added "DEFAULT" driver as the default for each of these machine types for now.

### DIFF
--- a/community/modules/compute/gke-node-pool/gpu_definition.tf
+++ b/community/modules/compute/gke-node-pool/gpu_definition.tf
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+## Required variables:
+#  guest_accelerator
+#  machine_type
+
+locals {
+  # example state; terraform will ignore diffs if last element of URL matches
+  # guest_accelerator = [
+  #   {
+  #     count = 1
+  #     type  = "https://www.googleapis.com/compute/beta/projects/PROJECT/zones/ZONE/acceleratorTypes/nvidia-tesla-a100"
+  #   },
+  # ]
+  accelerator_machines = {
+    "a2-highgpu-1g"  = { type = "nvidia-tesla-a100", count = 1 },
+    "a2-highgpu-2g"  = { type = "nvidia-tesla-a100", count = 2 },
+    "a2-highgpu-4g"  = { type = "nvidia-tesla-a100", count = 4 },
+    "a2-highgpu-8g"  = { type = "nvidia-tesla-a100", count = 8 },
+    "a2-megagpu-16g" = { type = "nvidia-tesla-a100", count = 16 },
+    "a2-ultragpu-1g" = { type = "nvidia-a100-80gb", count = 1 },
+    "a2-ultragpu-2g" = { type = "nvidia-a100-80gb", count = 2 },
+    "a2-ultragpu-4g" = { type = "nvidia-a100-80gb", count = 4 },
+    "a2-ultragpu-8g" = { type = "nvidia-a100-80gb", count = 8 },
+    "a3-highgpu-8g"  = { type = "nvidia-h100-80gb", count = 8 },
+    "g2-standard-4"  = { type = "nvidia-l4", count = 1 },
+    "g2-standard-8"  = { type = "nvidia-l4", count = 1 },
+    "g2-standard-12" = { type = "nvidia-l4", count = 1 },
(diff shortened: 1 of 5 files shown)


community/modules/compute/gke-node-pool/gpu_definition.tf
```diff
@@ -0,0 +1,52 @@
...
+
+locals {
+  # example state; terraform will ignore diffs if last element of URL matches
+  # guest_accelerator = [
+  #   {
+  #     count = 1
+  #     type  = "https://www.googleapis.com/compute/beta/projects/PROJECT/zones/ZONE/acceleratorTypes/nvidia-tesla-a100"
+  #   },
+  # ]
+  accelerator_machines = {
```
nick-stroud, line 29: Is it possible to share this code with other modules (using the dup/diff pre-commit). I realize that the GKE data is probably more rich than needed for vm-instance or Slurm, would it be possible to have all use the more detailed data and then just dumb it down when injecting into other modules. Other options would be to omit the default part from this file and add it in main.tf. It is already a pain having to keep this list updated manually, and I would rather not have to do it in 2 places.

nick-stroud (changes requested):

community/modules/compute/gke-node-pool/gpu_definition.tf
ankitkinra: Updated file and added to dup check

community/modules/compute/gke-node-pool/main.tf
```diff
@@ -158,6 +169,12 @@ resource "google_container_node_pool" "node_pool" {
       condition     = !(var.local_ssd_count_ephemeral_storage > 0 && var.local_ssd_count_nvme_block > 0)
       error_message = "Only one of local_ssd_count_ephemeral_storage or local_ssd_count_nvme_block can be set to a non-zero value."
     }
+
+    # non preattached gpu machine type should always have guest_accelerator config defined
+    precondition {
+      condition     = (local.preattached_gpu_machine_family && length(var.guest_accelerator != null ? var.guest_accelerator : []) >= 0) || (!local.preattached_gpu_machine_family && length(var.guest_accelerator != null ? var.guest_accelerator : []) > 0)
```
nick-stroud, line 175: This condition does not make sense to me: `length(var.guest_accelerator != null ? var.guest_accelerator : []) >= 0)`, it seems always true. 

var.guest_accelerator = null
var.guest_accelerator = []
var.guest_accelerator = [item]

all true.

nick-stroud, line 175: `!local.preattached_gpu_machine_family && length(var.guest_accelerator != null ? var.guest_accelerator : []) > 0`

I think this condition is incorrect, how about for `machine_type: c2-standard-60`. I would expect `var.guest_accelerator` to be `null`.

Eventually, the error message below is not true. I don't think we can actually do validation here because regardless of whether or not `preattached_gpu_machine_family`, we may or may not have `guest_accelerator`.

community/modules/compute/gke-node-pool/main.tf
nick-stroud, line 1: The implication of the way this PR is written is that supplying guest accelerator is an all or none activity. Let me give a few examples. What if I want to supply:

```
  - id: g2_pool
    source: community/modules/compute/gke-node-pool
    use: [gke_cluster]
    settings:
      machine_type: g2-standard-4
      guest_accelerator:
      - gpu_driver_installation_config:
        - gpu_driver_version: v5.54

# OR

  - id: n1_pool
    source: community/modules/compute/gke-node-pool
    use: [gke_cluster]
    settings:
      machine_type: n1-standard-8
      guest_accelerator:
      - type: nvidia-l4
        count: 1
```

In the first case I would want to populate the type and count, in the second case I would want to populate the nulls and default.

I believe this could be done using optionals in the variable (https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes), it would also take some refactoring of main and maybe the shared gpu definition file.

Logic would be something like:
- default to empty list instead of null
- if preattached type but list is empty, populate full object (covered currently)
- if preattached type but object is partial, fill in type and count (first case above)
- use optional with default of "DEFAULT" for `gpu_driver_version` (second case above)

I think this would be a really nice QOL feature if we did it right but let's discuss if you feel it will be too complex.

nick-stroud (changes requested):

community/modules/compute/gke-node-pool/main.tf
ankitkinra: Hmm, makes sense for the first validation, I think added that as for non preattached gpu families the second condition was becoming false but I agree it can be `local.preattached_gpu_machine_family || (!local.preattached_gpu_machine_family && length(var.guest_accelerator != null ? var.guest_accelerator : []) > 0)`

Additionally I had an assumption that the user will either use a pre-attached gpu family machine type or add `guest_accelerator` but if that assumption is incorrect, there is no point of adding this validation here and instead focus on making the defaults / optional work better.

community/modules/compute/gke-node-pool/main.tf
ankitkinra: Agreed this would be better and based on the offline sync I will keep this PR which does the first thing on the list but then focus on the rest in a separate PR.

nick-stroud (approved):

PR #2682, opened 2024-06-11
Add stand alone MIG module

mr0re1:


community/modules/compute/mig/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+  description = "Labels to add to the MIG"
+  type        = map(string)
+}
+
+variable "zone" {
+  description = "Compute Platform zone. Required, currently only zonal MIGs are supported"
+  type        = string
+}
+
+
```
cdunbar13, line 35: Remove extra line, here and below this variable

cdunbar13 (commented):
Looks good, only questions is if we should add the ability to add randomness to the mig name to prevent collisions.  I'm not sure if that's something that can easily happen, so I leave it to you.

cdunbar13 (approved):

mr0re1:
> Looks good, only questions is if we should add the ability to add randomness to the mig name to prevent collisions. I'm not sure if that's something that can easily happen, so I leave it to you.

We use deployment name as suffix by default. I think it provides sufficient uniqueness. Landing it now. Let's discuss offline if you have any concerns about collisions that aren't solved by this.

PR #2685, opened 2024-06-12
Expose variables in `vm-instance` module for setting reservation and resource policy names

rohitramu:


modules/compute/vm-instance/variables.tf
```diff
@@ -299,18 +299,36 @@ variable "bandwidth_tier" {
   }
 }
 
+variable "reservation_name" {
+  description = <<-EOT
+  If not null, use the reservation with this name.
```
mr0re1, line 304: What format of names are expected?
* `name`
* `projects/<project>/reservation/<name>`
* others?

modules/compute/vm-instance/variables.tf
```diff
@@ -299,18 +299,36 @@ variable "bandwidth_tier" {
...
+  This argument will be ignored if "resource_policy_name" is set.
   EOT
 
   type    = any # It's a workaround of lack of `optional` in Terraform 1.2
   default = null
   validation {
     condition     = var.placement_policy == null ? true : try(keys(var.placement_policy), null) != null
     error_message = <<-EOT
-    The var.placement_policy should be either unset/null or be a map/object with 
+    The var.placement_policy should be either unset/null or be a map/object with
```
mr0re1, line 331: Let's add validation that those two are exclusive

rohitramu:
No bandwidth to see this through at the moment.  Will reopen if this ends up getting assigned a higher priority.

PR #2687, opened 2024-06-13
Allow users to turn on access to cluster via GCP public IP address space

ankitkinra:
Allows user to enable access to GKE cluster via GCP public IP address space

* Added a new variable (false by default) in the GKE cluster module
* After this change user can access the GKE cluster (if this variable is turned on) if they are trying to access cluster from their GCP hosted workstation of GCP VM.


community/examples/ml-gke.yaml
nick-stroud, line 1: Please add description to PR. 

- Explain or link to documentation on what access this feature allows. It sounds like I could have another VM in google cloud and then I could access GKE API from that VM.
- What manual testing have you done on this change?

community/examples/ml-gke.yaml
```diff
@@ -22,10 +22,13 @@ vars:
   zones:
   - asia-southeast1-b  # g2 machine has better availability in this zone
 
-  # Cidr block containing the IP of the machine calling terraform.
-  # The following line must be updated for this example to work.
+  # Uncomment the following lines if you want to add
+  # cidr block containing the IP of the machine calling terraform.
   authorized_cidr: <your-ip-address>/32
```
nick-stroud, line 27: new comment above says "uncomment" but the authorized cidr is not commented out.

nick-stroud (changes requested):

community/examples/ml-gke.yaml
ankitkinra: Update the PR description.

For manual testing I added the variable in the blueprint, deployed the cluster and tested that my GCP cloud workstation was able to access the cluster without explicitly adding my workstation ip address

community/examples/ml-gke.yaml
ankitkinra: Updated and removed the changes to blueprint.

community/examples/ml-gke.yaml
nick-stroud: PS. My comment here was to actually edit the PR description to add this information.

nick-stroud (approved):

PR #2725, opened 2024-06-27
Exclude Filestore module from integration test

mr0re1:
* Add integration test  `pre_create_task` `remove_modules`;
* Use it on two tests to exclude filestore;
* Restore "static" network name added to those tests by #2530

### DIFF
--- a/tools/cloud-build/daily-tests/remove_module.py
+++ b/tools/cloud-build/daily-tests/remove_module.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


tools/cloud-build/daily-tests/remove_module.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import yaml # pip install pyyaml
+
+def delete_module(bp, mod_id: str):
```
rohitramu, line 18: If we print to console in this method, will it show up in CloudBuild logs?  If so, I think it would be really helpful to see in the logs exactly which modules were removed from the blueprint.  It would also be really helpful to see two more artifacts:
1. Original blueprint
2. Modified blueprint

Maybe we could save them in GCS and reference them in our CloudBuild yaml like this?
https://cloud.google.com/build/docs/building/store-artifacts-in-cloud-storage

At the very least, for this PR, it would be good to print the blueprint before and after modification to the log.

rohitramu (changes requested):

PR #2732, opened 2024-07-01
Make `cleanup_compute` resilient to delete failures

mr0re1:
* Ignore failures in `gcloud delete`;
* Increate batch size 10 -> 16.

**Ack.** that this could lead to endless loop. The possible conditions:
* User has permission to list, but not to delete;
* Instances are under "delete protection".

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
@@ -38,28 +38,40 @@ if ! type -P gcloud 1>/dev/null; then
 	exit 1
 fi
 
+compute_cmd="$API_ENDPOINT gcloud compute"
+
 echo "Deleting compute nodes"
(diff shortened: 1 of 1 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
```diff
@@ -32,25 +32,25 @@ fi
 echo "Deleting compute nodes"
 node_filter="labels.slurm_cluster_name=${cluster_name} AND labels.slurm_instance_role=compute"
```
tpdownes, line 33: 1. I think a `"NOT status:(STOPPING TERMINATED)"` filter would go a long way toward addressing the real cause of this problem.
2. I like the idea of retrying failures but not forever. I think something like

```
max_failures=5
retry_attempt=0
gcloud compute instances delete || [ $((retry_attempt++)) -gt "$max_failures"]]
```

tpdownes (changes requested):

PR #2746, opened 2024-07-08
Remove `enable_devel` from examples

mr0re1:
`enable_devel: false` is currently not working due to incompatibility between scripts in image and scripts on slurm repo head.

### DIFF
--- a/community/examples/hpc-build-slurm-image.yaml
+++ b/community/examples/hpc-build-slurm-image.yaml
@@ -70,6 +70,8 @@ deployment_groups:
           }
       - type: shell
         destination: install_slurm.sh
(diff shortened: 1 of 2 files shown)


community/examples/hpc-build-slurm-image.yaml
nick-stroud, line 1: Please add description explaining:
1. Why we are doing this, enable_devel: false is currently not working due to incompatibility between scripts in image and scripts on slurm repo head.
2. note that changes to scripts in the built image will not reflect in the deployed cluster and give a breadcrumb as to where scripts will be sourced from (version of slurm pointed to by ghpc module).

nick-stroud (approved):

PR #2740, opened 2024-07-03
Add parallelstore slurm blueprint and documentation

harshthakkar01:
This PR 
* Adds public blueprint to use parallelstore with Slurm cluster and with Client VM running debian 12. 
* Documents parallelstore module, usage, hydration.
* Adds integration test

### Submission Checklist

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cloud HPC Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #2747, opened 2024-07-08
Update parameters for parallelstore module as per feedback

harshthakkar01:
Update import data parameters after discussion.
Detail information about how to use private-service-access if it is turned on by default on VPC will be added in the documentation. 

Remove -o flag for mount-options. 

### Submission Checklist

### DIFF
--- a/modules/file-system/parallelstore/variables.tf
+++ b/modules/file-system/parallelstore/variables.tf
@@ -60,7 +60,11 @@ variable "mount_options" {
 }
 
 variable "private_vpc_connection_peering" {
-  description = "The name of the VPC Network peering connection."
+  description = <<-EOT
+    The name of the VPC Network peering connection.
+    If using new VPC, please use community/modules/network/private-service-access to create private-service-access and
+    If using existing VPC with private-service-access enabled, set this manually."
+    EOT
   type        = string
 }
 
@@ -76,14 +80,14 @@ variable "network_id" {
   }
 }
 
-variable "source_gcs_bucket_uri" {
+variable "import_gcs_bucket_uri" {
   description = "The name of the GCS bucket to import data from to parallelstore."
   type        = string
-  default     = ""
+  default     = null
 }
 
-variable "destination_hydration_parallelstore" {
+variable "import_destination_path" {
   description = "The name of local path to import data on parallelstore instance from GCS bucket."
   type        = string
-  default     = "/"
+  default     = null
 }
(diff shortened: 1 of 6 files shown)


modules/file-system/parallelstore/variables.tf
```diff
@@ -76,13 +80,13 @@ variable "network_id" {
   }
 }
 
-variable "source_gcs_bucket_uri" {
+variable "import_gcs_bucket_uri" {
   description = "The name of the GCS bucket to import data from to parallelstore."
   type        = string
   default     = ""
```
nick-stroud, line 86: null would be a better default, then check for null instead of empty string.

modules/file-system/parallelstore/variables.tf
```diff
@@ -76,13 +80,13 @@ variable "network_id" {
...
 }
 
-variable "destination_hydration_parallelstore" {
+variable "import_destination_path" {
   description = "The name of local path to import data on parallelstore instance from GCS bucket."
   type        = string
   default     = "/"
```
nick-stroud, line 92: I would default this to null. Then to a pre-condition, if this is set then source_gcs_bucket_uri must also be set. If import_gcs_bucket_uri is set and this is null then you can fall back to "/".

modules/file-system/parallelstore/variables.tf
```diff
@@ -60,7 +60,11 @@ variable "mount_options" {
 }
 
 variable "private_vpc_connection_peering" {
-  description = "The name of the VPC Network peering connection."
+  description = <<-EOT
+    The name of the VPC Network peering connection.
+    If using new VPC, please use community/modules/network/private-service-access to create private-service-access and
+    If using existing VPC with private-service-access enabled, set this manually."
```
nick-stroud, line 66: I would prefer a link to an example of documentation. If documentation is coming in separate PR then please add a link from here then.

nick-stroud: documentation should include how to find the name if I am setting it manually.

nick-stroud (changes requested):

modules/file-system/parallelstore/variables.tf
harshthakkar01: done.

modules/file-system/parallelstore/variables.tf
harshthakkar01: Done.

modules/file-system/parallelstore/variables.tf
harshthakkar01: Documentation in #2740 will cover information about how to use existing-vpc if it has private-service-access enabled.

nick-stroud (approved):

nick-stroud (approved):

nick-stroud (approved):

PR #2758, opened 2024-07-12
Move GKE Modules to Core

chengcongdu:
Moving the below GKE modules from community to core
* compute/gke-node-pool
* compute/gke-job-template
* scheduler/gke-cluster
* scheduler/pre-existing-gke-cluster

Update examples, tests, docs and badges to reflect the above migration


chengcongdu:
@nick-stroud PTAL

nick-stroud (approved):

PR #2761, opened 2024-07-15
Fix 'spot' and 'termination_action' variable passing in schedmd-slurm-gcp-v6-controller

ctk21:
This PR fixes the passing of 'spot' and 'termination_action' in schedmd-slurm-gcp-v6-controller when setting up the instance templates. 

The slurm v6 nodeset templates contain spot variables (for example [these](https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/ae08b5988ee9b2a67e2b6dee7cce6f36709446be/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf#L289-L301)), but without this fix those variables get dropped.


mr0re1:
Hi @ctk21  , thank you contribution! LGTM.

mr0re1 (approved):

PR #2770, opened 2024-07-17
Improve `pytest` flow

mr0re1:
Invoke `pytest` from root of the repo, without adding flags.
No changes to tests logic.

```sh
$ pwd
/usr/local/google/home/orlov/wp/hpc-toolkit_too
$ pytest
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.11.8, pytest-7.4.0, pluggy-1.2.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /usr/local/google/home/orlov/wp/hpc-toolkit_too
configfile: .pytest.ini
plugins: mock-3.14.0, anyio-3.7.1, benchmark-4.0.0, typeguard-2.13.3, xdist-3.3.1
collected 28 items

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_topology.py ..                                                                       [  7%]
community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py .....................                                                        [ 82%]
tests/test_maintenance.py .....                                                                                                                                                         [100%]

===================================================================================== 28 passed in 1.60s ======================================================================================
$
```

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
@@ -1,0 +1,53 @@
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+import sys
+from dataclasses import dataclass, field
+
+SCRIPTS_DIR = "community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts"
+if SCRIPTS_DIR not in sys.path:
+    sys.path.append(SCRIPTS_DIR)  # TODO: make this more robust
+
+
+# TODO: use "real" classes once they are defined (instead of NSDict)
+@dataclass
+class TstNodeset:
+    nodeset_name: str
+    node_count_static: int = 0
+    node_count_dynamic_max: int = 0
+
+
+@dataclass
+class TstCfg:
+    slurm_cluster_name: str = "m22"
+    nodeset: dict[str, TstNodeset] = field(default_factory=dict)
+    nodeset_tpu: dict[str, TstNodeset] = field(default_factory=dict)
+    output_dir: Optional[str] = None
+
+
+@dataclass
+class TstTPU:  # to prevent client initialization durint "TPU.__init__"
+    vmcount: int
+
+
+def make_to_hostnames_mock(tbl: Optional[dict[str, list[str]]]):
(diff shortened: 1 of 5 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
```diff
@@ -0,0 +1,53 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+import sys
+from dataclasses import dataclass, field
+
+SCRIPTS_DIR = "community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts"
```
cdunbar13, line 19: Is this supposed to be a relative path?  If so you might want to make it absolute with os.path.abspath.  I'm not sure if adding it to the path below will work without it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
```diff
@@ -0,0 +1,53 @@
...
+
+@dataclass
+class TstTPU:  # to prevent client initialization durint "TPU.__init__"
+    vmcount: int
+
+
+def make_to_hostnames_mock(tbl: Optional[dict[str, list[str]]]):
+    tbl = tbl or {}
+
+    def se(k: str) -> list[str]:
```
cdunbar13, line 48: does this `list` and the `dict` above work?  My understanding was you're supposed to import List and Dict from typing

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
```diff
@@ -0,0 +1,53 @@
...
+        if k not in tbl:
```
cdunbar13, line 49: Should we check if k exists?

cdunbar13 (approved):
A few small comments.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
mr0re1: Old version of Python complained about using `list[...]`, to work it around we use `typing List` in non-test code instead. Since test code runs on machines with reasonably fresh Python we can afford to use a modern way to do it (`list`)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
mr0re1: that's what we do here (`if k not in tbl`)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/common.py
mr0re1: Relative path works.

PR #2777, opened 2024-07-18
Simplify SlurmGCP logging

mr0re1:
Move addition and processing of common loging arguments to a function.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -523,52 +523,20 @@ def main():
...
-parser.add_argument(
-    "--trace-api",
-    "-t",
-    action="store_true",
-    help="Enable detailed api request output",
-)
-parser.add_argument(
-    "--force",
-    "-f",
-    action="store_true",
```
harshthakkar01, line 547: Why we are removing this argument ?

harshthakkar01 (approved):

PR #2778, opened 2024-07-18
Add `allow_automatic_updates` flag

rohitramu:
The "disable automatic updates" feature is implemented in HPC VM images built on or after July 12th 2024.  Currently, these are `hpc-centos-7-v20240712` and `hpc-rocky-linux-8-v20240712`.  The feature can be used by setting the `google_disable_automatic_updates=TRUE` metadata entry on a VM instance.

To make the feature easy to use for customers, this PR adds the `disable_automatic_updates` flag to the following modules:
- `modules/compute/vm-instance`
- `modules/scheduler/batch-job-template`
- `community/modules/compute/htcondor-execute-point`
- `community/modules/compute/schedmd-slurm-gcp-v5-node-group`
- `community/modules/compute/schedmd-slurm-gcp-v6-nodeset`
- `community/modules/compute/schedmd-slurm-gcp-v6-nodeset-dynamic`
- `community/modules/scheduler/htcondor-access-point`
- `community/modules/scheduler/htcondor-central-manager`
- `community/modules/scheduler/schedmd-slurm-gcp-v5-controller`
- `community/modules/scheduler/schedmd-slurm-gcp-v5-login`
- `community/modules/scheduler/schedmd-slurm-gcp-v6-controller`
- `community/modules/scheduler/schedmd-slurm-gcp-v6-login`

The following modules were not updated because references to updated modules shouldn't be made before the module being referenced has been included in an official release:
- `community/modules/remote-desktop/chrome-remote-desktop`
- `community/modules/compute/pbspro-execution`
- `community/modules/scheduler/pbspro-client`
- `community/modules/scheduler/pbspro-server`

I have tested the `vm-instance` implementation with the following blueprint:
```yaml
blueprint_name: disable-auto-updates-test

vars:
  deployment_name: disable-auto-updates-test
  region: us-central1
  zone: us-central1-a

deployment_groups:
- group: primary
  modules:
  - id: network1
    source: modules/network/vpc

  - id: startup-test-rocky
    settings:
      configure_ssh_host_patterns:
      - '*'
      runners:
        - type: shell
          destination: /tmp/updates_disabled_test.sh
          content: |
            #!/bin/bash

            status=\$(systemctl is-enabled dnf-automatic.timer)
            if [[ status != 'disabled' ]]; then
              exit 10
            fi
  - id: workstation-rocky
    use:
    - network1
    - startup-test-rocky
    settings:
      instance_count: 1
      name_prefix: test-rocky
      disable_automatic_updates: true
      instance_image:
        project: cloud-hpc-image-public
        name: hpc-rocky-linux-8-v20240712
  - id: wait-rocky
    settings:
      instance_names: $(workstation-rocky.name)

  - id: startup-test-centos
    settings:
      configure_ssh_host_patterns:
      - '*'
      runners:
        - type: shell
          destination: /tmp/updates_disabled_test.sh
          content: |
            #!/bin/bash

            status=\$(systemctl is-enabled yum-cron)
            if [[ status != 'disabled' ]]; then
              exit 10
            fi
  - id: workstation-centos
    use:
    - network1
    - startup-test-centos
    settings:
      instance_count: 1
      name_prefix: test-centos
      disable_automatic_updates: true
      instance_image:
        project: cloud-hpc-image-public
        name: hpc-centos-7-v20240712
  - id: wait-centos
    settings:
      instance_names: $(workstation-centos.name)

```


community/modules/compute/pbspro-execution/main.tf
```diff
@@ -68,7 +68,8 @@ module "execution_startup_script" {
 }
 
 module "pbs_execution" {
-  source = "github.com/GoogleCloudPlatform/hpc-toolkit//modules/compute/vm-instance?ref=v1.35.0&depth=1"
+  # source = "github.com/GoogleCloudPlatform/hpc-toolkit//modules/compute/vm-instance?ref=v1.35.0&depth=1"
+  source = "../../../../modules/compute/vm-instance"
```
tpdownes, line 72: This is not going to work reliably unfortunately.

tpdownes (changes requested):

community/modules/compute/pbspro-execution/main.tf
rohitramu: I asked in chat when I sent out this PR, but I'll ask here too: do we have guidelines or a standard strategy for referencing changes made in the same PR?

Update the reference in a separate PR and change the ref to "develop"?  Use relative references?  Or do we need to wait for the next release to update the reference in the "calling" module?

rohitramu: Since this issue is only occurring in the PBS Pro and Chrome Remote Desktop modules, I reverted the change there.  We can make this change after the next release, so the ref can be updated to the next label.

mr0re1 (approved):

modules/compute/vm-instance/compute_image.tf
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright 2022 Google LLC
```
harshthakkar01, line 2: 2024 ?

rohitramu: Fixed.

harshthakkar01:
lgtm for `vm-instance` and `batch-job-template` module.

harshthakkar01 (approved):

tpdownes (approved):
At nit level, I think you should:

1. add `nullable = false` to each variable declaration
2. Bump versions.tf for each module to require Terraform 1.1 (if it's not already above that)

The consequence of not doing this is somewhat above minor: customers who actively set this to `null` in a blueprint will probably see a crash on the ternary assignment not containing a boolean.

I would like to see more use of `nullable` in our code when there are benefits: in this case you can treat the boolean value as guaranteed to be a boolean and the code is both simply read and less crash-prone.

modules/compute/vm-instance/variables.tf
```diff
@@ -392,3 +392,13 @@ variable "allocate_ip" {
   })
   default = null
 }
+
+variable "disable_automatic_updates" {
```
nick-stroud, line 396: My preference would be for `run_automatic_package_updates: true` to be the default.
or maybe better `allow_automatic_package_updates: true`

This comes from our previous experience with naming variables with the prefix `disable_`. It can be confusing with the double negative when reading the variable documentation or as seen in blueprints if we were ever to change the default value.

I realize that the actual action we are taking is to disable the feature, and that using `disable_` may be a more accurate statement, but I believe the potential user confusion is more important.

When reading documentation we would either have
`disable_automatic_updates` (default `false`) OR
`allow_automatic_package_updates` (default `true`)
to me the second is clearer

This would be my vote. I am fine to be out-voted by other reviewers.

rohitramu:
> At nit level, I think you should:
> 
> 1. add `nullable = false` to each variable declaration
> 2. Bump versions.tf for each module to require Terraform 1.1 (if it's not already above that)
> 
> The consequence of not doing this is somewhat above minor: customers who actively set this to `null` in a blueprint will probably see a crash on the ternary assignment not containing a boolean.
> 
> I would like to see more use of `nullable` in our code when there are benefits: in this case you can treat the boolean value as guaranteed to be a boolean and the code is both simply read and less crash-prone.

@tpdownes Thanks for the suggestions, I have done both:
1. Added `nullable = false`
2. Ensured all versions.tf files require Terraform 1.1 or higher

modules/compute/vm-instance/variables.tf
rohitramu: I've updated all references to be `allow_automatic_updates` and specified in the variable description that this is specifically for system packages.

PR #2797, opened 2024-07-22
Support multivpc with GKE

sharabiani:
This PR adds support for multivpc module with GKE clusters, allowing blueprints to leverage existing Google Kubernetes Engine (GKE) clusters.

### [Multivpc support in gke-cluster] enable_multi_networking and additional_networks variable settings added to gke-cluster module:
**Purpose**: Enables multi networking feature, and to accept multivpc as "use" in blueprint. Applies Network objects to the cluster using a manifest
**Location**: community/modules/scheduler/gke-cluster

### [Multivpc support in pre-existing-gke-cluster] additional_networks variable added to pre-existing-gke-cluster module:
**Purpose**: to accept multivpc as "use" in blueprint. Applies Network objects to the cluster using a manifest.
**Location**:   modules/compute/pre-existing-gke-cluster

### [Multivpc support in gke-node-pool] additional_networks variable added to gke-node-pool module:
**Purpose**: to accept multivpc as "use" in blueprint. Adds multiple networks to the nodes.
**Location**:   modules/compute/gke-node-pool

### [Fix default values issues in gke-node-pool] ephemeral_storage_local_ssd_config and local_nvme_ssd_block_config will be added to configs only if they are assigned a number.
**Purpose**: GKE will decide about default values. Zero is not a correct default value.
**Location**:   modules/compute/gke-node-pool

### [Fix in multivpc module] network_name populated in the additional_networks output in multivpc module:
**Purpose**: to pass the names of the vpc networks created.
**Location**:   modules/network/multivpc

### an example of GKE added in a blueprint
**Location**:  examples/gke-a3-megagpu.yaml
**Location**:  examples/gka3-highgpu.yaml

## Manual testing:

1. Deployed examples/gke-multi-vpc-a3-megagpu-8g.yaml
2. Provisioned the deployment
3. Verified the successful creation of a cluster, vpcs, new node pool in the GKE cluster through Pantheon.

## Additional Notes:

- Made changes to relevant documentation.


ankitkinra (commented):
Started meta discussion on kubectl vs kubernetes-manifest on chat

modules/compute/gke-node-pool/README.md
```diff
@@ -73,6 +73,10 @@ use-cases). In this case, ensure that you turn off the
 [enable_secure_boot](#input\_enable\_secure\_boot) option to allow unsigned
 kernel modules to be loaded.
 
+To maximize GPU network bandwidth, nodepools accept multiple VPCs. Pass a multivpc module to gke-node-pool module, and [install GPUDirect and configure NCCL] (https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl).
```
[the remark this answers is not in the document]
sharabiani: I changed the description here. I believe most of the description will be done automatically by implementing the upcoming features.

modules/compute/gke-node-pool/README.md
```diff
@@ -168,6 +172,39 @@ Finally, the following is an example of using a GPU (with sharing config) attach
           gpu_sharing_strategy: "TIME_SHARING"
 ```
 
+Finally, the following is adding multivpc to a node pool (with sharing config) attached to an `n1` machine:
```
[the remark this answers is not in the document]
sharabiani: fixed. based on the new changes because of using kubectl provider in gke-cluster module there is no need for a second deployment.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+    use: [gke_cluster, gpunets]
+    settings:
+      machine_type: a3-megagpu-8g
+      guest_accelerator:
+      - type: nvidia-h100-mega-80gb
+        count: 8
+        gpu_driver_installation_config:
+        - gpu_driver_version: LATEST
+      disk_type: pd-balanced
+      autoscaling_total_min_nodes: 2
```
nick-stroud, line 73: I think the most common use case to start would be static nodes + reservation. I am looking and don't thing we have reservation support yet for GKE (in Toolkit), so this is probably fine for now.

@ankitkinra can we add a task to make sure we add reservation support, then close this comment.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+      zones: [$(vars.zone)]
+
+# We need to do the following here:
+# 1- Install the GPUDirect binary and configure NCCL: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl
+# 2- Deploy NRI device injector plugin: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#nri-device-injector
+# 3- Deploy a test workload: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#test-workload
+# 4- Use recommended NCCL configuration settings to improve performance: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#environment-variables-nccl
+# 5- Add GPUDirect to manifests: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#add-gpudirect-manifests
```
nick-stroud, line 81: Can you link to description so there is one source of truth for these instructions.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+      global_ip_address_range: 192.169.0.0/16
+      network_count: 8
+      subnetwork_cidr_suffix: 24
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network1, gpunets]
+    settings:
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
+      enable_multi_networking: true
```
nick-stroud, line 56: Is this needed since you are 'use'ing gpunets.
I think you could achieve this by having an output on multivpc of `enable_multi_netowrking: true`

modules/compute/gke-node-pool/README.md
```diff
@@ -168,6 +173,39 @@ Finally, the following is an example of using a GPU (with sharing config) attach
           gpu_sharing_strategy: "TIME_SHARING"
 ```
 
+Finally, the following is adding multivpc to a node pool:
+
+```yaml
+  - id: network
```
nick-stroud, line 180: I think you need more fields for a gke network, see your example.

modules/compute/gke-node-pool/README.md
```diff
@@ -168,6 +173,39 @@ Finally, the following is an example of using a GPU (with sharing config) attach
...
+  - id: a3-megagpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke-cluster, multinetwork]
+    settings:
+      machine_type: a3-megagpu-8g
+      guest_accelerator:
+      - type: nvidia-h100-mega-80gb
+        count: 8
+        gpu_driver_installation_config:
+        - gpu_driver_version: LATEST
```
nick-stroud, line 205: I think the guest accelerator is unneeded here (autopopulated)

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -296,3 +296,34 @@ variable "service_account" {
     error_message = "service_account is deprecated and replaced with service_account_email and scopes."
   }
 }
+
+variable "enable_multi_networking" {
+  description = "Enables [multi networking](https://cloud.google.com/kubernetes-engine/docs/how-to/setup-multinetwork-support-for-pods#create-a-gke-cluster). This setting is immutable on clusters and enables [Dataplane V2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2?hl=en)."
```
nick-stroud, line 301: Should we mention that enterprise is needed?

modules/scheduler/gke-cluster/README.md
```diff
@@ -42,6 +42,31 @@ are created in the VPC module. By default the `gke-cluster` module will look for
 ranges with the names `pods` and `services`. These names can be configured using
 the `pods_ip_range_name` and `services_ip_range_name` settings.
 
+### Multi-networking
+
+To create network objects in GKE cluster, you can pass a multivpc module to gke-cluster module instead of [applying a manifest manually](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment). Passing a multivpc module enables multi networking and [Dataplane V2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2?hl=en) on the cluster.
+
+```yaml
+  - id: network
+    source: modules/network/vpc
```
nick-stroud, line 51: I think you need more fields here to make it compatible with GKE. See your example.

modules/scheduler/gke-cluster/main.tf
```diff
@@ -30,6 +30,12 @@ locals {
   }]
 
   sa_email = var.service_account_email != null ? var.service_account_email : data.google_compute_default_service_account.default_sa.email
+
+  # additional VPCs enable multi networking 
+  derived_enable_multi_networking = length(var.additional_networks) > 0 ? true : var.enable_multi_networking
```
nick-stroud, line 35: I think you can avoid this by having multinetwork output the `enable_multi_networking: true` field

modules/scheduler/gke-cluster/main.tf
```diff
@@ -30,6 +30,12 @@ locals {
...
+
+  # multi networking needs enabled Dataplane v2
+  derived_enable_dataplane_v2 = local.derived_enable_multi_networking ? true : var.enable_dataplane_v2
```
nick-stroud, line 38: Here if user explicitly sets `enable_dataplane_v2` to false we silently ignore their setting. I do not like to do this.

I think we should do the following:
- change var.enable_dataplane_v2 to a string and use it like an enum
- have default be something like "SELECT_DEFAULT"
- provide explicit options for "ENABLED" and "DISABLED"
- if "SELECT_DEFUALT" is selected then we can override
- use precondition to ensure that we don't have conflicting options

nick-stroud (commented):
I have not finished review but want to send initial comments.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - id: a3-megagpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, gpunets]
+    settings:
+      machine_type: a3-megagpu-8g
+      guest_accelerator:
+      - type: nvidia-h100-mega-80gb
+        count: 8
+        gpu_driver_installation_config:
+        - gpu_driver_version: LATEST
```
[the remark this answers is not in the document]
ankitkinra: default version is "DEFAULT", let's try removing this and see if that works ?

examples/gke-multi-vpc-a3-megagpu-8g.yaml
ankitkinra: Created a task

modules/scheduler/gke-cluster/main.tf
ankitkinra: Great point, should we also add a validation somewhere that multi-networking requires `Dataplane_v2` so that it is not a silent failure.

modules/compute/gke-node-pool/main.tf
```diff
@@ -104,12 +104,18 @@ resource "google_container_node_pool" "node_pool" {
...
+      }
     }
 
-    local_nvme_ssd_block_config {
-      local_ssd_count = var.local_ssd_count_nvme_block
+    dynamic "local_nvme_ssd_block_config" {
+      for_each = var.local_ssd_count_nvme_block > 0 ? [1] : []
+      content {
+        local_ssd_count = var.local_ssd_count_nvme_block
+      }
```
nick-stroud, line 118: I don't quite see why these had to be made dynamic blocks? Is there a problem being explicit and assigning zero when the value is zero.

modules/scheduler/gke-cluster/main.tf
nick-stroud: @ankitkinra: yes.
my last bullet: `use precondition to ensure that we don't have conflicting options`

A pre-condition is the Terraform way of performing validation when you need access to more than 1 variable.

modules/scheduler/pre-existing-gke-cluster/main.tf
```diff
@@ -19,3 +19,57 @@ data "google_container_cluster" "existing_gke_cluster" {
   project  = var.project_id
   location = var.region
 }
+
+data "google_client_config" "default" {}
+
+provider "kubernetes" {
```
nick-stroud, line 25: Even if you can get away with using kubernetes provider here, I would use the same kubectl as gke-cluster module for similarity.

nick-stroud (changes requested):

modules/scheduler/gke-cluster/main.tf
```diff
@@ -315,3 +317,52 @@ module "workload_identity" {
...
+spec:
+  vpc: ${each.value.network}
+  vpcSubnet: ${each.value.subnetwork}
+  deviceMode: NetDevice
+YAML
+
+  provider = kubectl
+}
+
+resource "kubectl_manifest" "additional_nets" {
```
ankitkinra, line 349: Is this manifest always applied ? Or only when the machine type is a3-mega or a3-highgpu ?

examples/README.md
```diff
@@ -1472,6 +1473,28 @@ cleaned up when the job is deleted.
 
 [storage-gke.yaml]: ../examples/storage-gke.yaml
 
+### [gke-multi-vpc-a3-megagpu-8g.yaml] ![community-badge] ![experimental-badge]
```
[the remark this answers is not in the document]
sharabiani: Renamed to: gke-a3-megagpu.yaml
Added new example: gke-a3-highgpu.yaml

modules/scheduler/gke-cluster/main.tf
sharabiani: Whenever a multivpc is passed to the module.

modules/compute/gke-node-pool/main.tf
sharabiani: Fixed. Now if the value is not null the block will be added.

modules/scheduler/gke-cluster/main.tf
sharabiani: I kept it boolean but set null as default value. I also added a validation in the variable so that we prevent when enable_dataplane_v2 is explicitly set to false and enable_multinetworking is set to true.

sharabiani: I think we need to add a validation to ensure enable_multi_networking is not explicitly set to false when a multivpc is passed to the gke-cluster.

modules/scheduler/gke-cluster/main.tf
sharabiani: > I think you can avoid this by having multinetwork output the `enable_multi_networking: true` field

I'm not exactly sure what this means. Does it mean to add an output "enable_multi_networking" to multinetwork module? I wonder isn't multivpc a generic module and enable_multi_networking a gke-only setting?

examples/README.md
```diff
@@ -1472,6 +1473,28 @@ cleaned up when the job is deleted.
...
+
+This blueprint shows how to use multiple VPC networks with GKE in the toolkit.
+
+After provisioning the cluster and the nodepool, we need to do the following:
+1- Install the GPUDirect binary and configure NCCL: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl
+2- Deploy NRI device injector plugin: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#nri-device-injector
```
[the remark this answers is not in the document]
sharabiani: Resolving this as you we added relevant tasks for automating these steps.

examples/gke-a3-highgpu.yaml
```diff
@@ -0,0 +1,86 @@
...
+- group: primary
+  modules:
+  - id: network1
+    source: modules/network/vpc
+    settings:
+      subnetwork_name: gke-subnet
+      secondary_ranges:
+        gke-subnet:
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
```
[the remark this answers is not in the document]
sharabiani: Good question! Do you think requiring users to provide them as inputs is good ux?
What do you think @nick-stroud ?

examples/gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,86 @@
...
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gpunets
+    source: modules/network/multivpc
+    settings:
+      network_name_prefix: $(vars.deployment_name)-gpunet
+      global_ip_address_range: 192.169.0.0/16
```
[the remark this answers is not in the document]
sharabiani: This needs to modify multivpc module which is used in other places too. Do you think we can check impact and make changes as an independent task?

examples/README.md
```diff
@@ -1472,6 +1473,28 @@ cleaned up when the job is deleted.
...
+3- Deploy a test workload: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#test-workload
+4- Use recommended NCCL configuration settings to improve performance: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#environment-variables-nccl
+5- Add GPUDirect to manifests: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#add-gpudirect-manifests
```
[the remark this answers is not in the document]
sharabiani: I'm resolving this because we are going to remove these with the upcoming changes that automates the steps.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network1, gpunets]
+    settings:
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
+      enable_multi_networking: true
+      master_authorized_networks:
+      - cidr_block: $(vars.authorized_cidr)
+        display_name: "kubectl-access-network" # Allows your machine run kubectl command.
```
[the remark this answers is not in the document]
sharabiani: fixed

examples/gke-multi-vpc-a3-megagpu-8g.yaml
```diff
@@ -0,0 +1,91 @@
...
+    outputs: [instructions]
+
+  - id: a3-megagpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, gpunets]
+    settings:
+      machine_type: a3-megagpu-8g
+      guest_accelerator:
+      - type: nvidia-h100-mega-80gb
+        count: 8
```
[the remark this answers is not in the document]
sharabiani: h100-mega added to the gpu_definitions.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
sharabiani: Can you clarify which description, please?
We are going to automate these steps with the next tasks, by the way.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
sharabiani: I'm not exactly sure what this means. Does it mean to add an output "enable_multi_networking" to multinetwork module? I wonder isn't multivpc a generic module and enable_multi_networking a gke-only setting?

Btw, this line is removed because passing gpunets was enough to enable it.

modules/scheduler/gke-cluster/variables.tf
sharabiani: done

modules/scheduler/gke-cluster/main.tf
sharabiani: Validations replaced with preconditions. 
enable_multi_networking also was checked to prevent explicitly set "false" value when a multivpc is passed.

modules/scheduler/pre-existing-gke-cluster/main.tf
sharabiani: done

examples/README.md
```diff
@@ -1472,6 +1474,50 @@ cleaned up when the job is deleted.
 
 [storage-gke.yaml]: ../examples/storage-gke.yaml
 
+### [gke-a3-megagpu.yaml] ![community-badge] ![experimental-badge]
+
+This blueprint shows how to provision a GKE cluster with A3 Megagpu machines in the toolkit.
```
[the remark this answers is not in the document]
sharabiani: done

examples/gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - cidr_block: $(vars.authorized_cidr)  # Allows your machine run kubectl command. It's required for the multi-network setup.
+        display_name: "kubectl-access-network"
+    outputs: [instructions]
+
+  - id: a3-megagpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, gpunets]
+    settings:
+      machine_type: a3-megagpu-8g
+      disk_type: pd-balanced
```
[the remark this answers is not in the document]
sharabiani: removed

examples/gke-a3-highgpu.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - cidr_block: $(vars.authorized_cidr)  # Allows your machine run kubectl command. It's required for the multi-network setup.
+        display_name: "kubectl-access-network"
+    outputs: [instructions]
+
+  - id: a3-highgpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, gpunets]
+    settings:
+      machine_type: a3-highgpu-8g
+      disk_type: pd-balanced
```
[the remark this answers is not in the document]
sharabiani: removed

modules/compute/gke-node-pool/README.md
```diff
@@ -73,6 +73,11 @@ use-cases). In this case, ensure that you turn off the
 [enable_secure_boot](#input\_enable\_secure\_boot) option to allow unsigned
 kernel modules to be loaded.
 
+To maximize GPU network bandwidth, nodepools accept multiple VPCs. Pass a multivpc module to gke-node-pool module, and [take these steps] (https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl) to install GPUDirect, configure NCCL, use recommended settings, and add GPUDirect to your pods.
+
+> **_NOTE:_** You must [enable multi networking](https://cloud.google.com/kubernetes-engine/docs/how-to/setup-multinetwork-support-for-pods#create-a-gke-cluster) feature when creating the GKE cluster. Passing the multivpc module to a gke-cluster module enables multi networking on the cluster creation.
+> Passing the multivpc module to a gke-cluster or pre-existing-gke-cluster module creates required network objects on the cluster for multi networking. You can do so by [applying a manifest manually](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment), too.
```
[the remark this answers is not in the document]
sharabiani: fixed

modules/scheduler/gke-cluster/README.md
```diff
@@ -42,6 +42,39 @@ are created in the VPC module. By default the `gke-cluster` module will look for
 ranges with the names `pods` and `services`. These names can be configured using
 the `pods_ip_range_name` and `services_ip_range_name` settings.
 
+### Multi-networking
+
+To create network objects in GKE cluster, you can pass a multivpc module to gke-cluster module instead of [applying a manifest manually](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment). Passing a multivpc module enables multi networking and [Dataplane V2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2?hl=en) on the cluster.
```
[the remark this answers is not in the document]
sharabiani: The network objects are in the manifest on the linked page. Do you mean them or some more reference? @ankitkinra 

Reword applied.

modules/scheduler/gke-cluster/main.tf
ankitkinra: ack, now I see the for loop with additional_networks , thanks!

examples/gke-a3-megagpu.yaml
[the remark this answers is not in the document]
ankitkinra: Oh if it was already being used in other places, then let's skip that for now, but we can come back to this later. My intent was to make sure we don't ask users to provide values for which we already have intelligent defaults

examples/README.md
```diff
@@ -1536,9 +1537,9 @@ cleaned up when the job is deleted.
 
 [storage-gke.yaml]: ../examples/storage-gke.yaml
 
-### [gke-multi-vpc-a3-megagpu-8g.yaml] ![community-badge] ![experimental-badge]
+### [gke-a3-megagpu.yaml] ![community-badge] ![experimental-badge]
 
-This blueprint shows how to use multiple VPC networks with GKE in the toolkit.
+This blueprint shows how to provision a GKE cluster with A3 Megagpu machines in the toolkit.
```
ankitkinra: nit: "A3 Mega" instead of "A3 Megagpu"

examples/README.md
```diff
@@ -1556,7 +1557,29 @@ After provisioning the cluster and the nodepool, we need to do the following:
 > `--vars authorized_cidr=<your-ip-address>/32`.** You can use a service like
 > [whatismyip.com](https://whatismyip.com) to determine your IP address.
 
-[gke-multi-vpc-a3-megagpu-8g.yaml]: ../examples/gke-multi-vpc-a3-megagpu-8g.yaml
+[gke-a3-megagpu.yaml]: ../examples/gke-a3-megagpu.yaml
+
+### [gke-a3-highgpu.yaml] ![community-badge] ![experimental-badge]
+
+This blueprint shows how to provision a GKE cluster with A3 Highgpu machines in the toolkit.
```
ankitkinra: nit: "A3 High" instead of "A3 Highgpu"

modules/scheduler/pre-existing-gke-cluster/versions.tf
```diff
@@ -20,11 +20,15 @@ terraform {
...
+      version = ">= 1.7.0"
+    }
   }
 
   provider_meta "google" {
     module_name = "blueprints/terraform/hpc-toolkit:pre-existing-gke-cluster/v1.37.2"
   }
 
-  required_version = ">= 1.0.0"
+  required_version = ">= 1.3"
```
ankitkinra, line 33: Can we remove this and update the root terraform required version ?

modules/compute/gke-node-pool/variables.tf
```diff
@@ -107,7 +107,7 @@ variable "local_ssd_count_nvme_block" {
   
   EOT 
   type        = number
-  default     = 0
+  default     = null
```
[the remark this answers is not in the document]
sharabiani: Definitely, I can do that as a separate task if you that's okay.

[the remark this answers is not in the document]
sharabiani: Task added

modules/scheduler/gke-cluster/main.tf
tpdownes: 👍🏻 if similar data are needed in different formats, it's fine to have multiple output values.

examples/gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,84 @@
...
+      zones: [$(vars.zone)]
+
+# We need to do the following here:
+# 1- Install the GPUDirect binary and configure NCCL: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl
+# 2- Deploy NRI device injector plugin: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#nri-device-injector
+# 3- Deploy a test workload: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#test-workload
+# 4- Use recommended NCCL configuration settings to improve performance: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#environment-variables-nccl
+# 5- Add GPUDirect to manifests: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#add-gpudirect-manifests
+
+  - id: job-template
```
nick-stroud, line 76: Same thing, I am wondering if having the job template will cause more confusion than benefit at this point.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -96,7 +96,7 @@ variable "local_ssd_count_ephemeral_storage" {
   [See above](#local-ssd-storage) for more info.
   EOT 
   type        = number
-  default     = 0
+  default     = null
```
nick-stroud, line 99: null as an int can be a bit of an anti-pattern. I want to understand the driver for this. What behavior drove us to make this change.

examples/gke-a3-highgpu.yaml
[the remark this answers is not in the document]
ankitkinra: Created the task, and closing this conversation

examples/gke-a3-highgpu.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+  - id: job-template
+    source: modules/compute/gke-job-template
+    use: [a3-highgpu_pool]
+    settings:
+      image: nvidia/cuda:11.0.3-runtime-ubuntu20.04
+      command:
+      - nvidia-smi
+      node_count: 2
+    outputs: [instructions]
```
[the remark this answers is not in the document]
sharabiani: What's your suggestion? Should we remove it and put a description in output?

examples/gke-a3-megagpu.yaml
sharabiani: What's your suggestion? Should we remove it and put a description in output?

modules/scheduler/pre-existing-gke-cluster/versions.tf
sharabiani: Can you clarify this? I believe it is in the terraform root config.

modules/scheduler/gke-cluster/main.tf
```diff
@@ -85,7 +91,11 @@ resource "google_container_cluster" "gke_cluster" {
     autoscaling_profile = var.autoscaling_profile
   }
 
-  datapath_provider = var.enable_dataplane_v2 ? "ADVANCED_DATAPATH" : "LEGACY_DATAPATH"
+  datapath_provider = local.derived_enable_dataplane_v2 ? "ADVANCED_DATAPATH" : "LEGACY_DATAPATH"
+
+  enable_multi_networking = local.derived_enable_multi_networking
+
+  networking_mode = "VPC_NATIVE"
```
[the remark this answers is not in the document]
sharabiani: fixed

modules/scheduler/gke-cluster/main.tf
sharabiani: @nick-stroud while you are reviewing to see if I maybe solved this another way, I added enable_multi_networking output to multivpc module. It won't affect my changes on gke-cluster logic.

modules/scheduler/gke-cluster/README.md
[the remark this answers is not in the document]
ankitkinra: Sure we can close this now. Let's clean this up post automation.

modules/scheduler/pre-existing-gke-cluster/versions.tf
ankitkinra: Ack

examples/gke-a3-highgpu.yaml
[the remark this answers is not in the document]
sharabiani: removed.

examples/gke-a3-megagpu.yaml
sharabiani: removed.

modules/compute/gke-node-pool/variables.tf
sharabiani: description updated.

modules/scheduler/gke-cluster/main.tf
sharabiani: enable_multi_networking output removed from multivpc module.

examples/gke-multi-vpc-a3-megagpu-8g.yaml
sharabiani: linked

modules/compute/gke-node-pool/README.md
```diff
@@ -73,6 +73,11 @@ use-cases). In this case, ensure that you turn off the
 [enable_secure_boot](#input\_enable\_secure\_boot) option to allow unsigned
 kernel modules to be loaded.
 
+To maximize GPU network bandwidth, nodepools accept multiple VPCs. Pass a multivpc module to gke-node-pool module, and [take these steps] (https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl) to install GPUDirect, configure NCCL, use recommended settings, and add GPUDirect to your pods.
+
+> **_NOTE:_** You must [enable multi networking](https://cloud.google.com/kubernetes-engine/docs/how-to/setup-multinetwork-support-for-pods#create-a-gke-cluster) feature when creating the GKE cluster. Passing the multivpc module to a gke-cluster module enables multi networking on the cluster creation.
+> Passing the multivpc module to a gke-cluster or pre-existing-gke-cluster module [creates required network objects](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment) on the cluster for multi networking.
```
nick-stroud, line 79: Nit: Similarly, I would rewrite to say "When `gke-cluster` or `pre-existing-gke-cluster ` depends on `multivpc` (with the `use` keyword)," the [network objects](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment) required for multi networking will be created on the cluster."

In place of "Passing the multivpc module to a gke-cluster or pre-existing-gke-cluster module [creates required network objects](https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-gke-environment) on the cluster for multi networking."

modules/scheduler/gke-cluster/main.tf
```diff
@@ -30,6 +30,12 @@ locals {
   }]
 
   sa_email = var.service_account_email != null ? var.service_account_email : data.google_compute_default_service_account.default_sa.email
+
+  # additional VPCs enable multi networking 
+  derived_enable_multi_networking = length(var.additional_networks) > 0 ? true : coalesce(var.enable_multi_networking, false)
```
nick-stroud, line 35: I would reverse the logic on this. Since `enable_multi_networking` default is null, if it is set to either True or False then we know it is an explicit user action. We should not silently ignore this user action. 

Only if the variable is null then we should override with intuition based on additional network count.

modules/scheduler/gke-cluster/main.tf
```diff
@@ -30,6 +30,12 @@ locals {
...
+
+  # multi networking needs enabled Dataplane v2
+  derived_enable_dataplane_v2 = local.derived_enable_multi_networking ? true : coalesce(var.enable_dataplane_v2, false)
```
nick-stroud, line 38: Same as above, let us not silently ignore user input. Always default to taking user input and if it is invalid config then check with a pre-condition (as you do below).

modules/scheduler/gke-cluster/main.tf
```diff
@@ -168,6 +176,14 @@ resource "google_container_cluster" "gke_cluster" {
     ignore_changes = [
       node_config
     ]
+    precondition {
+      condition     = !(!coalesce(var.enable_dataplane_v2, true) && (coalesce(var.enable_multi_networking, false) || length(var.additional_networks) > 0))
```
nick-stroud, line 180: I believe pre-conditions can use local vars. I think better than testing the `var.` here and having coalesce in multiple places, it would be better to test your `local. derived_enable_dataplane_v2` and `local. derived_enable_multi_networking` which will actually be used to define the configuration. This in combination with always taking a user provided option first should make for clear configuration.

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -262,7 +262,7 @@ variable "authenticator_security_group" {
 variable "enable_dataplane_v2" {
   description = "Enables [Dataplane v2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2). This setting is immutable on clusters."
```
nick-stroud, line 263: Add null meaning to description. "If null, will default to false unless using multi-networking, in which case it will default to true"

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -296,3 +296,34 @@ variable "service_account" {
     error_message = "service_account is deprecated and replaced with service_account_email and scopes."
   }
 }
+
+variable "enable_multi_networking" {
+  description = "Enables [multi networking](https://cloud.google.com/kubernetes-engine/docs/how-to/setup-multinetwork-support-for-pods#create-a-gke-cluster) (Requires GKE Enterprise). This setting is immutable on clusters and enables [Dataplane V2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2?hl=en)."
```
nick-stroud, line 301: Add null meaning: "if null, will determine state based on if `additional_networks` are passed in."

examples/README.md
```diff
@@ -1563,7 +1568,7 @@ After provisioning the cluster and the nodepool, we need to do the following:
 
 This blueprint shows how to provision a GKE cluster with A3 High machines in the toolkit.
 
-After provisioning the cluster and the nodepool, we need to do the following:
+After provisioning the cluster and the nodepool, we need to do the [following](#gke-a3-high-setup-steps):
```
nick-stroud, line 1571: The link for "following" seems broken.

examples/README.md
```diff
@@ -1542,10 +1542,15 @@ cleaned up when the job is deleted.
 This blueprint shows how to provision a GKE cluster with A3 Mega machines in the toolkit.
 
 After provisioning the cluster and the nodepool, we need to do the following:
+
 1- Install the GPUDirect binary and configure NCCL: https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#install-gpudirect-tcpx-nccl
```
nick-stroud, line 1546: If you use `1.` instead of `1-` markdown will auto format into list, and you won't need white space between list items.

nick-stroud (changes requested):

examples/README.md
sharabiani: done

modules/scheduler/gke-cluster/main.tf
sharabiani: Here I used the preconditions as a replacement for cross-variables validation. I think here if we validate the input variables instead of processed local variables, we can give more to the point error description of the root problematic variables combinations. This validations also help us not to ignore user explicit actions; explicit settings are either acceptable or rejected here. What do you think?

modules/scheduler/gke-cluster/main.tf
sharabiani: if local.derived_enable_multi_networking is true (either via additional_networks or enable_multi_networking), enable_dataplane_v2 can be either null or explicitly True. Otherwise precondition blocks it.

modules/scheduler/gke-cluster/main.tf
sharabiani: if additional_networks is passed, enable_multi_networking can be either null or explicitly True. Otherwise precondition blocks it.

nick-stroud: `derived_enable_multi_networking = coalesce(var.enable_multi_networking, length(var.additional_networks) > 0 ? true : false)`

modules/scheduler/gke-cluster/main.tf
nick-stroud, line 180: nit: I would recommend the following

```
!(var.enable_dataplane_v2 == false && local.derived_enable_multi_networking)
```

The two changes being 1. direct comparison to false and 2. using the local value.

I know we agreed not to use local, but in this case I think we are just duplicating logic that you have already encoded into `local.derived_enable_multi_networking`. Furthermore, now that your logic for `local.derived_enable_multi_networking` always respects user inputs I think the case for using it is more clear cut. Furthermore, this pre-condition is really about validating `var.enable_dataplane_v2` which we are still using direct reference to input from customer.

---
**Not NIT**

An independent concern is that you currently have something like

`A & B | C` which evaluates to equivalent of `(A & B) | C`
This is not equivalent to `A & (B | C)` which I believe is what you are really after. This is a moot point if you make the change recommended above.

If you chose to stick with the current format I think you need an additional set of parens:

```
!(!coalesce(var.enable_dataplane_v2, true) && ((coalesce(var.enable_multi_networking, false) || length(var.additional_networks) > 0)))
```

nick-stroud (approved):

modules/scheduler/gke-cluster/main.tf
```diff
@@ -168,6 +176,14 @@ resource "google_container_cluster" "gke_cluster" {
...
+      error_message = "'enable_dataplane_v2' cannot be false when enabling multi networking."
+    }
+    precondition {
+      condition     = !(!coalesce(var.enable_multi_networking, true) && length(var.additional_networks) > 0)
+      error_message = "'enable_multi_networking' cannot be false when passing multivpc module."
```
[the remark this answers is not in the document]
sharabiani: done.

modules/scheduler/gke-cluster/main.tf
sharabiani: Updated to:
`!(!coalesce(var.enable_dataplane_v2, true) && local.derived_enable_multi_networking)`

This code `var.enable_dataplane_v2 == false` throws error if it is null.

sharabiani: About "Not NIT" part, Please correct me if I'm wrong. I think the additional parens were not needed because the suggestion would change `!(!A && (B | C))` to `!(!A && ((B | C)))` which do not seem different.

nick-stroud: Looking again, maybe I missed a parens.

PR #2805, opened 2024-07-24
Update `topology.conf` during `slurmsync`

mr0re1:
Naive implementation that always run `reconfigure` if any changes detected to topology.
No drawbacks since no changes are expected yet.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
@@ -44,6 +44,17 @@ def dict_to_conf(conf, delim=" ") -> str:
     )
 
 
+TOPOLOGY_PLUGIN_TREE = "topology/tree"
+
+def topology_plugin(lkp: util.Lookup) -> str:
+    """
+    Returns configured topology plugin, defaults to `topology/tree`.
+    """
+    cp, key = lkp.cfg.cloud_parameters, "topology_plugin"
+    if key not in cp or cp[key] is None:
+        return TOPOLOGY_PLUGIN_TREE
+    return cp[key]
+
 def conflines(lkp: util.Lookup) -> str:
     params = lkp.cfg.cloud_parameters
     def get(key, default):
@@ -113,7 +124,7 @@ def conflines(lkp: util.Lookup) -> str:
         "SuspendTimeout": get("suspend_timeout", 300),
         "TreeWidth": get("tree_width", default_tree_width),
         "JobSubmitPlugins": "lua" if any_tpu else None,
-        "TopologyPlugin": get("topology_plugin", "topology/tree"),
+        "TopologyPlugin": topology_plugin(lkp),
     }
     return dict_to_conf(conf_options, delim="\n")
 
@@ -497,10 +508,14 @@ def gen_topology(lkp: util.Lookup) -> TopologyBuilder:
     return bldr
 
 
-def gen_topology_conf(lkp: util.Lookup) -> None:
-    """generate slurm topology.conf from config.yaml"""
+def gen_topology_conf(lkp: util.Lookup) -> bool:
+    """
+    Generates slurm topology.conf.
+    Returns whether the topology.conf got updated.
+    """
     bldr = gen_topology(lkp).compress()
     conf_file = lkp.etc_dir / "cloud_topology.conf"
+    old_hash = util.hash_file(conf_file) if conf_file.exists() else ""
 
     with open(conf_file, "w") as f:
         f.writelines(FILE_PREAMBLE + "\n")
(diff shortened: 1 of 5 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -522,6 +529,11 @@ def main():
     except Exception:
         log.exception("failed to sync custom scripts")
 
+    try:
+        update_topology(lkp)
+    except Exception:
+        log.exception("failed to sync custom scripts")
```
nick-stroud, line 535: update error message

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
```diff
@@ -482,18 +482,24 @@ def gen_topology(lkp: util.Lookup) -> TopologyBuilder:
...
     conf_file = Path(lkp.cfg.output_dir or slurmdirs.etc) / "cloud_topology.conf"
+    old_hash = util.hash_file(conf_file) if conf_file.exists() else ""
+
     with open(conf_file, "w") as f:
         f.writelines(FILE_PREAMBLE + "\n")
         for line in bldr.render_conf_lines():
             f.write(line)
             f.write("\n")
         f.write("\n")
-    util.chown_slurm(conf_file, mode=0o600)
```
nick-stroud, line 495: Why is this removed? Seems unrelated.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -522,6 +529,11 @@ def main():
     except Exception:
         log.exception("failed to sync custom scripts")
 
+    try:
+        update_topology(lkp)
```
nick-stroud, line 533: We are now generating topology on every sync. Description says "no drawbacks since no changes expected".

1. What is cost of generating topology on large cluster, say 5000 nodes. Even though we may not call reconfigure on every sync could this impact sync run time? Should we guard this behind a flag? For example when topology/none is selected then this does not run?
2. Why do we expect it not to change? Probably just a lack of understanding of our current state but seems like whole reason for this feature is we expect it to change.

nick-stroud (changes requested):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -488,10 +488,11 @@ def reconfigure_slurm():
...
             log.info("Restarting slurmctld to make changes take effect.")
             try:
+                # TODO: consider removing "restart" since "reconfigure" should restart slurmctld as well
                 run("sudo systemctl restart slurmctld.service", check=False)
-                run(f"{lkp.scontrol} reconfigure", timeout=30)
-            except Exception as e:
-                log.error(e)
+                util.scontrol_reconfigure(lkp)
+            except Exception:
+                log.exception("failed to reconfigure slurmctld")
```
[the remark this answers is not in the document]
mr0re1: > You are swallowing exception details here. Why? 

No, observe:

```py
import logging
log = logging.getLogger()
def fn(): raise ValueError("gg")

print("=== try log.exception")
try:
  fn()
except Exception:
  log.exception("terrible event")

print("=== try log.error")
try:
  fn()
except Exception as e:
  log.error(e)
```

```sh
=== try log.exception
terrible event
Traceback (most recent call last):
  File "/usr/local/google/home/orlov/tmp/s.py", line 9, in <module>
    fn()
  File "/usr/local/google/home/orlov/tmp/s.py", line 5, in fn
    def fn(): raise ValueError("gg")
              ^^^^^^^^^^^^^^^^^^^^^^
ValueError: gg
=== try log.error
gg
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
mr0re1: `install_topology_conf` does it already. No need to duplicate.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
mr0re1: > What is cost of generating topology on large cluster, say 5000 nodes.

It's super (cheap < `0.01 sec`) currently, without using `physicalHost`

> Should we guard this behind a flag? For example when topology/none is selected then this does not run?

Good call, added condition to do it IFF `topology/tree` is used.

> Why do we expect it not to change?

Because only possible change (until we use `physicalHost`) is addition/removal of nodes, that in order should be cased by changes in configuration -> full blown reconfigure will happen before `update_topology`

nick-stroud (approved):

PR #2817, opened 2024-07-27
SlurmGCP. Fix broken `--trace-api` flag.

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/variables.tf
```diff
@@ -35,7 +35,7 @@ variable "enable_debug_logging" {
 
 variable "extra_logging_flags" {
```
harshthakkar01, line 36: Should we just remove this or update this extra_loggin_flags --> trace_api ?

harshthakkar01 (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/variables.tf
mr0re1: This is out of scope, let's not change interfaces

PR #2820, opened 2024-07-27
SlurmGCP. Rework logging

mr0re1:
* Don't initialize file-specific logger, instead use "root logger" configured with proper destination file;
* Don't pass `log` as an argument to the `setup_network_storage.py` functions, use "root logger" instead;
* Stop using excessive `argparse.ArgumentParser` initialization since none of scripts define `__doc__`.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
@@ -19,6 +19,7 @@ import os
 import sys
 import stat
 import time
+import logging
 
 import shutil
 from pathlib import Path
@@ -29,6 +30,8 @@ import util
 from util import lkp, run, cfg, dirs, separate
 from more_executors import Executors, ExceptionRetryPolicy
 
+
+log = logging.getLogger()
 
 def mounts_by_local(mounts):
     """convert list of mounts to dict of mounts, local_mount as key"""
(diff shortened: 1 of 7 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -580,9 +571,8 @@ def log_api_request(request):
     log.debug(f"{inspect.stack()[1].function}:\n{pretty_req}")
 
 
-def handle_exception(exc_type, exc_value, exc_trace):
+def _handle_exception(exc_type, exc_value, exc_trace):
     """log exceptions other than KeyboardInterrupt"""
-    # TODO does this work?
```
harshthakkar01, line 585: Was this here to verify something ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
```diff
@@ -30,6 +31,8 @@
 from more_executors import Executors, ExceptionRetryPolicy
 
 
+log = logging.getLogger()
```
harshthakkar01, line 34: for this root logger, what would be destination file ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
mr0re1: Yep, verified.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
mr0re1: It depends on what "main" uses it.

harshthakkar01 (approved):

PR #2843, opened 2024-07-31
Add `instance_properties` var to `nodeset`

mr0re1:
Add `instance_properties` var to `nodeset`

**Motivation:** to unblock advanced use cases not fully supported by SlurmGCP.


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -480,3 +480,15 @@ variable "network_storage" {
...
   default = []
 }
+
+
+variable "instance_properties" {
+  description = <<-EOD
+    Override the instance properties.
+    See https://cloud.google.com/compute/docs/reference/rest/v1/regionInstances/bulkInsert
+    If any sub-field (e.g. scheduling) is set, it will override the values computed by 
+    SlurmGCP and ignoring values of provided vars.
```
nick-stroud, line 490: What is the result of bad properties? What if I define key that does not exist? What if I define bad value? Will it error and where will it show.

I realize this is for experimental features. Really I am just wondering, should we say here, "errors generated by bad inputs can be found in /var/log/slurm/resume.log"?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -480,3 +480,15 @@ variable "network_storage" {
   }))
   default = []
 }
+
+
+variable "instance_properties" {
```
nick-stroud, line 485: How can we reiterate that this is experimental and will not receive full support (similar to what you put in pr description)?

Maybe add "used to test features not supported by Slurm GCP, recommended for advanced usage only."

nick-stroud (approved):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
mr0re1: > What if I define key that does not exist?

I've observed wrong key being silently ignored.

>  What if I define bad value? 
>  should we say here, "errors generated by bad inputs can be found in /var/log/slurm/resume.log"?

Bulk insert will fail and `srun` will spit out a nice error message, really apparent without digging logs.

PR #2853, opened 2024-08-01
Add `sort_nodes.py`

mr0re1:
````sh
$ /slurm/scripts/sort_nodes.py --help
usage: sort_nodes.py [-h] [--nodelist NODELIST] [--ntasks-per-node NTASKS_PER_NODE] [--output OUTPUT]

This script sorts nodes based on their `physicalHost`.

See https://cloud.google.com/compute/docs/instances/use-compact-placement-policies

You can reduce latency in tightly coupled HPC workloads (including distributed ML training)
by deploying them to machines that are located close together.
For example, if you deploy your workload on a single physical rack, you can expect lower latency
than if your workload is spread across multiple racks.
Sending data across multiple rack requires sending data through additional network switches.

Example usage:
``` my_sbatch.sh
#SBATCH --ntasks-per-node=8
#SBATCH --nodes=64

export SLURM_HOSTFILE=$(/slurm/scripts/sort_nodes.py)

srun -l hostname | sort
```

options:
  -h, --help            show this help message and exit
  --nodelist NODELIST   Slurm 'hostlist expression' of nodes to sort, if not set the value of SLURM_NODELIST environment variable will be used
  --ntasks-per-node NTASKS_PER_NODE
                        Number of times to repeat each node in resulting sorted list.
                        If not set, the value of SLURM_NTASKS_PER_NODE environment variable will be used,
                        if neither is set, defaults to 1
  --output OUTPUT       Output file to write, defaults to 'hosts.<uuid>'
````

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
@@ -67,28 +67,25 @@ def order(paths: List[List[str]]) -> List[str]:
                 n.children[v] = Vert(v, n)
             n = n.children[v]
 
-    def post_order_traversal(v: Vert, f: Callable[[Vert], None]) -> None:
-        # freeze order, since lex_key can be mutated
-        children = list(sorted(v.children.values(), key=lambda u: u.lex_key))
-        for u in children:
-            post_order_traversal(u, f)
-        f(v)
-    
     # propagate "lexicographical key" from the leaves to the root
-    def lex_key_prop(v: Vert) -> None:
+    def set_lex_key(v: Vert) -> None:
         if not v.children: # this is a node
             v.lex_key = name_idx[v.name]
         else:
+            for u in v.children.values():
+                set_lex_key(u)
             v.lex_key = min(u.lex_key for u in v.children.values())
-    post_order_traversal(root, lex_key_prop)
+    set_lex_key(root)
 
     # gather leafs
     result = []
-    def collect_nodes(v: Vert) -> None:
-        if not v.children: 
+    def gather_nodes(v: Vert) -> None:
+        if not v.children: # this is a node
             result.append(v.name)
-    post_order_traversal(root, collect_nodes)
+        for u in v.children.values():
+            gather_nodes(u)
     
+    gather_nodes(root)
     return result
 
 
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
@@ -37,28 +37,30 @@ srun -l hostname | sort
 import os
 import subprocess
 import uuid
-from typing import List, Optional, Dict, Callable
+from typing import List, Optional, Dict
+from collections import OrderedDict
 
 def order(paths: List[List[str]]) -> List[str]:
     """
-    Orders the leafs of the tree in a way that minimizes the sum of distance in between 
+    Orders the leaves of the tree in a way that minimizes the sum of distance in between 
     each pair of neighboring nodes in the resulting order.
     The resulting order will always start from the first node in the input list.
-    The ordering is "stable" with respect to the input order of the leafs i.e.
+    The ordering is "stable" with respect to the input order of the leaves i.e.
     given a choice between two nodes (identical in other ways) it will select "nodelist-smallest" one.
+
+    Returns a list of nodenames, ordered as described above.  
     """
     if not paths: return []
-    # lookup table for node names order
-    name_idx = {p[-1]: i for i, p in enumerate(paths)}
     class Vert:
         "Represents a vertex in a *network* tree."
         def __init__(self, name: str, parent: "Vert"):
             self.name = name
             self.parent = parent
-            self.children = {}
-            self.lex_key: int = -1
+            # Use `OrderedDict` to preserve insertion order
+            # TODO: once we move to Python 3.7+ use regular `dict` since it has the same guarantee
+            self.children = OrderedDict()
 
-    # build a tree
+    # build a tree, children are ordered by insertion order
     root = Vert("", None)
     for path in paths:
         n = root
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_topology.py [hunk 1]
```diff
@@ -89,3 +91,20 @@ def tpu_se(ns: TstNodeset) -> TstTPU:
     conf.gen_topology_conf(util.Lookup(cfg))
     want_written = PRELUDE + "\n".join(want_compressed) + "\n\n"
     assert open(cfg.output_dir + "/cloud_topology.conf").read() == want_written
+
```
samskillman, line 94: Suggest another test here demonstrating sorting of something like slurm-[10,5,12,7,22].

samskillman (changes requested):
Would suggest adding a test of numerical hostnames, and needs fix to get the lowest number first. Could be combined with a test of two groups of three nodes at depth 2, where the lowest number node is in the node with a higher first level string. something like 

x/slurm-1
x/slurm-2
x/slurm-3
y/slurm-0
y/slurm-4
z/slurm-10

I'd expect 0,4,10,1,2,3

mr0re1:
@samskillman  , for 
> something like
x/slurm-1
x/slurm-2
x/slurm-3
y/slurm-0
y/slurm-4
z/slurm-10
I'd expect 0,4,10,1,2,3

The order `0, 4, 1, 2, 3, 10` seems to be more natural is the particular reason you wrote `0,4,10,1,2,3`?

mr0re1:
* Addressed comments about name-order:
Don't use naive lexicographic order anywhere, always rely on order of names in `nodelist`.
Significantly refactor `def order`.

* Rename `topologicaly_sort_nodes -> sort_nodes` to reduce confusion, since order is not "topological".

samskillman:
> @samskillman , for
> 
> > something like
> > x/slurm-1
> > x/slurm-2
> > x/slurm-3
> > y/slurm-0
> > y/slurm-4
> > z/slurm-10
> > I'd expect 0,4,10,1,2,3
> 
> The order `0, 4, 1, 2, 3, 10` seems to be more natural is the particular reason you wrote `0,4,10,1,2,3`?

I agree - I meant for that z to be a y. I think either would be acceptable solutions in principle, but I agree 0,4,1,2,3,10 is better. My mistake on the typo.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
```diff
@@ -0,0 +1,183 @@
...
+    zone = f"zone_{inst.zone}"
+    if not inst.physical_host:  # node without physical host info (e.g. no placement policy)
+        return [zone, "unknown", node_name]
+
+    assert inst.physical_host.startswith("/"), f"Unexpected physicalHost: {inst.physical_host}"
+    parts = inst.physical_host[1:].split("/")
+    if len(parts) >= 4:
+        return [*parts, node_name]
+    elif len(parts) == 3:
+        return [zone, *parts, node_name]  # add zone
```
nick-stroud, line 111: It is not intuitive to me what is the significance of being >=4 or ==3. Do some physical hosts include zone? I am cautious about the special casing here since I expect physical host may change. At min can you add comment as to what you are checking for here, as you have done for other cases above.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
```diff
@@ -0,0 +1,183 @@
...
+from typing import List, Optional, Dict, Callable
+
+def order(paths: List[List[str]]) -> List[str]:
+    """
+    Orders the leafs of the tree in a way that minimizes the sum of distance in between 
+    each pair of neighboring nodes in the resulting order.
+    The resulting order will always start from the first node in the input list.
+    The ordering is "stable" with respect to the input order of the leafs i.e.
+    given a choice between two nodes (identical in other ways) it will select "nodelist-smallest" one.
+    """
```
nick-stroud, line 49: From description and signature, the interface is not clear to me. It took some digging to understand that it is returning a list of node names (leaves) and that the node names are expected to be the last element in a path, I had been thinking strictly physical_host.

Something like the following would be helpful:
```
return: a list containing each leaf of the input paths, ordered as described above.  
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
```diff
@@ -0,0 +1,183 @@
...
+    for path in paths:
+        n = root
+        for v in path:
+            if v not in n.children:
+                n.children[v] = Vert(v, n)
+            n = n.children[v]
+
+    # propagate "lexicographical key" from the leaves to the root
+    def set_lex_key(v: Vert) -> None:
+        if not v.children: # this is a node
```
nick-stroud, line 72: nit: s/node/leaf/

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
```diff
@@ -0,0 +1,183 @@
...
+```
+"""
+import os
+import subprocess
+import uuid
+from typing import List, Optional, Dict, Callable
+
+def order(paths: List[List[str]]) -> List[str]:
+    """
+    Orders the leafs of the tree in a way that minimizes the sum of distance in between
```
nick-stroud, line 44: s/leafs/leaves/

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # propagate "lexicographical key" from the leaves to the root
+    def set_lex_key(v: Vert) -> None:
+        if not v.children: # this is a node
+            v.lex_key = name_idx[v.name]
+        else:
+            for u in v.children.values():
+                set_lex_key(u)
+            v.lex_key = min(u.lex_key for u in v.children.values())
+    set_lex_key(root)
```
nick-stroud, line 78: I don't see where lex_key is ever used, which I think is what you want as dict maintains arg input order where desired.

nick-stroud (changes requested):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
mr0re1: Good catch. Looking into it, lead to refactoring, see the diff.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/sort_nodes.py
mr0re1: Made it more robust by removing assertion about length of `parts`.

nick-stroud (approved):

PR #2874, opened 2024-08-05
Add doc about DWS Flex Start

mr0re1:


docs/slurm-dws-flex.md
```diff
@@ -0,0 +1,24 @@
...
+        reservationAffinity:
+          consumeReservationType: NO_RESERVATION
+        scheduling:
+          maxRunDuration: { seconds: $(2 * 60 * 60) } # 2 hours
+          onHostMaintenance: TERMINATE
+          instanceTerminationAction: DELETE
+      # the rest of the settings, e.g. node_count_static, machine_type, additional_disks, etc.
+```
+
+**All** fields in `instanceProperties` should match provided values, except for `maxRunDuration`, which should be set to the desired duration in seconds (up to 604800 = 7 days).
```
nick-stroud, line 24: Add: 
Note: DWS Flex may not be available for all machine types. Check "here" to see if your machine type is supported.

nick-stroud, line 24: Add:
Warning: The use of the `instance_properties` setting directly overrides bulk insert api parameters. While the documented sample was tested at the time of publication, it is not regression tested and may cease to work based on changes in the build insert api.

mr0re1: @nick-stroud , is there a "here"? Please provide any public documentation if you know any.

mr0re1: There is no adequate public documentation to reference and any reasons to confidently imply limitations. Resolving

nick-stroud (approved):

PR #2882, opened 2024-08-06
Add setting `allow_automatic_updates: false` to examples

rohitramu:


harshthakkar01:
Can we make the default to false instead of updating all bps ?

rohitramu:
> Can we make the default to false instead of updating all bps ?

The idea is to ensure that disabling automatic updates is "opt-in" (i.e. not the default).  This is the same reason automatic updates are not disabled by default in the HPC VM images.  Additionally, this feature only works for certain VM images.  By setting `allow_automatic_updates: true` by default, we ensure the default value works for all images, avoiding confusing error messages for users unfamiliar with the relationship between this feature and the selected VM image.

harshthakkar01 (approved):

PR #2891, opened 2024-08-08
GKE Blueprints to support reservations

arajmane-g:
Toolkit users will be able to utilise reservations when deploying GKE clusters. They can use any reservation or specific reservations available in the zone.

Reservations are assumed to exist.

### Implementation
Introduce new blueprint settings for the module `modules/compute/gke-node-pool`: {`reservation_type`, `specific_reservation.key`, `specific_reservation.values`}. Pass these values to the provider as part of the `reservation_affinity` block which is part of `node_config`.

Use precondition to perform validations. Because, you cannot refer other variables in variable validations in Terraform versions older than 1.9. Here, we need to ensure the `reservation_type` and `specific_reservation` values are consistent. For example, if you don't want to use any reservation, you shouldn't be setting `specific_reservation`.

Few caveats wrt the validations:
- We cannot ensure `specific_reservation.values` to be a non empty array. We want `specific_reservation.values` to be optional. And, when it is not set along with `{NO_RESERVATION, ANY_RESERVATION}`, `Terraform plan` won't allow `length(null)`.
- We are not validating the existence of reservations.

### Testing
Since, no unit testing exists for Terraform modules at the moment, manually tested invalid values. To list a few tests

- `specific_reservation` should not be allowed with `{NO_RESERVATION, ANY_RESERVATION}`
- `specific_reservation.key` should not other than `compute.googleapis.com/reservation-name` when targeting `SPECIFIC_RESERVATION`
- `specific_reservation.values` should not null when targeting `SPECIFIC_RESERVATION`

For e2e testing, an integration test will be added as a fast follow up to the PR. Till then manual verification was done by deploying different blueprints:
- The change is backward compatible. Not providing the relevant settings will deploy the cluster as if no reservation is to be used
- When ANY_RESERVATION is targeted, any "Automatic" reservation in the zone is utilised by the cluster
- When SPECIFIC_RESERVATION is targeted, the provided specific reservation is utilised by the cluster

### Submission Checklist


modules/compute/gke-node-pool/variables.tf
```diff
@@ -265,3 +265,31 @@ variable "service_account" {
...
+    key    = string
+    values = list(string)
+  })
+  default = {
+    key    = null
+    values = null
+  }
+
+  validation {
+    condition     = (var.specific_reservation.key == "compute.googleapis.com/reservation-name" && length(var.specific_reservation.values) > 0) || (var.specific_reservation.key == null && var.specific_reservation.values == null)
```
arajmane-g, line 292: will check if the condition can be made more readable by spanning it over two lines.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -265,3 +265,31 @@ variable "service_account" {
...
+  default     = null
+
+  validation {
+    condition     = contains(["NO_RESERVATION", "ANY_RESERVATION", "SPECIFIC_RESERVATION"], var.reservation_type)
+    error_message = "Accepted values are: {NO_RESERVATION, ANY_RESERVATION, SPECIFIC_RESERVATION}"
+  }
+}
+
+variable "specific_reservation" {
+  description = "Reservation resources to consume when targeting SPECIFIC_RESERVATION. Specify `compute.googleapis.com/reservation-name` as the key and the list of reservation names as the value."
```
arajmane-g, line 281: Will check how to avoid long lines and in turn, text wrap

modules/compute/gke-node-pool/main.tf
```diff
@@ -150,6 +150,12 @@ resource "google_container_node_pool" "node_pool" {
         "net.ipv4.tcp_wmem" = "4096 16384 16777216"
       }
     }
+
+    reservation_affinity {
+      consume_reservation_type = var.reservation_type
```
ankitkinra, line 155: Does this allow customer to not specify reservation affinity at all ?

arajmane-g: Thanks for the comment. PR in its current state will not. 

Still working on the defaults considering possible scenarios. Specifically, treatment of null vs empty strings by the provider. (being my fist Terraform PR) going to rely on tests to come up with those.

For `consume_reservation_type`, `NO_RESERVATION` is a good default. It will take care of no reservation affinity at all (per the provider documentation) as long as `key`, `values` defaults are such that provider ignores them.

ankitkinra: I think there is this pattern here : https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/9e68ecc363ebf212c3f574db9b25aa2f86f6db97/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf#L141

```
count = var.cloudsql != null ? 1 : 0
```

Similar to this : https://github.com/hashicorp/terraform/issues/21953

Can you see if that helps ?

arajmane-g: This pattern is not applicable here. It is used to disable a "resource". What we are trying to do is to conditionally not include an "argument".

Arguments with `null` values are treated unset and omitted by Terraform ([documentation reference](https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements#conditionally-omitted-arguments:~:text=Terraform%20v0.12%20now%20allows%20assigning%20the%20special%20value%20null%20to%20an%20argument%20to%20mark%20it%20as%20%22unset%22.)). Checked this with by creating a terraform plan. Kept `key`, `values` null with `NO_RESERVATION` as the default value of `consume_reservation_type`. In the generated plan, only the `consume_reservation_type` argument is part of the `reservation_affinity` block

ankitkinra: Got, does this pattern work ? 

```
dynamic "gcfs_config" {
      for_each = var.enable_gcfs ? [1] : []
      content {
        enabled = true
      }
```

or https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/e01aa0d2de300ed3e9f7ec3eb484a5fe06546da0/modules/compute/gke-node-pool/main.tf#L131

I am also a bit new to terraform so if the above pattern also doesn't work, let's ask around. (Hopefully it is a not a huge change)

arajmane-g: > This pattern is not applicable here. It is used to disable a "resource". What we are trying to do is to conditionally not include an "argument".
> 
> Arguments with `null` values are treated unset and omitted by Terraform ([documentation reference](https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements#conditionally-omitted-arguments:~:text=Terraform%20v0.12%20now%20allows%20assigning%20the%20special%20value%20null%20to%20an%20argument%20to%20mark%20it%20as%20%22unset%22.)). Checked this with by creating a terraform plan. Kept `key`, `values` null with `NO_RESERVATION` as the default value of `consume_reservation_type`. In the generated plan, only the `consume_reservation_type` argument is part of the `reservation_affinity` block

I meant, the `null` arguments should work for us.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -265,3 +265,29 @@ variable "service_account" {
     error_message = "service_account is deprecated and replaced with service_account_email and scopes."
   }
 }
+
+variable "reservation_type" {
+  description = "Type of reservation to consume"
```
ankitkinra, line 270: Can we add here how to create this reservation ? And also mention that we assume this is a valid reservation ? 

Also what happens when this is not a valid reservation, does the GKE cluster creation throws error ? And is that error visible to the customer ?

ankitkinra (commented):
Mostly LGTM at this stage, just waiting on one update.

modules/compute/gke-node-pool/variables.tf
arajmane-g: > Can we add here how to create this reservation ? And also mention that we assume this is a valid reservation ?

Done. Updated the specific_reservation's description. I think, it's more relevant there.

> Also what happens when this is not a valid reservation, does the GKE cluster creation throws error ? And is that error visible to the customer ?

If it is not a valid reservation or the reservation does not have enough capacity, the cluster (and node pool) creation succeeds with problems**  which are not apparent in the Toolkit's deploy logs. 

**: Node pool creation succeeds but the GCE Instance creation fails in both of the mentioned cases. Failure reason is available in the node pool's status. I am not sure whether it's correct behaviour. We can check this with the GKE team if this is working as designed or a bug -- node pool creation succeeded but the nodes backing the pool cannot be created.

As far as Toolkit is concerned, do you think, the variable description asking to ensure the name and capacity of the reservation is sufficient? Ideally, the error should be visible in the Toolkit's logs. I will explore how to modify Toolkit's logs to use the node pool's status (it is available in the node-pools describe API)

I should also mention that as soon as a matching reservation is created or additional capacity is added (offline), the GCE Instance creation succeeds.

ankitkinra: Thanks for checking, let's add a task to throw an error if the reservation is not found at the time of "blueprint deployment" to unblock this reservation.

ankitkinra (approved):

PR #2895, opened 2024-08-08
Delete at most one resource policy at a time

mr0re1:
* Delete at most one resource policy at a time
* Don't fail if some policy can't be deleted

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
@@ -44,13 +44,9 @@ done
 
 echo "Deleting resource policies"
 policies_filter="name:${cluster_name}-*"
-while true; do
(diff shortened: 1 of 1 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
nick-stroud, line 1: In addition to running integration tests please manually test for a case with 2 active placement policies.

nick-stroud, line 1: If this is hotfix to main and there is code change then we also need to roll the patch version number.

nick-stroud (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/scripts/cleanup_compute.sh
mr0re1: done

mr0re1: ack

PR #2914, opened 2024-08-12
Install `script_nodes.py` along slurm binaries (`/usr/local/bin`)

mr0re1:


tpdownes (approved):
`/usr/local/sbin` (by default) will make this more accessible to users.

PR #2916, opened 2024-08-12
Add setting `allow_automatic_updates: false` to examples

rohitramu:


rohitramu:
This is a cherrypick of #2882 .  It contains a fix for the `cloud-batch` test which is currently broken in 'release-candidate'.

nick-stroud (approved):

PR #2917, opened 2024-08-12
SlurmGCP. Stop using global mutable `lkp`

mr0re1:
Replace `lkp` with `lookup() -> Lookup`.

**Motivation:**
This showcases the problem of updating (slurmsync & setup) global `lkp`:

```py
# util.py
lkp = { "value": "watermelon" }

# some.py
from util import lkp
def get_value(): return lkp["value"]

# setup.py
import util
import some

util.lkp = {"value": "apple"} 
print(some.get_value())
```

```sh
$ python showtime/setup.py
watermelon
```

The fix would be:

```py
# instead of
#from util import lkp
#def get_value(): return lkp["value"]
# do:
import util
def get_value(): return util.lkp["value"]
```

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
@@ -45,7 +45,7 @@ from util import (
     TPU,
     chunked,
 )
-from util import lkp, CONFIG_FILE
+from util import lookup, CONFIG_FILE
 from suspend import delete_instances
 from resume import start_tpu
 import conf
@@ -72,9 +72,9 @@ NodeStatus = Enum(
 
 
 def start_instance_op(inst):
-    return lkp.compute.instances().start(
-        project=lkp.project,
-        zone=lkp.instance(inst).zone,
+    return lookup().compute.instances().start(
+        project=lookup().project,
+        zone=lookup().instance(inst).zone,
         instance=inst,
     )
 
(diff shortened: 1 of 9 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -72,26 +72,24 @@
...
+        zone=lookup().instance(inst).zone,
         instance=inst,
     )
 
 
 def start_instances(node_list):
     log.info("{} instances to start ({})".format(len(node_list), ",".join(node_list)))
 
-    normal, tpu_nodes = separate(lkp.node_is_tpu, node_list)
-    invalid, valid = separate(lambda inst: bool(lkp.instance), normal)
```
harshthakkar01, line 86: why this is removed ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -87,16 +87,9 @@ def delete_tpu_instances(instances):
 
 def delete_instances(instances):
     """delete instances individually"""
-    invalid, valid = separate(lambda inst: bool(lkp.instance(inst)), instances)
```
harshthakkar01, line 90: why this is removed ?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
mr0re1: Because it's a nonsense
`bool(lkp.instance)` is attempt to cast function (not result of function call, but function itself to bool), see:

```py
>>> def const_false(): return False
...
>>> bool(const_false)
True
```

As result `invalid` is always empty, all items go into `valid`.
Overall this change does affect behavior, just removes garbage.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
mr0re1: This is one is a bit of stretch. Reverted to reduce the scope.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
mr0re1: Looking at code path, only "valid" (i.e. running) instances can be supplied here.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -96,7 +96,7 @@ def delete_instances(instances):
 
     requests = {inst: delete_instance_request(inst) for inst in valid}
 
-    log.info(f"delete {len(valid)} instances ({to_hostlist_fast(valid)})")
```
harshthakkar01, line 99: do we need to flip this as well ?

mr0re1: Yes, thank you for spotting !

PR #2918, opened 2024-08-12
SlurmGCP. Store "extra files" hash in config

mr0re1:
No changes in behavior other than making `config.yaml` creation/update to wait until other files are written.

**Motivation:**
* Enable `slurmsync` to perform informed checks & updates while handling changes in "extra-files";
* Pre-requisite to remove dependency on `slurm_file_checksum` from controller.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
@@ -40,6 +40,8 @@ resource "random_uuid" "cluster_id" {
 ##################
 
 locals {
+  tp = "${local.bucket_dir}/" # prefix to trim from the bucket path to get a "file name"
+
   config = {
     enable_slurm_gcp_plugins = var.enable_slurm_gcp_plugins
     enable_bigquery_load     = var.enable_bigquery_load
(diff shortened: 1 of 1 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
```diff
@@ -95,6 +97,29 @@ locals {
 
     # Providers
     endpoint_versions = var.endpoint_versions
+
+    # Extra-files MD5 hashes
+    # Makes config file creation depend on the files
+    # Allows for informed updates & checks on slurmsync side
+    slurm_gcp_scripts_md5 = google_storage_bucket_object.devel.md5hash,
```
tpdownes, line 104: Just confirming that these `*_md5` are to-be-used by https://github.com/GoogleCloudPlatform/slurm-gcp at some later point in time?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/outputs.tf
```diff
@@ -55,6 +55,7 @@ output "nodeset_tpu" {
 }
 
 output "checksum" {
-  description = "Checksum of all files written to the bucket."
-  value       = local.checksum
+  description = "Checksum of config written to the bucket."
+  # wrap into `md5` to get nice representation suitable for use in labels
```
tpdownes, line 59: `nit`: I would suggest being more explicit here about converstion from base 64 to hex. I believe issue is b64 contains characters that are not allowed in label values.

https://cloud.google.com/compute/docs/labeling-resources#requirements

tpdownes (approved):
I'd appreciate a response to the question (to confirm my understanding matches the description) and you can choose to address `nit` or not.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
mr0re1: yes, it's not explicitly used now, implicitly it's used for temporal ordering.

PR #2919, opened 2024-08-13
SlurmGCP. Don't fetch config in `util.py`

mr0re1:
* Only fetch config (from GCS) during `setup` and `slurmsync`.
* During `setup` wait indefinitely until  config is available.
* Use only "devel" checksum on controller labels, this would allow controller instance to start **before** config is stored in a bucket, though it will still wait for "devel.zip" to upload.

**Motivation:** allow for separate lifecycle of config and cluster.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
@@ -445,8 +445,18 @@ def setup_compute():
 
 def main():
     start_motd()
-    configure_dirs()
+    
+    sleep_seconds = 5
+    while True:
+        try:
+            _, cfg = util.fetch_config()
+            util.update_config(cfg)
+            break
+        except Exception as e:
+            log.exception(f"could not fetch config, sleeping for {sleep_seconds}s")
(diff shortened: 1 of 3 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/outputs.tf
```diff
@@ -55,6 +55,7 @@ output "nodeset_tpu" {
 }
 
 output "checksum" {
-  description = "Checksum of all files written to the bucket."
-  value       = local.checksum
```
harshthakkar01, line 59: what about checksum for other files ?

mr0re1: we don't need them. This output has been added in a first place, so we can order terraform resources creation, namely prevent controller startup before config is saved to GCS. After this change controller is "resilient enough" to  wait until config is created.

harshthakkar01:
PS test failure.

mr0re1:
Reverted Terraform changes, keeping python ones. There is another conflict, slurmsync is configured to run regardless of "setup status". Will need to fix it first before removing terraform dependency.

harshthakkar01 (approved):

PR #2920, opened 2024-08-13
Build is taking longer, extend timeout so test will pass while investigate

rohitramu:


rohitramu:
This is a cherry-pick of #2897.

nick-stroud (approved):

PR #2922, opened 2024-08-13
SlurmGCP. Minor refactoring.

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1732,39 +1713,34 @@ def machine_types(self, project=None):
...
             }
         else:
-            machines = self.machine_types(project=project)
-            machine_info = next(iter(machines[machine_type].values()), None)
-            if machine_info is None:
+            machines = self.machine_types()
+            if machine_type not in machines:
                 raise Exception(f"machine type {machine_type} not found")
+            per_zone = machines[machine_type]
+            assert per_zone
```
abbas1902, line 1734: Wouldn't the if-condition above always catch per_zone remaining False?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -289,21 +289,17 @@ def nodes_unknown():
...
+        NodeStatus.orphan: nodes_delete,
+        NodeStatus.power_down: nodes_power_down,
+        NodeStatus.preempted: lambda: (nodes_down(), nodes_restart()),
+        NodeStatus.restore: nodes_idle,
+        NodeStatus.resume: nodes_resume,
+        NodeStatus.terminated: nodes_down,
+        NodeStatus.unbacked: nodes_down,
+        NodeStatus.unchanged: lambda: None,
+        NodeStatus.unknown: nodes_unknown,
+    }[status]()
```
abbas1902, line 302: Just curious what this ```[status]()``` syntax does here

mr0re1: Here we create an anonymous map `NodeStatus -> function`, 
Then we perform a lookup function attributed to given `status`, by doing`[status]`
Finally we call found function `()`.

You can think of this construction as `dict(...).get(status).__call__()`;

The primary difference between old and new code is usage of `[status]` instead of `get(status)`.
If `X` IS NOT present in dict, `get(X)` will return `None`, while `[X]` will raise a `KeyError(X)` exception.

The exception is more preferable since it's more explicit about what went wrong `KeyError: NodeStatus.potatus"`. 
While `get` would successfully return `None`, but calling it `()` would fail with cryptic:
`TypeError: 'NoneType' object is not callable` 


See https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
mr0re1: `machines` is `Dict[str, Dict[str, something]]`, say 
```py
machines = {
  "sewing": {
     "albania": 33,
     "bobruysk": 77
   },
   "soda": {},
}

assert "sewing" in machines # OK
assert "soda" in machines   # OK
assert machines["sewing"]   # OK - not empty dict
assert machines["soda"]     # FAIL - empty dict
```

The reason I'm using crude `assert` without explanatory message for user, is that because I expect it to never fire. By the way of construction of `machines` by `machine_types` functions, we know that  if `machine_type` is present in result -> `machine_type` is present in at least one `zone`. This is assertion of code invariant, while also useful for calming type-checkers.

mr0re1: About "calming type-checkers". Consider:

```py
# zebu.py
from typing import Optional

def az(coin: int) -> Optional[str]:
  return "az" if coin % 2 else None

def gel():
  za = az(1)
  print(za.lower())
```

```sh
$ mypy zebu.py  --check-untyped-defs
zebu.py:8: error: Item "None" of "str | None" has no attribute "lower"  [union-attr]
Found 1 error in 1 file (checked 1 source file)
```

We can hint `mypy`:

```py
...
def gel():
  za = az(1)
  assert za is not None # <--
  print(za.lower())
```

```sh
$ mypy zebu.py  --check-untyped-defs
Success: no issues found in 1 source file
```

abbas1902: appreciate the example here, makes sense!

abbas1902 (approved):

Issue #2774, opened 2024-07-17
Using a newer version of Terraform can lead to controller replacement on reconfigure for Slurm GCP v6


nick-stroud:
### Describe the bug

For most changes, when reconfiguring a Slurm GCP v6 cluster, the controller should not be destroyed. This allows for operations such as adding a partition to be done on a running cluster.

It has been discovered that newer versions of Terraform introduce a bug such that these reconfigure operations cause the controller to be destroyed and recreated instead of being updated in place. 

This can be very disruptive to a running cluster as state may be lost such as current queue and accounting information.

Terraform version 1.4 is known to have the intended behavior. Terraform version 1.7 is known to exhibit the bad behavior. This bug is caused by a change in behavior of how Terraform treats drift between state and Terraform code.

The maintainers of this repository are aware of this issue and working to implement a shot and long term fix. If your workflow includes reconfiguring running Slurm GCP v6 clusters, please be advised to not upgrade beyond Terraform 1.4 until this bug has been addressed.

### Steps to reproduce

1. Install terraform 1.7
2. Deploy `examples/hpc-slurm.yaml`
3. Add a partition to the blueprint
4. Re-deploy the blueprint (`ghpc deploy -w`)

### Expected behavior

Without impacting queue, accounting db, or running jobs the partition is added to the cluster. The controller vm is modified in place and is not deleted.

### Actual behavior

The controller is deleted and a new controller VM is created.

### Blueprint

Any Slurm GCP v6 blueprint.

TimZaman:
Running into this while doing the slurm-on-gcp tutorial
```
./ghpc create examples/hpc-slurm.yaml -l ERROR --vars project_id=personal-235500
validator "test_tf_version_for_slurm" failed:
Error: using a newer version of Terraform can lead to controller replacement on reconfigure for Slurm GCP v6

Please be advised of this known issue: https://github.com/GoogleCloudPlatform/hpc-toolkit/issues/2774
Until resolved it is advised to use Terraform 1.4.0 with Slurm deployments.

To silence this warning, add flag: --skip-validators=test_tf_version_for_slurm

One or more blueprint validators has failed. See messages above for suggested
actions. General troubleshooting guidance and instructions for configuring
validators are shown below.

- https://goo.gle/hpc-toolkit-troubleshooting
- https://goo.gle/hpc-toolkit-validation

Validators can be silenced or treated as warnings or errors:

[...]

PR #2933, opened 2024-08-15
SlurmGCP. Perform cleanup per nodeset

mr0re1:
* Do clean up per nodeset instead of total:
  * solves problem "deletion of nodeset with networks causes total wipe";
  * solves problem "dynamic nodes getting deleted";
  * pre-requisite for reversing chain of dependencies between modules;
* Remove controller dependency on clean up - solves #2774 ;
* Do not participate in "race against controller":
  * Fetch list on instances to delete once;
  * Do not fail on failed attempt to delete;
  * Wait until `STOPPING` instances are deleted.

**Remaining problems:**
* Turning `enable_cleanup_compute`  off triggers clean up;
* "Slow to die" controller or removal of one partition only have an opportunity to leak some instances (or fail on policies deletion).


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/cleanup_compute.sh
```diff
@@ -0,0 +1,79 @@
...
+set -e -o pipefail
+
+project="$1"
+cluster_name="$2"
+nodeset_name="$3"
+universe_domain="$4"
+compute_endpoint_version="$5"
+gcloud_dir="$6"
+
+if [[ -z "${project}" || -z "${cluster_name}" || -z "${nodeset_name}" || -z "${universe_domain}" || -z "${compute_endpoint_version}" ]]; then
```
tpdownes, line 25: `nit` you might add an argument count check here instead. e.g.

https://stackoverflow.com/a/18568726

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/cleanup_compute.sh
```diff
@@ -0,0 +1,79 @@
...
+gcloud compute instances list --format="value(selfLink)" --filter="${running_nodes_filter}" >"$tmpfile"
+# Do 10 instances at a time
+while batch="$(head -n 10)" && [[ ${#batch} -gt 0 ]]; do
+	nodes=$(echo "$batch" | paste -sd " " -) # concat into a single space-separated line
+	# The lack of quotes around ${nodes} is intentional and causes each new space-separated "word" to
+	# be treated as independent arguments. See PR#2523
+	# shellcheck disable=SC2086
+	gcloud compute instances delete --quiet ${nodes} || echo "Failed to delete some instances"
+done <"$tmpfile" # have to use a temp file, since `< <(gcloud ...)` doesn't work nicely with `head`
+rm -f "$tmpfile"
```
tpdownes, line 58: I suggest a `trap` for this immediately after creating the file. e.g.

`trap 'rm -rf -- "$tmpfile"' EXIT`

(confirm this first, I'm doing a bit of cut-and-paste from example)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/variables.tf
```diff
@@ -0,0 +1,66 @@
...
+Enables automatic cleanup of compute nodes and resource policies (e.g.
+placement groups) managed by this module, when cluster is destroyed.
+
+*WARNING*: Toggling this off will impact the running workload.
+Deployed compute nodes will be destroyed.
+EOD
+  type        = bool
+}
+
+variable "universe_domain" {
```
tpdownes, line 39: It seems like we either need default values for these API configurations here or in the bash script. Am I not seeing where the default/simple case (GA API, standard universe) is being set?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/main.tf
```diff
@@ -14,36 +14,32 @@
...
+    cluster_name             = var.slurm_cluster_name
+    nodeset_name             = var.nodeset.nodeset_name
     universe_domain          = var.universe_domain
     compute_endpoint_version = var.endpoint_versions.compute
     gcloud_path_override     = var.gcloud_path_override
   }
 
   provisioner "local-exec" {
-    command = "/bin/bash ${path.module}/scripts/cleanup_compute.sh ${self.triggers.project_id} ${self.triggers.cluster_name} ${self.triggers.universe_domain} ${self.triggers.compute_endpoint_version} ${self.triggers.gcloud_path_override}"
+    command = "/bin/bash ${path.module}/cleanup_compute.sh ${self.triggers.project_id} ${self.triggers.cluster_name} ${self.triggers.nodeset_name} ${self.triggers.universe_domain} ${self.triggers.compute_endpoint_version} ${self.triggers.gcloud_path_override}"
```
tpdownes, line 39: I'd like to consider having this be in a subdirectory as it was before. Let me review best practices for storage of local-exec scripts.

tpdownes (changes requested):
- There are some minor bash script changes I'd like to see.
- A couple v6 integration tests were skipped, can you please run them?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/main.tf
mr0re1: Moved to `scripts/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/cleanup_compute/variables.tf
mr0re1: It's in `community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf`
I would prefer to have at most one place were defaults are defined, and touching controller module interface increases scope needlessly.

tpdownes: I agree, I just couldn't find it via GitHub UI easily. I also think this is a case where the sub-module is an internal implementation detail so exposing the default at the top-level is the right thing.

tpdownes (approved):

PR #2945, opened 2024-08-16
adding optional fields to redirect use of embedded modules to pull fr…

RachaelSTamakloe:
this pr adds optional fields(toolkit_modules_url and toolkit_modules_version) to redirect use of embedded modules to pull modules from github when provided github url and module version

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -685,10 +685,14 @@ func (bp *Blueprint) checkBlueprintName() error {
 // exclsuively supplied (i.e., one is present, but the other is missing).
 func (bp *Blueprint) checkToolkitModulesUrlAndVersion() error {
 	if bp.ToolkitModulesURL == "" && bp.ToolkitModulesVersion != "" {
-		return fmt.Errorf("toolkit_modules_url must be provided when toolkit_modules_version is specified")
+		return BpError{Root.ToolkitModulesURL, HintError{
+			Err:  errors.New("toolkit_modules_url must be provided when toolkit_modules_version is specified"),
+			Hint: "Specify toolkit_modules_url"}}
 	}
 	if bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion == "" {
-		return fmt.Errorf("toolkit_modules_version must be provided when toolkit_modules_url is specified")
+		return BpError{Root.ToolkitModulesVersion, HintError{
+			Err:  errors.New("toolkit_modules_version must be provided when toolkit_modules_url is specified"),
+			Hint: "Specify toolkit_modules_version"}}
 	}
 	return nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -685,12 +685,12 @@ func (bp *Blueprint) checkBlueprintName() error {
 // exclsuively supplied (i.e., one is present, but the other is missing).
 func (bp *Blueprint) checkToolkitModulesUrlAndVersion() error {
 	if bp.ToolkitModulesURL == "" && bp.ToolkitModulesVersion != "" {
-		return BpError{Root.ToolkitModulesURL, HintError{
+		return BpError{Root.ToolkitModulesVersion, HintError{
 			Err:  errors.New("toolkit_modules_url must be provided when toolkit_modules_version is specified"),
 			Hint: "Specify toolkit_modules_url"}}
 	}
 	if bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion == "" {
-		return BpError{Root.ToolkitModulesVersion, HintError{
+		return BpError{Root.ToolkitModulesURL, HintError{
 			Err:  errors.New("toolkit_modules_version must be provided when toolkit_modules_url is specified"),
 			Hint: "Specify toolkit_modules_version"}}
 	}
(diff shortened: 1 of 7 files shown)


pkg/config/expand.go
```diff
@@ -143,9 +144,24 @@ func (bp Blueprint) expandGroup(gp groupPath, g *Group) error {
...
+			if err != nil {
+				return fmt.Errorf("error constructing versioned module source: %w", err)
+			}
+			m.Source = newSource
+		}
+	}
 	return validateModuleInputs(mp, *m, bp)
 }
 
+// TODO: Add validation and error checks for baseURL and version
```
mr0re1, line 160: nit. This looks like a bad place to do validations. `ToolkitModulesURL` and `ToolkitModulesVersion` should be validated in advance, not during module processing.

You can save 5 lines by removing never-error handling

pkg/config/expand.go
```diff
@@ -143,9 +144,24 @@ func (bp Blueprint) expandGroup(gp groupPath, g *Group) error {
 func (bp Blueprint) expandModule(mp ModulePath, m *Module) error {
 	bp.applyUseModules(m)
 	bp.applyGlobalVarsInModule(m)
+	// Versioned Module Logic for Embedded Modules
+	if bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion != "" {
```
mr0re1, line 148: AFAIK, this is too late for substituting source, e.g.:
`validateModule` has already been executed by this point, validation will be done agains "non-doctored" source.

pkg/config/expand.go
```diff
@@ -143,9 +144,24 @@ func (bp Blueprint) expandGroup(gp groupPath, g *Group) error {
...
+		if strings.HasPrefix(m.Source, "modules/") || strings.HasPrefix(m.Source, "community/") {
```
tpdownes, line 149: @mr0re1 is there an opportunity here to create an `func bool isLocalModule(source)` helper? Looks like it might get used in `pkg/inspect/list.go`?

pkg/inspect/list.go
```diff
@@ -67,3 +67,7 @@ func LocalModules() ([]SourceAndKind, error) {
 	}
 	return ret, nil
 }
+
+func IsLocalModule(source string) bool {
```
mr0re1, line 71: We already have `sourcereader.IsEmbeddedPath` , please use it instead.

pkg/config/expand.go
```diff
@@ -103,9 +104,23 @@ func (bp *Blueprint) expandVars() error {
 	return nil
 }
 
+func (bp *Blueprint) substituteModuleSources() {
```
mr0re1, line 107: You can use `WalkModules` instead of writing nested loops.

pkg/config/expand.go
```diff
@@ -103,9 +104,23 @@ func (bp *Blueprint) expandVars() error {
...
+				m.Source = fmt.Sprintf("%s//%s?ref=%s&depth=1", bp.ToolkitModulesURL, m.Source, bp.ToolkitModulesVersion)
+			}
+		}
+	}
+}
+
 func (bp *Blueprint) expandGroups() error {
 	bp.addKindToModules()
-
+	if bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion != "" {
```
mr0re1, line 121: nit. I would avoid doing this check here, instead create a function:
```go
func (bp Blueprint) transformSource(s string) string {
  if sourcereader.IsEmbeddedPath(s) && bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion != "" {
    return fmt.Sprintf("%s//%s?ref=%s&depth=1", bp.ToolkitModulesURL, s, bp.ToolkitModulesVersion)
  }
  return s
}

...

bp.WalkModulesSafe(func(_ ModulePath, m *Module) {
   m.Source = bp.transformSource(m.Source)
})
```

mr0re1:
It would be also necessary to add unit tests.

pkg/config/config.go
```diff
@@ -677,6 +680,23 @@ func (bp *Blueprint) checkBlueprintName() error {
 	return nil
 }
 
+// checkToolkitModulesUrlAndVersion returns an error if either
+// toolkit_modules_url or toolkit_modules_version is
+// exclsuively supplied (i.e., one is present, but the other is missing).
+func (bp *Blueprint) checkToolkitModulesUrlAndVersion() error {
+	if bp.ToolkitModulesURL == "" && bp.ToolkitModulesVersion != "" {
+		return BpError{Root.ToolkitModulesURL, HintError{
```
mr0re1, line 688: It needs to be `Root.ToolkitModulesVersion`, since `bp.ToolkitModulesURL` is not defined there is no place in blueprint file to point to.

pkg/config/config.go
```diff
@@ -677,6 +680,23 @@ func (bp *Blueprint) checkBlueprintName() error {
...
+			Err:  errors.New("toolkit_modules_url must be provided when toolkit_modules_version is specified"),
+			Hint: "Specify toolkit_modules_url"}}
+	}
+	if bp.ToolkitModulesURL != "" && bp.ToolkitModulesVersion == "" {
+		return BpError{Root.ToolkitModulesVersion, HintError{
```
mr0re1, line 693: Same

mr0re1 (approved):
LGTM, please address comments about incorrect path in errors

PR #2772, opened 2024-07-17
Add validator for Terraform version and SlurmGCP6

mr0re1:
```sh
$ make && ./ghpc create tst.yaml -w
**************** building ghpc ************************
validator "test_tf_version_for_slurm" failed:
Error: Using a newer version of Terraform can lead to controller replacement on reconfigure for Slurm GCP v6

Please be advised of this known issue: https://github.com/GoogleCloudPlatform/hpc-toolkit/issues/2774
Until resolved it is advised to use Terraform 1.4.0 with Slurm deployments.

To silence this warning, add flag: --skip-validators=test_tf_version_for_slurm

One or more blueprint validators has failed. See messages above for suggested
actions. General troubleshooting guidance and instructions for configuring
validators are shown below.

- https://goo.gle/hpc-toolkit-troubleshooting
- https://goo.gle/hpc-toolkit-validation

Validators can be silenced or treated as warnings or errors:

- https://goo.gle/hpc-toolkit-validation-levels

validation failed due to the issues listed above
$ make && ./ghpc create tst.yaml -w --skip-validators=test_tf_version_for_slurm
**************** building ghpc ************************
Creating deployment folder "orlovtst" ...
To deploy your infrastructure please run:

./ghpc deploy orlovtst ...
```

PR #2962, opened 2024-08-21
Add machine type availability checks to slurm-gcp-v6-nodeset

annuay-google:
### Issue
Blueprint yaml files allow users to specify machine and zone combinations that do not exist. Infra is provisioned by Terraform during `./ghpc deploy` , but autoscaling may fail later if capacity is not found by bulk insert APIs in zones specified

### Approach
Created a Terraform precondition to verify that the machine type is available in at least one zone. In case they're not, there is no feasible way for bulk insert to allocate machines, and terraform will exit during plan/apply

### Testing
Ran `./ghpc create`, `./ghpc deploy` on this configuration. The specified machine type does not exist in both zones `us-central1-b` and `us-central1-c`. Verified that Terraform exits and error message is as expected

config:
```yaml
vars:
  deployment_name: hpc-slurm
  region: us-central1
  zone: us-central1-b

deployment_groups:
- group: primary
  modules:
  - id: h3_nodeset
    use: [network]
    settings:
      node_count_dynamic_max: 20
      # Note that H3 is available in only specific zones. https://cloud.google.com/compute/docs/regions-zones
      machine_type: h3-standard-88
      # H3 does not support pd-ssd and pd-standard
      # https://cloud.google.com/compute/docs/compute-optimized-machines#h3_disks
      disk_type: pd-balanced
      bandwidth_tier: gvnic_enabled
      allow_automatic_updates: false
      zones:
      - us-central1-c
```
output:
```
Testing if deployment group hpc-slurm/primary requires adding or changing cloud infrastructure
Error: exit status 1

Error: Resource precondition failed

  on modules/embedded/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf line 193, in resource "terraform_data" "machine_type_zone_validation":
 193:       condition     = length(local.zones_with_machine_type) > 0
    ├────────────────
    │ local.zones_with_machine_type is empty tuple

machine type h3-standard-88 is not available in any of the zones
["us-central1-b","us-central1-c"]. To list zones in which it is available, run:

gcloud compute machine-types list --filter="name=h3-standard-88"

Hint: terraform plan for deployment group hpc-slurm/primary failed
```
Added 1 valid zone (`us-central1-a`) and verified that infra is correctly provisioned

### Submission Checklist

### DIFF
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
@@ -77,7 +77,7 @@ locals {
     gpu                    = one(local.guest_accelerator)
 
     labels           = local.labels
-    machine_type     = var.machine_type
+    machine_type     = terraform_data.machine_type_zone_validation.output
     metadata         = local.metadata
     min_cpu_platform = var.min_cpu_platform
 
@@ -173,5 +173,30 @@ data "google_compute_reservation" "reservation" {
 
     # TODO: wait for https://github.com/hashicorp/terraform-provider-google/issues/18248
     # Add a validation that if reservation.project != var.project_id it should be a shared reservation
+  }
+}
+
+data "google_compute_machine_types" "machine_types_by_zone" {
+  for_each = local.zones
+  filter   = format("name = \"%s\"", var.machine_type)
+  zone     = each.value
+}
+
+locals {
+  machine_types_by_zone   = data.google_compute_machine_types.machine_types_by_zone
+  zones_with_machine_type = [for k, v in local.machine_types_by_zone : k if length(v.machine_types) > 0]
+}
+
+resource "terraform_data" "machine_type_zone_validation" {
+  input = var.machine_type
+  lifecycle {
+    precondition {
+      condition     = length(local.zones_with_machine_type) > 0
+      error_message = <<-EOT
+        machine type ${var.machine_type} is not available in any of the zones ${jsonencode(local.zones)}". To list zones in which it is available, run:
+
+        gcloud compute machine-types list --filter="name=${var.machine_type}"
+        EOT
+    }
   }
 }
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
@@ -130,7 +130,7 @@ modules. For support with the underlying modules, see the instructions in the
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.3 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.4 |
 | <a name="requirement_google"></a> [google](#requirement\_google) | >= 5.11 |
 
 ## Providers
@@ -138,6 +138,7 @@ modules. For support with the underlying modules, see the instructions in the
 | Name | Version |
 |------|---------|
 | <a name="provider_google"></a> [google](#provider\_google) | >= 5.11 |
+| <a name="provider_terraform"></a> [terraform](#provider\_terraform) | n/a |
 
 ## Modules
 
@@ -147,8 +148,10 @@ No modules.
 
 | Name | Type |
 |------|------|
+| [terraform_data.machine_type_zone_validation](https://registry.terraform.io/providers/hashicorp/terraform/latest/docs/resources/data) | resource |
 | [google_compute_default_service_account.default](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_default_service_account) | data source |
 | [google_compute_image.slurm](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_image) | data source |
+| [google_compute_machine_types.machine_types_by_zone](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_machine_types) | data source |
 | [google_compute_reservation.reservation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_reservation) | data source |
 | [google_compute_zones.available](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_zones) | data source |
 
--- a/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/versions.tf
+++ b/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/versions.tf
@@ -15,7 +15,7 @@
 */
 
 terraform {
-  required_version = ">= 1.3"
+  required_version = ">= 1.4"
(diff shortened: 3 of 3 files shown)


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
```diff
@@ -175,3 +175,21 @@ data "google_compute_reservation" "reservation" {
...
+  filter   = format("name = \"%s\"", local.nodeset.machine_type)
+  zone     = each.value
+}
+
+locals {
+  machine_types_by_zone      = data.google_compute_machine_types.machine_types_by_zone
+  zones_without_machine_type = [for k, v in local.machine_types_by_zone : k if length(v.machine_types) == 0]
+}
+
+check "is_machine_type_available_in_zones" {
```
mr0re1, line 190: Check is available starting Terraform 1.5.0 . AFAIK we are supporting terraform 1.2.0.

See `Makefile`: `MIN_TERRAFORM_VERSION=1.2 # for deploying modules`

annuay-google: Yes, checks are available with only TF 1.5+. I had discussed a similar concern regarding TF version with Tom (that turned out to be a non-issue). He had clarified then that we should be able to use Terraform 1.5

@tpdownes would it be okay for us to bump up the minimum version to 1.5?

mr0re1: LGTM, feel free to land if you get a green light in regard of terraform version.

mr0re1 (dismissed):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
annuay-google: ack, will wait for Tom

nick-stroud: I agree that TF 1.5 is acceptable, but we should wait to put this in until https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/2772 has been rolled back.

tpdownes: Completed in #2961

tpdownes, line 190: 1. You have the [wrong syntax for checks](https://developer.hashicorp.com/terraform/language/checks#syntax). The nice thing about checks is that data sources are "scoped" inside the check block which allows them to fail.

2. I think you are using the wrong condition. The assertion we want is "at least 1 zone has the machine type" not "there are 0 zones without the machine type". Both your examples should succeed.

3. Despite (1), it's interesting to me that the condition is failing in the check block rather than the data source. I would think that it would fail at the data source stage. If the google_compute_machine_types data source can successfully return the empty list, it would actually be more desirable to **fail** as a precondition than to warn with a check.

Can you please verify the behavior of the data source on a to confirm that it can return empty without failing?

e.g.

```hcl
data "google_compute_machine_types" "machine_types_by_zone" {
  for_each = toset(["us-central1-b", "us-central1-c"])
  filter   = "name = \"h3-standard-88\""
  zone     = each.value
}
```

Run `terraform state list` and `terraform state show data.google_compute_machine_types.machine_types_by_zone[\"us-central1-b\"]`. If this returns the empty list after `terraform apply`, then we are safe to fail on the data source instead of warning in a check block. So long as you fix the boolean logic (2).

tpdownes (changes requested):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
annuay-google: Verified that 3 works. The data source does not fail.

Updated the code to use `precondition` with `terraform_data` resource as discussed over the chat. If none of the zones have the machine type available, Terraform will throw an error and exit.

Tested the code on success and failure scenarios (updated the PR description).

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
```diff
@@ -175,3 +175,23 @@ data "google_compute_reservation" "reservation" {
...
+  zone     = each.value
+}
+
+locals {
+  machine_types_by_zone   = data.google_compute_machine_types.machine_types_by_zone
+  zones_with_machine_type = [for k, v in local.machine_types_by_zone : k if length(v.machine_types) > 0]
+}
+
+resource "terraform_data" "machine_type_zone_validation" {
+  lifecycle {
```
tpdownes, line 191: You should be chaining this resource into the graph dependencies.

https://developer.hashicorp.com/terraform/language/resources/terraform-data

I suggest adding:

```hcl
  input = var.machine_type
```

and then replace line 80 with

```
    machine_type = terraform_data.machine_type_zone_validation.output
```

tpdownes (changes requested):
To make the terraform_data resource and precondition useful, it should be inserted into the terraform resource graph.

tpdownes (changes requested):
Please also change versions.tf to require terraform 1.4

annuay-google:
> Please also change versions.tf to require terraform 1.4

Done

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/main.tf
```diff
@@ -175,3 +175,24 @@ data "google_compute_reservation" "reservation" {
...
+  machine_types_by_zone   = data.google_compute_machine_types.machine_types_by_zone
+  zones_with_machine_type = [for k, v in local.machine_types_by_zone : k if length(v.machine_types) > 0]
+}
+
+resource "terraform_data" "machine_type_zone_validation" {
+  input = var.machine_type
+  lifecycle {
+    precondition {
+      condition     = length(local.zones_with_machine_type) > 0
+      error_message = "machine type ${var.machine_type} is not available in any of the zones ${jsonencode(local.zones)}"
```
tpdownes, line 195: A good suggestion was made to me to help the user to the right answer. I suggest the following:


 to guide the user to select
```suggestion
      error_message = <<-EOT
        machine type ${var.machine_type} is not available in any of the zones ${jsonencode(local.zones)}". To list zones in which it is available, run:

        gcloud compute machine-types list --filter="name=${var.machine_type}"
        EOT
```

tpdownes (changes requested):
Please add one final helpful hint to the user. Then squash your commits and this should be ready!

annuay-google:
> Please add one final helpful hint to the user. Then squash your commits and this should be ready!

Done. Verified the error as well
```
machine type h3-standard-88 is not available in any of the zones
["us-central1-b"]". To list zones in which it is available, run:

gcloud compute machine-types list --filter="name=h3-standard-88"
```

tpdownes (approved):
This is a big improvement in user experience. Thank you for the submission!

PR #2966, opened 2024-08-22
Add `NodeState` enum

mr0re1:
Add proper enum to reduce usage of error-prone string literals (follow up action).

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 import pytest
+import itertools
 import common # needed to import util
 import util
 from google.api_core.client_options import ClientOptions  # noqa: E402
+
+from util import NodeState
 
 # Note: need to install pytest-mock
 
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
```diff
@@ -130,3 +133,24 @@ def test_create_client_options(
     ud_mock.return_value = "googleapis.com"
     ep_mock.return_value = ep_ver
     assert util.create_client_options(api).__repr__() == expected.__repr__()
+
+
+def test_NodeState():
+    
+    for l,r in itertools.product((NodeState.DOWN, NodeState("DOWN"), "DOWN"), repeat=2):
```
abbas1902, line 140: Just curious (from a high level) if this file is tested against often, as this seems like functionality that would rarely change (assuming the NodeState also does not change).

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
```diff
@@ -130,3 +133,24 @@ def test_create_client_options(
...
+    gecko = NodeState("GECKO")  # unknown state
+    assert gecko == NodeState("GECKO")
+    assert gecko == "GECKO"
+    assert gecko != NodeState.DOWN
+
+    
+    assert NodeState.DOWN != NodeState.POWER_DOWN
+    assert NodeState.DOWN != NodeState("dOwN") # case sensitive
+
+    assert f"{NodeState.DOWN}" == "NodeState.DOWN"
```
abbas1902, line 152: This assertion seems to be failing in pre-commit, yet works fine while testing on my local env (based on the error message seems to be comparing this as 'DOWN' == 'NodeState.Down'). Maybe try running it again?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1917,3 +1917,52 @@ def update_config(cfg: NSDict) -> None:
...
+        o._name_ = value
+        return o
+
+    def _generate_next_value_(name, *_):
+        return name
+    
+    def __repr__(self):
+        return self.__str__()
+    
+    # *** Base states
```
abbas1902, line 1950: Will other node states like unknown or end be added or are they not needed? (from https://github.com/SchedMD/slurm/blob/master/slurm/slurm.h#L980)

PR #2968, opened 2024-08-22
Fix local_ssd_config issue that forces node-pool recreation

sharabiani:
This PR fixes the issue that when ephemeral_storage_local_ssd_config is not set, it forces node-pool recreation on re-deploy.

### disk_definitions implemented:
It contain default values for ephemeral_storage_local_ssd_config and local_ssd_count_nvme_block per machine type. It's defined based on the values [here](https://cloud.google.com/compute/docs/disks/local-ssd?_gl=1*103lsfs*_ga*MTczMTAzMDM4My4xNzIyMzc1ODU5*_ga_WH2QY8WWF5*MTcyMzYzODIzOS4xOC4xLjE3MjM2MzgzNTEuNTAuMC4w#choose_number_local_ssds).

#### Selection priority order:
1. variable local_ssd_count_ephemeral_storage and local_ssd_count_nvme_block if any is not null
2. disk_definitions if not empty for the machine_type
3. default to null value for both local_ssd_count_ephemeral_storage and local_ssd_count_nvme_block

## Manual testing:

Scenario 1:
1. Deployed examples/gke-a3-highgpu.yaml
2. Run the deploy again.
3. Verified **NO** destruction and recreation of the node-pool is required.

Scenario 2:
1. Deployed examples/gke-a3-highgpu.yaml
2. Set local_ssd_count_ephemeral_storage to 16 explicitly
3. Run the deploy again.
4. Verified **NO** destruction and recreation of the node-pool is required.

Scenario 2:
1. Deployed examples/gke-a3-highgpu.yaml
2. Removed "a3-highgpu-8g" from disk_definitions
3. Run the deploy again.
4. Verified destruction and recreation of the node-pool **is** required.

## Additional Notes:

- Made changes to relevant documentation.


modules/compute/gke-node-pool/disk_definitions.tf
```diff
@@ -0,0 +1,36 @@
...
+    "a3-highgpu-8g" = { local_ssd_count_ephemeral_storage = 16, local_ssd_count_nvme_block = null },
+    "a3-megagpu-8g" = { local_ssd_count_ephemeral_storage = 16, local_ssd_count_nvme_block = null },
+  }
+
+  generated_local_ssd_config = lookup(local.local_ssd_machines, var.machine_type, { local_ssd_count_ephemeral_storage = null, local_ssd_count_nvme_block = null })
+
+  # Select in priority order:
+  # (1) var.local_ssd_count_ephemeral_storage and var.local_ssd_count_nvme_block if any is not null
+  # (2) local.local_ssd_machines if not empty
+  # (3) default to null value for both local_ssd_count_ephemeral_storage and local_ssd_count_nvme_block
```
nick-stroud, line 34: Should we perhaps default to 0 instead of null and always include the blocks in the main terraform.

My thinking is that if it really falls all the way through and we use null, then we are exposed to destroying node pools on re-creation. 

I would rather just set to zero and risk an error on creation due to bad config.

The first class of problem is a hidden bug that could do damage to a running cluster later (high cost). The second class of problem will show on deployment and can be worked around with no code change by adding the value to the blueprint.

sharabiani: 1. Given `local_ssd_count_ephemeral_storage = 16` and `local_ssd_count_nvme_block = 0` and `including both blocks` throws error:
`Error: error creating NodePool: googleapi: Error 400: At most one Local SSD configurations can be specified.`

2. In my tests, the only acceptable scenarios for a3 machines were as follows:

- Set no blocks for ephemeral and nvme: creates node-pool successfully but re-deploy forces recreation
- Set only one block for eithe ephermal or nvme with acceptable values for the machine type: creates node-pool successfully and does not force recreation on re-deploy

3. As I learned from you, we cannot consider 0 as block exclusion because we may ignore explicit user actions.


Therefore, considering 0 as default and adding both blocks doesn't seem to be practically possible here. Please let me know if I'm missing something here.

sharabiani: I'm resolving this as we agreed on the current approach over the chat.

ankitkinra (approved):

modules/compute/gke-node-pool/variables.tf
```diff
@@ -104,7 +104,7 @@ variable "local_ssd_count_nvme_block" {
   description = <<-EOT
   The number of local SSDs to attach to each node to back block storage.  
   Uses NVMe interfaces.  Must be supported by `machine_type`.
-  When set to null, GKE decides about default value.
+  When set to null,  default value either is [set based on machine_type](https://cloud.google.com/compute/docs/disks/local-ssd?_gl=1*103lsfs*_ga*MTczMTAzMDM4My4xNzIyMzc1ODU5*_ga_WH2QY8WWF5*MTcyMzYzODIzOS4xOC4xLjE3MjM2MzgzNTEuNTAuMC4w#choose_number_local_ssds) or GKE decides about default value.
```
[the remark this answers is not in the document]
sharabiani: done

modules/compute/gke-node-pool/variables.tf
```diff
@@ -93,7 +93,7 @@ variable "local_ssd_count_ephemeral_storage" {
   description = <<-EOT
   The number of local SSDs to attach to each node to back ephemeral storage.  
   Uses NVMe interfaces.  Must be supported by `machine_type`.
-  When set to null, GKE decides about default value.
+  When set to null,  default value either is [set based on machine_type](https://cloud.google.com/compute/docs/disks/local-ssd?_gl=1*103lsfs*_ga*MTczMTAzMDM4My4xNzIyMzc1ODU5*_ga_WH2QY8WWF5*MTcyMzYzODIzOS4xOC4xLjE3MjM2MzgzNTEuNTAuMC4w#choose_number_local_ssds) or GKE decides about default value.
```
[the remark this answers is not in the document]
sharabiani: done

PR #2969, opened 2024-08-22
Support named placements in GKE node pools

arajmane-g:
Toolkit users will be able to utilise named compact placement policy in their node pools when deploying GKE clusters. 

### Implementation
Introduce a new setting `placement_policy` by deprecating `compact_placement`. `type` and `name` are two nested settings. The latter is optional.

Placement policy is assumed to exist.

### Testing
Integration test will be added as a follow-up.

Following things were tested from the dev environment:
- Not setting either of the settings should not include the relevant block in the generated plan
- Setting `compact_placement` should fail at the Terraform plan stage with a message indicating deprecation
- It was verified that named, unnamed compact placement policies work using the new setting
- Terraform apply should fail for an unknown policy
- Terraform apply should succeed for a known compact placement policy. It was verified that the policy is attached to the VMs in the provisioned node pool

### Submission Checklist


modules/compute/gke-node-pool/variables.tf
```diff
@@ -293,3 +293,13 @@ variable "specific_reservation" {
     values = null
   }
 }
+
+variable "compact_placement_policy" {
```
ankitkinra, line 297: Can we instead make placement policy generic and allow customer to pass in which type of placement they would like ? 
Today the only option is "COMPACT" based on https://cloud.google.com/kubernetes-engine/docs/how-to/compact-placement#create_a_compact_placement_policy_in_a_new_cluster

However there might be more options coming.

modules/compute/gke-node-pool/main.tf
```diff
@@ -66,13 +66,25 @@ resource "google_container_node_pool" "node_pool" {
...
+  # Those blueprints will stop working if we remove this block.
   dynamic "placement_policy" {
     for_each = var.compact_placement ? [1] : []
     content {
       type = "COMPACT"
     }
   }
 
+  dynamic "placement_policy" {
+    for_each = (!var.compact_placement && try(contains(["COMPACT"], var.placement_type), false)) ? [1] : []
```
ankitkinra, line 81: nit: If we are checking that COMPACT is the only valid placement_type we are accepting in the variables, then we don't need to verify that again.

modules/compute/gke-node-pool/main.tf
```diff
@@ -66,13 +66,25 @@ resource "google_container_node_pool" "node_pool" {
     max_unavailable = 1
   }
 
+  # NOTE: `placement_type` and `placement_policy` together can support compact placement of nodes in a node pool.
+  # So, one might wonder if we can remove this block. But to support backward compatibility this needs to be kept.
+  # By backward compatibility we mean that there maybe blueprints that might be using `compact_placement`.
```
ankitkinra, line 71: Let's deprecate this variable following team guidelines.

modules/compute/gke-node-pool/main.tf
```diff
@@ -202,6 +214,15 @@ resource "google_container_node_pool" "node_pool" {
       On the other hand, with SPECIFIC_RESERVATION you must set `specific_reservation.key` and `specific_reservation.values` to `compute.googleapis.com/reservation-name` and a list of reservation names respectively.
       EOT
     }
+    precondition {
+      condition     = var.compact_placement || var.placement_type == null || try(contains(["COMPACT"], var.placement_type), false)
+      error_message = "`COMPACT` is the only supported value for `placement_type`."
+    }
```
ankitkinra, line 220: Can we add another precondidtion that if var.compact_placement is true then var.placement_type should be null/empty ?

modules/compute/gke-node-pool/variables.tf
```diff
@@ -172,11 +172,36 @@ variable "spot" {
...
+  EOT
+  type        = string
+  default     = null
+
+}
+
+variable "placement_policy" {
+  description = <<-EOT
+  Name of the placement policy to use when `placement_type` is set.
+  It is assumed that the specified policy exists. To create a placement policy refer to https://cloud.google.com/sdk/gcloud/reference/compute/resource-policies/create/group-placement.
```
ankitkinra, line 198: Perhaps as a follow up item, we should also test whether this policy exists or not (if that is possible by getting this resource)

modules/compute/gke-node-pool/main.tf
arajmane-g: Ok. Let me check. I agree this can be removed as long as it's redundant.

modules/compute/gke-node-pool/variables.tf
arajmane-g: Not validating it separately as the node pool provision operation is supposed to fail for an unknown policy.

But, rethinking about it. Validating and failing early would be a good experience than failing at the time of provisioning the node pool (which happens after cluster creation). Will address as a follow-up 👍

modules/compute/gke-node-pool/main.tf
arajmane-g: deprecating `compact_placement` obviates this

modules/compute/gke-node-pool/main.tf
```diff
@@ -67,9 +67,10 @@ resource "google_container_node_pool" "node_pool" {
   }
 
   dynamic "placement_policy" {
-    for_each = var.compact_placement ? [1] : []
+    for_each = var.placement_policy.type != null ? [1] : []
     content {
-      type = "COMPACT"
+      type        = var.placement_policy.type
```
annuay-google, line 72: If the [only supported value](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#placement_policy) of type is "COMPACT", do we need to take it from the user as an input? I'm wondering if we can pass it to the provider directly from this module

modules/compute/gke-node-pool/variables.tf
```diff
@@ -171,10 +171,36 @@ variable "spot" {
   default     = false
 }
 
+# tflint-ignore: terraform_unused_declarations
 variable "compact_placement" {
-  description = "Places node pool's nodes in a closer physical proximity in order to reduce network latency between nodes."
+  description = "DEPRECATED: Use `placement_policy`"
```
annuay-google, line 176: Is there a reason why we don't want to support backward compatibility for now? Also, do we know how frequently our users use this option? If it's fairly low, we should be good

arajmane-g: Yes, the usage is low. It was discussed offline.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -171,10 +171,36 @@ variable "spot" {
...
+    condition     = var.compact_placement == null
+    error_message = "`compact_placement` is deprecated. Use `placement_policy` instead"
+  }
+}
+
+variable "placement_policy" {
+  description = <<-EOT
+  Group placement policy to use for the node pool's nodes. `COMPACT` is the only supported value for `type` currently. `name` is the name of the placement policy.
+  It is assumed that the specified policy exists. To create a placement policy refer to https://cloud.google.com/sdk/gcloud/reference/compute/resource-policies/create/group-placement.
+  Beware of the restrictions for placement policies https://cloud.google.com/compute/docs/instances/placement-policies-overview#restrictions-compact-policies
```
ankitkinra, line 189: nit: Reword to

```
Note:  Placement policies have the [following](https://cloud.google.com/compute/docs/instances/placement-policies-overview#restrictions-compact-policies) restrictions
```

modules/compute/gke-node-pool/variables.tf
```diff
@@ -171,10 +171,36 @@ variable "spot" {
...
+  type = object({
+    type = string
+    name = optional(string)
+  })
+  default = {
+    type = null
+    name = null
+  }
+  validation {
+    condition     = var.placement_policy.type == null || try(contains(["COMPACT"], var.placement_policy.type), false)
```
ankitkinra, line 201: nit: Should the validation look like following, as placement_policy can be null as well ? 

```
var.placement_policy == null || var.placement_policy.type == null || try(contains(["COMPACT"], var.placement_policy.type), false)
```

ankitkinra (approved):
Approved with some nits, please fix them before merging!

modules/compute/gke-node-pool/variables.tf
arajmane-g: `placement_policy` cannot null. Its default is an object. 
```
  default = {
    type = null
    name = null
  }
```

PR #2220, opened 2024-02-08
Adopt empty string as default value for maintenance_interval

tpdownes:
The default value of null cannot be set as a deployment variable; this
will allow the value to be set at the top of a blueprint with a comment to the
user to modify to "PERIODIC" as necessary.

### Submission Checklist

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cloud HPC Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #2971, opened 2024-08-23
Expose maintenance interval as a blueprint setting for node pools in GKE

annuay-google:
### Request
Expose host_maintenance_interval as a user configurable setting in the blueprint

### Approach
Expose as an optional terraform variable. If passed, we will set this under node_pool.node_config, using a dynamic block. Possible values are PERIODIC and AS_NEEDED

### Test plan
- [x] Update hpc-gke.yaml example with settings.host_maintenance_interval = PERIODIC. [Verified](https://bit.googleplex.com/#/annuay/6396960309248000) this is reflected in Terraform plan and on gcloud cli
- [x] Update hpc-gke.yaml example with settings.host_maintenance_interval = AS_NEEDED. [Verified](https://bit.googleplex.com/#/annuay/6181882909687808) this is reflected in Terraform plan and on gcloud cli
- [x] Keep hpc-gke.yaml as is (no settings.host_maintenance_interval key in yaml). [Verified](https://bit.googleplex.com/#/annuay/6340693855305728) host maintenance interval is unset on gcloud cli

### Deployment
Merge to develop. No additional action needed as of now

### Submission Checklist


modules/compute/gke-node-pool/variables.tf
```diff
@@ -319,3 +319,16 @@ variable "specific_reservation" {
     values = null
   }
 }
+
+variable "host_maintenance_interval" {
+  description = "Specifies the frequency of planned maintenance events."
+  type        = string
+  default     = "PERIODIC"
+  validation {
+    condition = var.host_maintenance_interval != null ? contains(
```
ankitkinra, line 328: Looking at this similar change https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/2220 , we did not allow `null` as an allowed value. 

It would be great if we can stay consistent and use empty string as the "null" value.

annuay-google: Makes sense, updated

annuay-google:
I've tested all scenarios (see description for details). Please review @ankitkinra

modules/compute/gke-node-pool/main.tf
```diff
@@ -162,6 +162,13 @@ resource "google_container_node_pool" "node_pool" {
       key                      = var.specific_reservation.key
       values                   = var.specific_reservation.values
     }
+
+    dynamic "host_maintenance_policy" {
+      for_each = true ? [1] : []
```
arajmane-g, line 167: I am guessing this condition is yet to be modified.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -319,3 +319,14 @@ variable "specific_reservation" {
     values = null
   }
 }
+
+variable "host_maintenance_interval" {
+  description = "Specifies the frequency of planned maintenance events."
+  type        = string
+  default     = ""
```
arajmane-g, line 326: AFAIU, Terraform doesn't omit empty strings but only nulls. If the default is an empty string, for a blueprint that doesn't provide the relevant setting the `host_maintenance_policy` block would be included in the generated plan, unless the condition in the /gke-node-pool/main.tf#167 checks for valid values.

arajmane-g (commented):
re: Test Plan

I think, ml-gke would be the right blueprint to use for testing

modules/compute/gke-node-pool/main.tf
annuay-google: Thanks for catching this. Fixed

modules/compute/gke-node-pool/variables.tf
annuay-google: Fixed the conditional, so this will not be included unless not ""

modules/compute/gke-node-pool/main.tf
```diff
@@ -162,6 +162,13 @@ resource "google_container_node_pool" "node_pool" {
       key                      = var.specific_reservation.key
       values                   = var.specific_reservation.values
     }
+
+    dynamic "host_maintenance_policy" {
```
ankitkinra, line 166: nit: I could not find host_maintenance_policy in https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool but I do see that it was added https://github.com/hashicorp/terraform-provider-google/pull/15347 , so I might not be looking at the right place.

annuay-google: I couldn't find host_maintenance_policy in the docs or Pantheon either. As you mentioned, it is available in the provider repository and has been used in the ai-infra-cluster-provisioning [code](https://github.com/GoogleCloudPlatform/ai-infra-cluster-provisioning/blob/main/a3-mega/terraform/modules/cluster/gke/main.tf#L228). I've applied the changes and verified that these changes are reflected in the gcloud cli response (attached in PR desc.)

modules/compute/gke-node-pool/main.tf
```diff
@@ -162,6 +162,13 @@ resource "google_container_node_pool" "node_pool" {
...
+      for_each = var.host_maintenance_interval == "" ? [] : [1]
```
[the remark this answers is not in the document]
annuay-google: Fixed

ankitkinra (approved):

annuay-google:
Added maintenance policy to a daily test. Verified that the [changes were reflected](https://bit.googleplex.com/#/annuay/6339632562176000) in the created node pool (ubuntu)

arajmane-g (approved):

PR #2982, opened 2024-08-29
Improved serial port collection tool

cdunbar13:
Slightly improved serial port tool retrieval.  

It is suggested to eventually remove the `capture_serial.sh` script as it does not attempt to continue collection from the last line retrieved and does not have at timeout.

### DIFF
--- a/tools/serial_port_collector.py
+++ b/tools/serial_port_collector.py
@@ -1,0 +1,69 @@
+#!/bin/python3
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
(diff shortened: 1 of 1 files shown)


tools/serial_port_collector.py
```diff
@@ -0,0 +1,70 @@
...
+from google.cloud import compute_v1
+from argparse import ArgumentParser, RawTextHelpFormatter
+
+"""This tool collects serial port output and prints it to the terminal until
+the VM is deleted or it hits the timeout (300s).  It takes in, project, vm_name
+and zone as arguments.  The script should only print each line once, using the
+line number of the previous serial port retrieval as the starting point of the
+next request.
+
+usage: serial_port_collector.py [-h] -p PROJECT -v VM_NAME -z ZONE
```
mr0re1, line 26: Will not argparse add "usage" section as well?

tools/serial_port_collector.py
```diff
@@ -0,0 +1,70 @@
...
+    
+    parser.add_argument("-z", "--zone", required=True, type=str,
+                    help="The zone the vm is located in")
+
+    args = parser.parse_args()
+
+    next=0
+    sleep_timer = 2
+    timeout_cntr = 0
+    while True and timeout_cntr < 300:
```
mr0re1, line 61: Do I read it correctly that if for 5 minutes there were no outputs the tool exits?
There a cases where startup script runs for long time without output. Do we need this termination condition at all?

mr0re1 (approved):

tools/serial_port_collector.py
cdunbar13: Yes that is correct.  I've moved the timeout to be an optional input and default to no timeout at all.

PR #2985, opened 2024-08-30
kubernetes provider added to gke-cluster module

sharabiani:
### Fixed missing kubernetes provider in gke-cluster module
The gke-cluster module is implicitly depended on the kubernetes provider ([because of workload-ldentity module](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/workload-identity/)) and the provider was mistakenly removed before.
This PR adds the provider to gke-cluster module.


arajmane-g (approved):
LGTM

ankitkinra (approved):
Please run pr-ml-gke test as well

PR #2997, opened 2024-09-03
Revisit the Reservation Interface for GKE Blueprints

arajmane-g:
Replace the settings {`reservation_type`, `specific_reservation.key`, `specific_reservation.values`} by one setting `reservation_affinity` that encapsulates all relevant settings.

#### Why?

There are issues with this interface 
- Specific_reservation.key does not have to be user input. Its value is always [compute.googleapis.com/reservation-name](http://compute.googleapis.com/reservation-name)
- reservation_type and specific_reservation don’t have to be two different settings and can be grouped / nested inside an object.
- Specific_reservation.values is an array. If we were to keep the old interface, to allow project_id (to support shared reservations) we need to introduce another nested array under specific_reservation (no other suitable place) which would contain projects. This is error prone as users have to ensure the order of reservation and project is correct.
- Doesn’t map to the [provider interface](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#nested_reservation_affinity) well. reservation_type is a top level setting

### Testing
Since, no unit testing exists for Terraform modules at the moment, manually tested all combinations of consume_reservation_type and specific_reservations.

For e2e testing, an integration test will be added as a fast follow up to the PR. Till then manual verification was done by deploying different blueprints:
- The change is backward compatible. Not providing the relevant settings will deploy the cluster as if no reservation is to be used
- When ANY_RESERVATION is targeted, any "Automatic" reservation in the zone is utilised by the cluster
- When SPECIFIC_RESERVATION is targeted, the provided specific reservation is utilised by the cluster

### Submission Checklist


modules/compute/gke-node-pool/variables.tf
```diff
@@ -318,34 +318,31 @@ variable "additional_networks" {
...
+    consume_reservation_type = "NO_RESERVATION"
+    specific_reservations    = []
+  }
+  validation {
+    condition     = contains(["NO_RESERVATION", "ANY_RESERVATION", "SPECIFIC_RESERVATION"], var.reservation_affinity.consume_reservation_type)
+    error_message = "Accepted values are: {NO_RESERVATION, ANY_RESERVATION, SPECIFIC_RESERVATION}"
   }
 }
 
+
```
ankitkinra, line 345: nit: delete

modules/compute/gke-node-pool/main.tf
```diff
@@ -159,9 +161,9 @@ resource "google_container_node_pool" "node_pool" {
     }
 
     reservation_affinity {
-      consume_reservation_type = var.reservation_type
-      key                      = var.specific_reservation.key
-      values                   = var.specific_reservation.values
+      consume_reservation_type = var.reservation_affinity.consume_reservation_type
```
ankitkinra, line 164: did you also test the test case of passing `NO_RESERVATION` ? Also can we trigger the existing e2e ML GKE tests ?

arajmane-g: Yes.

ankitkinra (approved):

PR #3002, opened 2024-09-04
Updating image builder for internal provider

cdunbar13:
With the use of the `google-private` internal provider, we should update the test infrastructure.  This will point tests that require `google-private` to the correct bucket to get the binary.  Access to the binary should be restricted to whitelisted users including anyone in the google.com org.

### DIFF
--- a/tools/cloud-build/images/test-runner/terraformrc
+++ b/tools/cloud-build/images/test-runner/terraformrc
@@ -1,0 +1,12 @@
+# .terraformrc setup for google network mirror
+provider_installation {
+  network_mirror {
+    url = "https://storage.googleapis.com/network-mirror-private-provider/"
+    include = ["hashicorp/google-private"]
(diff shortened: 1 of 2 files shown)


tools/cloud-build/images/test-runner/terraformrc
```diff
@@ -0,0 +1,12 @@
+# .terraformrc setup for google network mirror
+provider_installation {
+  network_mirror {
+    url = "https://storage.googleapis.com/network-mirror-private-provider/"
+    include = ["hashicorp/google-private"]
+    exclude = ["hashicorp/null", "hashicorp/local", "hashicorp/random", "hashicorp/google", "hashicorp/google-beta"]
```
mr0re1, line 6: What is motivation to use both `include` and `exclude` here?

mr0re1 (approved):

tpdownes (approved):

tools/cloud-build/images/test-runner/terraformrc
cdunbar13: Just to be extra careful about what we're pulling from the network mirror.

PR #3017, opened 2024-09-09
Adds option to allow Kueue/Jobset to be installed on a GKE cluster via blueprints

ankitkinra:
Allows blueprint to install Kueue/Jobset for a GKE cluster which reduces the manual steps required by the user to configure these frameworks.

In this PR:
* Downloaded versions for kueue / jobset and updated the manifest to add tolerations for the system pool taint.

Testing 
* Manually tested that a3 high cluster comes up with kueue and jobset controllers

Example yaml in a blueprint

```yaml
- id: workload_component_install
    use: [gke_cluster]
    settings:
      kueue:
        install: true
        config_path: $(ghpc_stage("manifests/user-provided-kueue-config.yaml"))
      jobset:
        install: true
```

Future enhancements 
* Automatic basic configuration for Kueue

### Submission Checklist

### DIFF
--- a/examples/gke-a3-highgpu.yaml
+++ b/examples/gke-a3-highgpu.yaml
@@ -66,3 +66,12 @@ deployment_groups:
       autoscaling_total_min_nodes: 2
       zones: [$(vars.zone)]
     outputs: [instructions]
+
+  - id: workload_component_install
+    source: modules/management/kubectl-apply
+    use: [gke_cluster]
+    settings:
+      kueue:
+        install: true
+      jobset:
+        install: true
--- a/examples/gke-a3-megagpu.yaml
+++ b/examples/gke-a3-megagpu.yaml
@@ -66,3 +66,12 @@ deployment_groups:
       autoscaling_total_min_nodes: 2
       zones: [$(vars.zone)]
     outputs: [instructions]
+
+  - id: workload_manager_install
+    source: modules/management/kubectl-apply
+    use: [gke_cluster]
+    settings:
+      kueue:
+        install: true
+      jobset:
+        install: true
--- a/modules/management/kubectl-apply/main.tf
+++ b/modules/management/kubectl-apply/main.tf
@@ -23,6 +23,11 @@ locals {
   apply_manifests_map = tomap({
     for index, manifest in var.apply_manifests : index => manifest
   })
+
+  install_kueue         = try(var.kueue.install, false)
+  install_jobset        = try(var.jobset.install, false)
+  kueue_install_source  = format("${path.module}/manifests/kueue-%s.yaml", try(var.kueue.version, ""))
+  jobset_install_source = format("${path.module}/manifests/jobset-%s.yaml", try(var.jobset.version, ""))
 }
 
 data "google_container_cluster" "gke_cluster" {
@@ -33,7 +38,7 @@ data "google_container_cluster" "gke_cluster" {
 
 data "google_client_config" "default" {}
 
-module "kubectl" {
+module "kubectl_apply_manifests" {
   for_each = local.apply_manifests_map
   source   = "./kubectl"
 
@@ -42,6 +47,42 @@ module "kubectl" {
   template_vars     = each.value.template_vars
   server_side_apply = each.value.server_side_apply
   wait_for_rollout  = each.value.wait_for_rollout
+
+  providers = {
+    kubectl = kubectl
+    http    = http.h
+  }
+}
+
+module "install_kueue" {
+  source            = "./kubectl"
+  source_path       = local.install_kueue ? local.kueue_install_source : null
+  server_side_apply = true
+
+  providers = {
+    kubectl = kubectl
+    http    = http.h
+  }
+}
+
+module "install_jobset" {
+  source            = "./kubectl"
+  source_path       = local.install_jobset ? local.jobset_install_source : null
+  server_side_apply = true
+
+  providers = {
+    kubectl = kubectl
+    http    = http.h
+  }
+}
+
+module "configure_kueue" {
+  source      = "./kubectl"
+  source_path = local.install_kueue ? try(var.kueue.config_path, "") : null
+  depends_on  = [module.install_kueue]
+
+  server_side_apply = true
+  wait_for_rollout  = true
 
   providers = {
     kubectl = kubectl
--- a/modules/management/kubectl-apply/manifests/kueue-v0.8.1.yaml
+++ b/modules/management/kubectl-apply/manifests/kueue-v0.8.1.yaml
@@ -1,0 +1,12735 @@
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/name: kueue
+    control-plane: controller-manager
+  name: kueue-system
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.15.0
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/name: kueue
+    control-plane: controller-manager
+  name: admissionchecks.kueue.x-k8s.io
+spec:
+  group: kueue.x-k8s.io
+  names:
+    kind: AdmissionCheck
+    listKind: AdmissionCheckList
+    plural: admissionchecks
+    singular: admissioncheck
+  scope: Cluster
+  versions:
+  - name: v1beta1
+    schema:
+      openAPIV3Schema:
+        description: AdmissionCheck is the Schema for the admissionchecks API
+        properties:
+          apiVersion:
+            description: |-
+              APIVersion defines the versioned schema of this representation of an object.
+              Servers should convert recognized schemas to the latest internal value, and
+              may reject unrecognized values.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+            type: string
+          kind:
+            description: |-
+              Kind is a string value representing the REST resource this object represents.
+              Servers may infer this from the endpoint the client submits requests to.
+              Cannot be updated.
+              In CamelCase.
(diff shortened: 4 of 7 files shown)


examples/gke-a3-highgpu.yaml
```diff
@@ -56,6 +56,10 @@ deployment_groups:
       master_authorized_networks:
       - cidr_block: $(vars.authorized_cidr)  # Allows your machine run kubectl command. It's required for the multi-network setup.
         display_name: "kubectl-access-network"
+      system_node_pool_taints:
+      - key: "components.gke.io/gke-managed-components"
+        value: true
+        effect: "PREFER_NO_SCHEDULE" # Allows systems like Kueue to be installed on the system default pool
```
nick-stroud, line 62: I suspect this will be problematic and in edge cases we will end up with workloads getting placed on the system pool. Instead of changing the taint, can we add a toleration to the kueue pods?

modules/management/kubectl-apply/main.tf
```diff
@@ -23,6 +23,11 @@ locals {
   apply_manifests_map = tomap({
     for index, manifest in var.apply_manifests : index => manifest
   })
+
+  install_kueue         = try(var.kueue.install, false)
+  install_jobset        = try(var.jobset.install, false)
```
nick-stroud, line 28: IIUC, jobset is a dependency for Kueue to work. Why not automatically install jobset if queue is requested. 

```
  install_jobset        = try(var.jobset.install, local.install_kueue)
```

There is precedence for this sort of thing in startup-script module where we automatically install ansible if the user wants to run an ansible script.

Note: I still think there is value to have a separate variable just for jobset (if you only want to install jobset, or control the version)

modules/management/kubectl-apply/variables.tf
```diff
@@ -36,3 +36,22 @@ variable "apply_manifests" {
   }))
   default = []
 }
+
+variable "kueue" {
+  description = "Install and configure Kueue workload scheduler, more on this here https://kueue.sigs.k8s.io/docs/overview/"
+  type = object({
+    install     = optional(bool, false)
+    version     = optional(string, null)
```
nick-stroud, line 44: nit: I am on the fence between picking a known good version as a default or always taking the latest which I assume is what you are doing here.

chengcongdu: +1 on stable

modules/management/kubectl-apply/variables.tf
```diff
@@ -36,3 +36,22 @@ variable "apply_manifests" {
...
+    config_path = optional(string, null)
```
chengcongdu, line 45: are we going to provide a default config in toolkit?

examples/gke-a3-highgpu.yaml
chengcongdu: I would prefer to have kueue system pod in the system NP, mainly for isolation, stability and maintenance consideration.
PREFER_NO_SCHEDULE make it possible for either case, and would be hard to maintain. Toleration for Kueue system pod seems a better fit.

modules/management/kubectl-apply/variables.tf
```diff
@@ -36,3 +36,22 @@ variable "apply_manifests" {
   }))
   default = []
 }
+
+variable "kueue" {
+  description = "Install and configure Kueue workload scheduler, more on this here https://kueue.sigs.k8s.io/docs/overview/"
```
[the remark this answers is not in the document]
chengcongdu: [] () without space

examples/gke-a3-highgpu.yaml
chengcongdu: resourceFlavor might be use for workload taint/toleration setting https://kueue.sigs.k8s.io/docs/reference/kueue.v1beta1/#kueue-x-k8s-io-v1beta1-ResourceFlavorSpec

ankitkinra: At the moment we are downloading pod installation manifest directly from github without any modifications.
I do not want to modify that file as that has a bunch of things in there, so let me see if we can find if we can apply add tolerations after the manifest has been applied.

modules/management/kubectl-apply/variables.tf
ankitkinra: I have added the known good version in the local block where I am getting the source url for the download. I would want us not to use the LATEST version without doing some sort of integration test.

examples/gke-a3-highgpu.yaml
ankitkinra: I think there are two things to note:
* kueue-controller-manager which is installed via the downloaded github script, IIUC this doesn't use the ResourceFlavor based tolerations those are for workloads that are being admitted via Kueue
* Kueue workloads which in this case we might not want on the system pod as they would be for the GPU nodepool.

I agree if adding tolerations to `kueue-controller-manager` is possible that is the best way forward.

modules/management/kubectl-apply/main.tf
```diff
@@ -23,6 +23,11 @@ locals {
   apply_manifests_map = tomap({
     for index, manifest in var.apply_manifests : index => manifest
   })
+
+  install_kueue         = try(var.kueue.install, false)
+  install_jobset        = anytrue([try(var.jobset.install, false), local.install_kueue])
```
sharabiani, line 28: - AFAIK Jobset is supported in Kueue but it is not a requirement for that.
- However, if we are going to install it with Kueue, it's better not to ignore user's explicit `false` input. Providing false can mean uninstalling Jobset.
- And if we are going to make Kueue's installation depended on Jobset, it's better to control it with precondition.

modules/management/kubectl-apply/main.tf
```diff
@@ -48,3 +53,38 @@ module "kubectl" {
...
+
+  providers = {
+    kubectl = kubectl
+    http    = http.h
+  }
+}
+
+module "configure_kueue" {
+  source      = "./kubectl"
+  source_path = try(var.kueue.config_path, "")
```
sharabiani, line 81: I'd control if local.install_kueue is true, because it will apply configuration manifest if it's provided but `install` setting is false.

modules/management/kubectl-apply/variables.tf
sharabiani: Targeting the latest version is tricky because whenever a new version releases deployment will uninstall/install the kueue or jobset for the customers.
I believe changes in the blueprints or upgrading to a new version of gcluster (probably?) should be the expected reasons for Kueue/Jobset updates.

examples/gke-a3-highgpu.yaml
ankitkinra: Downloaded the manifest and added tolerations

modules/management/kubectl-apply/variables.tf
ankitkinra: It is a pinned version and as we would need to update the tolerations we can only accept the versions of kueue/jobset that we have modified, hence added a known version validation

ankitkinra: +1 on not picking LATEST by default.

modules/management/kubectl-apply/main.tf
```diff
@@ -23,6 +23,11 @@ locals {
   apply_manifests_map = tomap({
     for index, manifest in var.apply_manifests : index => manifest
   })
+
+  install_kueue         = try(var.kueue.install, false)
+  install_jobset        = anytrue([try(var.jobset.install, false), local.install_kueue])
+  kueue_install_source  = format("https://github.com/kubernetes-sigs/kueue/releases/download/%s/manifests.yaml", coalesce(try(var.kueue.version, ""), "v0.8.0"))
```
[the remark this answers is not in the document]
ankitkinra: Done, PTAL

modules/management/kubectl-apply/main.tf
ankitkinra: In order to unblock this PR, I have made kueue and jobset installation independent. We can do this improvement in the next iteration.

modules/management/kubectl-apply/main.tf
ankitkinra: Done

modules/management/kubectl-apply/manifests/kueue-v0.8.1.yaml
chengcongdu, line 1: I thought we will be pulling in the config like [this](https://kueue.sigs.k8s.io/docs/installation/#install-a-custom-configured-released-version), and have some logic to add the toleration. Instead of maintaining a local version, is this feasible?
btw, the PR for adding generic toleration to Kueue got turn down. The SIG laying toward letting user to customize the config after ward instead of providing a generic toleration

examples/gke-a3-highgpu.yaml
```diff
@@ -65,4 +65,12 @@ deployment_groups:
       machine_type: a3-highgpu-8g
       autoscaling_total_min_nodes: 2
       zones: [$(vars.zone)]
-    outputs: [instructions]
```
chengcongdu, line 68: please revert this deletion

curious, are you branching from main instead of develop?

examples/gke-a3-megagpu.yaml
```diff
@@ -65,4 +65,12 @@ deployment_groups:
       machine_type: a3-megagpu-8g
       autoscaling_total_min_nodes: 2
       zones: [$(vars.zone)]
-    outputs: [instructions]
```
chengcongdu, line 68: revert deletion

chengcongdu (approved):
please address the comment before merging

modules/management/kubectl-apply/variables.tf
```diff
@@ -36,3 +36,32 @@ variable "apply_manifests" {
...
+  type = object({
+    install     = optional(bool, false)
+    version     = optional(string, "v0.8.1")
+    config_path = optional(string, null)
+  })
+  default = {}
+
+  validation {
+    condition     = !var.kueue.install || contains(["v0.8.1"], var.kueue.version)
+    error_message = "Supported versions of Kueue are [v0.8.1]"
```
sharabiani, line 51: nit: Supported **_version_** of Kueue **_is_** [v0.8.1]

modules/management/kubectl-apply/variables.tf
```diff
@@ -36,3 +36,32 @@ variable "apply_manifests" {
...
+  description = "Install [Jobset](https://github.com/kubernetes-sigs/jobset) which manages a group of K8s [jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) as a unit."
+  type = object({
+    install = optional(bool, false)
+    version = optional(string, "v0.5.2")
+  })
+  default = {}
+
+  validation {
+    condition     = !var.jobset.install || contains(["v0.5.2"], var.jobset.version)
+    error_message = "Supported versions of Jobset are [v0.5.2]"
```
sharabiani, line 65: nit: Supported _**version**_ of Jobset **_is_** [v0.5.2]

modules/management/kubectl-apply/main.tf
```diff
@@ -48,3 +53,38 @@ module "kubectl" {
     http    = http.h
   }
 }
+
+module "install_kueue" {
+  source            = "./kubectl"
+  source_path       = local.install_kueue ? local.kueue_install_source : null
+  server_side_apply = true
+
```
sharabiani, line 61: nit: maybe we explicitly set `wait_for_rollout = true`? to keep it safe from any changes on default values in kubectl module.

sharabiani (approved):

PR #3021, opened 2024-09-11
Update to RDMA VPC for reducing repeated code

cdunbar13:
Reducing the size of the rdma-vpc module to focus just on the RDMA use case.  This is also implementing feature to reduce repeated code.

Testing done with blueprint which will be shared offline.


tpdownes (approved):
I think we'll want to do a design doc on the rdma-vpc module and how/if we can integrate the features into the standard vpc module. But, for experimental purposes, this reduces crud.

PR #3022, opened 2024-09-11
Create default service account resource in controller only if service_account_email is null

alyssa-sm:
This PR limits the creation of the `google_compute_default_service_account` resource in the controller to only get created when `service_account_email` is null.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
@@ -25,7 +25,7 @@ locals {
     }
   ]
 
-  service_account_email = coalesce(var.service_account_email, data.google_compute_default_service_account.default.email)
+  service_account_email = (var.service_account_email == null || var.service_account_email == "") ? data.google_compute_default_service_account.default[0].email : var.service_account_email
 
   # can't rely on `email=null` as it's used to instantiate `cloudsql_secret_accessor`
   service_account = {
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
```diff
@@ -29,6 +29,8 @@ locals {
 }
 
 data "google_compute_default_service_account" "default" {
+  count = var.service_account_email == null ? 1 : 0
```
mr0re1, line 32: Or empty string, to match behavior of `coalesce`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -25,7 +25,7 @@ locals {
     }
   ]
 
-  service_account_email = coalesce(var.service_account_email, data.google_compute_default_service_account.default.email)
+  service_account_email = coalesce(var.service_account_email, data.google_compute_default_service_account.default[0].email)
```
mr0re1, line 28: Please test with non-empty `var.service_account_email`, to make sure that `coalesce` does "short-circuiting" and will not fail with "index out of bounds"

mr0re1 (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/main.tf
alyssa-sm: Changed it to check for null or empty string.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
alyssa-sm: Good catch. It did not exit the statement and failed with an index error when `var.service_account_email` was non-empty. I changed it to an if statement that checks for null or empty string instead of using `coalesce`.

PR #3024, opened 2024-09-12
Validate Reservations in GKE Blueprints

arajmane-g:
### About
Validate GKE Node Pool Reservations

### Why
We want to ensure that the [requirements wrt consuming reservations](https://cloud.google.com/compute/docs/instances/reservations-overview#requirements) are identified early. Otherwise, the deployed Node Pool won’t consume reservations which would require troubleshooting. Essentially, validations will save users' time

### Testing
Existing integration tests will be modified as a follow-up. Before the test to create a reservation, deploy a cluster to use it, assert the consumption and destroy the reservation during clean up

### Submission Checklist

### DIFF
--- a/modules/compute/gke-node-pool/README.md
+++ b/modules/compute/gke-node-pool/README.md
@@ -264,6 +264,7 @@ limitations under the License.
 | [null_resource.enable_tcpxo_in_workload](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [null_resource.install_dependencies](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [google_compute_default_service_account.default_sa](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_default_service_account) | data source |
+| [google_compute_reservation.specific_reservations](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_reservation) | data source |
 
 ## Inputs
 
--- a/modules/compute/gke-node-pool/main.tf
+++ b/modules/compute/gke-node-pool/main.tf
@@ -30,10 +30,8 @@ locals {
     effect = "NO_SCHEDULE"
   }] : []
 
-  autoscale_set                  = var.autoscaling_total_min_nodes != 0 || var.autoscaling_total_max_nodes != 1000
-  static_node_set                = var.static_node_count != null
-  reservation_resource_api_label = "compute.googleapis.com/reservation-name"
-  specific_reservations_count    = try(length(var.reservation_affinity.specific_reservations), 0)
+  autoscale_set   = var.autoscaling_total_min_nodes != 0 || var.autoscaling_total_max_nodes != 1000
+  static_node_set = var.static_node_count != null
 }
 
 data "google_compute_default_service_account" "default_sa" {
@@ -162,8 +160,8 @@ resource "google_container_node_pool" "node_pool" {
 
     reservation_affinity {
       consume_reservation_type = var.reservation_affinity.consume_reservation_type
-      key                      = local.specific_reservations_count != 1 ? null : local.reservation_resource_api_label
-      values                   = local.specific_reservations_count != 1 ? null : [for reservation in var.reservation_affinity.specific_reservations : reservation.name]
+      key                      = length(local.verified_specific_reservations) != 1 ? null : local.reservation_resource_api_label
+      values                   = length(local.verified_specific_reservations) != 1 ? null : [for r in local.verified_specific_reservations : "projects/${r.project}/reservations/${r.name}"]
     }
 
     dynamic "host_maintenance_policy" {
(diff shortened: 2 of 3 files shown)


modules/compute/gke-node-pool/main.tf
```diff
@@ -162,8 +160,8 @@ resource "google_container_node_pool" "node_pool" {
 
     reservation_affinity {
       consume_reservation_type = var.reservation_affinity.consume_reservation_type
-      key                      = local.specific_reservations_count != 1 ? null : local.reservation_resource_api_label
-      values                   = local.specific_reservations_count != 1 ? null : [for reservation in var.reservation_affinity.specific_reservations : reservation.name]
+      key                      = local.input_specific_reservations_count != 1 ? null : local.reservation_resource_api_label
+      values                   = local.input_specific_reservations_count != 1 ? null : [for reservation in var.reservation_affinity.specific_reservations : reservation.name]
```
arajmane-g, line 164: use `local. verified_specific_reservations` instead so and pass in the full qualified name of the reservation so that shared reservations work.

```
values = length(local.verified_specific_reservations) != 1 ? null : [for reservation in local.verified_specific_reservations : reservation.id]
```

arajmane-g:
PR-test-ml-gke-e2e failed with the error. Will fix it

```
Error: exit status 1

Error: Invalid object key

  on modules/embedded/modules/compute/gke-node-pool/reservation_definitions.tf line 58, in locals:
  58:     "guest_accelerators" : { for acc in local.guest_accelerator : acc.type => acc.count },
    ├────────────────
    │ acc.type is null
```

modules/compute/gke-node-pool/reservation_definitions.tf
```diff
@@ -0,0 +1,69 @@
...
+
+data "google_compute_reservation" "specific_reservations" {
+  for_each = (
+    local.input_specific_reservations_count == 0 ?
+    {} :
+    {
+      for pair in flatten([
+        for zone in var.zones : [
+          for reservation in var.reservation_affinity.specific_reservations : {
+            key : "${coalesce(reservation.project, var.project_id)}/${zone}/${reservation.name}"
```
ankitkinra, line 25: nit: It looks like we are checking that specific reservation could be present in any of the zones present in the blueprint, I am wondering if we additionally ask user for zone in the reservation input, that will make this logic simpler and also user might catch their own mistakes easily.

arajmane-g: > It looks like we are checking that specific reservation could be present in any of the zones present in the blueprint

Yes.

> I am wondering if we additionally ask user for zone in the reservation input, that will make this logic simpler and also user might catch their own mistakes easily.

Actually, it was a deliberate choice. 

The question is whether to make the new variable (`var.reservation_affinity.specific_reservations[].zone`) a required variable  or not. If it were optional, our fallback would still be var.zones and in turn, we would require nested loops.

I chose to not introduce the variable at all in favor of keeping blueprints simple (one less variable) and not asking for input unless absolutely necessary.

One might question why was `var.reservation_affinity.specific_reservations[].projec`t variable introduced then? The project input is different from the zone input in that for making shared reservations work, we have no way of knowing the project which created the reservation. Whereas for zone, its value will always be one of `var.zones`.

tools/cloud-build/daily-tests/blueprints/ml-gke-e2e.yaml
```diff
@@ -57,7 +57,9 @@ deployment_groups:
       name: g2-latest-driver
       machine_type: g2-standard-4
       guest_accelerator:
-      - gpu_driver_installation_config:
+      - type: nvidia-tesla-t4
+        count: 2
```
ankitkinra, line 61: Is this allowed, I thought we could add the type of GPU only for `n series` machine families ?   
Did you mean `nvidia-l4` ? 
Also just for context, how is this change important to this PR ?

modules/compute/gke-node-pool/reservation_definitions.tf
ankitkinra: Thanks for the explanation!

modules/compute/gke-node-pool/README.md
ankitkinra, line 1: Can you please add the testing combinations for this PR as integration test will be added in a subsequent PR.

tools/cloud-build/daily-tests/blueprints/ml-gke-e2e.yaml
arajmane-g: Context: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3024#issuecomment-2346525779

Wanted to see if the any `type`, `count` keys help with the breaking test. 

You are right about the compatibility of the machine type and accelerator. Will use `nvidia-l4`

modules/compute/gke-node-pool/README.md
arajmane-g: Updated the PR description

tools/cloud-build/daily-tests/blueprints/ml-gke-e2e.yaml
ankitkinra: Accelerator type is meant to be optional for a known machine family, i.e. our code should be able to extract that `g2-standard-4` corresponds to `1 nvidia-l4` . 

It now seems that based on this change, we are regressing that feature.

ankitkinra (commented):
Overall LGTM, but left a comment on why we need to explicitly specify type of the accelerator machines

tools/cloud-build/daily-tests/blueprints/ml-gke-e2e.yaml
arajmane-g: > It now seems that based on this change, we are regressing that feature.

Right. 

> .. i.e. our code should be able to extract that g2-standard-4 corresponds to 1 nvidia-l4 .

That was my understanding as well. I suspect there might be a bug here https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/modules/compute/gke-node-pool/gpu_definition.tf#L56

When the `var.guest_accelerator` is specified it overrides `local.generated_guest_accelerator`. But, if the specified value does not have type, count then the `local.guest_accelerator` won't have those keys either.

So, we either fix the gpu_definition itself to have those keys or the `reservation_vm_properties` map construction handles it. I will form an opinion and made changes accordingly.

ankitkinra (approved):

PR #3025, opened 2024-09-12
GKE A3 Mega Integration Test

cdunbar13:
Initial integration test for the gke-a3-megagpu example blueprint.

Does a simple build and destroy of the deployment.  Tested using cloud build.


tpdownes (changes requested):
- Doesn't appear to include destroy
- The "post_deploy_tests" loop depends upon "remote_node"

tpdownes (approved):
Please make sure you follow the internal process to actually enable these tests. Advise the team that they will see some failures in tests unless they rebase off your PR.

Great improvement!

PR #3049, opened 2024-09-18
add additional vpc validation for a3/a3mega machine

chengcongdu:
For A3/A3Mega machine to work properly, additional vpc is needed. This change validate the required vpc is configured when creating nodepool with a3/a3mega machine

Failure on validation won't block the deployment, but will provide warning for user.

sample output when validation fail:
Warning: Check block assertion failed

  on modules/embedded/modules/compute/gke-node-pool/gpu_direct.tf line 54, in check "gpu_direct_check_multi_vpc":
  54:     condition = !(var.machine_type == "a3-megagpu-8g" && length(var.additional_networks) != 8)
    ├────────────────
    │ var.additional_networks is list of object with 7 elements
    │ var.machine_type is "a3-megagpu-8g"

To achieve optimal performance for a3-megagpu-8g machine, 8 additional vpc is
recommended. You could configure it in the blueprint through
modules/network/multivpc with network_count set as 8


modules/compute/gke-node-pool/gpu_direct.tf
```diff
@@ -44,3 +44,14 @@ locals {
...
 }
+
+check "gpu_direct_check_multi_vpc" {
+  assert {
+    condition     = !(var.machine_type == "a3-highgpu-8g" && length(var.additional_networks) != 4)
+    error_message = "To achieve optimal performance for ${var.auto_upgrade} machine, 4 additional vpc is recommended. You could configure it in the blueprint through modules/network/multivpc with network_count set as 4"
+  }
+  assert {
+    condition     = !(var.machine_type == "a3-megagpu-8g" && length(var.additional_networks) != 8)
+    error_message = "To achieve optimal performance for ${var.auto_upgrade} machine, 8 additional vpc is recommended. You could configure it in the blueprint through modules/network/multivpc with network_count set as 8"
```
ankitkinra, line 55: Replace `${var.auto_upgrade}` with `${var.machine_type}`

modules/compute/gke-node-pool/gpu_direct.tf
```diff
@@ -44,3 +44,14 @@ locals {
     }
   }
 }
+
+check "gpu_direct_check_multi_vpc" {
+  assert {
+    condition     = !(var.machine_type == "a3-highgpu-8g" && length(var.additional_networks) != 4)
+    error_message = "To achieve optimal performance for ${var.auto_upgrade} machine, 4 additional vpc is recommended. You could configure it in the blueprint through modules/network/multivpc with network_count set as 4"
```
ankitkinra, line 51: Replace `${var.auto_upgrade}` with `${var.machine_type}`

ankitkinra (approved):

modules/compute/gke-node-pool/gpu_direct.tf
chengcongdu: fixed

modules/compute/gke-node-pool/gpu_direct.tf
```diff
@@ -44,3 +44,14 @@ locals {
     }
   }
 }
+
+check "gpu_direct_check_multi_vpc" {
+  assert {
+    condition     = !(var.machine_type == "a3-highgpu-8g" && length(var.additional_networks) != 4)
```
sharabiani, line 50: - I'd add 'valid_additional_networks_count' to the `gpu_direct_settings` and would have only one general check/condition to validate additional networks count based on machine type.
- Do we need to control 'var.additional_networks' is not null before checking it's length?

tpdownes: One could add `nullable = false` to `variable "additional_networks" { ... }` to prevent its value from ever being `null`. IMO this would be a good idea.

IMO it's better to have separate `assert` with different error messages.

chengcongdu: the default value for additional_networks is [], so technically it should never be null. Adding the `nullable = false` for further guarding.
I would prefer to have separate assertion for easier understanding as well.

PR #3068, opened 2024-09-24
Add initial_node_count support to gke-node-pool

sharabiani:
Support for initial_node_count added to gke-node-pool module.
Set it to provision the number of desired nodes right away after node-pool creation without the need for manual resize using cloud console.

The setting will be considered only when creating a node-pool and in the next deployments it will be ignored.

### Manual tests:
- Added the initial_node_count setting to a node-pool in a blueprint 
- Provisioned the blueprint
- Verified the number of provisioned nodes on Cloud Console


ankitkinra (commented):
If we ignore this setting update, then wouldn't our subsequent configurations using this node count be incorrect ? 

Does this setting work with autoscaling ? How does this relate to `static_node_count` ?

sharabiani:
@ankitkinra 

> If we ignore this setting update, then wouldn't our subsequent configurations using this node count be incorrect ?
It's recommended to ignore it for updates because if the node_pool resized manually, this setting will cause node_pool recreation: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool.html#initial_node_count

> Does this setting work with autoscaling ? How does this relate to `static_node_count` ?

node_count and initial_node_count cannot be set together:
https://github.com/hashicorp/terraform-provider-google/blob/6be57bdad382050af99b07d4b35ede995bfd97fe/google/services/container/resource_container_node_pool.go#L940
I added two precondition to control node_count and min/max total node count in autoscaling.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -354,3 +354,9 @@ variable "host_maintenance_interval" {
     error_message = "Invalid host_maintenance_interval value. Must be PERIODIC, AS_NEEDED or the empty string"
   }
 }
+
+variable "initial_node_count" {
+  description = "The initial number of nodes for the pool. In regional clusters, this is the number of nodes per zone. Changing this setting after node pool creation will not make any effect. It cannot be set with static_node_count and must be set to a value between autoscaling_total_min_nodes and autoscaling_total_max_nodes."
```
ankitkinra, line 359: Just to ask a basic question, is initial node count only relevant for the autoscaling case, if yes can we add that  ? 
Also if there is a GKE documentation on how to use this, can we please refer to that ?

ankitkinra (approved):

modules/compute/gke-node-pool/variables.tf
sharabiani: I did not find any comprehensive documentation about it and that's why I referred to its source code to understand how it works.
[This is the document for api's nodepool object](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters.nodePools#NodePool) and it has some description about initial_node_count.

And regarding if it's bound to autoscaling, I did not find such binding in the source code. But I made this constraint in our module because if static_node_count is not set, we enable autoscaling; and I wanted to have the initial_node_count something between autoscaling range to prevent any possible issue.

```
  dynamic "autoscaling" {
    for_each = local.static_node_set ? [] : [1]
    content {
      total_min_node_count = var.autoscaling_total_min_nodes
      total_max_node_count = var.autoscaling_total_max_nodes
      location_policy      = "ANY"
    }
  }
```

PR #3069, opened 2024-09-24
Update wording about DWS Flex

mr0re1:


docs/slurm-dws-flex.md
```diff
@@ -1,5 +1,10 @@
 # Obtaining SlurmGCP nodes with DWS Flex
 
+> [!NOTE]
+> DWS Flex Start is currently in early development and undergoing extensive testing. While it
```
ankitkinra, line 4: nit: is it possible to add a warning ?

ankitkinra (approved):

docs/slurm-dws-flex.md
mr0re1: I don't think it merits to be a warning. Machines would still be provisioned.

PR #3073, opened 2024-09-25
Allow users to provide a commit hash instead of git tag for Spack and Ramble installations

rohitramu:
This is a bugfix for Spack and Ramble installation.

The git commands are being run from outside the git directory, so `--git-dir` is set.  It seems this is insufficient, and it's possible that `--work-tree` also needs to be set.

[A new argument `-C`](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt) was introduced in git v1.8.5 (in 2013), and it sets both `--git-dir` and `--work-tree`, when given the repository's local directory path.


rohitramu:
I have successfully tested this change with the blueprint used for HPC Image testing, as we first identified the problem there.

nick-stroud (approved):

PR #3075, opened 2024-09-26
NeMo readme instructions for preloading gpt2 tokenizer

koallison:
Addresses error message:
```
ValueError: Unable to instantiate HuggingFace AUTOTOKENIZER for gpt2. Exception: [Errno 2] No such file or directory: '/home/<user>/.cache/huggingface/hub/models--gpt2/snapshots/607a30d783dfa663caf39e06633721c8d4cfcd7e/tokenizer.json'
```

Adding ${HOME}/.cache to container_mounts will download the tokenizer files on the first run but fail the run. The second run succeeds. \
We can preload the tokenizer files by creating an AutoTokenizer object in a short script, allowing the pre-training script to succeed on the first run.

### DIFF
--- a/examples/machine-learning/a3-highgpu-8g/nemo-framework/README.md
+++ b/examples/machine-learning/a3-highgpu-8g/nemo-framework/README.md
@@ -25,6 +25,16 @@ README
 
 3. Run an example NeMo Framework Pre-Training
 
+   First, prepare the cache. This will download several files to the 
+   ~/.cache/huggingface folder which are needed to load the tokenizer for 
+   training.
+   
+   ```shell
+   pip install transformers
+   python -c "from transformers import AutoTokenizer; \
+       AutoTokenizer.from_pretrained('gpt2')"
+   ```
+
    This will run an example of training a 5B parameter GPT3 model for 10 steps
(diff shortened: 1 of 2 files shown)


samskillman (dismissed):
Confirmed that these changes work when combined with updates to use NeMo 24.07 in #3089 


Manual testing was performed on A3-Mega:
* Ran `rm -rf ~/.cache/huggingface`, executed new instructions, and confirmed the same files end up back in that location. Submitting an 8 node job afterwards runs successfully. 
* Did not explicitly test on A3-High, but approach/solution is identical.

examples/machine-learning/a3-highgpu-8g/nemo-framework/README.md
nick-stroud, line 1: Would it be possible to put in the PR description a summary of what manual testing was done?

tpdownes: I am comfortable with documentation in https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3075#pullrequestreview-2338502992

tpdownes (approved):

PR #3077, opened 2024-09-26
Update Toolkit release to v1.40.0

RachaelSTamakloe:
Set release-candidate to version 1.40.0


tpdownes (approved):
Approved. All integration tests passed except the A3 Mega test which failed because our testing capacity was moved to another region. This was addressed by #3086.

PR #3080, opened 2024-09-27
gke-topology-scheduler module implemented

sharabiani:
### Changes:

This module enables topology on a Google Kubernetes Engine cluster.
This is implemented based on sources and instructions explained [here](https://github.com/GoogleCloudPlatform/container-engine-accelerators/tree/master/gpudirect-tcpxo/topology-scheduler).

### DIFF
--- a/community/modules/compute/gke-topology-scheduler/main.tf
+++ b/community/modules/compute/gke-topology-scheduler/main.tf
@@ -1,0 +1,27 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module "kubectl_apply" {
+  source = "../../../../modules/management/kubectl-apply"
+
+  cluster_id = var.cluster_id
+  project_id = var.project_id
+
+  apply_manifests = [
+    { source = "${path.module}/manifests/topology-scheduler-scripts.yaml" },
+    { source = "${path.module}/manifests/service-account.yaml" },
+    { source = "${path.module}/manifests/label-nodes-daemon.yaml" },
+    { source = "${path.module}/manifests/schedule-daemon.yaml" }
+  ]
+}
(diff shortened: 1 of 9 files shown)


community/modules/compute/gke-topology-scheduler/main.tf
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module "kubectl_apply" {
+  source = "../../../../modules/management/kubectl-apply"
+
+  cluster_id = var.cluster_id
+  project_id = var.project_id
+
+  apply_manifests = [
```
ankitkinra, line 21: Is there a way to verify that node pool has compact policies applied ?

community/modules/compute/gke-topology-scheduler/manifests/schedule-daemon.yaml
```diff
@@ -0,0 +1,48 @@
...
+  selector:
+    matchLabels:
+      app: topology-scheduler
+  template:
+    metadata:
+      labels:
+        app: topology-scheduler
+    spec:
+      tolerations:
+      - key: "node-role.kubernetes.io/control-plane"
```
ankitkinra, line 33: Should we add a toleration similar to kueue to allow it to land on system nodepool ? 

```
tolerations:
      - effect: NoSchedule
        key: components.gke.io/gke-managed-components
        operator: Equal
        value: "true"
```

community/modules/compute/gke-topology-scheduler/main.tf
sharabiani: I'm not completely sure which nodepool?
We are not passing any node-pool module to this module.

ankitkinra: Ah okay , I thought topology scheduler only worked when we have nodes giving their placement information .  So if that is not found, topology scheduler is not helpful. But if that is not the case, feel free to disregard.

community/modules/compute/gke-topology-scheduler/manifests/schedule-daemon.yaml
ankitkinra: Which nodepool does Topology scheduler gets provisioned on ? By default our blueprints have two nodepools:
* system (Does not allow any thing to get provisioned until we add the toleration i was describing)
* GPU (we only want GPU workload to land here)

So did you add another nodepool to your blueprint for testing ?

community/modules/compute/gke-topology-scheduler/main.tf
ankitkinra: Followed up offline, we will work on this toleration if it is required as part of blueprint that is using this topology scheduler

community/modules/compute/gke-topology-scheduler/manifests/schedule-daemon.yaml
ankitkinra: Followed offline, this testing will be done with full blueprint

ankitkinra (approved):

community/modules/compute/gke-topology-scheduler/main.tf
sharabiani: Resolving this conversion. Will be tested with the xpk-a3 blueprints and modified if reuired.

PR #3085, opened 2024-09-30
inlude nccl test instruction in output for sample workload

chengcongdu:
provide instruction on how to run nccl test in sample workload for a3-high and a3-mega machines.


modules/compute/gke-node-pool/outputs.tf
```diff
@@ -80,6 +80,14 @@ locals {
 
     You can use the following commands to submit the sample job:
       kubectl create -f ${abspath(local.gpu_direct_setting.updated_workload_path)}
+    After submitting the sample job, you can validate the GPU performance by initiating NCCL test included in the sample workload:
+      NCCL test can be initiated from any one of the sample job Pods and coordinate with the peer Pods:
+      export POD_NAME=$(kubectl get pods -l job-name=my-sample-job -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' | head -n 1)
+      export PEER_POD_IPS=$(kubectl get pods -l job-name=my-sample-job -o go-template='{{range .items}}{{.status.podIP}}{{" "}}{{end}}')
+      kubectl exec --stdin --tty --container=nccl-test $POD_NAME -- /scripts/allgather.sh $PEER_POD_IPS
+    Depends on the Msg size used for transmission in the test, the busbw would different a bit.
```
ankitkinra, line 88: Can we leave this out for now, unless we are super confident in these numbers. Ideally reference benchmarks numbers will be published elsewhere.

tpdownes: Do not include any statement about performance.

tpdownes (dismissed):
No statement about performance expectation should be made. Remove any related text.

ankitkinra (approved):

PR #3100, opened 2024-10-07
Batch BigQuery load of sacct data

### LINKED ISSUE #3008: Logging to BigQuery can fail if number of rows to insert is too large
Pulled from #2989. I ran into an issue where load_bq.py would fail as the number of rows was too large for [insert_rows](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/1e38ce0c2353a3b14520043ab27514d8defd2d71/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py#L270). My workaround was to batch over slices of the `jobs` list with a batch size set to 10000 (which I think is the upper limit for insert_rows).

I can send a PR with the patch but I was a bit hesitant given the lack of unit tests for the script and adding those would be a bit of work (faking the client etc)

fdmalone:
Resolves #3008, chunks BigQuery loads into batches of 10000 rows


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_load_bq.py
```diff
@@ -0,0 +1,39 @@
...
+    print(num_batches)
+    load_cache = []
+    if jobs:
+        start_job_idx = 0
+        end_job_idx = BQ_MAX_ROW_LOAD_SIZE
+        for _ in range(num_batches):
+            load_cache.append(jobs[start_job_idx:end_job_idx])
+            start_job_idx = end_job_idx
+            end_job_idx += BQ_MAX_ROW_LOAD_SIZE
+    if jobs:
```
mr0re1, line 33: nit. If I read it correctly, `if` can be removed 
```py
sum([]) == 0
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_load_bq.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+
+# test the chunking logic if not the BigQuery loads themselves
+@pytest.mark.parametrize("num_jobs_to_load", (0, 11, 10001, 51131, 104321))
+def test_chunked_bq_load(num_jobs_to_load: int):
```
mr0re1, line 20: This code doesn't cover anything from `scripts/load_bq.py`.
I would argue that it doesn't have any utility. Please either refactor `scripts/load_bq.py` in a such way that it can be tested or remove this test file.

mr0re1:
Thank you for contribution! Overall LGTM.
Please address comment about tests.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
```diff
@@ -321,8 +321,16 @@ def main():
     # on failure, an exception will cause the timestamp not to be rewritten. So
     # it will try again next time. If some writes succeed, we don't currently
     # have a way to not submit duplicates next time.
+    print(f"loading BigQuery data in batches of size : {BQ_MAX_ROW_LOAD_SIZE}")
+    num_batches = (len(jobs) // BQ_MAX_ROW_LOAD_SIZE) + 1
```
mr0re1, line 325: Looks like this logic will not work if 
```py
len(jobs) == BQ_MAX_ROW_LOAD_SIZE  => 
num_batches == 2 => 
start_job_idx goes out of bound of jobs list
```

While not a big deal , i suppose `bq_client.insert_rows` can handle empty `table` arg (please confirm if you want to keep current logic), I would advice to either:

* Fix `num_batches` computation, (e.g. use `math.ceil` or add `1` to the length);
* Write it without `num_batches`, e.g.:

```py
...
  while start_job_idx < len(jobs):
  ...
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
```diff
@@ -282,6 +284,26 @@ def bq_submit(jobs):
...
+):
+    """Submit sacct data in batches of size bq_row_batch_size
+
+    Args:
+        jobs: A list of dictionaries of sacct accounting data.
+        submit_function: The method to submit the jobs to BigQuery with. Defaults to bq_submit.
+        bq_row_batch_size: The accounting data will be submitted to BigQuery in
+            batches of this size.
+    """
+    num_batches = int(math.ceil(len(jobs) / bq_row_batch_size))
```
mr0re1, line 298: nit. `math.ceil` already returns `int`, no need to cast.

fdmalone: I would hold off on review for a minute. going to make a bigger change.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
```diff
@@ -282,6 +284,26 @@ def bq_submit(jobs):
     print(f"successfully loaded {len(jobs)} jobs")
 
 
+def batched_bq_submit(
+    jobs, submit_function=bq_submit, bq_row_batch_size=BQ_ROW_BATCH_SIZE
```
mr0re1, line 288: No need for such complexity if only reason for this testability. Did you consider using mock patchers instead?
https://docs.python.org/3/library/unittest.mock.html#the-patchers

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_load_bq.py
mr0re1, line 33: No sure if this condition is requires , `sum([]) == 0`  so just checking

```
sum([sum(x) for x in load_cache]) == num_jobs_to_load * (num_jobs_to_load + 1) // 2
```

Should be sufficient.

fdmalone: I'm updating the tests, sorry I just pushed didn't mean to trigger re-review :)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
fdmalone: yeah I think patching requires a bit of refactoring, currently the client etc are all global in load_bq

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
fdmalone: whoops, fixed to a for loop

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
fdmalone: done.

fdmalone:
I removed the test. I think testing this cleanly requires some refactoring to move the client and table instances into main() and passed as arguments (or cached as singletons) so that we could mock them. Currently importing from load_bq will try to connect to a cloud project etc.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/load_bq.py
```diff
@@ -322,7 +323,13 @@ def main():
     # it will try again next time. If some writes succeed, we don't currently
     # have a way to not submit duplicates next time.
     if jobs:
-        bq_submit(jobs)
+        num_batches = (len(jobs) - 1) // BQ_ROW_BATCH_SIZE + 1
+        print(
+            f"loading {num_batches} batches of BigQuery data in batches of size : {BQ_ROW_BATCH_SIZE}"
+        )
+        for indx in range(0, len(jobs), BQ_ROW_BATCH_SIZE):
+            print(f"loading BigQuery data batch {indx} of {num_batches}")
```
mr0re1, line 331: Misleading message: "loading BigQuery data batch 15000  of 3"

mr0re1: could be written as 
`f"loading BigQuery data batch {indx // BQ_ROW_BATCH_SIZE} of {num_batches}"`

or simply don't use `num_batches`

mr0re1 (approved):

PR #3101, opened 2024-10-07
SlurmGCP. Do not put job_submit script into config.

mr0re1:
**Motivation:**
* Reduce clutter in user-visible config;
* Move to the same lifecycle as other scripts. 

**Changes:**
* Make `job_submit` a part of "devel" zip;
* Move from `etc` to `scripts`;
* Remove `slurm_files` variable;
* Apply auto-formatting, no other changes to the `job_submit.lua.tpl`.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
```diff
@@ -360,19 +360,20 @@ def install_cgroup_conf(lkp: util.Lookup) -> None:
...
     ):
-        conf_options = {
-            "scripts_dir": lkp.cfg.slurm_scripts_dir or dirs.scripts,
-        }
-        conf = lkp.cfg.jobsubmit_lua_tpl.format(**conf_options)
-
-        conf_file = lkp.etc_dir / "job_submit.lua"
-        conf_file.write_text(conf)
-        util.chown_slurm(conf_file, 0o600)
+        return # No TPU partitions, no need for job_submit.lua
```
abbas1902, line 368: For my own learning, is doing it this way is preferable since we'd expect most deployments to end up using this control flow and hence can take advantage of the early return?

abbas1902 (approved):
Overall LGTM, just a question about one of the stylistic choices made.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
mr0re1: It's not about performance gains, but readability.

PR #3104, opened 2024-10-07
Retry `wait-for-startup` script on internal error

rohitramu:
The command `gcloud compute instances get-serial-port-output` occasionally fails with an internal error, unrelated to the client's calling code.  An example of the error:
```
ERROR: (gcloud.compute.instances.get-serial-port-output) Could not fetch
serial port output: Internal error. Please try again or contact Google
Support. (Code: '623E40738DA88.30A0B2F.2313B7F8')

Error: exit status 1

Error: local-exec provisioner error

  with module.wait-for-instances.null_resource.wait_for_startup[0],
  on modules/embedded/community/modules/scripts/wait-for-startup/main.tf line 33, in resource "null_resource" "wait_for_startup":
  33:   provisioner "local-exec" {

Error running command '/bin/bash
modules/embedded/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh':
exit status 1
```

This PR adds error handling for what should be "non-fatal errors", and defines "Internal error" as one of them.

### DIFF
--- a/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
+++ b/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
@@ -60,14 +60,29 @@ FINISH_LINE="startup-script exit status"
 # Match string for failures on the new guest agent
 FINISH_LINE_ERR="Script.*failed with error:"
 
+NON_FATAL_ERRORS=(
+	"Internal error"
+)
+
 until [[ now -gt deadline ]]; do
 	ser_log=$(
 		set -o pipefail
 		${fetch_cmd} 2>"${error_file}" |
 			c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}"
 	) || {
-		cat "${error_file}"
-		exit 1
+		err=$(cat "${error_file}")
+		echo "$err"
+		fatal_error="true"
+		for e in "${NON_FATAL_ERRORS[@]}"; do
+			if [[ $err = *"$e"* ]]; then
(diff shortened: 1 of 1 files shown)


community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -60,14 +60,29 @@ FINISH_LINE="startup-script exit status"
...
 		set -o pipefail
 		${fetch_cmd} 2>"${error_file}" |
 			c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}"
 	) || {
-		cat "${error_file}"
-		exit 1
+		err=$(cat "${error_file}")
+		echo "$err"
+		fatal_error="true"
+		for e in "${NON_FATAL_ERRORS[@]}"; do
```
mr0re1, line 76: Looks like we are adding a risk of running an endless-loop of "NON_FATAL_ERRORS". I guess it's not a big problem since there is a global timeout.

mr0re1 (approved):

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
rohitramu: Doesn't this just loop through the list of strings?  It's just trying to determine whether the error message contains one of the substrings in $NON_FATAL_ERRORS.

Is there a possibility of an infinite loop that I'm missing elsewhere?

PR #3107, opened 2024-10-08
upgrading tpg from 5.x to 6.x

RachaelSTamakloe:
### Submission Checklist

### DIFF
--- a/pkg/config/expand.go
+++ b/pkg/config/expand.go
@@ -199,11 +199,11 @@ func getDefaultGoogleProviders(bp Blueprint) map[string]TerraformProvider {
 	return map[string]TerraformProvider{
 		"google": {
 			Source:        "hashicorp/google",
-			Version:       ">= 4.84.0, < 6.7.0",
+			Version:       ">= 4.84.0, < 6.8.0",
 			Configuration: gglConf},
 		"google-beta": {
 			Source:        "hashicorp/google-beta",
-			Version:       ">= 4.84.0, < 6.7.0",
+			Version:       ">= 4.84.0, < 6.8.0",
 			Configuration: gglConf}}
 }
 
--- a/pkg/config/expand_test.go
+++ b/pkg/config/expand_test.go
@@ -93,10 +93,10 @@ func (s *zeroSuite) TestExpandProviders(c *C) {
 		c.Check(g.TerraformProviders, DeepEquals, map[string]PR{
 			"google": TerraformProvider{
 				Source:  "hashicorp/google",
-				Version: ">= 4.84.0, < 6.7.0"},
+				Version: ">= 4.84.0, < 6.8.0"},
 			"google-beta": TerraformProvider{
 				Source:  "hashicorp/google-beta",
-				Version: ">= 4.84.0, < 6.7.0"}})
+				Version: ">= 4.84.0, < 6.8.0"}})
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/config/expand.go
```diff
@@ -199,11 +199,11 @@ func getDefaultGoogleProviders(bp Blueprint) map[string]TerraformProvider {
 	return map[string]TerraformProvider{
 		"google": {
 			Source:        "hashicorp/google",
-			Version:       ">= 4.84.0, < 5.45.0",
+			Version:       ">= 4.84.0, < 6.7.0",
```
tpdownes, line 202: It's not obvious to me whether our changes may be backwards compatible to 4.84.0. The safest claim is that the final few releases of the 5.x branch are compatible because they were given the 6.x behavior while supporting the old behavior with deprecation warnings. They are still doing some backports on the 5.x release. I believe the appropriate target releases are now:

- https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.44.2
- https://github.com/hashicorp/terraform-provider-google/releases/tag/v6.7.0

i.e. 
```suggestion
			Version:       ">= 5.44.2, < 6.8.0",
```

pkg/config/expand.go
```diff
@@ -199,11 +199,11 @@ func getDefaultGoogleProviders(bp Blueprint) map[string]TerraformProvider {
...
 			Configuration: gglConf},
 		"google-beta": {
 			Source:        "hashicorp/google-beta",
-			Version:       ">= 4.84.0, < 5.45.0",
+			Version:       ">= 4.84.0, < 6.7.0",
```
tpdownes, line 206: Same.

```suggestion
			Version:       ">= 5.44.2, < 6.8.0",
```

tpdownes (changes requested):
IMO we should also be updating the versions.tf of every module you've touched to require `>= 5.44.2`.

tpdownes:
Outwardly it appears that the 3rd party DAOS module claims only compatibility with 4.84 series. That is probably inaccurate as anything that is 4.84-compatible should also be 5.x compatible.

https://github.com/daos-stack/google-cloud-daos/blob/43f7fe3641e7fc6a6da5476fcec91484b5f9cd97/terraform/modules/daos_server/versions.tf

FYI, you can find the root of provider version constraint mismatches with the `terraform providers` command. This one is relatively easy to read from the output, though.

https://developer.hashicorp.com/terraform/cli/commands/providers

Options (not necessarily exclusive):

1. We now have 1st party DAOS in the form of Parallelstore. We can transition customers to that via a deprecation plan for these blueprints.
2. Work with Intel to update the version constraint. It does not follow recommended Terraform best practices because it is imposing an upper bound.

tpdownes (changes requested):
Please test and confirm that the code as-written is compatible with the current target TPG version. I believe the issue is when the intent is to have no GPU accelerators so simply testing that case directly should be enough.

When I read the release notes, there doesn't appear to be a late 5.x release which explicitly supports both formats. So it makes me think the new config is entirely backwards compatible. So I'm just looking for confirmation.

tpdownes (approved):
LGTM. Discuss was had to leave 4.84 minimum bound in place as we develop a plan to migrate our DAOS module users to a Parallelstore solution.

Please squash the last two commits into one so that the git history is more sensible.

PR #3108, opened 2024-10-08
Implement xpk-gke-a3-megagpu blueprint

sharabiani:
### xpk-gke-a3-megagpu blueprint implemented
- Provisions a gke-cluster with multiple node-pools with A3Mega machine types
- Sets multiple group placement policies for each node-pool
- Installs Topology Aware Scheduler
- Installs and Configures Kueue and Jobset
- Creates a ConfigMap


community/examples/xpk-gke-a3-megagpu-files/kueue-credentials.yaml.tftpl
ankitkinra, line 1: i think we can rename this file as kueue-xpk-configuration.yaml.tftpl

community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,118 @@
...
+      jobset:
+        install: true
+
+  - id: topology_aware_scheduler_install
+    source: community/modules/compute/gke-topology-scheduler
+    use: [gke_cluster]
+
+  - id: workload_manager_config
+    source: modules/management/kubectl-apply
+    use: [gke_cluster]
```
ankitkinra, line 112: should we add a dependency on `workload_manager_install` which installs kueue ?

community/examples/xpk-gke-a3-megagpu-files/kueue-credentials.yaml.tftpl
sharabiani: done

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: I'm trying make [these changes](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3111) and it will satisfy this dependency need.

Anyhow, is there anyway to explicitly make dependencies between these blocks?

community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,120 @@
...
+      network_name_prefix: $(vars.deployment_name)-gpunet
+      global_ip_address_range: 192.169.0.0/16
+      network_count: 8
+      subnetwork_cidr_suffix: 24
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network1, gpunets]
+    settings:
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
```
ankitkinra, line 55: Why is this required ? Can we not use the previous method of `authorized_cidr` to work ?

community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,120 @@
...
+      master_authorized_networks:
+      - cidr_block: $(vars.authorized_cidr)  # Allows your machine run kubectl command. It's required for the multi-network setup.
+        display_name: "kubectl-access-network"
+      #min_master_version: "1.29.6-gke.1326000" # (TODO: Ask MaxText the reason to set this) Couldn't find this version in the valid master versions in australia-southeast1-c. Can be left unset to be set by GKE to the version of the most recent official release.
```
ankitkinra, line 59: What is the xpk default ?

community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -0,0 +1,120 @@
...
+  - id: group_placement_0
+    source: modules/compute/resource-policy
+    settings:
+      name: $(vars.deployment_name)-gp-np-0
+      group_placement_max_distance: 2
+
+  - id: group_placement_1
+    source: modules/compute/resource-policy
+    settings:
+      name: $(vars.deployment_name)-gp-np-0
```
ankitkinra, line 72: Name conflict with the previous one policy ?

community/examples/xpk-gke-a3-megagpu.yaml
ankitkinra: I realized that there is no way to add dependency unless we have input/output relationship via `use`. So if we want to add dependency this will need to be in a separate deployment group .

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: I tried to make the blueprint's output as close as to xpk's. XPk doesn't support private cluster yet.
I assume the reason is if one creates a cluster using Cluster Toolkit, and then push workloads from different machines, xpk must receive multiple cidr and apply it. Our current single `authorized_cidr` might not be sufficient either, and they would need to add more authorized_cidrs using GCP Console.
What is your opinion?

ankitkinra: Sounds good , perhaps we can add that in the comments next to this variable ?

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: GKE sets the default version if it is not provided by user. my comment is to ask why did they wanted to have that specific version.

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: A last-minute modification mistake. Fixed.

community/examples/xpk-gke-a3-megagpu.yaml
ankitkinra: Please update this and otherwise I am going to LGTM

ankitkinra (approved):

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: I remove the `enable_priate_rndpoint: false` for now. They can open access for all ips by setting CIDR 0.0.0.0/0 . We are creating a private cluster in Cluster Toolkit and only this variable is not enough to make it public. [Link](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#gke_private_clusters_1-)

community/examples/xpk-gke-a3-megagpu.yaml
sharabiani: removed the comment. We can add later if needed.

PR #3114, opened 2024-10-09
Experimental Networking Features for GKE

arajmane-g:
Move the experimental GKE networking features to the experimental branch. 

### Submission Checklist


modules/compute/gke-node-pool/main.tf
```diff
@@ -44,6 +44,7 @@ resource "google_container_node_pool" "node_pool" {
   name           = var.name == null ? var.machine_type : var.name
   cluster        = var.cluster_id
   node_locations = var.zones
+  version        = var.is_gke_sandbox ? var.node_version : null
```
ankitkinra, line 47: We should not add `is_gke_sandbox` here, can we not expose the version in all cases ?

modules/compute/gke-node-pool/variables.tf
```diff
@@ -354,3 +354,17 @@ variable "host_maintenance_interval" {
     error_message = "Invalid host_maintenance_interval value. Must be PERIODIC, AS_NEEDED or the empty string"
   }
 }
+
+# REMOVE_ME: It's a temporary variable used in internal testing
```
ankitkinra, line 358: remove this ?

modules/compute/gke-node-pool/variables.tf
```diff
@@ -354,3 +354,17 @@ variable "host_maintenance_interval" {
...
+variable "is_gke_sandbox" {
+  description = "Temporary variable to identify the GKE sandbox environment"
+  default     = false
+  type        = bool
+}
+
+# REMOVE_ME: It's a temporary variable used in internal testing
```
ankitkinra, line 365: Can it be exposed always ?

modules/scheduler/gke-cluster/main.tf
```diff
@@ -47,7 +85,7 @@ resource "google_container_cluster" "gke_cluster" {
 
   project         = var.project_id
   name            = local.name
-  location        = var.region
+  location        = var.is_gke_sandbox ? var.zone : var.region
```
ankitkinra, line 88: Can we instead name this variable as `create_multi_zonal_cluster` ?

modules/scheduler/gke-cluster/main.tf
```diff
@@ -59,6 +97,10 @@ resource "google_container_cluster" "gke_cluster" {
 
   network    = var.network_id
   subnetwork = var.subnetwork_self_link
+  # Note: Though the default value of VPC_NATIVE is sufficient to enable IP Aliasing,
```
ankitkinra, line 100: My counter argument which you may not agree to is, if GKE decides that default mode is not `VPC_NATIVE` but something else based on the other variables that are passed in, maybe we do want to default to that ?

modules/scheduler/gke-cluster/main.tf
```diff
@@ -196,9 +238,12 @@ resource "google_container_node_pool" "system_node_pools" {
   provider = google-beta
   count    = var.system_node_pool_enabled ? 1 : 0
 
-  project = var.project_id
-  name    = var.system_node_pool_name
-  cluster = google_container_cluster.gke_cluster.self_link
+  project  = var.project_id
+  name     = var.system_node_pool_name
+  cluster  = var.is_gke_sandbox ? google_container_cluster.gke_cluster.name : google_container_cluster.gke_cluster.self_link
```
ankitkinra, line 243: Can we create explicit variable called `use_gke_self_link` (and then set in blueprint). Again my intent not exposing sandbox in providers ?

modules/scheduler/gke-cluster/main.tf
```diff
@@ -196,9 +238,12 @@ resource "google_container_node_pool" "system_node_pools" {
...
+  version  = var.min_master_version
+  location = var.is_gke_sandbox ? var.zone : null
```
ankitkinra, line 245: ditto as above

modules/scheduler/pre-existing-gke-cluster/main.tf
```diff
@@ -20,26 +20,51 @@ data "google_container_cluster" "existing_gke_cluster" {
...
         source        = "${path.module}/templates/network-object.yaml.tftpl",
-        template_vars = { name = "vpc${idx + 1}" }
+        template_vars = { name = network_info.subnetwork }
       }
     ]
   ])
 }
+
+module "kubectl_apply" {
+  source = "../../management/kubectl-apply" # can point to github
```
ankitkinra, line 64: nit: del `# can point to github`

community/modules/network/rdma-vpc/main.tf
```diff
@@ -125,6 +125,23 @@ locals {
       alias_ip_range     = []
     }
   ]
+
+  # FIX_ME(arajmane): There is a concern about this not working in a shared VPC environment. 
+  # To unblock experimental testing, we decided to go ahead with this.
```
tpdownes, line 130: I believe that this choice (`subnetwork_project = null`) will run into problems in any environment that does not have a provider set at the project level.

It should not be merged as-is unless you can demonstrate that adding `subnetwork_project = var.project_id` causes known problems.

tpdownes (changes requested):

community/modules/network/rdma-vpc/main.tf
tpdownes: Specifically your code is going to error when there is no default project specified in the provider. It's better not to fall back to that.

It will hit this:

https://github.com/hashicorp/terraform-provider-google/blob/main/google/tpgresource/field_helpers.go#L397

after hitting this:

https://github.com/hashicorp/terraform-provider-google/blob/00a52318354fe60774305a53606a7c2f7ff6a9d3/google/services/compute/compute_instance_helpers.go#L466

modules/scheduler/gke-cluster/main.tf
arajmane-g: Introducing `cluster_availability_type` which takes the values: `{REGIONAL, ZONAL}`

modules/scheduler/gke-cluster/main.tf
arajmane-g: Introducing `cluster_reference_type` which takes the values: `{SELF_LINK, NAME}` with `SELF_LINK` being the default

community/modules/network/rdma-vpc/main.tf
arajmane-g: @tpdownes yes, will update the `subnetwork_project = var.project_id`. 

Ideally, I understand we should be using self_link. But, looks like the GKE sandbox environment limitation to me, ATM. 

That said, I am going to check with the GKE team to see why in the sandbox environment self_link isn't working for subnetwork. In fact, similarly when creating the `system_node_pools` we are needing to pass the cluster name instead of self link.

arajmane-g: see modules/scheduler/gke-cluster/main.tf in this PR

cdunbar13 (approved):
LGTM for now.  There is at least one issue that should be addressed in another PR.

Fixing `subnetwork_name = "${var.rdma_subnetwork_name_prefix}-${idx}",` in the gke-cluster and pre-existing-gke-cluster.  This should use the subnetwork name that comes from the networks that are passed in (either require subnetworks to be populated in the list or get the information from a data source)

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -334,16 +334,20 @@ variable "rdma_subnetwork_name_prefix" {
...
-  description = "Temporary variable to identify the GKE sandbox environment"
-  default     = false
-  type        = bool
+variable "cluster_reference_type" {
+  description = "How the google_container_node_pool.system_node_pools refers to the cluster. Possible values are: {SELF_LINK, NAME}"
+  default     = "SELF_LINK"
+  type        = string
+}
+
+variable "cluster_availability_type" {
```
ankitkinra, line 343: nit: I would just use the boolean `is_multi_zonal` as I do not think this will change that often.
but this can be done later in later PR, if you agree

ankitkinra (approved):

tpdownes (approved):

modules/scheduler/gke-cluster/variables.tf
cdunbar13: Resolving this to merge.

PR #3046, opened 2024-09-18
Fix multivpc missing region

wiktorn:
This helps avoid errors like:
```
Error: cannot determine self_link for subnetwork "***": Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.
```

When just `use`-ing the generated networks.

### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #3115, opened 2024-10-09
Revert PR#3046 and add more line breaks for readability

ankitkinra:
Reverts https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3046 as that was causing GKE blueprints to fail.
Confirmed with @wiktorn that the revert is okay and does not break any functionality.

GKE blueprint error:

```
Error: Invalid count argument

  on modules/embedded/modules/management/kubectl-apply/kubectl/main.tf line 50, in data "kubectl_path_documents" "yamls":
  50:   count   = local.directory != null ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

Theory is as follows:

vpc.subnetwork_name is know before apply and vpc.subnetwork_self_link is not known until after apply. If the name is an input then it should be determined prior to apply time.

If this is the case then there was a count (branching logic) that before change was known at apply but after change is unknown until vpc is applied

### DIFF
--- a/modules/network/multivpc/main.tf
+++ b/modules/network/multivpc/main.tf
@@ -24,7 +24,11 @@ locals {
   maximum_subnetworks   = pow(2, local.subnetwork_new_bits)
   additional_networks = [
     for vpc in module.vpcs :
-    merge(var.network_interface_defaults, { network = vpc.network_name, subnetwork = vpc.subnetwork_name, subnetwork_project = var.project_id })
+    merge(var.network_interface_defaults, {
+      network            = vpc.network_name
+      subnetwork         = vpc.subnetwork_name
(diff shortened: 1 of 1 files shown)


wiktorn (approved):

nick-stroud (dismissed):

tpdownes (approved):

nick-stroud (approved):

ankitkinra:
Merging as most of the required tests have passed `PR-test-ml-a3-highgpu-slurm (hpc-toolkit-dev) `

PR #3123, opened 2024-10-11
Create and use non-default service accounts in GKE

annuay-google:
## Approach
- Allow user to create new SA from blueprint and use the same in cluster and node pool
- Do not fetch default SA from a data source, instead construct using project id
- User can pass a pre existing SA with appropriate permissions. Checking for the right permissions is not in the scope of this PR and will be covered in a follow up PR

## Testing
- [x] Deployed hpc-gke.yaml with non default new SA and verified provisioning and destruction, and that SA with appropriate roles was created. Verified all node pools used the SA specified
- [x] Deployed hpc-gke.yaml with non default pre-existing SA and verified provisioning and destruction, and that no new SA was created. Verified all node pools used the SA specified
- [x] Deployed hpc-gke.yaml with no SA passed. Verified default SA was used in all node pools
- [x] Verified all GKE integration tests passed
- [x] Verified logs are pushed from k8s pods and visible in GCP logging from Pantheon

### Create and use new SA (one per node pool) - recommended
Check updated example blueprints
### Use pre-existing SA
```
  - id: gke_cluster
    source: modules/scheduler/gke-cluster
    use: [network1]
    settings:
      enable_private_endpoint: false  # Allows for access from authorized public IPs
      service_account_email: ml-01-gke-sa@ns-playground-a.iam.gserviceaccount.com
    outputs: [instructions]

  - id: compute_pool
    source: modules/compute/gke-node-pool
    use: [gke_cluster]
    settings:
      service_account_email: ml-01-gke-sa@ns-playground-a.iam.gserviceaccount.com
```
### Use default SA
```
  - id: gke_cluster
    source: modules/scheduler/gke-cluster
    use: [network1]
    settings:
      enable_private_endpoint: false  # Allows for access from authorized public IPs
    outputs: [instructions]

  - id: compute_pool
    source: modules/compute/gke-node-pool
    use: [gke_cluster]
```

### Submission Checklist

### DIFF
--- a/examples/ml-gke.yaml
+++ b/examples/ml-gke.yaml
@@ -40,19 +40,32 @@ deployment_groups:
         - range_name: services
           ip_cidr_range: 10.0.32.0/20
 
+  - id: gke_service_account
+    source: community/modules/project/service-account
+    settings:
+      name: gke-sa
+      project_roles:
+      - logging.logWriter
+      - monitoring.metricWriter
+      - monitoring.viewer
+      - stackdriver.resourceMetadata.writer
+      - storage.objectViewer
+      - artifactregistry.reader
+
   - id: gke_cluster
     source: modules/scheduler/gke-cluster
-    use: [network1]
+    use: [network1, gke_service_account]
     settings:
       enable_private_endpoint: false  # Allows for access from authorized public IPs
       master_authorized_networks:
       - display_name: deployment-machine
         cidr_block: $(vars.authorized_cidr)
+      configure_workload_identity_sa: true
     outputs: [instructions]
 
   - id: g2_pool
     source: modules/compute/gke-node-pool
-    use: [gke_cluster]
+    use: [gke_cluster, gke_service_account]
     settings:
       disk_type: pd-balanced
       machine_type: g2-standard-4
--- a/examples/storage-gke.yaml
+++ b/examples/storage-gke.yaml
@@ -38,9 +38,33 @@ deployment_groups:
         - range_name: services
           ip_cidr_range: 10.0.32.0/20
 
+  - id: gke_service_account
+    source: community/modules/project/service-account
+    settings:
+      name: gke-sa
+      project_roles:
+      - logging.logWriter
+      - monitoring.metricWriter
+      - monitoring.viewer
+      - stackdriver.resourceMetadata.writer
+      - storage.objectViewer
+      - artifactregistry.reader
+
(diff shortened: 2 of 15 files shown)


examples/ml-gke.yaml
```diff
@@ -40,19 +40,32 @@ deployment_groups:
         - range_name: services
           ip_cidr_range: 10.0.32.0/20
 
+  - id: gke_service_account
+    source: community/modules/project/service-account
+    settings:
+      name: gke-service-account
+      project_roles:
+      - logging.logWriter
```
ankitkinra, line 48: Would we not need more permissions like compute.createResources, networking permissions, etc?

annuay-google: AFAIK, that's not required. We're passing the service to the GKE cluster (and GKE node pool), which uses this SA in the node_config argument. This is the SA the nodes will use. [parameter definition](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#service_account). I'm running all GKE integration tests to see if there's something I've not caught

modules/compute/gke-node-pool/main.tf
```diff
@@ -20,8 +20,6 @@ locals {
 }
 
 locals {
-  sa_email = var.service_account_email != null ? var.service_account_email : data.google_compute_default_service_account.default_sa.email
```
sharabiani, line 23: Are we going to still use the default_service_account if the gke_service_account not provided?
If yes I believe we should keep the current logic and resources.

examples/storage-gke.yaml
```diff
@@ -38,9 +38,33 @@ deployment_groups:
         - range_name: services
           ip_cidr_range: 10.0.32.0/20
 
+  - id: gke_service_account
```
sharabiani, line 41: Isn't `gcp_service_account` more descriptive? GKE service accounts are to manage k8s resources while the account we need is GCP IAM sa.

examples/ml-gke.yaml
ankitkinra: Ack, thanks!

ankitkinra (approved):

modules/compute/gke-node-pool/main.tf
sharabiani: please ignore

examples/storage-gke.yaml
sharabiani: please ignore

PR #3124, opened 2024-10-11
GKE A3 high integration test

ighosh98:
Initial integration test for the gke-a3-highgpu example blueprint.

Does a simple build and destroy of the deployment. Tested using cloud build.

### DIFF
--- a/tools/cloud-build/daily-tests/builds/gke-a3-highgpu.yaml
+++ b/tools/cloud-build/daily-tests/builds/gke-a3-highgpu.yaml
@@ -1,0 +1,66 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+tags:
+- m.gke-cluster
+- m.gke-node-pool
+- m.vpc
+- m.multivpc
+- m.service-account
+- m.kubectl-apply
+- gke
+
+timeout: 14400s  # 4hr
+steps:
+- id: gke-a3-highgpu
+  name: us-central1-docker.pkg.dev/$PROJECT_ID/hpc-toolkit-repo/test-runner
+  entrypoint: /bin/bash
+  env:
+  - "ANSIBLE_HOST_KEY_CHECKING=false"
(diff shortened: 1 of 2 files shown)


annuay-google:
Looks like there are some A3 mega file changes. If it's not intended, can you please clean those up?

ighosh98:
The a3 mega files are needed as PR-OFE-test was failing without it being added

cdunbar13:
> The a3 mega files are needed as PR-OFE-test was failing without it being added

Did you rebase your branch on the newest develop?

ighosh98:
Yes. When the test failed, I ran a rebase(`git pull --rebase`) and pushed the changes.

annuay-google (approved):
LGTM. Asking @cdunbar13 to take a look as well, since I've reviewed the differences between this PR and his primarily

tools/cloud-build/daily-tests/builds/gke-a3-high.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - -c
+  - |
+    set -x -e
+    cd /workspace && make
+    BUILD_ID_FULL=$BUILD_ID
+    BUILD_ID_SHORT=$${BUILD_ID_FULL:0:6}
+    EXAMPLE_BP=examples/gke-a3-highgpu.yaml
+
+    # Replacing the static subnet name to prevent collisions
+    sed  -i "s/gke-subnet-a3-high/gke-subnet-a3-high-$${BUILD_ID_SHORT}/" $${EXAMPLE_BP}
```
cdunbar13, line 42: Was this tested?  Looking at the a3high gke blueprint, it looks like this will match weirdly.  Notice that it's `gke-subnet-a3-highgpu` and you're missing the `gpu` part.  The `gpu` part was not in the mega blueprint, which is why this match is probably wrong.

cdunbar13 (changes requested):
Can you leave a comment with a build that ran successfully and address the other comment I have.  After those two, it should be good to go.

tools/cloud-build/daily-tests/builds/gke-a3-high.yaml
ighosh98: I had tested it out once before raising the CR. Let me share the test artifacts with you offline. I am ok making the change you have proposed. Can discuss it offline.

ighosh98:
Build ran successfully.

tools/cloud-build/daily-tests/builds/gke-a3-high.yaml
ighosh98: I looked at the [a3-mega template code](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/tools/cloud-build/daily-tests/builds/gke-a3-megagpu.yaml#L43). The high-gpu code follows the structure as defined here. If we are making changes in this template, should a change be made in the mega template as well? Let me know which convention we should follow. 

I made a commit with the gpu change. Testing with the gpu change was causing the pre-commit to fail as well. The integration test passes in both the cases.

cdunbar13 (approved):

PR #3127, opened 2024-10-11
gke-node-pool default name conflict fixed

sharabiani:
This PR fixes the default name conflicts when multiple gke-node-pool is added to a blueprint with a same machine type and without specifying a name.

The machine_type used to be considered as name for the node-pool so in the above scenario multiple modules were trying to create the same node-pool. Now the default name will be machine_type + a random string.

### DIFF
--- a/modules/compute/gke-node-pool/main.tf
+++ b/modules/compute/gke-node-pool/main.tf
@@ -39,10 +39,14 @@ data "google_compute_default_service_account" "default_sa" {
   project = var.project_id
 }
 
+resource "random_id" "nodepool_name_suffix" {
+  byte_length = 8
+}
+
 resource "google_container_node_pool" "node_pool" {
   provider = google-beta
 
-  name           = var.name == null ? var.machine_type : var.name
+  name           = var.name == null ? "${var.machine_type}-${random_id.nodepool_name_suffix.hex}" : var.name
   cluster        = var.cluster_id
   node_locations = var.zones
 
--- a/modules/compute/gke-node-pool/variables.tf
+++ b/modules/compute/gke-node-pool/variables.tf
@@ -31,7 +31,7 @@ variable "zones" {
 }
 
 variable "name" {
-  description = "The name of the node pool. If left blank, will default to the machine type."
+  description = "The name of the node pool. If left blank, will default to the machine type and a suffix with a random string."
   type        = string
   default     = null
 }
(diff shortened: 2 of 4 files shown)


modules/compute/gke-node-pool/variables.tf
```diff
@@ -31,11 +31,19 @@ variable "zones" {
...
-  description = "The name of the node pool. If left blank, will default to the machine type."
+  description = <<-EOD
+    The name of the node pool. If not set, automatically populated by machine type and module id (unique blueprint-wide) as suffix.
+    If setting manually, ensure a unique value across all gke-node-pools.
+    EOD
   type        = string
   default     = null
 }
 
+variable "ineternal_filed_to_be_used_by_secret_coven_of_mages_do_not_touch" {
```
mr0re1, line 42: typo `s/filed/field/`
also name can be les pompous, e.g. `internal_ghpc_module_id`

modules/compute/gke-node-pool/main.tf
```diff
@@ -33,6 +33,8 @@ locals {
   autoscale_set    = var.autoscaling_total_min_nodes != 0 || var.autoscaling_total_max_nodes != 1000
   static_node_set  = var.static_node_count != null
   initial_node_set = try(var.initial_node_count > 0, false)
+
+  module_unique_id = replace(lower(var.ineternal_filed_to_be_used_by_secret_coven_of_mages_do_not_touch), "/[^a-z0-9]/", "")
```
mr0re1, line 37: seems to be too strict, aren't `-` also allowed?

modules/compute/gke-node-pool/main.tf
```diff
@@ -42,7 +44,7 @@ data "google_compute_default_service_account" "default_sa" {
 resource "google_container_node_pool" "node_pool" {
   provider = google-beta
 
-  name           = var.name == null ? var.machine_type : var.name
+  name           = var.name == null ? "${var.machine_type}-${local.module_unique_id}" : var.name
```
mr0re1, line 47: Try `coalesce` function instead.

modules/compute/gke-node-pool/variables.tf
sharabiani: done

modules/compute/gke-node-pool/main.tf
sharabiani: done

modules/compute/gke-node-pool/main.tf
sharabiani: done

ankitkinra (approved):

modules/compute/gke-node-pool/variables.tf
```diff
@@ -31,11 +31,19 @@ variable "zones" {
...
+  description = <<-EOD
+    The name of the node pool. If not set, automatically populated by machine type and module id (unique blueprint-wide) as suffix.
+    If setting manually, ensure a unique value across all gke-node-pools.
+    EOD
   type        = string
   default     = null
 }
 
+variable "internal_ghpc_module_id" {
+  description = "Populates with module id (unique blueprint-wide)."
```
mr0re1, line 43: Please add a note that warns users from setting it.

sharabiani: done

PR #3128, opened 2024-10-11
Remove mention of `./[community/]modules` from docs and examples

mr0re1:
* Remove mention of `./[community/]modules` from docs and examples;
* Added line about not using it with toolkit modules;
* Clean up outdated mentions of "role";
* Update recommendations for `startup-script.source` to use `ghpc_stage`.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1730,10 +1730,9 @@ vars:
 deployment_groups:
 - group: groupName
   modules:
-
-  # Local source, prefixed with ./ (/ and ../ also accepted)
+  # Embedded module (part of the toolkit), prefixed with `modules/` or `community/modules`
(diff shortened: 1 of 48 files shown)


examples/README.md
```diff
@@ -1730,10 +1730,9 @@ vars:
 deployment_groups:
 - group: groupName
   modules:
-
-  # Local source, prefixed with ./ (/ and ../ also accepted)
+  # Embedded module (part of the toolkit), prefixed with `modules/` or `community/modules`
```
tpdownes, line 1733: I suggest capitalizing "Toolkit"

tpdownes (approved):
This PR does not change functionality of Toolkit but steers users away from using the `./` notation for our internal modules. This only works when invoking `gcluster` from the root of the Toolkit working copy of the repository anyhow.

PR #3136, opened 2024-10-16
Include deployment name in Spack and Ramble bucket names (like startup-script)

rohitramu:
This will make it easier to know which buckets are tied to a deployment.

### DIFF
--- a/community/modules/scripts/ramble-setup/main.tf
+++ b/community/modules/scripts/ramble-setup/main.tf
@@ -73,7 +73,7 @@ locals {
   }
 
   bucket_md5  = substr(md5("${var.project_id}.${var.deployment_name}"), 0, 8)
-  bucket_name = "ramble-scripts-${local.bucket_md5}"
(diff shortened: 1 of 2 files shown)


community/modules/scripts/ramble-setup/main.tf
nick-stroud, line 1: Can you trim the deployment name to make sure length of bucket name does not exceed limit. I think it is 63 chars.

nick-stroud, line 1: Please run spack tests with green before merging.

nick-stroud (approved):

community/modules/scripts/ramble-setup/main.tf
rohitramu: Good point - done.

Doc for context: https://cloud.google.com/storage/docs/buckets#naming

rohitramu: The `spack-gromacs` test has now succeeded.

PR #3144, opened 2024-10-17
Allow customization of Parallelstore mounts

wiktorn:
Add two customization options:
* daos_agent_config - to allow providing additional configuration in `/etc/daos/daos_agent.yml`
* dfuse_environment - to allow providing environment variables for `dfuse` process

## Topics for discussion:
* [x] whether using templates is proper approach here (it gives a lot of flexibility)
* [x] support for whitespaces in mount points (currently not supported, not sure if it was supported earlier)
* [x] handling of `/etc/daos/daos_agent.yml` - currently anyway everything is commented, and we just uncomment specific lines with `sed`. The other approach is either to create a new file (as implemnted), or just make sure that everything is commented, and just add configuration at the end. I think this aproach is cleaner in terms of what is expected result / file content.
* [x] usage of systemctl to start dfuse on first call. I did not implemented as such for now, as I didn't find a good way to fail `systemctl start` command if dfuse failed to start and this is current behavior
* [x] Adding `Restart=always` to systemd unit. With my testing, one needs to run first `umount $mount_point`, but if this indeed is necessary, this could be wrapped in script together with `$mount_command`, and this would make sure, that even if `dfuse` crashes, it will restart

## TODO
* [x] align those changes with `modules/file-system/pre-existing-network-storage`

This PR also addresses bug in startup-script module,
```
startup-script: Mon Dec 16 08:44:43 -0500 2024 Info [4589]: === start executing runner: early_run_hotfixes.sh-b383 ===
startup-script: /slurm/custom_scripts/controller.d/ghpc_daos_mount.sh: line 373: .//tmp/tmp.g1yoDyA6T5/early_run_hotfixes.sh: No such file or directory
startup-script: Mon Dec 16 08:44:43 -0500 2024 Info [4589]: === early_run_hotfixes.sh-b383 finished with exit_code=127 ===
startup-script: Mon Dec 16 08:44:43 -0500 2024 Error [4589]: === execution of early_run_hotfixes.sh-b383 failed, exiting ===


If re-running the script, using for example: DEBUG=1 google_metadata_script_runner startup
```

### DIFF
--- a/modules/file-system/parallelstore/templates/mount-daos.sh.tftpl
+++ b/modules/file-system/parallelstore/templates/mount-daos.sh.tftpl
@@ -20,59 +20,48 @@ OS_VERSION=$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"/
 OS_VERSION_MAJOR=$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"//g' -e 's/\..*$//')
 
 if ! {
-	{ [[ "${OS_ID}" = "rocky" ]] || [[ "${OS_ID}" = "rhel" ]]; } && { [[ "${OS_VERSION_MAJOR}" = "8" ]] || [[ "${OS_VERSION_MAJOR}" = "9" ]]; } ||
-		{ [[ "${OS_ID}" = "ubuntu" ]] && [[ "${OS_VERSION}" = "22.04" ]]; } ||
-		{ [[ "${OS_ID}" = "debian" ]] && [[ "${OS_VERSION_MAJOR}" = "12" ]]; }
+	{ [[ "$${OS_ID}" = "rocky" ]] || [[ "$${OS_ID}" = "rhel" ]]; } && { [[ "$${OS_VERSION_MAJOR}" = "8" ]] || [[ "$${OS_VERSION_MAJOR}" = "9" ]]; } ||
+		{ [[ "$${OS_ID}" = "ubuntu" ]] && [[ "$${OS_VERSION}" = "22.04" ]]; } ||
+		{ [[ "$${OS_ID}" = "debian" ]] && [[ "$${OS_VERSION_MAJOR}" = "12" ]]; }
 }; then
-	echo "Unsupported operating system ${OS_ID} ${OS_VERSION}. This script only supports Rocky Linux 8, Redhat 8, Redhat 9, Ubuntu 22.04, and Debian 12."
+	echo "Unsupported operating system $${OS_ID} $${OS_VERSION}. This script only supports Rocky Linux 8, Redhat 8, Redhat 9, Ubuntu 22.04, and Debian 12."
 	exit 1
 
 fi
 
-# Parse local_mount, mount_options from argument.
-# Format mount-options string to be compatible to dfuse mount command.
-# e.g. "disable-wb-cache,eq-count=8" --> --disable-wb-cache --eq-count=8.
-for arg in "$@"; do
-	if [[ $arg == --access_points=* ]]; then
-		access_points="${arg#*=}"
-	fi
-	if [[ $arg == --local_mount=* ]]; then
-		local_mount="${arg#*=}"
-	fi
-	if [[ $arg == --mount_options=* ]]; then
-		mount_options="${arg#*=}"
-		mount_options="--${mount_options//,/ --}"
-	fi
-done
-
 # Edit agent config
 daos_config=/etc/daos/daos_agent.yml
-sed -i "s/#.*transport_config/transport_config/g" $daos_config
-sed -i "s/#.*allow_insecure:.*false/  allow_insecure: true/g" $daos_config
-sed -i "s/.*access_points.*/access_points: $access_points/g" $daos_config
 
+# rewrite $daos_config from scratch
+mv $${daos_config} $${daos_config}.orig
+
+exclude_fabric_ifaces=""
 # Get names of network interfaces not in first PCI slot
 # The first PCI slot is a standard network adapter while remaining interfaces
 # are typically network cards dedicated to GPU or workload communication
-if [[ "$OS_ID" == "debian" ]] || [[ "${OS_ID}" = "ubuntu" ]]; then
+if [[ "$${OS_ID}" == "debian" ]] || [[ "$${OS_ID}" = "ubuntu" ]]; then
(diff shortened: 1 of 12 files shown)


modules/file-system/parallelstore/variables.tf
```diff
@@ -24,6 +24,20 @@ variable "deployment_name" {
   type        = string
 }
 
+variable "daos_agent_config" {
```
harshthakkar01, line 27: I am not sure if adding new variables would be a better choice. Since the underlying thought process behind toolkit is, how you can use same interface for all file systems, like I just need to change 1 line to deploy parallelstore from filestore in blueprint.

mr0re1: Frankly, I see no problem here. Added variables are parallelstore specific and go into parallelstore module. @harshthakkar01 what would be a better alternative?

harshthakkar01: Had an offline discussion and we agreed on moving on with this approach.

harshthakkar01:
> Adding Restart=always to systemd unit. With my testing, one needs to run first umount $mount_point, but if this indeed is necessary, this could be wrapped in script together with $mount_command, and this would make sure, that even if dfuse crashes, it will restart

In this case, if you `enable` systemd unit, it will run for every restart (this is the current behavior, wdyt ?).  For the case, you mentioned, I also observed you need to umount it first.

harshthakkar01:
>> usage of systemctl to start dfuse on first call. I did not implemented as such for now, as I didn't find a good way to fail systemctl start command if dfuse failed to start and this is current behavior

Current behavior we dont use systemd unit to mount the instance. We manually mount it using bash and `enable` the service to be run for every restart.

tpdownes:
> > Adding Restart=always to systemd unit. With my testing, one needs to run first umount $mount_point, but if this indeed is necessary, this could be wrapped in script together with $mount_command, and this would make sure, that even if dfuse crashes, it will restart
> 
> In this case, if you `enable` systemd unit, it will run for every restart (this is the current behavior, wdyt ?). For the case, you mentioned, I also observed you need to umount it first.

I strongly recommend `Restart=on-failure` for almost any SystemD unit unless `always` provides some benefit you can clearly articulate. I believe `always` isn't allowed for oneshots.

wiktorn:
> > > usage of systemctl to start dfuse on first call. I did not implemented as such for now, as I didn't find a good way to fail systemctl start command if dfuse failed to start and this is current behavior
> 
> Current behavior we dont use systemd unit to mount the instance. We manually mount it using bash and `enable` the service to be run for every restart.

Using systemd always would simplify the code, as there would only one way to mount the filesystem, whether this is the first startup of the instance, or the next one. The only drawback is that I did not find a way, to fail `systemctl enable --now` (or similar command) when service did not start successfully.

I can add a check after the call, with some sleeps to check if either service started successfully or the mountpoint is mounted and fail the startup script if those checks fail. WDYT?

harshthakkar01:
You may need to rebase this after https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3256 (this one was high priorty). I have discussed with Ivan and he will follow up with on this as I am ooo for 3 weeks from today.

wiktorn:
> You may need to rebase this after #3256 (this one was high priorty). I have discussed with Ivan and he will follow up with on this as I am ooo for 3 weeks from today.
> 
> Btw, we agreed on moving forward with adding customization for mounts. Thanks for this PR.

Thanks for letting me know. For now, I rebased this PR on top of #3256, once it is in, I can skip a few commits from here.

I aligned now `pre-existing-network-storage` with `parallelstore` in terms of scripts and APIs and fixed some smaller issues I faced (like installation failing on RHEL 8, but started to work well after `dnf clean all` :shrug: )

modules/scripts/startup-script/templates/startup-script-custom.tftpl
```diff
@@ -31,7 +31,7 @@ stdlib::runner() {
   stdlib::info "=== start executing runner: $object ==="
   case "$1" in
     ansible-local) stdlib::run_playbook "$destpath/$filename" "$args";;
-    shell) chmod u+x /$destpath/$filename && ./$destpath/$filename $args;;
+    shell) chmod u+x /$destpath/$filename && $destpath/$filename $args;;
```
wiktorn, line 34: This change is to fix failures like this:
```
startup-script: Mon Dec 16 08:44:43 -0500 2024 Info [4589]: === start executing runner: early_run_hotfixes.sh-b383 ===
startup-script: /slurm/custom_scripts/controller.d/ghpc_daos_mount.sh: line 373: .//tmp/tmp.g1yoDyA6T5/early_run_hotfixes.sh: No such file or directory
startup-script: Mon Dec 16 08:44:43 -0500 2024 Info [4589]: === early_run_hotfixes.sh-b383 finished with exit_code=127 ===
startup-script: Mon Dec 16 08:44:43 -0500 2024 Error [4589]: === execution of early_run_hotfixes.sh-b383 failed, exiting ===
```

harshthakkar01:
Please resolve conflicts

modules/file-system/parallelstore/templates/mount-daos.sh.tftpl
```diff
@@ -20,59 +20,55 @@ OS_VERSION=$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"/
...
+  allow_insecure: true
+$exclude_fabric_ifaces
+${daos_agent_config}
+
+EOF
 
 # Get names of network interfaces not in first PCI slot
 # The first PCI slot is a standard network adapter while remaining interfaces
 # are typically network cards dedicated to GPU or workload communication
 if [[ "$OS_ID" == "debian" ]]; then
```
harshthakkar01, line 60: Will this be needed now ?

wiktorn: good catch, this is in the oven for too long :-)

harshthakkar01: Sorry about that. Its mainly from my side.

modules/scripts/startup-script/templates/startup-script-custom.tftpl
harshthakkar01: Ack, thanks for pointing this out and fixing it.

harshthakkar01 (approved):

PR #3145, opened 2024-10-17
Add logging output during Packer failures

cdunbar13:
This PR adds an output to the Packer module that prints a custom gcloud logging command of the format

`gcloud logging --project <project> read 'logName=("projects/<project>/logs/GCEMetadataScripts" OR "projects/<project>/logs/google_metadata_script_runner") AND resource.labels.instance_id=<instance #>' --format="table(timestamp, resource.labels.instance_id, jsonPayload.message)" --order=asc`

when Packer fails during the startup scripts.  It also updates packer to create a instance name variable for the instance so that can be more easily tracked in the module.  It varies from the command in the README in that it should not need the freshness option as it is specific to one instance.  It also does not look for `"^startup-script: "` as that can be a bit too specific of a query.

This PR also removes some redundant image name outputs, and updates the golden copies.

This was tested on a simple Packer blueprint that introduced an script that only ran `exit 1` and cause the Packer module to end with an error.  It was also tested with a clean build to confirm no new issues were introduced.


tpdownes (changes requested):
Please review my concerns about the shell of the cleanup provisioner. I'd rather have seen these as 3 separate, simple commits. If that's easy to do at this point, please do.

PR #3150, opened 2024-10-21
Initial commit for new logging output

cdunbar13:
This PR adds an output to the Packer module that prints a custom gcloud logging command of the format

gcloud logging --project <project> read 'logName=("projects/<project>/logs/GCEMetadataScripts" OR "projects/<project>/logs/google_metadata_script_runner") AND resource.labels.instance_id=<instance #>' --format="table(timestamp, resource.labels.instance_id, jsonPayload.message)" --order=asc

when Packer fails during the startup scripts. It also updates packer to create a instance name variable for the instance so that can be more easily tracked in the module. It varies from the command in the README in that it should not need the freshness option as it is specific to one instance. It also does not look for "^startup-script: " as that can be a bit too specific of a query.

This PR also updates the golden copies.

This was tested on a simple Packer blueprint that introduced an script that only ran exit 1 and cause the Packer module to end with an error. It was also tested with a clean build to confirm no new issues were introduced.

### DIFF
--- a/modules/packer/custom-image/image.pkr.hcl
+++ b/modules/packer/custom-image/image.pkr.hcl
@@ -21,6 +21,9 @@ locals {
   image_name_default = "${local.image_family}-${formatdate("YYYYMMDD't'hhmmss'z'", timestamp())}"
   image_name         = var.image_name != null ? var.image_name : local.image_name_default
 
+  # construct vm image name for use when getting logs
+  instance_name = "packer-${substr(uuidv4(), 0, 6)}"
+
   # default to explicit var.communicator, otherwise in-order: ssh/winrm/none
   shell_script_communicator      = length(var.shell_scripts) > 0 ? "ssh" : ""
   ansible_playbook_communicator  = length(var.ansible_playbooks) > 0 ? "ssh" : ""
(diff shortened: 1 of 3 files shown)


modules/packer/custom-image/image.pkr.hcl
```diff
@@ -197,4 +201,25 @@ build {
...
+
+  # If there is an error during image creation, print out command for getting packer VM logs
+  error-cleanup-provisioner "shell-local" {
+    environment_vars = [
+      "PRJ_ID=${var.project_id}",
+      "INST_NAME=${local.instance_name}",
+      "ZONE=${var.zone}",
+    ]
+    inline = [
+      "#!/bin/bash -e",
```
tpdownes, line 213: I think you want to use `execute_command` parametmer and `set -e` inside the script. The `provisioner` block above is an example (I believe  error-cleanup-provisioner implements all the parameters)

https://developer.hashicorp.com/packer/docs/provisioners/shell-local#execute_command

tpdownes (changes requested):

modules/packer/custom-image/image.pkr.hcl
cdunbar13: The execute command is specifically for scripts that are linked, I believe.  I've added the `inline_shebang` to specify the shell command.

tpdownes (approved):

PR #3157, opened 2024-10-23
SlurmGCP. "All or nothing" bulk insert on requests with placements

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -134,8 +130,14 @@ def create_instances_request(nodes, partition_name, placement_group, job_id=None
     log.debug(f"create_instances_request: {model} placement: {placement_group}")
 
     body = NSDict()
+
     body.count = len(nodes)
-    body.minCount = 1
+
+    if placement_group:
```
cdunbar13, line 136: Isn't it simpler to just use
```
# Only set minCount if we need placement.
if not placement_group:
     body.minCount
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -120,11 +120,7 @@ def per_instance_properties(node):
 
 def create_instances_request(nodes, partition_name, placement_group, job_id=None):
     """Call regionInstances.bulkInsert to create instances"""
-    assert len(nodes) > 0
-    if placement_group:
-        assert len(nodes) <= min(PLACEMENT_MAX_CNT, BULK_INSERT_LIMIT)
-    else:
-        assert len(nodes) <= BULK_INSERT_LIMIT
+    assert 0 < len(nodes) <= BULK_INSERT_LIMIT
```
cdunbar13, line 123: Testing the second assert later on will end up with some weird debug issues and I'm not sure what the model line will end up doing.

```
assert 0 < len(nodes) <= placement_group ? min(PLACEMENT_MAX_CNT, BULK_INSERT_LIMIT) : BULK_INSERT_LIMIT
```

cdunbar13 (approved):
It probably is fine but I have some comments.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
mr0re1: I want to explicitly spell out "no op" branch.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
mr0re1: ack

PR #3167, opened 2024-10-25
Packer Logging README update

cdunbar13:
Updating Packer README to include new tip about a logging command that will be printed in the case that Packer fails.

### DIFF
--- a/modules/packer/custom-image/README.md
+++ b/modules/packer/custom-image/README.md
@@ -210,25 +210,14 @@ to the console. For example:
 ==> example.googlecompute.toolkit_image: Startup script, if any, has finished running.
 ```
 
-### Viewing Startup Script Logs
-
-The recommended method for debugging the image build process is to use Cloud
-Logging. This can be done by either searching for the VM instance in the Cloud
-Console or using the following template command with the variables `PROJECT_ID`
-(e.g. `test_project_1`) and `INSTANCE_ID` (note: unique numerical id, not
-instance name) specified:
-
-```shell
-gcloud logging --project <PROJECT_ID> read 'logName=("projects/<PROJECT_ID>/logs/GCEMetadataScripts" OR "projects/<PROJECT_ID>/logs/google_metadata_script_runner") AND resource.labels.instance_id=<INSTANCE_ID>' --format="table(timestamp, resource.labels.instance_id, jsonPayload.message) --order=asc
-```
+### Debugging startup-script failures
 
 > [!NOTE]
 > There can be a delay in the propagation of the logs from the instance to
 > Cloud Logging, so it may require waiting a few minutes to see the full logs.
 
-If the Packer image build fails, the module will output the command above
-with the variables specified and can be used in a terminal without
-modification.
+If the Packer image build fails, the module will output a `gcloud` command
+that can be used directly to review startup-script execution.
 
 ## License
 
--- a/modules/packer/custom-image/README.md
+++ b/modules/packer/custom-image/README.md
@@ -210,41 +210,25 @@ to the console. For example:
 ==> example.googlecompute.toolkit_image: Startup script, if any, has finished running.
 ```
 
-To monitor the progress of the startup script, first gather the name and zone
-of the instance using Cloud Console or the following command:
+### Viewing Startup Script Logs
+
+The recommended method for debugging the image build process is to use Cloud
+Logging. This can be done by either searching for the VM instance in the Cloud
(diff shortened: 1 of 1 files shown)


modules/packer/custom-image/README.md
```diff
@@ -210,24 +210,41 @@ to the console. For example:
...
-Linux utility `tac`.
+This will produce a list of VMs starting with `packer` along with other
+information including which zone the VMs are in.  If there is more than one
+Packer VM you will need to determine which is the one that you wish to monitor
+using criteria such as zone or machine-type.
+
+Once the VM name is determined, you can either check the serial port output in
+Cloud Console or by running the command:
+
+```shell
```
tpdownes, line 228: The serial port API is not a reliable method of collecting this data -- you note one reason (it goes away when the VM does) -- and I do not want us to recommend it.

modules/packer/custom-image/README.md
```diff
@@ -210,24 +210,41 @@ to the console. For example:
 ==> example.googlecompute.toolkit_image: Startup script, if any, has finished running.
 ```
 
-Using the default value for \[var.scopes\]\[#input_scopes\], the output of
-startup script execution will be stored in Cloud Logging. It can be examined
-using the [Cloud Logging Console][logging-console] or with a
-[gcloud logging read][logging-read-docs] command (substituting `<<PROJECT_ID>>`
-with your project ID):
+To monitor the progress of the startup script, first gather the name and zone
```
tpdownes, line 213: Please remove the text beginning at line 213 and and ending at line 234

modules/packer/custom-image/README.md
```diff
@@ -210,24 +210,41 @@ to the console. For example:
...
+Cloud Console or by running the command:
+
+```shell
+gcloud compute instances get-serial-port-output <VM_NAME> --port 1 --zone <ZONE> --project <PROJECT_ID>
+```
+
+The serial port output of the Packer VM will contain the startup script logs.
+This output will only be available while Packer is running.
+
+### Gathering Startup Script Logs After Failure
```
tpdownes, line 235: Please remove "after failure" and alter the text so that the logic is:

- we recommend Cloud Logging to review startup script progress
- note that there may be a delay in cloud logging entries appearing
- here's the pattern the command takes
- this command is automatically printed with correct values when startup-script execution fails

We could consider the logic of the last step (i.e. always print it vs. on-error). I am satisfied with on-error for now.

tpdownes (changes requested):

modules/packer/custom-image/README.md
cdunbar13: Removed.

modules/packer/custom-image/README.md
cdunbar13: Removed

modules/packer/custom-image/README.md
```diff
@@ -210,24 +210,25 @@ to the console. For example:
 ==> example.googlecompute.toolkit_image: Startup script, if any, has finished running.
 ```
 
-Using the default value for \[var.scopes\]\[#input_scopes\], the output of
-startup script execution will be stored in Cloud Logging. It can be examined
-using the [Cloud Logging Console][logging-console] or with a
-[gcloud logging read][logging-read-docs] command (substituting `<<PROJECT_ID>>`
-with your project ID):
+### Viewing Startup Script Logs
```
tpdownes, line 213: The focus of #3150 is improving the debugging experience. I don't think relying on the user to fill in the blanks and browse to the Cloud Console in a timely fashion helps that message. I want to see this become a simple section:

```md
## Debugging startup-script failures

> [!NOTE]
> There can be a delay in the propagation of the logs from the instance to
> Cloud Logging, so it may require waiting a few minutes to see the full logs.

If the Packer image build fails, the module will output a `gcloud` command
that can be used directly to review startup-script execution.
```

I suggest createinga follow-up task that uses a `shell-local` provisioner that prints out only a link to the Cloud Logging for live viewing.  For example:

```
https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%22562843389805978984%22?project=hpc-toolkit-dev
```

tpdownes (changes requested):

modules/packer/custom-image/README.md
cdunbar13: Done

tpdownes (approved):

PR #3175, opened 2024-10-28
Update MTU for a3 mega for GKE based on best practices 

ankitkinra:
Based on the https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-vpcs-subnets, where the VPCs are created 

```
gcloud compute networks create PROJECT_ID-net-$N \
    --subnet-mode=custom \
    --mtu=8244

```

### DIFF
--- a/examples/gke-a3-megagpu.yaml
+++ b/examples/gke-a3-megagpu.yaml
@@ -33,6 +33,7 @@ deployment_groups:
     source: modules/network/vpc
     settings:
       subnetwork_name: gke-subnet-a3-mega
+      mtu: 8244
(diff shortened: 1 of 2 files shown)


examples/gke-a3-megagpu.yaml
```diff
@@ -33,6 +33,7 @@ deployment_groups:
     source: modules/network/vpc
     settings:
       subnetwork_name: gke-subnet-a3-mega
+      mtu: 8244
```
chengcongdu, line 36: can you update the same for a3high

PR #3192, opened 2024-10-30
XPK blueprint updates to make it more compatible with the tool

ankitkinra:
# Minor fixes for xpk compatibility 

* Update the naming convention of the node pools
* Name the config map with deployment name

### DIFF
--- a/community/examples/xpk-gke-a3-megagpu.yaml
+++ b/community/examples/xpk-gke-a3-megagpu.yaml
@@ -33,6 +33,7 @@ deployment_groups:
     source: modules/network/vpc
     settings:
       subnetwork_name: xpk-gke-a3-megagpu-subnet
+      mtu: 8244
       secondary_ranges:
         xpk-gke-a3-megagpu-subnet:
         - range_name: pods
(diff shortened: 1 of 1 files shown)


community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -70,25 +72,40 @@ deployment_groups:
...
+      name: system-pool
+      machine_type: "e2-standard-16"
+      autoscaling_total_min_nodes: 2
+      autoscaling_total_max_nodes: 8
+      initial_node_count: 2
+      zones: [$(vars.zone)]
+      host_maintenance_interval: PERIODIC
+    outputs: [instructions]
+
+  - id: a3_megagpu_pool_np_0
```
pawloch00, line 88: why do we have two nodepools?

pawloch00:
Should TAS be left here? @ankitkinra

community/examples/xpk-gke-a3-megagpu.yaml
ankitkinra: The second modepool has no default taints and can handle system software which has not been modified by our team.

community/examples/xpk-gke-a3-megagpu.yaml
```diff
@@ -33,6 +33,7 @@ deployment_groups:
     source: modules/network/vpc
     settings:
       subnetwork_name: xpk-gke-a3-megagpu-subnet
+      mtu: 8244
```
annuay-google, line 36: Curious, any reason to set the MTU to 8244?

annuay-google (approved):
LGTM, comment is non blocking

community/examples/xpk-gke-a3-megagpu.yaml
ankitkinra: this is from the best practices from GKE from https://cloud.google.com/kubernetes-engine/docs/how-to/gpu-bandwidth-gpudirect-tcpx#create-vpcs-subnets

ankitkinra:
> Should TAS be left here? @ankitkinra

We can potentially update in a follow up PR

PR #3199, opened 2024-10-31
Fix SSH issues with integration tests

cdunbar13:
A number of the integration tests have been failing lately due to a host being "UNREACHABLE" via ssh.  This PR creates a wait-for-host task that checks both the ansible connection and that port 22 is accessible from the builder.  

The PR should also resolve at least one issue where the playbook would completely halt because of an unreachable host and leave behind resources to be cleaned up.

This is being tested on the chrome-remote-desktop integration test.  I was able to reproduce the error by removing the firewall rule creation task in the test.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -93,7 +93,7 @@
       ansible.builtin.debug:
         var: remote_ip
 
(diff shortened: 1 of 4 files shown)


rohitramu (approved):
I've triggered the following tests:
- `slurm-gcp-v6-rocky8`
- `slurm-gcp-v6-debian`
- `htcondor`

This PR LGTM, but please wait for these three to succeed before submitting.

PR #3200, opened 2024-10-31
Gke update default taints for user node pools

ankitkinra:
Removing the default user node pool taint of "user-workload" allows the user workloads jobs to land on the GPU node pools without adding tolerations to all user jobs.
Default taint was added to user node pools to prevent non GPU workloads from landing there.
However, GKE adds a `gpu` taint already to GPU nodepools https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#taint which makes these taints redundant and complicate user workflows.


modules/scheduler/gke-cluster/variables.tf
```diff
@@ -170,11 +170,7 @@ variable "system_node_pool_taints" {
     value  = any
     effect = string
   }))
-  default = [{
```
chengcongdu, line 173: are we removing the system np taint as well? I thought we were only discussing removing the manual gpu taint

ankitkinra: Sounds good, keeping this for now.

chengcongdu (approved):

PR #3209, opened 2024-11-01
SlurmGCP. Escape GCP error reasons that may cause malformed CLI args

mr0re1:
Bug observed in a wild, but no exact value of reason has been recovered. Here is a reproduction:

```py
reason = "rock'n roll"
args = f"scontrol update nodename=nodelist state=down reason='{reason}'"
shlex.split(args)

File /usr/lib/python3.11/shlex.py:191, in shlex.read_token(self)
    189         print("shlex: I see EOF in quotes state")
    190     # XXX what error should be raised here?
--> 191     raise ValueError("No closing quotation")
    192 if nextchar == self.state:
    193     if not self.posix:

ValueError: No closing quotation
```


harshthakkar01:
This will affect since bash escapes certain special characters right ? 

Like $, \ ? Will this affect other places as well ?

mr0re1:
> This will affect since bash escapes certain special characters right ?
> Like $, \ ? Will this affect other places as well ?

Can you come up with example of reasonable reason (lol) ?
IMHO nothing should be interpolated here, so safe escaping should be a no-brainer. WDYT?

harshthakkar01:
This was just for my understanding, like if this is to escape bash character. Happy to lgtm.

harshthakkar01 (approved):

PR #3215, opened 2024-11-04
SlurmGCP. Add `set -e` to prolog mux

mr0re1:
**Motivation:** To allow fail-scenario


tpdownes (approved):
We have tested this on an internal cluster with the standard prolog/epilog solutions for the A3 VM family. We have not observed any unanticipated failures that cause undesired drains of nodes.

We will need to document this change for end users in our next release notes, perhaps with a link to a Discussion. This can be handled at release

PR #3219, opened 2024-11-05
Adding retries to a3mega image blueprint

cdunbar13:
The A3 Mega image creation blueprint has a transient failure on downloading the kernel headers within the Packer module.  This adds retries to the task and a delay of a minute to alleviate that issue.

Testing using the PR-test-ml-a3-megagpu-slurm integration test.

### DIFF
--- a/examples/machine-learning/a3-megagpu-8g/slurm-a3mega-image.yaml
+++ b/examples/machine-learning/a3-megagpu-8g/slurm-a3mega-image.yaml
@@ -83,6 +83,11 @@ deployment_groups:
               ansible.builtin.get_url:
                 url: "{{ package_url }}"
                 dest: "{{ package_filename }}"
+              retries: 3
+              delay: 60
+              register: result
+              until: result is success
+              failed_when: result is failure
             - name: Install kernel headers
               ansible.builtin.apt:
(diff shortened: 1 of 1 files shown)


examples/machine-learning/a3-megagpu-8g/slurm-a3mega-image.yaml
```diff
@@ -83,6 +83,11 @@ deployment_groups:
...
                 url: "{{ package_url }}"
                 dest: "{{ package_filename }}"
+              retries: 3
+              delay: 60
+              register: result
+              until: result is success
+              failed_when: result is failure
             - name: Install kernel headers
               ansible.builtin.apt:
                 deb: "{{ package_filename }}"
```
tpdownes, line 93: nit: I think it's safer to use a less generic name than `result` because (I believe) that they are not scoped within the task. Will be issue if we do something similar.

tpdownes (approved):

examples/machine-learning/a3-megagpu-8g/slurm-a3mega-image.yaml
cdunbar13: Fixed, changed to `knl_dl`

PR #3224, opened 2024-11-06
Cleanup Always when SSH Fails

arajmane-g:
### About the Change
Ansible has a way to ignore unreachable host errors and let the play continue: [ignore_unreachable: true](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#ignoring-unreachable-host-errors). 

Slurm integration test is relying on the same: code [ref](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/b857af00ff4d5210bc49fb781264ebb1f671cb19/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml#L191)

Since, the ml-gke-e2e test uses the [base-integration-test.yml](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml) playbook, this change is required therein. It will fix all tests that use this playbook ([github code search](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fcluster-toolkit+%22base-integration-test.yml%22+path%3Atools%2Fcloud-build%2Fdaily-tests&type=code))

### Testing
I am not sure how to test this. So, thinking if it is fine to merge the fix and see if this recurs in our daily tests.

### Submission Checklist

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -183,7 +183,6 @@
 
     - name: Run Integration tests for Cluster Toolkit
       ansible.builtin.include_tasks: "{{ test }}"
-      ignore_unreachable: false # end the play when the host is unreachable
       vars:
         remote_node: "{{ remote_node }}"
         deployment_name: "{{ deployment_name }}"
--- a/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml
@@ -169,6 +169,7 @@
   vars:
     startup_timeout_seconds: 600  # 10 minutes
   gather_facts: false
+  ignore_unreachable: true  # ensure always block will run even if SSH fails
   tasks:
   - name: Remote Test Block
     vars:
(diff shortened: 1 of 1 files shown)


tpdownes (changes requested):
This PR causes the build to fail with the error

```
ERROR! 'ignore_unreachable' is not a valid attribute for a TaskInclude
```

~I believe you've used the wrong name for the setting.~

It looks like you L186 is not valid. I do think you may have identified potentially ignored failures. For the moment, I prefer that you remove L186 and document the potential gap as a task. The first line change gets us to the point where we are cleaning up after SSH timeouts.

arajmane-g:
@tpdownes 

> It looks like you L186 is not valid.

Looks like it. This line was inspired by [a snippet from the documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#ignoring-unreachable-host-errors:~:text=And%20at%20the%20playbook%20level%3A) whic specifies the flag in a task. Whereas, we are trying to specify it with `ansible.builtin.include_tasks`. So, makes sense.

> I do think you may have identified potentially ignored failures. For the moment, ..., document the potential gap as a task.

My concern was since SSH failed, the subsequent tasks in the play would fail. So, wanted to end the play explicitly. But, if that's not a valid attribute for `ansible.builtin.include_tasks`, it's fine, I think. 

> The first line change gets us to the point where we are cleaning up after SSH timeouts.

👍  As such the [Slurm test play](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/b857af00ff4d5210bc49fb781264ebb1f671cb19/tools/cloud-build/daily-tests/ansible_playbooks/slurm-integration-test.yml#L191) does only that much

tpdownes (approved):

PR #3227, opened 2024-11-06
Add future reservation support

abbas1902:
This PR allows users to add their future reservations into nodesets. Node state is handled by slurmsync and redeployments should no longer be necessary for pre/post reservation nodes.
### Submission Checklist


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1765,21 +1777,28 @@ def _get_reservation(self, project: str, zone: str, name: str) -> object:
...
         return self.compute.reservations().get(
             project=project, zone=zone, reservation=name).execute()
+    
+    @lru_cache()
+    def _get_future_reservation(self, project:str, zone:str, name: str) -> object:
+        """See https://cloud.google.com/compute/docs/reference/rest/v1/futureReservations"""
+        return self.compute.futureReservations().get(project=project, zone=zone, futureReservation=name).execute()
+    
+    
+    def nodeset_reservation(self, reservation_name:str, zone_policy_allow: Optional[list]) -> Optional[ReservationDetails]:
```
mr0re1, line 1787: There is nothing left about `nodeset` in this function. I would recommend to keep it as it was:
```py
def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
```

And add separate internal `def _get_reservation(name: str, zone)`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1765,21 +1777,28 @@ def _get_reservation(self, project: str, zone: str, name: str) -> object:
         """See https://cloud.google.com/compute/docs/reference/rest/v1/reservations"""
         return self.compute.reservations().get(
             project=project, zone=zone, reservation=name).execute()
+    
+    @lru_cache()
+    def _get_future_reservation(self, project:str, zone:str, name: str) -> object:
+        """See https://cloud.google.com/compute/docs/reference/rest/v1/futureReservations"""
+        return self.compute.futureReservations().get(project=project, zone=zone, futureReservation=name).execute()
```
mr0re1, line 1784: What happens with FutureReservation that reached its "end of life"?
Will it throw reasonable expectation if reservation doesn't exist? shall we handle it specifically, e.g. return null?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
             zone=zone,
             name=name,
             policies=policies,
-            bulk_insert_name=nodeset.reservation_name)
+            bulk_insert_name=reservation_name)
+    
+    def future_reservation(self, futureReservation:str, zone_policy_allow: Optional[list]) -> Optional[FutureReservation]:
```
mr0re1, line 1816: Under what condition it will return `None`, if never, why `Optional`?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -54,4 +54,19 @@ output "nodeset" {
...
     error_message = "Cannot use DWS Flex with `enable_placement`."
   }
+
+  precondition {
+    condition     = var.reservation_name == "" || var.future_reservation == null
+    error_message = "Cannot use reservations and future reservations in the same nodeset"
+  }
+
+  precondition {
+    condition     = var.node_count_dynamic_max == 0 || var.future_reservation == null
```
mr0re1, line 64: Why add this constraint?

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -463,6 +463,15 @@ variable "reservation_name" {
   }
 }
 
+variable "future_reservation" {
+  description = <<-EOD
+  Allows for the use of future reservations. Input can either be the future reservation name or full selfLink.
```
mr0re1, line 468: Please describe expected format of "full selfLink"

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -15,18 +15,16 @@
...
 import logging
 import re
 import sys
+from datetime import datetime, timedelta
 from enum import Enum
 from itertools import chain
 from pathlib import Path
 import yaml
-import datetime as dt
-from datetime import datetime
```
mr0re1, line 29: thank you.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -99,9 +97,8 @@ def _find_dynamic_node_status() -> NodeStatus:
     return NodeStatus.unchanged  # don't touch dynamic nodes
 
 
-def _find_tpu_node_status(nodename, state):
```
mr0re1, line 102: Why change this interface?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -404,6 +406,17 @@ def sync_slurm():
     for status, nodes in node_statuses.items():
         do_node_update(status, nodes)
 
+def future_reservation_sync(nodeset:object, nodename:str, state:object)-> None:
+    if future_reservation := lookup().future_reservation(nodeset.future_reservation, nodeset.zone_policy_allow):
```
mr0re1, line 410: nit. `s/future_reservation/fr/`. Long variable names don't improve readability.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -404,6 +406,17 @@ def sync_slurm():
...
+        currentTime = datetime.utcnow()
```
mr0re1, line 411: Don't increase nesting.

Instead of 

```py
def fn();
  if happy_path:
    # 50 lines of code
```

Do 

```
def fn();
  if not happy_path: 
    return
  # 50 lines of code
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+        activeReservation = None
+        endTime = None
+        startTime = None
+        
+        regex = re.compile(r'/futureReservations/(?P<name>[^/]+)(/.*)?$')
+
+        if match := regex.search(futureReservation):
+            name = match.group("name")
+
+        if reservation := self._get_future_reservation(project, zone, name):
```
mr0re1, line 1834: Under what condition this `if` doesn't trigger? Shall we return malformed `FutureReservation` in this case?

mr0re1, line 1834: nit. `s/reservation/fr/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+            startString = reservation.get("timeWindow").get("startTime")
+            endString = reservation.get("timeWindow").get("endTime")
```
mr0re1, line 1836: `fr.get("timeWindow").get("startTime")` - this is bad pattern.

 If reservation doesn't have `timeWindow` this line will fail with "None object doesn't have method get".
 
 While `fr["timeWindow"]["startTime"]` would fail with "fr doesn't have timeWindow" this is better.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+            startTime = datetime.fromisoformat(startString[:len(startString)-1])
+            endTime = datetime.fromisoformat(endString[:len(endString)-1])
```
mr0re1, line 1838: * Why trim, what at the end?
* There is a better way to trim last character
```py
>>> "hello"[:-1]
'hell'
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
             policies=policies,
-            bulk_insert_name=nodeset.reservation_name)
+            bulk_insert_name=reservation_name)
+    
+    def future_reservation(self, futureReservation:str, zone_policy_allow: Optional[list]) -> Optional[FutureReservation]:
+        zones = list(zone_policy_allow or [])
+        assert len(zones) == 1, "Only single zone is supported if using a reservation"
+        zone = zones[0]
+        
+        project = self.project
```
mr0re1, line 1821: `class FutureReservation` is not declared as frozen. 
Either declare it as `dataclass(frozen=True)`
Or use it as an accumulator of fields instead of declaring all those fields as individual variables.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1792,7 +1811,49 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+        name = futureReservation
+        nodesAreActive = False
+        specificReservationRequired = None
+        activeReservation = None
+        endTime = None
+        startTime = None
+        
+        regex = re.compile(r'/futureReservations/(?P<name>[^/]+)(/.*)?$')
+
+        if match := regex.search(futureReservation):
```
mr0re1, line 1831: `regex` used once, compiling it doesn't help to performance, you can save some lines:
```py
if match := re.search(r'/futureReservations/(?P<name>[^/]+)(/.*)?$', futureReservation)
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1461,6 +1461,18 @@ class ReservationDetails:
     policies: List[str] # names (not URLs) of resource policies
     bulk_insert_name: str # name in format suitable for bulk insert (currently identical to user supplied name in long format)
 
+@dataclass
+class FutureReservation:
+    project: str
+    zone: str
+    name: str
+    nodesAreActive: bool
```
mr0re1, line 1469: please use snake_case instead of camelCase

mr0re1, line 1469: `s/nodesAreActive/active/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
mr0re1, line 411: `s/currentTime/now/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -404,6 +406,17 @@ def sync_slurm():
...
+        if future_reservation.nodesAreActive is False and future_reservation.startTime>currentTime and state.base != "DOWN":
+            run(f"{lookup().scontrol} update nodename={nodename} state=down reason='Waiting for reservation to start at {future_reservation.startTime}'")
+
+        if future_reservation.nodesAreActive is False and future_reservation.endTime<currentTime and state.base != "DOWN":
+            run(f"{lookup().scontrol} update nodename={nodename} state=down reason='Reservation is over'")
+
+        if future_reservation.nodesAreActive and state.base == "DOWN":
+            run(f"{lookup().scontrol} update nodename={nodename} state=idle")
```
mr0re1, line 419: Transition from `down -> idle` should be done with `state=resume`. This is tribal knowledge, and I don't have good answer "why" ) We can dig into it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -404,6 +406,17 @@ def sync_slurm():
     for status, nodes in node_statuses.items():
         do_node_update(status, nodes)
 
+def future_reservation_sync(nodeset:object, nodename:str, state:object)-> None:
+    if future_reservation := lookup().future_reservation(nodeset.future_reservation, nodeset.zone_policy_allow):
+        currentTime = datetime.utcnow()
+        if future_reservation.nodesAreActive is False and future_reservation.startTime>currentTime and state.base != "DOWN":
+            run(f"{lookup().scontrol} update nodename={nodename} state=down reason='Waiting for reservation to start at {future_reservation.startTime}'")
+
+        if future_reservation.nodesAreActive is False and future_reservation.endTime<currentTime and state.base != "DOWN":
```
mr0re1, line 415: `s/ future_reservation.nodesAreActive is False/ not fr.active/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: Sounds good, just used the existing `_get_reservation` and created the reservation instance in line

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
abbas1902: Discussed offline, removing constraint

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
abbas1902: The callee now has access to each nodeset, before this function would find the nodeset but now there is no need since it can just be passed into this function directly

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: The trim is because the timestamp ends with a Z (to indicate UTC) but datetime doesn't handle the Z portion (I think in their python 3.12 package they add this feature) so I just trim it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: declared it as frozen

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
abbas1902: It appears to work with idle but it also appears to have similar behaviour to `state=resume`.

->Will usage of state=resume cause nodes to be provisioned (go through resume.py) (unwanted)
In the sense that `srun` causes resume script to run? If so yes (for both idle and resume) but isn't this what we'd like? Or are you asking if `state=idle` automatically triggers `resume.py` which is not the case (nor for `state=resume`). Also wondering on your thoughts about `state=down` vs `state=power_down` considering last week's discussion. Functionality wise it seems like the use of `state=power_down` always calls `suspend.py`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: The future reservation seems to last quite a long time even after fulfillment (over 4 months), if the reservation doesn't exist it spits out a 400 error, not sure if that's something we even want to handle because user in that scenario would probably want to redeploy anyway, I'd say no.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: Related to the previous conversation but doesn't seem like a real fr will return `None` so removing Optional.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: Doesn't appear that this if state will be false for a real fr, if fr is malformed/doesn't exist then we just let it error out is my view for redeployment.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -463,6 +463,15 @@ variable "reservation_name" {
   }
 }
 
+variable "future_reservation" {
+  description = <<-EOD
+  Allows for the use of future reservations. Input can be either the future reservation name or its selfLink in the format '[.../projects/PROJECT_ID/zones/ZONE/futureReservations/FUTURE_RESERVATION_NAME]'
```
mr0re1, line 468: The format definition is not clear to me. 
What `...` stands for?
`[]` usually (shell args lingo) means optional.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -104,9 +91,35 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
...
+            f"reservation {reservation.bulk_insert_name} is being used with policies {props.resourcePolicies}"
+        )
+    else:
+        props.resourcePolicies = []
+        log.info(
+            f"reservation {reservation.bulk_insert_name} is being used without any policies"
+        )
+
+def use_future_reservation(props:object, nodeset:object) -> None:
+    if (fr := lookup().future_reservation(nodeset)) and fr.active and fr.specific:
```
mr0re1, line 120: If feature reservation is not "active" we shouldn't even attempt to provision VM. Your code just silently omits `reservationAffinity`

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
abbas1902: wrote it to mean you can optionally include the `https://google/compute/beta...` (however it goes), but considering we use regex anyway I'll just remove this portion

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: Moving this function in-line in callee

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -54,4 +54,14 @@ output "nodeset" {
     condition     = !var.enable_placement || !var.dws_flex.enabled
     error_message = "Cannot use DWS Flex with `enable_placement`."
   }
+
+  precondition {
+    condition     = var.reservation_name == "" || var.future_reservation == null
+    error_message = "Cannot use reservations and future reservations in the same nodeset"
+  }
+
+  precondition {
```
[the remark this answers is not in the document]
abbas1902: I don't mind removing it, but don't we want to prevent the formation of placement groups?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -104,9 +91,31 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
 
     # Override with properties explicit specified in the nodeset
     props.update(nodeset.get("instance_properties") or {})
-    
+
     return props
 
+def update_reservation_props(reservation:Optional[object], props:object) -> None:
```
mr0re1, line 97: `reservation` should not be `None`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -78,23 +78,10 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
...
-            log.info(
-                f"reservation {reservation.bulk_insert_name} is being used with policies {props.resourcePolicies}"
-            )
-        else:
-            props.resourcePolicies = []
-            log.info(
-                f"reservation {reservation.bulk_insert_name} is being used without any policies"
-            )
+    if (fr := lookup().future_reservation(nodeset)) and fr.active and fr.specific:
+        update_reservation_props(fr.active_reservation, props)
```
mr0re1, line 84: This change doesn't address my comment. 
We need to handle case of non-active FR before we decide to issue a request for this instance. 
Your current code just pretends that there is no reservation at all.

abbas1902: Ah ok I initially misunderstood, yes hopefully this gap has now been addressed (in `resume_nodes` I purged all inactive nodes from the resume list). Behaviourally the node will go into configuring state for a few seconds before spitting out: 

```
srun: error: Job allocation 1 has been revoked
srun: Force Terminated JobId=1
srun: error: Node failure on whpcslurm-computenodeset-0

File: community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf [hunk 23]
```diff
@@ -463,6 +463,16 @@ variable "reservation_name" {
...
 
+variable "future_reservation" {
+  description = <<-EOD
+  If set, will make use of the future reservation for the nodeset. Input can be either the future reservation name or its selfLink in the format 'projects/PROJECT_ID/zones/ZONE/futureReservations/FUTURE_RESERVATION_NAME'.
+  See https://cloud.google.com/compute/docs/instances/future-reservations-overview
+  EOD
+  type        = string
+  default     = ""
+  nullable    = false
+}
```
mr0re1, line 474: Please add `validation`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1793,6 +1811,50 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
             name=name,
             policies=policies,
             bulk_insert_name=nodeset.reservation_name)
+    
+    def future_reservation(self, nodeset:object) -> Optional[FutureReservation]:
+        if not nodeset.future_reservation:
+            return None
+
+        zones = list(nodeset.zone_policy_allow or [])
```
mr0re1, line 1819: Exract zone from selfLink

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1793,6 +1811,50 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+            name = match.group("name")
+
+        fr = self._get_future_reservation(project, zone, name)
+
+        start_time = datetime.fromisoformat(fr["timeWindow"]["startTime"][:-1])
+        end_time = datetime.fromisoformat(fr["timeWindow"]["endTime"][:-1])
+
+        if "autoCreatedReservations" in fr["status"] and (fr_res:=fr["status"]["autoCreatedReservations"][0]):
+            if (match := re.search(r'/reservations/(?P<name>[^/]+)(/.*)?$',fr_res)) and (res := self._get_reservation(project, zone, match.group('name'))) and (start_time<=datetime.utcnow()<=end_time):
+                active_reservation = ReservationDetails(project=project,zone=zone,name=res["name"],policies=[],bulk_insert_name=res["name"])
```
mr0re1, line 1841: Could you pleas share (DM me, not here) the example of `res`, my concern is using it as `bulk_insert_name`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1461,6 +1461,18 @@ class ReservationDetails:
     policies: List[str] # names (not URLs) of resource policies
     bulk_insert_name: str # name in format suitable for bulk insert (currently identical to user supplied name in long format)
 
+@dataclass(frozen=True)
+class FutureReservation:
+    project: str
+    zone: str
+    name: str
+    active: Optional[bool]
```
mr0re1, line 1469: Is it possible to have `active != (active_reservation != None)`?
If not, than it's pure derivative, consider adding 

```py
@property 
def active(self) -> bool:
  return self.active_reservation != None
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1461,6 +1461,18 @@ class ReservationDetails:
...
+    specific: Optional[bool]
```
mr0re1, line 1470: Why are those fields , except `active_reservation`, are `Optional`. Under what conditions they will not be set?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
mr0re1: My lazy recommendation would be to fetch actual reservation from API, create a function shared with `def nodeset_reservation` to make request and compose `ReservationDetails`

abbas1902: I will send you this as requested, let's sync offline about this point

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -54,4 +54,14 @@ output "nodeset" {
     condition     = !var.enable_placement || !var.dws_flex.enabled
     error_message = "Cannot use DWS Flex with `enable_placement`."
   }
+
+  precondition {
+    condition     = var.reservation_name == "" || var.future_reservation == null
```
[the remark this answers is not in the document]
abbas1902: Thanks forgot default was changed

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -337,6 +341,11 @@ def start_tpu(data):
 
 def resume_nodes(nodes: List[str], resume_data=None):
     """resume nodes in nodelist"""
+    # Prevent non-active nodes associated with a future reservation from being resumed
+    for i in range(len(nodes)):
+        if lookup().future_reservation(lookup().node_nodeset(nodes[i])).active is False:
```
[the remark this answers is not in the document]
abbas1902: Made it check that fr is non-None. I think this mitigates the issue but should error log still be added for something else?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -159,21 +156,26 @@ def _find_tpu_node_status(nodename, state):
...
 
     # split below is workaround for VMs whose hostname is FQDN
     inst = lookup().instance(nodename.split(".")[0])
     power_flags = frozenset(
         ("POWER_DOWN", "POWERING_UP", "POWERING_DOWN", "POWERED_DOWN")
     ) & (state.flags if state is not None else set())
+   
+    if ns.future_reservation:
+        fr_sync(ns, nodename, state)
+        return NodeStatus.unchanged
```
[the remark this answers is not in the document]
abbas1902: Refactored a bit so we could take action in bulk, now in `slurm_sync` we call this function and similar to `resume.py` will filter out all fr nodes and manipulate its state in bulk.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: realized `active`  and `active_reservation` both track essentially the same thing. So `active` has been dropped.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: Changed the parameters now only `active_reservation` will ever be `Optional`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -404,6 +406,29 @@ def sync_slurm():
...
+    if not nodes:
+        return
+
+    fr_nodes = []
+    for i in range(len(nodes)):
+        if (lookup().node_is_fr(nodes[i])):
+            fr_nodes.append(nodes[i])
+
+    fr = lookup().future_reservation(lookup().node_nodeset(fr_nodes[0])) #we only need one fr for the whole nodeset
+    shared_status = lookup().slurm_node(fr_nodes[0]).base #status should be shared amongst these nodes
```
mr0re1, line 419: > status should be shared amongst these nodes

This is false assumption, node state can change independently for multiple reasons.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
abbas1902: Added shared function

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
abbas1902: Refactored to move away from this assumption, fr nodes only have special actions if they are outside of the bounds of the res, other we pass and let slurmsync bring handle their actions.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -463,6 +463,21 @@ variable "reservation_name" {
...
+  See https://cloud.google.com/compute/docs/instances/future-reservations-overview
+  EOD
+  type        = string
+  default     = ""
+  nullable    = false
+
+  validation {
+    condition     = length(regexall("^(projects/([a-z0-9-]+)/zones/([a-z0-9-]+)/futureReservations/([a-z0-9-]+))?$", var.future_reservation)) > 0 || length(regexall("^([a-z0-9-]+)$", var.future_reservation)) > 0
+    error_message = "Future reservation must be either the future reservation name or its selfLink in the format 'projects/PROJECT_ID/zone/ZONE/futureReservations/FUTURE_RESERVATION_NAME'."
+  }
```
mr0re1, line 478: We need to add validation that: (nodeset has single zone) AND (FR zone ==  nodeset zone)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -132,9 +118,28 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
...
+    }
+
+    if reservation.dense:
+        props.scheduling.provisioning_model = "RESERVATION_BOUND"
+
+    # Figure out `resourcePolicies`
+    if reservation.policies: # use ones already attached to reservations
+        props.resourcePolicies = reservation.policies
+    elif reservation.dense: # use once created by Slurm
+        props.resourcePolicies = [placement_group]
```
mr0re1, line 137: There is no `placement_group` variable in this function.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -132,9 +118,28 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
...
     return props
 
+def update_reservation_props(reservation:object, props:object) -> None:
+    props.reservationAffinity = {
+        "consumeReservationType": "SPECIFIC_RESERVATION",
+        "key": f"compute.{util.universe_domain()}/reservation-name",
+        "values": [reservation.bulk_insert_name],
+    }
+
+    if reservation.dense:
```
mr0re1, line 130: It should be `if reservation.dense OR reservation_comes_from_DWS_calendar`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -320,6 +324,13 @@ def start_tpu(data):
 
 def resume_nodes(nodes: List[str], resume_data: Optional[ResumeData]):
     """resume nodes in nodelist"""
+    # Prevent inactive nodes associated with a future reservation from being resumed
+    active_nodes = nodes[:]
```
mr0re1, line 328: This is bad way to filter list, both performance and readability wise.
Consider instead following:

```py
dormant_fr_nodes, nodes = util.separate(nodes, is_dormant_fr_node)

if dormant_fr_nodes:
  # do stuff here
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -320,6 +324,13 @@ def start_tpu(data):
...
+    for i in range(len(nodes)):
+        if (fr := lookup().future_reservation(lookup().node_nodeset(nodes[i]))) and not fr.active_reservation:
+            del active_nodes[i]
+
```
mr0re1, line 332: We shouldn't just skip them:
* log.warning
* down_nodes_notify_jobs(dormant_fr_nodes, "bla-bla reservation is not here")

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -152,6 +150,13 @@ def _find_dynamic_node_status() -> NodeAction:
     # * delete orhpaned instances
     return NodeActionUnchanged()  # don't touch dynamic nodes
 
+def get_fr_action(nodename:str, state:NodeState) -> Optional[NodeAction]:
+    fr = lookup().future_reservation(lookup().node_nodeset(nodename))
+    now = datetime.utcnow()
+    if fr and now<fr.start_time:
+        return NodeActionDown(reason=f"Waiting for reservation to start at {fr.start_time}")
+    if fr and now>fr.end_time:
+        return NodeActionDown(reason="Reservation is after end-time")
```
mr0re1, line 159: add reservation name (short one) to the message

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1586,6 +1597,9 @@ def partition_is_tpu(self, part: str) -> bool:
     def node_is_tpu(self, node_name=None):
         nodeset_name = self.node_nodeset_name(node_name)
         return self.cfg.nodeset_tpu.get(nodeset_name) is not None
+    
+    def node_is_fr(self, node_name=None) -> bool:
+        return self.future_reservation(self.node_nodeset(node_name)) is not None
```
mr0re1, line 1602: nit. `return self.node_nodeset(node_name).future_reservation is not None`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1784,19 +1818,35 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
-        # Converts policy URLs to names, e.g.:
-        # projects/111111/regions/us-central1/resourcePolicies/zebra -> zebra
-        policies = [u.split("/")[-1] for u in reservation.get("resourcePolicies", {}).values()]
+        active_reservation = None
+        match = re.search(r'^projects/(?P<project>[^/]+)/zones/(?P<zone>[^/]+)/futureReservations/(?P<name>[^/]+)(/.*)?$', nodeset.future_reservation)
+        project, zone, name = match.group("project","zone","name")
+        fr = self._get_future_reservation(project,zone,name)
 
-        return ReservationDetails(
+        start_time = datetime.fromisoformat(fr["timeWindow"]["startTime"][:-1])
```
mr0re1, line 1832: Please add a TODO to remove trimming hack after moving to 3.11
Also you can provide context: https://discuss.python.org/t/parse-z-timezone-suffix-in-datetime/2220/30

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1784,19 +1818,35 @@ def nodeset_reservation(self, nodeset: object) -> Optional[ReservationDetails]:
...
+        end_time = datetime.fromisoformat(fr["timeWindow"]["endTime"][:-1])
+
+        if "autoCreatedReservations" in fr["status"] and (fr_res:=fr["status"]["autoCreatedReservations"][0]):
+            if (match := re.search(r'projects/(?P<project>[^/]+)/zones/(?P<zone>[^/]+)/reservations/(?P<name>[^/]+)(/.*)?$',fr_res)) and (start_time<=datetime.utcnow()<=end_time):
```
mr0re1, line 1836: It should always match, if it doesn't it's an error and should be thrown

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: Added but going to sync offline about some questions I have with this.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -54,4 +54,28 @@ output "nodeset" {
...
+
+  precondition {
+    condition     = var.future_reservation == "" || length(var.zones) == 0
+    error_message = <<-EOD
+      If a future reservation is specified, `var.zones` should be empty.
+    EOD
+  }
+
+  precondition {
+    condition     = var.future_reservation == "" || length(regexall("/zones/${var.zone}/", local.nodeset.future_reservation)) > 0
```
mr0re1, line 76: nit. can be simplified: `local.fr_zone == local.zones[0]`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -132,9 +118,28 @@ def instance_properties(nodeset:object, model:str, placement_group:Optional[str]
 
     # Override with properties explicit specified in the nodeset
     props.update(nodeset.get("instance_properties") or {})
-    
     return props
 
+def update_reservation_props(reservation:object, props:object, placement_group:Optional[str], reservation_from_fr:bool = False) -> None:
```
mr0re1, line 123: nit. Don't use default values.

mr0re1 (approved):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -152,6 +150,13 @@ def _find_dynamic_node_status() -> NodeAction:
     # * delete orhpaned instances
     return NodeActionUnchanged()  # don't touch dynamic nodes
 
+def get_fr_action(nodename:str, state:NodeState) -> Optional[NodeAction]:
+    fr = lookup().future_reservation(lookup().node_nodeset(nodename))
+    now = datetime.utcnow()
```
mr0re1, line 155: If reservation is dormant and node is DOWN return NodeActionUnchanged

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
abbas1902: Noted, I can't use local.zones directly since its a set and hence unordered. But since we already check `var.zones` to be empty then we can use compared against `var.zone`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
abbas1902: Added, now its if reservation is dormant and node is not DOWN make it go DOWN otherwise return NodeActionUnchanged

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -152,6 +151,13 @@ def _find_dynamic_node_status() -> NodeAction:
     # * delete orhpaned instances
     return NodeActionUnchanged()  # don't touch dynamic nodes
 
+def get_fr_action(fr: FutureReservation, nodename:str, state:NodeState) -> Optional[NodeAction]:
+    now = datetime.utcnow()
```
mr0re1, line 155: nit. too many too long lines, consider:

```py
if  fr.start_time < now  < fr.end_time:
  return None # handle us any other node
if state.base is "DOWN":
  return NodeActionUnchanged()

if fr.start_time >= now  
  msg = f"Waiting for reservation:{fr.name} to start at {fr.start_time}" 
else:
  msg = f"Reservation:{fr.name} is after its end-time"
return NodeActionDown(reason=msg)


```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -217,6 +223,10 @@ def get_node_action(nodename: str) -> NodeAction:
     """Determine node/instance status that requires action"""
     state = lookup().slurm_node(nodename)
 
+    if fr := lookup().node_is_fr(nodename):
+        if fr_action := get_fr_action(nodename, state):
```
mr0re1, line 227: nit. `s/fr_action/action/`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -152,6 +151,13 @@ def _find_dynamic_node_status() -> NodeAction:
     # * delete orhpaned instances
     return NodeActionUnchanged()  # don't touch dynamic nodes
 
+def get_fr_action(fr: FutureReservation, nodename:str, state:NodeState) -> Optional[NodeAction]:
+    now = datetime.utcnow()
+    if fr and now<fr.start_time:
+        return NodeActionDown(reason=f"Waiting for reservation:{fr.name} to start at {fr.start_time}") if state.base != "DOWN" else NodeActionUnchanged()
+
+    if fr and now>fr.end_time:
```
mr0re1, line 159: `fr` is never `None`, otherwise it would be typed as `Optional[FutureReservation]` )

abbas1902: Yes I forgot to update `node_is_fr` to return fr is it exists. Changed that so now we know it is never None

mr0re1 (approved):

PR #3230, opened 2024-11-07
Make gke-node-pool compatible with TPG 6.x

tpdownes:
The gke-node-pool module uses older "attribute" syntax for the
    GPU-related arguments that has been removed in the google Terraform
    plugin 6.x. This commit replaces attribute syntax with block syntax.
    The key to understanding this change is that a dynamic block iterating
    over a list is equivalent to null when the list is empty (no dynamic
    blocks are inserted).

    The gpu_sharing_config and gpu_driver_installation_config settings are
    not (and never were) list(object) in the Terraform plugin. They could only
    ever taken on length 0 or 1. These are therefore being converted to object
    format as they are in the API.

    - https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#nested_guest_accelerator
    - https://developer.hashicorp.com/terraform/language/attr-as-blocks

### Submission Checklist

### DIFF
--- a/modules/compute/gke-node-pool/variables.tf
+++ b/modules/compute/gke-node-pool/variables.tf
@@ -79,16 +79,32 @@ variable "guest_accelerator" {
   type = list(object({
     type  = optional(string)
     count = optional(number, 0)
-    gpu_driver_installation_config = optional(list(object({
+    gpu_driver_installation_config = optional(object({
       gpu_driver_version = string
-    })))
+    }), { gpu_driver_version = "DEFAULT" })
     gpu_partition_size = optional(string)
-    gpu_sharing_config = optional(list(object({
-      gpu_sharing_strategy       = optional(string)
-      max_shared_clients_per_gpu = optional(number)
-    })))
+    gpu_sharing_config = optional(object({
+      gpu_sharing_strategy       = string
+      max_shared_clients_per_gpu = number
+    }))
   }))
-  default = null
+  default  = []
+  nullable = false
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.count != null])
+    error_message = "var.guest_accelerator[*].count cannot be null"
+  }
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.count >= 0])
+    error_message = "var.guest_accelerator[*].count must never be negative"
+  }
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.gpu_driver_installation_config != null])
+    error_message = "var.guest_accelerator[*].gpu_driver_installation_config must not be null; leave unset to enable GKE to select default GPU driver installation"
+  }
 }
 
 variable "image_type" {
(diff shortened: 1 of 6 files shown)


modules/compute/gke-node-pool/variables.tf
```diff
@@ -79,16 +79,32 @@ variable "guest_accelerator" {
...
+  }
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.count >= 0])
+    error_message = "var.guest_accelerator.count must never be negative"
+  }
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.gpu_driver_installation_config != null])
+    error_message = "var.guest_accelerator.gpu_driver_installation_config must not be set to null; use default value of empty list or set explicitly"
```
tpdownes, line 106: I see a problem with this error message that we should resolve before merging. I have 3 tests running on the current code and I'd like to see their success/failure first.

ankitkinra: are these new validations ? Also maybe you want to update this as the test ran ?

tpdownes: Yes, they are new, but they are valuable. In order of value:

## Protection against count = null

I have made the choice to refactor as little as possible in main.tf where it is not necessary. I think this code was written without a clear understanding of the behavior of `try(null, any, other, set, of values)`. The preceding code will return `null`. For example, this code (present in the PR and before the PR) will fail if count is null:

```hcl
  count = coalesce(try(ga.value.count, 0) > 0 ? ...
```

Proof:

```
~❯❯❯ terraform console
> try(null, 0) > 0
╷
│ Error: Operation failed
│
│
│ Error during operation: argument must not be null.
```

## Protection against count being negative

This is just sanity checking and I am doing it "while we are at it"

## Protection against gpu_driver_installation_config being null

The current code makes the choice to always install some driver, implicitly inserting a request for "DEFAULT" to GKE. I am choosing to make that default value more obvious in the variable definition, but otherwise the intent of this PR is to be a direct refactor for TPG 6.x without any behavioral changes.

IMO, we should allow the user to set this to null (so that no driver is installed) but that should be a small, focused PR.

modules/compute/gke-node-pool/variables.tf
```diff
@@ -79,16 +79,32 @@ variable "guest_accelerator" {
...
-  default = null
+  default  = []
+  nullable = false
+
+  validation {
+    condition     = alltrue([for ga in var.guest_accelerator : ga.count != null])
+    error_message = "var.guest_accelerator[*].count cannot be null"
+  }
+
+  validation {
```
ankitkinra, line 99: nit: is this possible ? can it be negative ?

ankitkinra (approved):
left a nit , other wise LGTM

modules/compute/gke-node-pool/variables.tf
tpdownes: well, yes, the user could accidentally supply a negative value in a blueprint. It just seems like something to validate early vs late.

PR #3251, opened 2024-11-12
Don't install Lustre in A3 tests

rohitramu:
Lustre clients do not seem to be available on their "latest" channel for Ubuntu 20.04, and this is causing integration test failures.

This PR is a temporary workaround to prevent failures while we debug the issue.


cdunbar13 (approved):
Wait for tests to complete, otherwise LGTM

rohitramu:
> Wait for tests to complete, otherwise LGTM

Thanks, both `ml-a3-highgpu-slurm` tests succeeded so I'll merge this PR.

PR #3253, opened 2024-11-13
Remove CentOS 7 from list of supported images

rohitramu:
CentOS 7 has been deprecated since July 2024.  This PR updates the list of supported VM images in the Cluster Toolkit documentation.

### DIFF
--- a/docs/vm-images.md
+++ b/docs/vm-images.md
@@ -105,20 +105,9 @@ project and the new image name in the `instance_image` field discussed in
 
 ## Cluster Toolkit Supported Images
 
-### HPC CentOS 7
-
-The Cluster Toolkit has officially supported the [HPC CentOS 7 VM Image][hpcimage]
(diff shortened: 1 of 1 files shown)


tpdownes:
This is rendering with "half an extra column" on the right-hand-side of the table:

https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/4641c36bd4d81028d75c398cdfd3828e7f35a65c/docs/vm-images.md

There doesn't seem to be anything similar in the commit on main.

rohitramu:
> This is rendering with "half an extra column" on the right-hand-side of the table:
> 
> https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/4641c36bd4d81028d75c398cdfd3828e7f35a65c/docs/vm-images.md
> 
> There doesn't seem to be anything similar in the commit on main.

Good catch.  It was the blank separator row that still had an extra column.  I've fixed this now.

tpdownes (approved):

PR #3279, opened 2024-11-18
add support for enable DCGM monitoring in GKE

chengcongdu:
- add support for enable DCGM monitoring in GKE
- add logging config with default component enabled
- remove monitoring_service and logging_service since these will conflict with the new monitoring_config and logging_config
- default component in monitoring_config and logging_config will retain the current setting of monitoring_service and logging_service


chengcongdu:
starting gke related checks

modules/scheduler/gke-cluster/main.tf
```diff
@@ -37,6 +37,23 @@ locals {
 
   # multi networking needs enabled Dataplane v2
   derived_enable_dataplane_v2 = coalesce(var.enable_dataplane_v2, local.derived_enable_multi_networking)
+
+  default_monitoring_component = [
```
ankitkinra, line 41: Are these all free by default ?

modules/scheduler/gke-cluster/main.tf
```diff
@@ -37,6 +37,23 @@ locals {
...
+    "DAEMONSET",
+    "DEPLOYMENT",
+    "STATEFULSET",
+    "STORAGE",
+    "HPA",
+    "CADVISOR",
+    "KUBELET"
+  ]
+
+  default_logging_component = [
```
ankitkinra, line 53: Also are these free by default ?

ankitkinra (approved):
PTAL a look at comments, but otherwise looks great.

modules/scheduler/gke-cluster/main.tf
chengcongdu: there is always cost for storage, but these are the minimal observability a cluster should have. And it is the default previously when the toolkit is using monitoring_service and logging_service as well.

chengcongdu: free for GKE Enterprise https://cloud.google.com/stackdriver/pricing?_gl=1*10s754z*_ga*MTE3Mjk1NjYyMy4xNzMxOTU3NDc1*_ga_WH2QY8WWF5*MTczMTk2MzcwNS4yLjEuMTczMTk2Nzk3Ny40Ny4wLjA.#anthos-costs

chengcongdu:
validation check pass https://screenshot.googleplex.com/72e3M4cvjURpD7n

PR #3288, opened 2024-11-19
Promote the new nic-types in vm-instance

cdunbar13:
Adding MRDMA and IRDMA to the vm-instance nic type list.  Requires version 6.13.0 of google-beta.


annuay-google (approved):
LGTM, please wait for tests using VM to pass

PR #3293, opened 2024-11-20
Updating RDMA-VPC to use v9.3 CFT modules

cdunbar13:
Enables compatibility with google/google-beta plugin providers v6.x. Tested using basic blueprint with just RDMA-VPC.

Used beta APIs for google-beta.


tpdownes (approved):
Please accept the suggestion and merge.

PR #3298, opened 2024-11-20
Node pool changes a3u

annuay-google:
For node pool support:

- Extended reservations
- Supports A3 Ultra NPI

### Submission Checklist


modules/compute/gke-node-pool/variables.tf
```diff
@@ -386,6 +386,12 @@ variable "gke_version" {
   type        = string
 }
 
+variable "node_version" {
```
annuay-google, line 389: Will remove this and use `gke_version` instead. Will make the changes tomorrow IST

annuay-google: Node version is not required on prod clusters, removing this

modules/compute/gke-node-pool/reservation_definitions.tf
```diff
@@ -14,18 +14,28 @@
  * limitations under the License.
 */
 
+# Split the input into three different lists where the details of a given reservation are at the same index across these lists. 
+locals {
+  # Specific block of an extended reservation can be targeted with exr-one/reservationBlocks/exr-one-block-1
+  # Data source needs to be queried with the reservation name only. So, we extract the reservation name
+  input_reservation_names    = [for r in try(var.reservation_affinity.specific_reservations, []) : split("/", r.name)[0]]
```
tpdownes, line 21: nit: I think you would be better off with default values for specific_reservations of `[]` and validation blocks that protect against it being set to null. e.g. in the variables.tf file for guest_accelerator.

Then you can avoid the `try` function here and the intent becomes clearer. I would create a task to look at this for future work.

```
  validation {
    condition     = alltrue([for ga in var.guest_accelerator : ga.gpu_driver_installation_config != null])
    error_message = "var.guest_accelerator[*].gpu_driver_installation_config must not be null; leave unset to enable GKE to select default GPU driver installation"
  }
```

tpdownes (commented):
1. A `nit` for future work.
2. I've triggered all GKE tests to confirm they pass.

modules/compute/gke-node-pool/reservation_definitions.tf
annuay-google: Done, have tagged you in the task as reporter

PR #3299, opened 2024-11-20
Cluster changes for a3u

annuay-google:
For cluster and pre-existing cluster, support:
1. RDMA Networking
2. Cluster availability and reference types
3. Multi-zonal clusters

### Submission Checklist


modules/scheduler/gke-cluster/main.tf
```diff
@@ -65,7 +65,7 @@ resource "google_container_cluster" "gke_cluster" {
 
   project         = var.project_id
   name            = local.name
-  location        = var.region
+  location        = var.cluster_availability_type == "MULTI_ZONAL" ? var.zone : var.region
```
annuay-google, line 68: This conditional looks a bit sketchy, I'll dig deeper tomorrow IST

modules/scheduler/gke-cluster/main.tf
```diff
@@ -209,6 +209,15 @@ resource "google_container_cluster" "gke_cluster" {
       condition     = !(!coalesce(var.enable_multi_networking, true) && length(var.additional_networks) > 0)
       error_message = "'enable_multi_networking' cannot be false when using multivpc module, which passes additional_networks."
     }
+    precondition {
```
tpdownes, line 212: nit: these would be better as simple validation blocks on the variables themselves. I believe this will also fail pretty early, but a validation block in variable is guaranteed to fail at "terraform plan" or "terraform apply" before anything is created. For example:

https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/1a1e22a252c8b5bf0be57fa44f46bf72ca0363bf/modules/compute/vm-instance/variables.tf#L230-L235

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -358,6 +358,24 @@ variable "additional_networks" {
...
 }
 
+variable "cluster_reference_type" {
+  description = "How the google_container_node_pool.system_node_pools refers to the cluster. Possible values are: {SELF_LINK, NAME}"
+  default     = "SELF_LINK"
+  type        = string
+}
+
+variable "cluster_availability_type" {
+  description = "Type of cluster availability. Possible values are: {REGIONAL, MULTI_ZONAL}"
```
tpdownes, line 368: I believe the terminology being used here is wrong and needs to be corrected. I believe the possible availability types are "REGIONAL" and "ZONAL". For either case, you can have node pools that are "ZONAL" or "MULTI_ZONAL" but that's not what the variable is specifying.

- https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-zonal-cluster#single-zone_versus_multi-zonal
- https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-regional-cluster#create-regional-single-zone-nodepool

tpdownes (changes requested):

modules/scheduler/gke-cluster/variables.tf
annuay-google: Correct, this variable is to specify the cluster location

annuay-google: Updated the terminology

modules/scheduler/gke-cluster/main.tf
annuay-google: Ack, done

modules/scheduler/gke-cluster/main.tf
annuay-google: With the terminology fixed to "ZONAL", this looks good

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -362,18 +368,20 @@ variable "cluster_reference_type" {
   description = "How the google_container_node_pool.system_node_pools refers to the cluster. Possible values are: {SELF_LINK, NAME}"
   default     = "SELF_LINK"
   type        = string
+  validation {
```
tpdownes, line 371: Add `nullable = false` here otherwise the `contains` test can Error (stacktrace):

```
❯ terraform console
>  contains(["SELF_LINK", "NAME"], null)
╷
│
│
```

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -362,18 +368,20 @@ variable "cluster_reference_type" {
...
+    condition     = contains(["SELF_LINK", "NAME"], var.cluster_reference_type)
+    error_message = "`cluster_reference_type` must be one of {SELF_LINK, NAME}"
+  }
 }
 
 variable "cluster_availability_type" {
-  description = "Type of cluster availability. Possible values are: {REGIONAL, MULTI_ZONAL}"
+  description = "Type of cluster availability. Possible values are: {REGIONAL, ZONAL}"
```
tpdownes, line 378: Add nullable = false here otherwise the contains test can Error (stacktrace):

```
❯ terraform console
>  contains(["SELF_LINK", "NAME"], null)
╷
│
│
```

tpdownes (approved):
Please make the minor `nullable` changes requested, then squash and merge.

Thanks!

modules/scheduler/gke-cluster/variables.tf
annuay-google: Done

modules/scheduler/gke-cluster/variables.tf
annuay-google: Done

PR #3303, opened 2024-11-24
Simplify providing additional_networks for Slurm nodesets

wiktorn:
Simplify definition of `additional_networks`, before the change one had to write for example
```yaml
           additional_networks:
             - network: ""
               subnetwork: $(network2.subnetwork_self_link)
               subnetwork_project: ""
               network_ip: ""
               nic_type: VIRTIO_NET
               stack_type: ""
               queue_count: null
               access_config: []
               ipv6_access_config: []
               alias_ip_range: []
```

Now its possible to write just:
```yaml
          additional_networks:
            - subnetwork: $(network2.subnetwork_self_link)
              nic_type: VIRTIO_NET
```


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -414,24 +414,24 @@ variable "additional_networks" {
...
-    network            = string
-    subnetwork         = string
-    subnetwork_project = string
-    network_ip         = string
-    nic_type           = string
-    stack_type         = string
-    queue_count        = number
-    access_config = list(object({
+    network            = optional(string)
+    subnetwork         = optional(string)
```
mr0re1, line 418: nit. I would recommend to make `subnetwork` non-optional since it's an "intended" way to pass "VPC", though name is misleading due to historic reasons (should be `subnetwork_self_link`).

mr0re1 (approved):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
wiktorn: done

PR #3310, opened 2024-11-26
Updating default vars for image to use managed-notebooks base image

JamesDuncanNz:
Updating the default Instance Image project & family to leverage the Images used by Vertex AI Workbench. 

The previous images were resulting in Jupyter not starting and the Notebook not making it out of the provisioning stage.


jrossthomson (commented):
LGTM

jrossthomson:
@mr0re1 

I have approved this.

PR #3322, opened 2024-11-29
GKE clusters can now be duplicated by changing only the deployment name

annuay-google:
**Note: All changes are backward compatible. The older variable "secondary_ranges" has been retained for backward compatibility.**

This will help us to use deployment name as a prefix to differentiate across VPCs, subnets, cluster and node pools

This PR lets you do something like
```
secondary_ranges_list:
- subnetwork_name: $(vars.deployment_name)-sub-0
  ranges:
  - range_name: pods
    ip_cidr_range: 10.4.0.0/14
  - range_name: services
    ip_cidr_range: 10.0.32.0/20
```
which wasn't possible earlier, as $(vars.x) can't be used for a yaml key

This in turn will let you create a new identical cluster for GKE by changing a single variable. Currently, we need to change 4-8 different names based on the cluster type

## Testing
- Applied with current code, planned with this branch and validated there is no diff. Current code used secondary_ranges and the new one used secondary_ranges_list
- Verified secondary CIDRs on pantheon
- Ran all GKE integration tests- all pass
- Passed both secondary_ranges and secondary_ranges_list and verified "expected only one" error was thrown
- Verified secondary ranges created are identical when using secondary_ranges or secondary_ranges_list

### Submission Checklist


ighosh98:
Question: You mentioned this would `help us to use deployment name as a prefix to differentiate across VPCs, subnets, cluster and node pools`. Could you share a code pointer where you are adding the deployment name prefix dynamically(as you explained in the example)?

Could you add examples/ screenshots to showcase the difference in the name/id of networks being created before and after this change?

ighosh98 (commented):
Question: You mentioned `this would help us to use deployment name as a prefix to differentiate across VPCs, subnets, cluster and node pools`. Could you share a code pointer where you are adding the deployment name prefix dynamically(as you explained in the example)?

Could you add examples/ screenshots to showcase the difference in the name/id of networks being created before and after this change?

modules/network/vpc/variables.tf
```diff
@@ -148,9 +148,9 @@ variable "additional_subnetworks" {
 }
 
 variable "secondary_ranges" {
```
ankitkinra, line 150: What happens if a customer doesn't update their existing blueprint to adapt to this change? Can we give them an example on how to modify this ?

examples/hpc-gke.yaml
```diff
@@ -30,7 +30,8 @@ deployment_groups:
     settings:
       subnetwork_name: gke-subnet
```
ankitkinra, line 31: Do the names on line 31 and 33 have to always same ? If yes how can we prevent user error by allowing them to specify different names ( should we add a validation to prevent this ) ?

ighosh98:
> Question: You mentioned `this would help us to use deployment name as a prefix to differentiate across VPCs, subnets, cluster and node pools`. Could you share a code pointer where you are adding the deployment name prefix dynamically(as you explained in the example)?
> 
> Could you add examples/ screenshots to showcase the difference in the name/id of networks being created before and after this change?

resolved offline

examples/hpc-gke.yaml
annuay-google: This doesn't always need to be the same. There can be multiple subnetworks, and each subnetwork may or may not contain secondary CIDR ranges

On preventing user error: for reference blueprints, we can do something like this to keep both in sync:
```
      subnetworks:
      - subnet_name: $(vars.deployment_name)-sub-0
        subnet_region: $(vars.region)
        subnet_ip: 192.168.0.0/18
      secondary_ranges:
      - subnetwork_name: $(vars.deployment_name)-sub-0
        ranges:
        - range_name: pods
          ip_cidr_range: 10.4.0.0/14
        - range_name: services
          ip_cidr_range: 10.0.32.0/20
```
This is also the purpose of this PR

modules/network/vpc/variables.tf
annuay-google: I have updated the description+readme to add an example on how to perform the migration from old to new template

modules/network/vpc/variables.tf
```diff
@@ -147,10 +147,33 @@ variable "additional_subnetworks" {
   }
 }
 
+
 variable "secondary_ranges" {
-  type        = map(list(object({ range_name = string, ip_cidr_range = string })))
-  description = "Secondary ranges that will be used in some of the subnets. Please see https://goo.gle/hpc-toolkit-vpc-deprecation for migration instructions."
-  default     = {}
+  type        = list(object({ subnetwork_name = string, ranges = list(object({ range_name = string, ip_cidr_range = string })) }))
```
mr0re1, line 152: nit. Please use multi-line format. This one-liner is hard to read.

modules/network/vpc/variables.tf
annuay-google: Made the changes backward compatible. Introduced new parameter secondary_ranges_list and retained the older one

modules/network/vpc/variables.tf
annuay-google: Done

modules/network/vpc/variables.tf
```diff
@@ -149,10 +149,26 @@ variable "additional_subnetworks" {
 
 variable "secondary_ranges" {
   type        = map(list(object({ range_name = string, ip_cidr_range = string })))
-  description = "Secondary ranges that will be used in some of the subnets. Please see https://goo.gle/hpc-toolkit-vpc-deprecation for migration instructions."
+  description = <<-EOT
+  "Secondary ranges associated with the subnets.
+  This will be deprecated in favour of secondary_ranges_list going forward.
+  Please migrate to using the same."
```
ighosh98, line 155: Let's define an end date for the migration.

modules/network/vpc/variables.tf
```diff
@@ -250,3 +266,12 @@ variable "firewall_log_config" {
     error_message = "var.firewall_log_config must be set to \"DISABLE_LOGGING\", or enable logging with \"INCLUDE_ALL_METADATA\" or \"EXCLUDE_ALL_METADATA\""
   }
 }
+
+resource "terraform_data" "secondary_ranges_validation" {
+  lifecycle {
+    precondition {
+      condition     = length(var.secondary_ranges) == 0 || length(var.secondary_ranges_list) == 0
+      error_message = "Only one of var.secondary_ranges or var.secondary_ranges_list should be specified"
```
ighosh98, line 274: Can we encourage customers to migrate and remove var.secondary_range from the error message? If needed, maybe link to an example on how the secondary_ranges_list needs to be defined.

annuay-google: This message is only for users who erroneously try to use both the list and the map. This can't be a warning, has to be an error

modules/network/vpc/variables.tf
annuay-google: As discussed, this is not common practice within toolkit. Created a backlog item to enforce a "check" to warn the user when we deprecate this field and stage for removal

ighosh98 (approved):

PR #3327, opened 2024-12-01
Add reservations to vm-instance

cdunbar13:
Added new setting `reservation_name` for vm-instance module.  This was tested using the A3U reservation on hpc-toolkit-gsc.

Will run appropriate integration tests against this.


modules/compute/vm-instance/variables.tf
```diff
@@ -409,3 +409,22 @@ variable "allow_automatic_updates" {
...
+  description = <<-EOD
+    Name of the reservation to use for VM resources, should be in one of the following formats:
+    - projects/PROJECT_ID/reservations/RESERVATION_NAME[/SUFF/IX]
+    - RESERVATION_NAME[/SUFF/IX]
+
+    Must be a "SPECIFIC" reservation
+    Set to empty string if using no reservation or automatically-consumed reservations
+  EOD
+  type        = string
+  default     = ""
```
rohitramu, line 423: Why not make it nullable to distinguish between "unset" and "empty string"?

modules/compute/vm-instance/README.md
```diff
@@ -231,6 +231,7 @@ limitations under the License.
 | <a name="input_placement_policy"></a> [placement\_policy](#input\_placement\_policy) | Control where your VM instances are physically located relative to each other within a zone.<br/>See https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_resource_policy#nested_group_placement_policy | `any` | `null` | no |
 | <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Project in which the HPC deployment will be created | `string` | n/a | yes |
 | <a name="input_region"></a> [region](#input\_region) | The region to deploy to | `string` | n/a | yes |
+| <a name="input_reservation_name"></a> [reservation\_name](#input\_reservation\_name) | Name of the reservation to use for VM resources, should be in one of the following formats:<br/>- projects/PROJECT\_ID/reservations/RESERVATION\_NAME[/SUFF/IX]<br/>- RESERVATION\_NAME[/SUFF/IX]<br/><br/>Must be a "SPECIFIC" reservation<br/>Set to empty string if using no reservation or automatically-consumed reservations | `string` | `""` | no |
```
rohitramu, line 234:
```suggestion
| <a name="input_reservation_name"></a> [reservation\_name](#input\_reservation\_name) | Name of the reservation to use for VM resources, should be in one of the following formats:<br/>- projects/PROJECT\_ID/reservations/RESERVATION\_NAME[/SUFF/IX]<br/>- RESERVATION\_NAME[/SUFF/IX]<br/><br/>Must be a "SPECIFIC" reservation<br/>Do not set if using no reservation or automatically-consumed reservations | `string` | `""` | no |
```

modules/compute/vm-instance/main.tf
```diff
@@ -261,6 +261,17 @@ resource "google_compute_instance" "compute_vm" {
     }
   }
 
+  dynamic "reservation_affinity" {
+    for_each = var.reservation_name == "" ? [] : [1]
+    content {
+      type = "SPECIFIC_RESERVATION"
+      specific_reservation {
+        key    = "compute.googleapis.com/reservation-name"
+        values = [var.reservation_name]
```
mr0re1, line 270: Did you test it with both short name and `projects/PROJECT_ID/reservations/RESERVATION_NAME` formats?

AFAIK, the `projects/PROJECT_ID/reservations/RESERVATION_NAME` is BulkInsert-specific format and not a proper selfLink (zone is missing), so it may not work here.

modules/compute/vm-instance/variables.tf
cdunbar13: Because then the logic of determining if a user used a null or "" to indicate no reservation becomes more complicated.

modules/compute/vm-instance/main.tf
cdunbar13: Yes It has been tested with both formats and works.

modules/compute/vm-instance/README.md
cdunbar13: Not changing this as it a recommendation if the user has the setting in their blueprint.

modules/compute/vm-instance/variables.tf
rohitramu: I don't think it significantly changes the complexity, you only need to wrap `var.reservation_name` in a `coalesce()` call:
```
length(regexall("^((projects/([a-z0-9-]+)/reservations/)?([a-z0-9-]+))?$", coalesce(var.reservation_name, ""))) > 0
```

It may be useful in the future to know with a reasonable degree of confidence whether or not the user set a value or left it unset.  I say "reasonable degree of confidence", because it's always possible for users to explicitly set it to `null`, but that seems generally unlikely to happen.

modules/compute/vm-instance/main.tf
rohitramu, line 270: Just curious - does the API (or underlying Terraform resource) actually take an array as input?  If so, shouldn't we change our interface to expose it as an array?

cdunbar13: It does take an array, but given that it's a specific array it only takes one value.  I think the idea is that `values` is a standardized input for key-value pairs, but in this case there's only going to be one value.

modules/compute/vm-instance/variables.tf
cdunbar13: This is to your comment below: Why give the user more options here?  Null and empty-string will end up evaluating the same way, so you're just adding unnecessary complexity if you allow for null.

Additionally this is how it's handled in the slurm-gcp modules (where I took the code)

rohitramu: Ahh if it's already standardized elsewhere, that makes sense.  Thanks for the explanation!

rohitramu (approved):

PR #3333, opened 2024-12-03
Parallelstore striping config

dgouju:
ParallelStore has different striping configurations possible for file-level and directory-level that impact [performance depending on the use case](https://cloud.google.com/parallelstore/docs/performance). This PR allows you to set those parameters on instance creation.
Possible values are described in the [Terraform module](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/parallelstore_instance) and [REST API](https://cloud.google.com/parallelstore/docs/reference/rest/v1/projects.locations.instances#filestripelevel)


modules/file-system/parallelstore/variables.tf
```diff
@@ -91,3 +91,33 @@ variable "import_destination_path" {
...
+      "FILE_STRIPE_LEVEL_MAX",
+    ], var.file_stripe)
+    error_message = "var.file_stripe must be set to \"FILE_STRIPE_LEVEL_UNSPECIFIED\", \"FILE_STRIPE_LEVEL_MIN\", \"FILE_STRIPE_LEVEL_BALANCED\",  or \"FILE_STRIPE_LEVEL_MAX\""
+  }
+}
+
+variable "directory_stripe" {
+  description = "The parallelstore stripe level for directories."
+  type        = string
+  default     = "DIRECTORY_STRIPE_LEVEL_UNSPECIFIED"
```
harshthakkar01, line 113: Curious, how do we come up with these default value ? Or these are set at downstream level ?

dgouju: Unless I misunderstood what "If not set, FileStripeLevel will default to FILE_STRIPE_LEVEL_BALANCED" meant ? I understood the `_UNSPECIFIED` as an alias?
- https://cloud.google.com/parallelstore/docs/reference/rest/v1/projects.locations.instances#filestripelevel
- https://cloud.google.com/parallelstore/docs/reference/rest/v1/projects.locations.instances#directorystripelevel

harshthakkar01: As per the document, if we don't set it will be default to `FILE_STRIPE_LEVEL_BALANCED`. My question is, do we have a reason to explicitly set default to `DIRECTORY_STRIPE_LEVEL_UNSPECIFIED` ?

harshthakkar01:
Needs to rebase.

harshthakkar01 (approved):

Issue #3343, opened 2024-12-04
GKE node pool upgrade settings are not configurable


chajath:
### Describe the bug

In toolkit, GKE node pool upgrade settings are hardcoded: https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/26fafe0d5a53138b807e490d4d6597544ad658d1/modules/compute/gke-node-pool/main.tf#L62-L66

This stops us from efficiently upgrading nodes in-place. Without change, each node upgrade can take up 9+ minutes, which makes maintaining big node pools unrealistic.

### Steps to reproduce

1. Trigger in-place node pool upgrade

### Expected behavior

You have an option to make sure multiple nodes are made unavailable time to minimize the downtime.

### Actual behavior

You don't have any option to upgrade more than one node at a time.

ankitkinra:
Quick question, you are okay with `SURGE` strategy , but want to configure more than one node `unavailable` at a time ?

chajath:
yes. In our particular case, being able to set higher value for `max_unavailable` while keeping everything else as is, would be enough.

chajath:
Could you cherrypick #3359 into `experimental` branch so we can use with GKE provision?

PR #3359, opened 2024-12-06
make upgrade settings configurable

ighosh98:
### Submission Checklist
Resolves issue #3343 

This PR adds limited support for surge strategy. We would re-evaluate if [blue green support as specified by terraform](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#node_pool_soak_duration-1) is needed in a future PR

In case, customers enter invalid configurations, we would throw an error before cluster provisioning starts by leveraging the preconditions.

##### Why coalesce over try?
[try](https://developer.hashicorp.com/terraform/language/functions/try#try-function) does not treat undefined variables as an error state as it catches for errors in the evaluation process. [Coalesce](https://developer.hashicorp.com/terraform/language/functions/coalesce) is a better fit for our use case

### Testing Proof:

Successfully provisioned a cluster with default parameters.

Screenshot 2024-12-07 at 12 06 35 AM


modules/compute/gke-node-pool/variables.tf
```diff
@@ -391,3 +391,25 @@ variable "max_pods_per_node" {
...
+    max_surge       = 0,
+    max_unavailable = 1
+  }
+  validation {
+    condition = (
+      contains(["SURGE"], var.upgrade_settings.strategy) &&
+      var.upgrade_settings.max_unavailable >= 0 &&
+      var.upgrade_settings.max_surge >= 0 &&
+    (var.upgrade_settings.max_unavailable > 0 || var.upgrade_settings.max_surge > 0))
+    error_message = "upgrade settings are invalid"
```
annuay-google, line 413: Please throw specific, actionable error messages when possible. One error message per and operand would be nice here

ighosh98: The issue with one error message per "and" is that user will have to figure out which one is causing the failure. Should we instead link the terraform doc instead? 

Proposed error_message:
"Only surge strategy is supported. Refer to https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#upgrade_settings-1 for more details"

modules/compute/gke-node-pool/main.tf
```diff
@@ -283,6 +283,22 @@ resource "google_container_node_pool" "node_pool" {
...
+      error_message = "max_unavailable should be set to 0 or greater"
+    }
+    precondition {
+      condition     = var.upgrade_settings.max_surge >= 0
+      error_message = "max_surge should be set to 0 or greater"
+    }
+    precondition {
+      condition     = (var.upgrade_settings.max_unavailable > 0 || var.upgrade_settings.max_surge > 0)
+      error_message = "At least one of max_unavailable or max_surge must greater than 0"
+    }
```
annuay-google, line 301: Isn't it possible that these parameters 
```
    max_surge       = optional(number),
    max_unavailable = optional(number)
```
can be null? Is that case handled gracefully?

ighosh98: This seems to be poorly documented. Will add test results for `{strategy:"SURGE"}`.

ighosh98: Ran multiple tests. There seems to be odd behaviors observed when one of the parameters get set to null. Based on offline discussions. If the values are not specified, we would use default parameters. Description would be updated to reflect the same. Some of the edge cases have been handled in the [kubernetes code base](https://github.com/kubernetes/kubernetes/blob/a499facee693a1a83daadb82d88f7b51d324ffc5/pkg/controller/daemon/update.go#L612-L615)

modules/compute/gke-node-pool/main.tf
```diff
@@ -283,6 +291,22 @@ resource "google_container_node_pool" "node_pool" {
...
+      error_message = "max_unavailable should be set to 0 or greater"
+    }
+    precondition {
+      condition     = local.upgrade_settings.max_surge >= 0
+      error_message = "max_surge should be set to 0 or greater"
+    }
+    precondition {
+      condition     = local.upgrade_settings.max_unavailable > 0 || local.upgrade_settings.max_surge > 0
+      error_message = "At least one of max_unavailable or max_surge must greater than 0"
+    }
```
annuay-google, line 309: Nit: error messages end with a "." from what I see in the module and across the repo

annuay-google (approved):

PR #3362, opened 2024-12-07
Untangle `exclusive` and `enable_placement`, update examples.

mr0re1:
**Motivations:**
* Enable usage of placements with dense reservations.

* Encourage usage of placements by making them to always work regardless of nodeset configuration.

**Done:**
Remove constraints:
* "If a reservation is specified, `var.enable_placement` must be `false`" - dense reservations support external placements, non-dense reservation will ignore Slurm-created placement.

* "If any non-static nodesets has `enable_placement`, `var.exclusive` must be set true" - nodeset-wise (non-exclusive) placement can be used with dynamic nodes.

Remove `enable_placement: false` from SlurmV6 examples.

**Reasons for users to set `enable_placement=false`**:
* If non-dense reservation is used, user can avoid extra-cost of creating placement policies;
* If customer wants to avoid "all or nothing" VM provisioning behaviour;
* If customer wants to intentionally have "spread" VMs (e.g. for reliability reasons)


alyssa-sm:
There is a error in `PR-validation` with one of the `validate_golden_copy` failing that should be addressed but otherwise LGTM

alyssa-sm (approved):

PR #3363, opened 2024-12-07
Add TTL to SSH key

mr0re1:
Add 1 hour TTL to os login keys to avoid:

```
ERROR: (gcloud.compute.os-login.ssh-keys.add) INVALID_ARGUMENT: Login profile size exceeds 32 KiB. Delete profile values to make additional space.
```


ankitkinra (approved):
Can you please add to description how TTL prevents this error ?

PR #3382, opened 2024-12-10
Updating to Slurm-GCP 6.8.7 to add nic-type to additional-networks

cdunbar13:
Updating slurm-gcp ref so it is compatible with A3U changes (primarily accepting new nic-types (MRDMA, IRDMA) in additional network field of nodeset).

Running against Slurm Integration tests.


mr0re1:
Not needed after https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3380 
Your changes are in

PR #3384, opened 2024-12-11
Slurmsync. Safeguard against nodes missing from Slurm state.

mr0re1:
`NodeState` is None => Slurm doesn't know this node, there are two reasons:
 * happy: 
   * node belongs to removed nodeset
   * node belongs to downsized portion of nodeset
   * dynamic node that didn't register itself
 * unhappy:
   * there is a drift in Slurm and SlurmGCP configurations
   * `slurm_nodes` function failed to handle `scontrol show nodes`,
      TODO: make `slurm_nodes` robust by using `scontrol show nodes --json`
      
In either of "unhappy" cases it's too dangerous to proceed - abort slurmsync.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Optional, Type
+
 import pytest
 from mock import Mock
 from common import TstNodeset, TstCfg # needed to import util
 import util
+from util import NodeState
 from datetime import timedelta
 from google.api_core.client_options import ClientOptions  # noqa: E402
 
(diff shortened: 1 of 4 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1700,8 +1701,36 @@ def make_node_tuple(node_line):
...
+            ns = self.node_nodeset(nodename)
+        except:
+            return None # Can't find nodeset, may be belongs to removed nodeset
+        
+        if self.node_is_dyn(nodename):
+            return None # we can't make any judjment for dynamic nodes
+        
+        cnt = sum(self.static_dynamic_sizes(ns))
+        if self.node_index(nodename) >= cnt:
+            return None # node belongs to downsized nodeset
```
nick-stroud, line 1730: Can we turn comments on 1730, 1726, and 1723 into logging. It would be useful in future to be able to observe the difference between None and POWERED_DOWN and to understand why None was returned.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
```diff
@@ -308,3 +311,36 @@ def test_nodeset_reservation_ok(nodeset, policies, expected):
...
+
+
+@pytest.mark.parametrize(
+    "node,state,want",
+    [
+        ("c-n-2", NodeState("DOWN", {}), NodeState("DOWN", {})), # happy scenario
+        ("c-d-vodoo", None, None), # dynamic nodeset
+        ("c-x-44", None, None), # unknown(removed) nodeset
+        ("c-n-7", None, None), # Out of bounds: c-n-[0-4] - downsized nodeset
+        ("c-t-7", None, None), # Out of bounds: c-t-[0-4] - downsized nodeset TPU
```
nick-stroud, line 324: Why not test c-n-5 on one of these to hit the limit case, protect against off by one error.

nick-stroud (approved):

romxero:
thanks yall

PR #3389, opened 2024-12-11
An option added to disable/enable workload script execution

sharabiani:
An option added to disable/enable workload script execution for A3-High and A3-Mega machine types.

The variable defines whether execute the script to create a sample workload and inject rxdm sidecar into workload. Currently, implemented for A3-Highgpu and A3-Megagpu only.


pawloch00 (approved):

annuay-google (approved):
LGTM

PR #3396, opened 2024-12-12
TopologyAwareScheduling enabled by default for Kueue v0.9.1

sharabiani:
This PR enables TopologyAwareScheduling by default for Kueue v0.9.1 installation.
It is already enabled for its v0.9.0.

### DIFF
--- a/modules/management/kubectl-apply/manifests/kueue-v0.9.1.yaml
+++ b/modules/management/kubectl-apply/manifests/kueue-v0.9.1.yaml
@@ -12331,6 +12331,7 @@ spec:
       - args:
         - --config=/controller_manager_config.yaml
         - --zap-log-level=2
+        - --feature-gates=TopologyAwareScheduling=true
         command:
         - /manager
         image: registry.k8s.io/kueue/kueue:v0.9.1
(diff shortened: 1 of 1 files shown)


modules/management/kubectl-apply/manifests/kueue-v0.9.1.yaml
```diff
@@ -12331,6 +12331,7 @@ spec:
       - args:
         - --config=/controller_manager_config.yaml
         - --zap-log-level=2
+        - --feature-gates=TopologyAwareScheduling=true
```
ighosh98, line 12334: Is TopologyAwareScheduling not enabled if we don't have the feature flag? Can you share a code pointer/documentation which explains this behavior?

annuay-google:
Can you please expand the PR description with:
- What are the consequences of enabling TAS by default?
- Why do we need to this enabled by default?
- How have we verified this change is safe?

modules/management/kubectl-apply/manifests/kueue-v0.9.1.yaml
sharabiani: yes, by default TAS is disabled. https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/#admin-facing-apis

PR #3406, opened 2024-12-16
Add kubectl provider in root module for blueprint with GKE cluster module setup

mohitchaurasia91:
## Why we need this change ?

When managing Kubernetes manifests with multiple blueprints using the kubectl-apply module, completely removing a kubectl-apply block from a blueprint results in a "Error: Provider configuration not present" error.

```
  - id: workload_manager_install
    use: [gke_cluster]
    settings:
      kueue:
        install: true
      jobset:
        install: true

  - id: workload_manager_config
    use: [gke_cluster]
    settings:
      apply_manifests:
      - source: $(ghpc_stage("maxtext-gke-a3-files"))/config-map.yaml.tftpl
        template_vars: {name: "a3plus-benchmark-resources-configmap", num_nodes: "1"}
      - source: $(ghpc_stage("maxtext-gke-a3-files"))/kueue-credentials.yaml.tftpl
        template_vars: {num_chips: "8"}
```
Similar issue is observed for PersistentVolume and PersistentVolumeClaims, when provisioning these resources using gke-persistent-volume module, which internally uses kubectl-apply module for k8s manifest deployment.

```
  - id: data-bucket-pv
    source: modules/file-system/gke-persistent-volume
    use: [gke_cluster, data-bucket]
    settings: {capacity_gb: 5000}
    
 - id: shared-filestore-pv
    source: modules/file-system/gke-persistent-volume
    use: [gke_cluster, filestore]
```
Removing workload_manager_config or persistent_volume and recreating and redeploying gives us the error:
```
Error: Provider configuration not present

To work with
module.workload_manager_config.module.kubectl_apply_manifests["1"].kubectl_manifest.apply_doc["2"]
(orphan) its original provider configuration at
module.workload_manager_config.provider["registry.terraform.io/gavinbunney/kubectl"]
is required, but it has been removed. This occurs when a provider
configuration is removed while objects created by that provider still exist
in the state. Re-add the provider configuration to destroy
module.workload_manager_config.module.kubectl_apply_manifests["1"].kubectl_manifest.apply_doc["2"]
(orphan), after which you can remove the provider configuration again.
```

## RCA

The issue stems from how the provider is defined. Unlike other Terraform providers which are typically defined at the root level, the kubectl-apply provider is defined within the modules/management/kubectl-apply/providers.tf file inside the module itself.

This becomes problematic when deleting a blueprint configuration that uses this module.  While the module's folder remains, Terraform effectively loses track of it and its associated provider. Consequently, Terraform can't determine the correct provider (kubectl_apply_manifests) to use when attempting to delete the resources associated with the removed module.

This behavior is due to Terraform's handling of child modules and their providers. You can find a more detailed explanation in this Stack Overflow answer: https://stackoverflow.com/a/58403262

## Solution

Moving kubectl provider to the root module (under `TerraformProviders` block) of the blueprint group, if the group contains dependent module of `gke-cluster` or `pre-existing-gke-cluster` module.

## Backward Compatibility Test

Tested this out for example blueprint (`storage-gke.yaml`, `gke-a3-highgpu.yaml`), following are the steps.
- on develop branch, perform gcluster deploy for example blueprint.
- switched over to feature branch and performed `make` to update gcluster binary.
- performed gcluster deploy, got a warning of past deployment was performed on diff gcluster version, suggesting to use `--force` based on user discretion.
```
Creating deployment folder "gke-a3-highgpu" ...
Error: ghpc_version has changed from "45068640-dirty" to "90aa74bb-dirty", using different versions of GHPC to update a live deployment is not officially supported
Hint: Use `--force` to overwrite the deployment anyway. Proceed at your own risk.
```
- used `--force` flag and did the deployment using feature branch and it worked as expected.
- Only recreate of `kubectl-apply` module related resource were required.
- Scenario of workload_manager_config and persistent_volume module removal and recreate worked as expected.

## How did we test this out ?
- [x] Use the same setup as in issue description. Remove the workload_manager_config module and recreate, redeploy. Verify that the error does not occur and infra is removed as expected
- [x] Use the same setup as in issue description. Remove the persistent_volume module and recreate, redeploy. Verify that the error does not occur and infra is removed as expected
- [x] Test for backward compatibility
- [x] PR cloud build for all GKE related blueprint should have `OK` status

### Submission Checklist


ankitkinra (commented):
Thanks for testing this out thoroughly, can you please add if you performed the following test:

Example blueprint with PV (storage)

Case 1
* First deploy of the blueprint on the `develop` branch (without the changes) with PV 
* Switch branch and do make
* Second deploy of the blueprint without any changes , which would typically not result in any diff

Result: Did it cause recreation of `kubectl` created modules ? Specifically if blueprint had storage, did it cause data loss ? 

Case 2
* First deploy of the blueprint on the `develop` branch (without the changes) with PV 
* Switch branch and do make
* Second deploy of the blueprint with only nodepool scaleup , which would typically not result in recreation of any module.

Result: Did it cause recreation of `kubectl` created modules ? Specifically if blueprint had storage, did it cause data loss ? 

Case 2: 
Case 1
* First deploy of the blueprint on the `develop` branch (without the changes) with PV 
* Switch branch and do make
* Second deploy of the blueprint without any changes , which would typically not result in any diff
* Third deploy of the blueprint of the above blueprint after completion of second deploy

Result 
Did the third deploy of blueprint cause any updates / loss of data ?

pkg/config/expand.go
```diff
@@ -186,6 +187,33 @@ func (bp Blueprint) expandBackend(grp *Group) {
 	}
 }
 
+func kubectlProviderRequiredModule(grp *Group) (bool, Module) {
+	for _, mod := range grp.Modules {
+		if strings.Contains(mod.Source, "gke-cluster") || strings.Contains(mod.Source, "pre-existing-gke-cluster") {
```
ankitkinra, line 192: Can we always add kubectl provider to all the blueprints ? What is the downside of this action ?

mohitchaurasia91: It can lead to an invalid terraform in non-gke case, as kubectl provider depend on gke-cluster / pre-existing-gke-cluster module output, for example.
```
provider "kubectl" {
  apply_retry_count      = 15
  cluster_ca_certificate = module.gke_cluster.cluster_ca_certificate
  host                   = module.gke_cluster.host_endpoint
  load_config_file       = false
  token                  = module.gke_cluster.access_token
}
```

mohitchaurasia91:
> Thanks for testing this out thoroughly, can you please add if you performed the following test:
> 
> Example blueprint with PV (storage)
> 
> Case 1
> 
> * First deploy of the blueprint on the `develop` branch (without the changes) with PV
> * Switch branch and do make
> * Second deploy of the blueprint without any changes , which would typically not result in any diff
> 
> Result: Did it cause recreation of `kubectl` created modules ? Specifically if blueprint had storage, did it cause data loss ?
> 
> Case 2
> 
> * First deploy of the blueprint on the `develop` branch (without the changes) with PV
> * Switch branch and do make
> * Second deploy of the blueprint with only nodepool scaleup , which would typically not result in recreation of any module.
> 
> Result: Did it cause recreation of `kubectl` created modules ? Specifically if blueprint had storage, did it cause data loss ?
> 
> Case 2: Case 1
> 
> * First deploy of the blueprint on the `develop` branch (without the changes) with PV
> * Switch branch and do make
> * Second deploy of the blueprint without any changes , which would typically not result in any diff
> * Third deploy of the blueprint of the above blueprint after completion of second deploy
> 
> Result Did the third deploy of blueprint cause any updates / loss of data ?

Thanks for adding a detailed description of test case to further consider. Following are the result from the test.
Case 1: No diff observed (except for local file output of `gke-job-template` module, where we are using random_id provider), k8s object like (PV and PVC) and data (filestore and gcs) persisted across the migration, no data loss was observed.
```
Summary of proposed changes: Plan: 6 to add, 0 to change, 2 to destroy.
Running terraform apply on deployment group storage-gke-03/primary
module.ephemeral-storage-job.random_id.resource_name_suffix: Destroying... [id=pgE]
module.shared-fs-job.random_id.resource_name_suffix: Destroying... [id=Zd4]
module.shared-fs-job.random_id.resource_name_suffix: Destruction complete after 0s
module.ephemeral-storage-job.random_id.resource_name_suffix: Destruction complete after 0s
module.ephemeral-storage-job.random_id.resource_name_suffix: Creating...
module.shared-fs-job.random_id.resource_name_suffix: Creating...
module.ephemeral-storage-job.random_id.resource_name_suffix: Creation complete after 0s [id=VEM]
module.shared-fs-job.random_id.resource_name_suffix: Creation complete after 1s [id=kFA]
module.data-bucket-pv.local_file.debug_file: Creating...
module.data-bucket-pv.local_file.debug_file: Creation complete after 0s [id=d69386e3400928e6f3956408ce8f9a4b69360112]
module.shared-filestore-pv.local_file.debug_file: Creating...
module.shared-filestore-pv.local_file.debug_file: Creation complete after 0s [id=4f05710324270aa5b85fc45a937d278b1552086d]
module.shared-fs-job.local_file.job_template: Creating...
module.shared-fs-job.local_file.job_template: Creation complete after 0s [id=242d0b096cea5345338b0602f1d562148ada3df5]
module.ephemeral-storage-job.local_file.job_template: Creating...
module.ephemeral-storage-job.local_file.job_template: Creation complete after 0s [id=76871c6ea3e38c3d7a5e9b21970665f0d4f23511]
```
Case 2: Same behavior observed as defined for Case 1
Case 3: Same behavior observed as defined for Case 1

pkg/config/expand.go
```diff
@@ -186,6 +187,33 @@ func (bp Blueprint) expandBackend(grp *Group) {
...
+			return true, mod
+		}
+	}
+	return false, Module{}
+}
+
+func getModuleKubectlProviders(mod Module) map[string]TerraformProvider {
+	kubectlConf := Dict{}
```
sharabiani, line 200: Some questions:
- Are we considering multiple gke-cluster, pre-existing-gke-cluster blocks in a single deployment group? It's rare but possible. (I was thinking probably we'll have an "alias" setting in the provider block representing the targeted cluster. It also gives us the ability to have multiple blocks of a same provider.)
- Does this work in a fresh deployment that no gke-cluster is provisioned and in consequence we don't have cluster info/credentials while creating the deployment? 
- Have we tried a scenario that we have a network, a multivpc, only a gke-cluster (a3 that applies some manifest inside) and nothing else; deploy the blueprint then remove the gke-cluster?

pkg/config/expand.go
```diff
@@ -186,6 +187,33 @@ func (bp Blueprint) expandBackend(grp *Group) {
...
+	for s, v := range map[string]string{
+		"cluster_ca_certificate": "cluster_ca_certificate",
+		"host":                   "host_endpoint",
+		"token":                  "access_token"} {
+		kubectlConf = kubectlConf.With(s, ModuleRef(mod.ID, v).AsValue())
+	}
+	kubectlConf = kubectlConf.
+		With("apply_retry_count", cty.NumberIntVal(15)).
+		With("load_config_file", cty.BoolVal(false))
```
sharabiani, line 209: Can we avoid hardcoded `load_config_file`, `apply_retry_count` and read it from the internal provider if they exist there?

pkg/config/expand.go
```diff
@@ -186,6 +187,33 @@ func (bp Blueprint) expandBackend(grp *Group) {
...
+	return map[string]TerraformProvider{
+		"kubectl": {
+			Source:        "gavinbunney/kubectl",
+			Version:       ">= 1.7.0",
```
sharabiani, line 213: Same here with avoiding hardcoded `Version`?

mohitchaurasia91: Referred an [already existing root module provider setup](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/pkg/config/expand.go#L190-L207), so not sure if we can avoid this.

pkg/config/expand.go
mohitchaurasia91: These provider config are created in root module on gcluster create/deploy ops, so there seems to be no internal provider which can be referred.

pkg/config/expand.go
mohitchaurasia91: > Some questions:
> 
> * Are we considering multiple gke-cluster, pre-existing-gke-cluster blocks in a single deployment group? It's rare but possible. (I was thinking probably we'll have an "alias" setting in the provider block representing the targeted cluster. It also gives us the ability to have multiple blocks of a same provider.)
> * Does this work in a fresh deployment that no gke-cluster is provisioned and in consequence we don't have cluster info/credentials while creating the deployment?
> * Have we tried a scenario that we have a network, a multivpc, only a gke-cluster (a3 that applies some manifest inside) and nothing else; deploy the blueprint then remove the gke-cluster?

Please find response in line.
- This change don't consider multiple gke-cluster, pre-existing-gke-cluster blocks in single deployment group, for simplicity multi deployment group could help as it aligns well with how current cluster-toolkit been setup.
- In this scenario, we expect to usage of pre-existing-gke-cluster module, in order to build the required cluster info/credentials while creating the deployment.
- Tried this scenario, was not able to remove the gke-cluster, as the kubectl provider depends on the gke-cluster module output. The gke-cluster module has an [internally defined](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/scheduler/gke-cluster/main.tf#L345-L368) kubectl-apply module, so not sure if this scenario is valid in current state. Meanwhile tried gcluster destroy and it worked fine.

pkg/config/expand.go
tpdownes: Re: downside. Provider plugins are not typically small. They are probably Go which is statically linked. The providers directory for the gke-a3-highgpu.yaml example is 377M in size. I also don't have any idea whether there is a guarantee that they will be available forever so I wouldn't add them unless necessary.

I think Terraform makes you add required_providers explicitly for a reason and we're tripping over it. For now, saying "every GKE cluster triggers the requirement for this provider" seems reasonable. In medium-term, I think the modulereader.ModuleInfo might contain this info.

sharabiani, line 192: `strings.Contains(mod.Source, "gke-cluster")` covers both "gke-cluster" and  "pre-existing-gke-cluster".

I would suggest to check full module path ("modules/scheduler/...") to prevent any name similarities later. ex. `gke-cluster-something`

pkg/config/expand.go
sharabiani: - Okay then, thanks. Probably we need to reflect in the doc that we have this single gke-cluster limit per deployment group.

- The below was @nick-stroud 's concern:
> Tried this scenario, was not able to remove the gke-cluster, as the kubectl provider depends on the gke-cluster module output. The gke-cluster module has an [internally defined](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/scheduler/gke-cluster/main.tf#L345-L368) kubectl-apply module, so not sure if this scenario is valid in current state. Meanwhile tried gcluster destroy and it worked fine.

pkg/config/expand.go
sharabiani: Isn't `gke_cluster` in `module.gke_cluster...` the id that we get from blueprint's block? If yes, I think we cannot assume it's name always will be "gke_cluster". Am I missing something?

modules/compute/gke-node-pool/variables.tf
```diff
@@ -24,6 +24,12 @@ variable "cluster_id" {
   type        = string
 }
 
+variable "gke_cluster_exists" {
```
sharabiani, line 27: is this automatically set when we have a gke-cluster or pre-existing-gke-cluster?

pkg/config/expand.go
mohitchaurasia91: so we are referring `module.Source` map to module source path and `module.ID` map to user defined unique module id, adding ref to [struct](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/pkg/config/config.go#L230-L240) which hold these value.

modules/compute/gke-node-pool/variables.tf
mohitchaurasia91: Yup, gke-cluster or pre-existing-gke-cluster both have `gke_cluster_exists` as a module output.

pkg/config/expand.go
mohitchaurasia91: Thanks for pointing this out, change pushed.

tpdownes: There is existing code that solves roughly this problem here: 894520281b60c0b93e856d0df5d2e1d9f20bdd16

sharabiani: So the below settings are dependent on user defined module id, right? My point is, we cannot always add the provider because we cannot assume the id for module is always `gke_cluster`.
```
  cluster_ca_certificate = module.gke_cluster.cluster_ca_certificate
  host                   = module.gke_cluster.host_endpoint
  token                  = module.gke_cluster.access_token
```

sharabiani (commented):
LGTM
if @tpdownes @ankitkinra agree with the limit of only one gke-cluster per deployment group:  https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3406#discussion_r1890432583

pkg/config/expand.go
mohitchaurasia91: Please let me clarify, had shared that as an example of kubectl provider when `gke-cluster` module id is `gke_cluster`, we have a dynamic setup out here. Please refer this [code block](https://github.com/mohitchaurasia91/cluster-toolkit/blob/fix-gke-module-provider/pkg/config/expand.go#L199-L206).
```
cluster_ca_certificate = module.gke_cluster.cluster_ca_certificate
host                   = module.gke_cluster.host_endpoint
token                  = module.gke_cluster.access_token
```

ankitkinra (approved):

PR #3414, opened 2024-12-16
Move a3u blueprints to develop branch

ankitkinra:
Move a3u blueprints along with all required files to develop branch to enable integration testing


tpdownes (approved):
When development on `develop` begins, please move the staged files to a subdirectory. A common point of confusion for customers is "what YAML is a blueprint and what's just YAML?"

ighosh98:
Quick question: Why are we moving away from the older format of having the blueprint in examples folder directly? For instance, the way [a3-highgpu](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/a3ultra-preview/examples/gke-a3-highgpu.yaml) blueprints are organised. Is this a long term refactoring that we would follow for future releases as well?

ankitkinra:
> Quick question: Why are we moving away from the older format of having the blueprint in examples folder directly? For instance, the way [a3-highgpu](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/a3ultra-preview/examples/gke-a3-highgpu.yaml) blueprints are organised. Is this a long term refactoring that we would follow for future releases as well?

Yes we should refactor to allow for saving the nccl tests as well, however we may want to dedup the nccl files in a single folder for GKE so we need to reorganize for sure.

PR #3430, opened 2024-12-19
Update python test `deployment_name` to `BUILD_ID`

alyssa-sm:
This PR updates the python test build files to export the `BUILD_ID` so it can be used as the `deployment_name` for the tests. Currently all tests use the same `deployment_name` causing them to share state.

Manually ran the three tests which all passed.

Example from successful test showing the deployment name was the shortened build id:
```
gcloud storage cp gs://daily-tests-tf-state/f4c987/f4c987.tgz .
PUSH
DONE
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
ID                                    CREATE_TIME                DURATION  SOURCE                                                                                         IMAGES  STATUS
f4c98735-27e7-4645-8b9c-93cd078df94a  2024-12-19T03:41:37+00:00  8M5S      gs://hpc-toolkit-dev_cloudbuild/source/1734579691.841393-9e096ec8fdae4f80a8cca91a1a7fa669.tgz  -       SUCCESS
```

### DIFF
--- a/tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
+++ b/tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
@@ -31,4 +31,6 @@ steps:
   - |
     set -x -e
     cd /workspace && make
+    export BUILD_ID="${BUILD_ID}"
+
     python3 tools/python-integration-tests/slurm_reconfig_size.py
--- a/tools/python-integration-tests/deployment.py
+++ b/tools/python-integration-tests/deployment.py
@@ -17,6 +17,7 @@ import shutil
 import os
 import subprocess
 import yaml
+import uuid
 
 class Deployment:
     def __init__(self, blueprint: str):
@@ -37,7 +38,6 @@ class Deployment:
     def parse_blueprint(self, file_path: str):
         with open(file_path, 'r') as file:
             content = yaml.safe_load(file)
-        self.deployment_name = content["vars"]["deployment_name"]
         self.zone = content["vars"]["zone"]
 
     def get_posixAccount_info(self):
@@ -50,8 +50,16 @@ class Deployment:
                 self.project_id = account['accountId']
                 self.username = account['username']
 
+    def generate_uniq_deployment_name(self):
+        BUILD_ID = os.environ.get('BUILD_ID')
+        if BUILD_ID:
+            return BUILD_ID[:6]
+        else:
+            return str(uuid.uuid4())[:6]
+
     def set_deployment_variables(self):
         self.workspace = os.path.abspath(os.getcwd().strip())
+        self.deployment_name = self.generate_uniq_deployment_name()
         self.parse_blueprint(self.blueprint_yaml)
         self.get_posixAccount_info()
         self.instance_name = self.deployment_name.replace("-", "")[:10] + "-slurm-login-001"
--- a/tools/cloud-build/daily-tests/builds/slurm-gcp-v6-simple-job-completion.yaml
+++ b/tools/cloud-build/daily-tests/builds/slurm-gcp-v6-simple-job-completion.yaml
@@ -31,4 +31,6 @@ steps:
   - |
     set -x -e
     cd /workspace && make
+    export BUILD_ID="${BUILD_ID}"
+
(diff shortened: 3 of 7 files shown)


tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
```diff
@@ -31,4 +31,8 @@ steps:
   - |
     set -x -e
     cd /workspace && make
+    BUILD_ID_FULL=$BUILD_ID
```
mr0re1, line 34: Let's not duplicate it, shortening can be done by python in one place.

tools/python-integration-tests/deployment.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, blueprint: str):
         self.workspace = None
         self.instance_name = None
         self.username = None
-        self.deployment_name = None
+        self.deployment_name = os.environ.get('BUILD_ID_SHORT')
```
mr0re1, line 29: We need to account for non-CloudBuild invocation. Let's default to short uuid if `BUILD_ID` is not present.

alyssa-sm: Done

tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
alyssa-sm: Moved shortening to `deployment.py`

mr0re1: Can we to not add those lines at all, isn't `BUILD_ID` already available to python executable?

tools/python-integration-tests/deployment.py
```diff
@@ -50,8 +50,16 @@ def get_posixAccount_info(self):
...
+    def set_deployment_name(self):
+        BUILD_ID = os.environ.get('BUILD_ID_FULL')
+        if BUILD_ID:
+            self.deployment_name = BUILD_ID[:6]
+        else:
+            self.deployment_name = str(uuid.uuid4())[:6]
+
     def set_deployment_variables(self):
         self.workspace = os.path.abspath(os.getcwd().strip())
+        self.deployment = self.set_deployment_name()
```
mr0re1, line 62: This is very strange chain of execution., why do you override `deployment` with `None`? Either:

Recommended:
```py
self.deployment_name = generate_uniq_deployment_name()
```

OR (unrecommended)

```py
self.set_deployment_name()
```

tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
alyssa-sm: I tried after removing all the lines, but BUILD_ID is not one of the environment variables. I checked by printing with `printenv` and didn't see it listed. Is there an alternative way to call the variable other than `os.environ.get()`?

tools/python-integration-tests/deployment.py
alyssa-sm: Ah yes this is very strange. I had multiple path in my head when I wrote this. Cleaned up according to recommended.

tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
```diff
@@ -31,4 +31,6 @@ steps:
   - |
     set -x -e
     cd /workspace && make
+    export BUILD_ID="${BUILD_ID}"
```
mr0re1, line 34: Do you still need to explicitly export it? AFAIK `BUILD_ID` is already set as environment variable. 
If it's not, land it, otherwise please remove re-exporting.

mr0re1 (approved):

tools/cloud-build/daily-tests/builds/slurm-gcp-v6-reconfig-size.yaml
alyssa-sm: Yes, I needed to explicitly export. Without the export, it isn't an accessible environment variable

PR #3439, opened 2024-12-19
Add compact placement validations

parulbajaj01:
Compact placement in GKE Standard node pools has certain limitations.
Added validations for 2nd and 5th limitation mentioned here: https://cloud.google.com/kubernetes-engine/docs/how-to/compact-placement#limitations-standard
3rd and 4th limitation can be changed over time, so not adding any validations for them.


ighosh98 (commented):
Add a description to provide proper context on the fix. Please link the relevant documentation that covers the rules

parulbajaj01:
> Add a description to provide proper context on the fix. Please link the relevant documentation that covers the rules
Updated now

modules/compute/gke-node-pool/main.tf
```diff
@@ -307,6 +307,14 @@ resource "google_container_node_pool" "node_pool" {
       condition     = local.upgrade_settings.max_unavailable > 0 || local.upgrade_settings.max_surge > 0
       error_message = "At least one of max_unavailable or max_surge must greater than 0"
     }
+    precondition {
+      condition     = var.placement_policy.type != "COMPACT" || length(var.zones) == 1
```
ighosh98, line 311: Add an explanation on how you arrived at this for the zones check as its not intuitive when one reads the code. You could add it as a reply to the comment or update the description.

parulbajaj01: Single zone validation: Added 2 conditions separated by an OR -> if the first one evaluates to TRUE, meaning type is not COMPACT, than we don't need to check the second condition of whether it's in a single zone or not.
But if it evaluates to false, meaning placement policy type is COMPACT, then we need to check the second condition also that it should be in single zone only, otherwise it will throw an error.

Blue-green upgrade validation: Similarly here also, added 2 conditions separated by an OR -> if the first one evaluates to TRUE, meaning type is not COMPACT, than we don't need to check the second condition whether it's a  BLUE_GREEN upgrade strategy. But if it evaluates to false, meaning placement policy type is COMPACT, then second condition should hold TRUE, upgrade strategy should not be BLUE_GREEN. Otherwise it will throw error.

ighosh98 (approved):

ankitkinra (approved):

PR #3448, opened 2024-12-20
Replace duplicated `gpu_definition.tf` by shared internal module

mr0re1:
Add module `modules/internal/gpu-definition` to handle default values for GPUs list.

NOTE: keep ones in `schedmd-slurm-gcp-v5*` due to soon removal.


tpdownes (approved):
I would like to review whether use of `null` (vs empty list) would help us avoid the need for this file, but I agree that a simple internal module is a better approach than code duplication.

PR #3452, opened 2024-12-20
Add integration test for A3U / A3U blueprint updates

cdunbar13:
This integration test has been tested on a cluster of one node and was successful.  The blueprint needed to be updated for the integration test to work (opening internal traffic using a firewall).


tpdownes (approved):
We have not yet deployed the test to run against develop branch nightly. We reviewed an internal test build and observed success but with a failure of the image deletion step. Documenting that as a next step to fix as the nightly test has large value.

PR #3469, opened 2024-12-27
Pin A3U blueprint module with commit ref from develop branch

mohitchaurasia91:
#### What changed ?
- Pinned A3U blueprint module to [e17bb15](https://github.com/GoogleCloudPlatform/cluster-toolkit/commit/e17bb153b36984a6dae6df9496928408f4db6a8e) commit hash from develop branch.

#### Why we require this change ?
- This change is required for A3U GA readiness.
- Prior plan of using `toolkit_modules_url` and `toolkit_modules_version` stalled, as we identified that we have no support for using commit hash (only works for branch and tag name).
- Its due to the hardcoding of git `depth` field in the current setup, while [expanding source path for each module](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/pkg/config/expand.go#L116-L121).

#### Why not pinned to latest [81e63e6](https://github.com/GoogleCloudPlatform/cluster-toolkit/commit/81e63e6be3defa2eab22988ebb649c0ea46ba80c) commit hash from develop branch?
- As it contain changes related to `gcluster` binary, which wont be available on main until next release.

#### How this has been tested ?
- Tested locally, using example A3U blueprint with diff, worked as expected.

### Submission Checklist

### DIFF
--- a/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
+++ b/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
@@ -33,25 +33,42 @@ vars:
   system_node_pool_disk_size_gb: 200
   a3ultra_node_pool_disk_size_gb: 100
 
+terraform_providers:
+  google:
+    source: hashicorp/google
+    version: 6.13.0
+    configuration:
+      project: $(vars.project_id)
+      region: $(vars.region)
+      zone: $(vars.zone)
+  google-beta:
+    source: hashicorp/google-beta
+    version: 6.13.0
+    configuration:
+      project: $(vars.project_id)
+      region: $(vars.region)
+      zone: $(vars.zone)
+
 deployment_groups:
 - group: primary
   modules:
   - id: gke-a3-ultra-net-0
-    source: github.com/GoogleCloudPlatform/cluster-toolkit.git//modules/network/vpc?ref=e0c690b
+    source: github.com/GoogleCloudPlatform/cluster-toolkit.git//modules/network/vpc?ref=e17bb15
     settings:
-      network_name: gke-a3-ultra-net-0
+      network_name: $(vars.deployment_name)-net-0
       subnetworks:
-      - subnet_name: gke-a3-ultra-sub-0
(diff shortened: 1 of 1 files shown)


examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
```diff
@@ -80,20 +97,20 @@ deployment_groups:
         - protocol: icmp
 
   - id: gke-a3-ultra-rdma-net
-    source: github.com/GoogleCloudPlatform/cluster-toolkit.git//community/modules/network/rdma-vpc?ref=98c49fe
+    source: github.com/GoogleCloudPlatform/cluster-toolkit.git//modules/network/gpu-rdma-vpc?ref=e17bb15
```
ighosh98, line 100: Doubt: Are we sure this is the final version we want to pin the ref to. Also, why do we want to pin this to a certain ref ?

ighosh98 (commented):
Can we discuss the possibility of adding the mentioned commit to the release and moving the commit id forward or removing the ref entirely? I also foresee other changes coming in. Alternatively, we could change the pinning ref later.

examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
mohitchaurasia91: We want to point in to the latest valid commit from develop branch. If we expect to have move change to be added in, then this should be updated further.

mohitchaurasia91:
> Discussed offline. LGTM.
> 
> Can you please add the explanation why we went ahead with the ref solution over leveraging the alternate solution before merging the PR?
> 
> ```
> toolkit_modules_url: github.com/GoogleCloudPlatform/cluster-toolkit
> toolkit_modules_version: COMMIT_HASH_FROM_DEVELOP
> ```

Thanks, added an explanation to the PR description.

annuay-google (approved):

ankitkinra:
Merging on mohit's behalf

PR #3476, opened 2024-12-30
add GKE support for managed hyperdisk

chengcongdu:
add GKE hyperdisk support to the gke-storage module to handle the creation of storage class and persistent volume claim.

The new module output persistent_volume_claims which can be consumed by the gke-job-template module.

cloud build: https://pantheon.corp.google.com/cloud-build/builds/294767c6-8972-4dda-a9cf-eb41107ce9de?project=hpc-toolkit-dev&e=GkeGettingStartedCloudShellSolutionsLaunch::GkeGettingStartedCloudShellSolutionsEnabled,GkeOnboardingOverviewTabLaunch::GkeOnboardingOverviewTabEnabled&mods=monitoring_api_prod

### DIFF
--- a/examples/gke-managed-hyperdisk.yaml
+++ b/examples/gke-managed-hyperdisk.yaml
@@ -1,0 +1,218 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+blueprint_name: gke-storage-hyperdisk
+vars:
+  project_id:  ## Set GCP Project ID Here ##
+  deployment_name: gke-storage-hyperdisk
+  region: us-central1
+  zone: us-central1-c
+
+  # Cidr block containing the IP of the machine calling terraform.
+  # The following line must be updated for this example to work.
+  authorized_cidr: <your-ip-address>/32
+
+deployment_groups:
+- group: primary
+  modules:
+  - id: network
+    source: modules/network/vpc
+    settings:
+      subnetwork_name: gke-subnet-hyperdisk
+      secondary_ranges:
+        gke-subnet-hyperdisk:
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network]
+    settings:
+      enable_persistent_disk_csi: true # enable Hyperdisk for the cluster
+      configure_workload_identity_sa: true
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
+      master_authorized_networks:
+      - display_name: deployment-machine
+        cidr_block: $(vars.authorized_cidr)
+    outputs: [instructions]
+
+  ### Set up storage class and persistent volume claim for Hyperdisk ###
+  - id: hyperdisk-balanced-setup
+    source: modules/file-system/gke-storage
+    use: [gke_cluster]
+    settings:
+      storage_type: Hyperdisk-balanced
+      access_mode: ReadWriteOnce
+      sc_volume_binding_mode: Immediate
+      sc_reclaim_policy: Delete
+      sc_topology_zones: [$(vars.zone)]
+      pvc_count: 1
+      capacity_gb: 100
+
+  - id: hyperdisk-throughput-setup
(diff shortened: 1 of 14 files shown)


examples/gke-managed-hyperdisk.yaml
```diff
@@ -0,0 +1,218 @@
...
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network]
+    settings:
+      enable_persistent_disk_csi: true # enable Hyperdisk for the cluster
```
ankitkinra, line 44: Is this correct comment ? seems like we are enabling persistent disk csi ?

examples/gke-managed-hyperdisk.yaml
```diff
@@ -0,0 +1,218 @@
...
+        value: 1000
+      - key: runAsGroup
+        value: 100
+      - key: fsGroup
+        value: 100
+      command:
+      - bash
+      - -c
+      - |
+        pip install transformers datasets
```
ankitkinra, line 118: Should we create a virtual environment for these python dependencies ?

examples/gke-managed-hyperdisk.yaml
```diff
@@ -0,0 +1,218 @@
...
```
ankitkinra, line 159: ditto ask as above here

examples/gke-managed-hyperdisk.yaml
```diff
@@ -0,0 +1,218 @@
+# Copyright 2024 Google LLC
```
ankitkinra, line 1: Meta: As this is a fairly complicated blueprint, can we do the following to potentially make this easier for the user to digest:
* Move all the params that the user is expected to change in the vars section, so things like node count, disk space etc. 
* Add to README how to run the multiple training jobs and what is the expected sample output.

examples/gke-managed-hyperdisk.yaml
```diff
@@ -0,0 +1,218 @@
...
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  - id: gke_cluster
+    source: modules/scheduler/gke-cluster
+    use: [network]
+    settings:
+      enable_persistent_disk_csi: true # enable Hyperdisk for the cluster
+      configure_workload_identity_sa: true
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
```
ankitkinra, line 46: Should we add the RAPID channel / maintenance exclusion similar to a3u blueprints?

chengcongdu: RAPID and MaintenanceExclusion is not a requirement for using Hyperdisk/Parallelstore. But I can add it if you think it is for consistent sake?

examples/gke-managed-hyperdisk.yaml
chengcongdu: This is the same blueprint for Parallelstore, all use need to do is to apply the job manifest as instructed by the output instruction.

examples/gke-managed-hyperdisk.yaml
chengcongdu: This is a one off environment and I don't see a need for specifically set up the separate environment for it. 
It is following the tensorflow example provided here https://cloud.google.com/parallelstore/docs/tensorflow-sample

examples/gke-managed-hyperdisk.yaml
chengcongdu: yes, PD and HyperDisk use the same csi

examples/gke-managed-hyperdisk.yaml
ankitkinra: My concern , cluster toolkit output is not durable. Do you think this is intended more as a POC than a durable environment where GKE jobs template we created also serve as samples to the customer ?

examples/gke-managed-hyperdisk.yaml
ankitkinra: If this is intended as a POC then maybe not required, however for a system which we expect customers to run for longer duration we should have those in there to make sure there is no unncessary disruption .

examples/gke-managed-hyperdisk.yaml
ankitkinra: I am not sure if this is a one-off environment, it would be the same environment as customers run their other cluster toolkit workflows. Is it not ?

examples/gke-managed-hyperdisk.yaml
chengcongdu: added RAPID and MaintenanceExclusion

examples/gke-managed-hyperdisk.yaml
ankitkinra: Hmm, I am not sure, perhaps we can take this offline. 

As cluster toolkit we want to make sure we try to minimize the dependencies we are asking customers to install or make it super easy for them.

examples/gke-managed-hyperdisk.yaml
ankitkinra: Ack, let's take this offline.

ankitkinra (approved):

PR #3477, opened 2024-12-30
Add NeMo framework example to a3-Ultra

akiki-liang0:
- Added appropriate environment variables, container mounts, and example command for running NeMo on a3ultra
- ensure NCCL_NET=gIB is being used

Testing:
- 10 consecutive successful runs using example command in README.md


tpdownes (approved):
LGTM. Thanks for adding testing information.

PR #3487, opened 2025-01-02
Add support for Redhat 7, 8 and 9 to startup-script Anisble

wiktorn:
Add support to startup-script for RedHat 7, 8 and 9 and Rocky Linux 9.

### DIFF
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -76,16 +76,28 @@ get_python_minor_version() {
 # newly installed packaged.
 install_python3_yum() {
 	major_version=$(rpm -E "%{rhel}")
-	enable_repo=""
-	if [ "${major_version}" -eq "7" ]; then
-		enable_repo="base,epel"
+	set -- "--disablerepo=*" "--enablerepo=baseos,appstream"
+
+	if grep -qi 'ID="rhel"' /etc/os-release && {
+		[ "${major_version}" -eq "7" ] || [ "${major_version}" -eq "8" ] ||
+			[ "${major_version}" -eq "9" ]
+	}; then
+		# Do not set --disablerepo / --enablerepo on RedHat, due to complex repo names
+		# clear array
+		set --
+	elif [ "${major_version}" -eq "7" ]; then
+		set -- "--disablerepo=*" "--enablerepo=base,epel"
 	elif [ "${major_version}" -eq "8" ]; then
-		enable_repo="baseos"
+		# use defaults
+		true
+	elif [ "${major_version}" -eq "9" ]; then
+		# use defaults
+		true
 	else
 		echo "Unsupported version of centos/RHEL/Rocky"
 		return 1
 	fi
-	yum install --disablerepo="*" --enablerepo="${enable_repo}" -y python3 python3-pip python3-venv
+	yum install "$@" -y python3 python3-pip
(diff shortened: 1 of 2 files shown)


examples/image-custom-slurm.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+---
+
+# See instructions at
+# https://github.com/GoogleCloudPlatform/hpc-toolkit/tree/main/examples#image-custom-slurmyaml-
+
+blueprint_name: image-custom-slurm
+
+vars:
+  project_id: wns-gsk-hpc # ## Set GCP Project ID Here ##
```
ankitkinra, line 23: Should we leave this blank ? So customers have to populate it ?

examples/image-custom-slurm.yaml
ankitkinra, line 1: Should we move this to community folder ?

examples/image-custom-slurm.yaml
wiktorn: Good catch, thanks, will change that.

examples/image-custom-slurm.yaml
wiktorn: I just noticed `community/examples/hpc-build-slurm-image.yaml` - I'll adapt this image instead.

There are also other examples, that could benefit from splitting the build into two packer steps, such as:
* examples/machine-learning/a3-highgpu-8g/v5-legacy/ml-slurm-a3-1-image-v5-legacy.yaml
* examples/machine-learning/a3-highgpu-8g/ml-slurm-a3-1-image.yaml
* examples/machine-learning/a3-megagpu-8g/slurm-a3mega-image.yaml

Should I change those too?

RachaelSTamakloe:
@wiktorn I notice that in the pr description you mention adding an example that shows how to build custom Slurm GCP image. But the mentioned example has been removed from this PR. Could you update the PR description?

modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -76,16 +76,28 @@ get_python_minor_version() {
...
+		true
+	elif [ "${major_version}" -eq "9" ]; then
+		# use defaults
+		true
 	else
 		echo "Unsupported version of centos/RHEL/Rocky"
 		return 1
 	fi
-	yum install --disablerepo="*" --enablerepo="${enable_repo}" -y python3 python3-pip python3-venv
+	yum install "$@" -y python3 python3-pip
```
RachaelSTamakloe, line 100: is there a reason for omitting the installation of `python3-venv`?

wiktorn: `python3-venv` package is not available on Redhat / Rocky 9, and it looks like it was not necessary anyway, as there are no failures creating virtual environment. So I saw two solutions:
1. Adapt list of installed packages depending on the version
2. Remove installation of `python3-venv`

I choose the latter, as makes code simpler.

RachaelSTamakloe: okay, understood. With Python 3.3 and later, the `venv` module is included in the default Python standard library so that's why creating virtual environment is successful without the explicit installation.

RachaelSTamakloe (approved):

wiktorn:
Hey @RachaelSTamakloe , I do not have permissions to merge this in myself. Can you do it for me? PR description is already updated.

PR #3489, opened 2025-01-03
update README for managed-hyperdiska nd managed-parallelstore example blueprint

chengcongdu:
add instruction on how to deploy sample workload in README for managed-hyperdiska nd managed-parallelstore


ankitkinra (approved):
Thank you!

PR #3506, opened 2025-01-07
Add blueprint name before build ID

alyssa-sm:
This PR updates the `deployment_name` to have the `blueprint_name` appended to the beginning. This solves the issue of the `deployment_name` not being recognized as an string when the build id starts with an int.


mr0re1 (approved):

mr0re1:
Please run affected tests before landing

PR #3509, opened 2025-01-08
Add module providing mount scripts for WEKA filesystems

wiktorn:
Support for mounting WEKA filesystems.

This module doesn't support creating WEKA storage, as this can be done by directly calling [weka-gcp](https://github.com/weka/terraform-gcp-weka/) module.

Because WEKA require gathering environment information from Metadata server during mount it's not possible to create a fstab entry that will work on following mounts. Hence, the mount is done using SystemD unit.

Things for consideration / cleanup:
* [ ] whether to keep  `network_storage`, `client_install_runner` and  `mount_runner`  outputs, or shall those be dropped, to prevent accidental use
* [ ] there is quite a few `shellcheck disable`, because shellcheck doesn't understand terraform templates, to be decided should shellcheck investigate templates as now, or whether add exclusion of template files
* [ ] ~~should this PR include a blueprint that provisions the WEKA cluster and uses this module~~

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -105,6 +105,8 @@ Modules that are still in development and less stable are labeled with the
   persistent volumes and persistent volume claims for shared storage.
 * **[nfs-server]** ![community-badge] ![experimental-badge] : Creates a VM and
   configures an NFS server that can be mounted by other VM.
+* **[weka-client]** ![community-badge] ![experimental-badge] : Installs client
+  and mounts [WEKA](https://www.weka.io/) filesystems.
 
 [filestore]: file-system/filestore/README.md
 [parallelstore]: file-system/parallelstore/README.md
@@ -114,6 +116,7 @@ Modules that are still in development and less stable are labeled with the
 [nfs-server]: ../community/modules/file-system/nfs-server/README.md
 [cloud-storage-bucket]: ../community/modules/file-system/cloud-storage-bucket/README.md
 [gke-persistent-volume]: ../modules/file-system/gke-persistent-volume/README.md
+[weka-client]: ../community/modules/file-system/weka-client/README.md
 
 ### Monitoring
 


community/modules/file-system/weka-client/outputs.tf
```diff
@@ -0,0 +1,87 @@
...
+
+#output "network_storage" {
+#  description = "Describes a remote network storage to be mounted by fs-tab."
+#  value = {
+#    server_ip             = var.server_ip
+#    remote_mount          = var.remote_mount
+#    local_mount           = var.local_mount
+#    fs_type               = var.fs_type
+#    mount_options         = var.mount_options
+#    client_install_runner = local.client_install_runner
```
cboneti, line 58: Here I am wondering, have you tried to define only the fs_type, client_install_runner and the mount runner, but leave the other variables blank? I am wondering if that could work. That would have as a side effect that almost all toolkit modules would "just work".

community/modules/file-system/weka-client/outputs.tf
```diff
@@ -0,0 +1,87 @@
...
+#    mount_runner          = local.mount_runner
+#  }
+#}
+#
+#output "client_install_runner" {
```
cboneti: I am wondering why you left these commented. In principle, could you not have all the runners below run in the client install (with the systemd service set as disabled), and have the "mount_runner" simply enable the service?

community/modules/file-system/weka-client/outputs.tf
wiktorn: All `network_storage` entries are processed during [setup by this loop](community/modules/file-system/nfs-server/metadata.yaml). The problem with WEKA is, that mount options are specific to the network configuration of the instance when using DPDK (which provides better performance).

`client_install_runner` and `mount_runner` are ignored by Slurm.

From what I can tell, `mount_runner` is also executed only once in a lifetime of VM, so it should follow the same logic to create a SystemD unit. 

So let me work on playbooks to install and mount WEKA for other compute solutions than Slurm.

community/modules/file-system/weka-client/outputs.tf
wiktorn: Given how `client_install_runner` and `mount_runner` are used so far they are anyway just executed one after another, so division of the tasks between those two can be arbitrary. I'll do that when providing support for other than Slurm modules, though it won't solve the issue for Slurm.

cboneti: It is somewhat arbitrary, but it does add clarity to know what installs and what mounts. 
Furthermore, our pattern is that the client_install_runner can be called during image creation, whereas the mount runner could be called during vm startup (making it easy for the same image to support multiple file systems).

community/modules/file-system/weka-client/outputs.tf
cboneti: network storage is used in multiple places (see https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/docs/network_storage.md. -- which is clearly outdated -- a bug was created to update it).  
Slurm handles storage as somewhat of a special case, which we considered removing many many times in the past.  The more unified way is what VM [instances](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/compute/vm-instance/startup_from_network_storage.tf), and [batch](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/scheduler/batch-job-template/startup_from_network_storage.tf) use. They automatically inject the client_install_runner and the mount_runner fields of the network storage if the file system type is not "natively supported", which is the case of Weka.  Native support is defined by file systems that are handled as special cases (nfs and gcs in batch, for instance), nfs, lustre, gcs in the case of Slurm. 

While my comment here is intended more like a guidance, not implementing compatibility with vm instance could be leaving value on the table.

wiktorn: Added `client_install_runner` and `mount_runner`, but I'm keeping `network_storage` commented out, just added a comment above with rationale

wiktorn:
Thanks @cboneti , updated the README.md.

As for the testing, I tested this module with a Slurm cluster on login nodes and compute nodes, as well as with compute-vm module. I did install a WEKA cluster  version `4.4.2.113`. 

Tests were done using `c2-standard-8` instance, with 3 VIRTIO NICs  dedicated to WEKA, and 3 cores.  I did some simple smoke performance tests and confirmed, that I get 1GB/s performance using single `dd` with 1GB blocksize.

SwarnaBharathiMantena (approved):

PR #3513, opened 2025-01-09
Enable hierarchical namespace support in cloud-storage-bucket module

SwarnaBharathiMantena:
Enable hierarchical namespace support in cloud-storage-bucket module. 
Boolean variable "enable_hierarchical_namespace" with default 'false', that links to this setting in the Terraform resource.

### Submission Checklist

### DIFF
--- a/community/modules/file-system/cloud-storage-bucket/README.md
+++ b/community/modules/file-system/cloud-storage-bucket/README.md
@@ -135,6 +135,7 @@ No modules.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_deployment_name"></a> [deployment\_name](#input\_deployment\_name) | Name of the HPC deployment; used as part of name of the GCS bucket. | `string` | n/a | yes |
+| <a name="input_enable_hierarchical_namespace"></a> [enable\_hierarchical\_namespace](#input\_enable\_hierarchical\_namespace) | If true, enables hierarchical namespace for the bucket. | `bool` | `false` | no |
 | <a name="input_force_destroy"></a> [force\_destroy](#input\_force\_destroy) | If true will destroy bucket with all objects stored within. | `bool` | `false` | no |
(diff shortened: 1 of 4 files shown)


samskillman (approved):

tpdownes (changes requested):
The documentation says this option must be enabled at creation?

https://cloud.google.com/blog/products/storage-data-transfer/understanding-new-cloud-storage-hierarchical-namespace?e=0

Can you document the UX for the following steps:

1. Create a bucket _without_ HNS?
2. Add some objects to it manually (outside terraform)
3. Using terraform, enable HNS on the existing bucket

My hope is that, either the API fails entirely with a clear-ish error message or, at least, the bucket destruction fails because there are existing objects in the bucket that Terraform doesn't know about.

Additionally add some short text to the variable description saying that the option must be set at first creation of the bucket.

SwarnaBharathiMantena:
The UX is as follows.
Steps:
1. Create a bucket without HNS?
2. Add some objects to it manually (outside terraform)
3. Using terraform, enable HNS on the existing bucket

On step 3, the existing bucket _without_ HNS got destroyed and was replaced by a new bucket with HNS enabled. The document uploaded on step 2 is lost. No exclusive warning or error observed.

tpdownes (approved):
I have some concerns about the lifecycle of changing this settings, but let's document and prioritize that as follow-up work.

tpdownes:
> I have some concerns about the lifecycle of changing this settings, but let's document and prioritize that as follow-up work.

Some further examination suggests that force_destroy was set to true during this test, so the observed behavior is expected. We will follow up to confirm that objects outside Terraform are not destroyed if force_destroy is set to false (default value).

SwarnaBharathiMantena:
Tested out the UX when the force_destroy is set to false, and it gave out an error.

Error: exit status 1
Error: Error trying to delete bucket bucket-without-hns-ux-hns-ux-testing-2-e9644b45 containing objects without force_destroy set to true

PR #3520, opened 2025-01-09
Use version prefix in conjunction with release channels

annuay-google:
Use min master version in conjunction with Rapid release channel
- Pick the latest version on Rapid release channel with prefix 1.31
- Maintenance exclusion and auto upgrade stays as it is today

Testing
- Verified that the correct version (latest from Rapid) is picked by the cluster
- Verified nodes are running the above latest version

### Submission Checklist

### DIFF
--- a/modules/scheduler/gke-cluster/main.tf
+++ b/modules/scheduler/gke-cluster/main.tf
@@ -68,6 +68,16 @@ data "google_project" "project" {
   project_id = var.project_id
 }
 
+data "google_container_engine_versions" "version_prefix_filter" {
+  provider       = google-beta
+  location       = var.cluster_availability_type == "ZONAL" ? var.zone : var.region
+  version_prefix = var.version_prefix
+}
+
+locals {
+  master_version = var.min_master_version != null ? var.min_master_version : data.google_container_engine_versions.version_prefix_filter.latest_master_version
+}
+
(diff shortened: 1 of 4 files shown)


modules/scheduler/gke-cluster/main.tf
```diff
@@ -68,6 +68,16 @@ data "google_project" "project" {
   project_id = var.project_id
 }
 
+data "google_container_engine_versions" "version_prefix_filter" {
+  provider       = google-beta
```
ankitkinra, line 72: In the next NPI preview, would we ever need to update this to private provider ?

modules/scheduler/gke-cluster/variables.tf
```diff
@@ -91,6 +91,12 @@ variable "min_master_version" {
   default     = null
 }
 
+variable "version_prefix" {
+  description = "If provided, Terraform will only return versions that match the string prefix. For example, `1.31.` will match all `1.31` series releases. Since this is just a string match, it's recommended that you append a `.` after minor versions to ensure that prefixes such as `1.3` don't match versions like `1.30.1-gke.10` accidentally."
+  type        = string
+  default     = "1.31."
```
chengcongdu, line 97: I don't think we should have a default value for this, GKE is continually upgrading and it has its own default version when customer not specifying

PR #3528, opened 2025-01-13
Fixing validation block error in Filestore module

RachaelSTamakloe:
a validation block introduced in this [pr](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3418/files) causes an error during `terraform init` as a result of referencing another variable. In this PR, the validation block is replaced with a lifecycle precondition block.  

error:
```
'terraform init' failed, 'terraform validate' skipped: modules/file-system/filestore
Terraform encountered problems during initialisation, including problems
with the configuration, described below.

The Terraform configuration must be valid before initialization so that
Terraform can determine which modules and providers need to be installed.
╷
│ Error: Invalid reference in variable validation
│ 
│   on variables.tf line 100, in variable "filestore_tier":
│  100:     condition     = !(var.protocol == "NFS_V4_1" && !contains(["HIGH_SCALE_SSD", "ZONAL", "REGIONAL", "ENTERPRISE"], var.filestore_tier))
│ 
│ The condition for variable "filestore_tier" can only refer to the variable
│ itself, using var.filestore_tier.
╵
```
### Submission Checklist

### DIFF
--- a/modules/file-system/filestore/main.tf
+++ b/modules/file-system/filestore/main.tf
@@ -104,5 +104,10 @@ resource "google_filestore_instance" "filestore_instance" {
         BASIC_HDD or BASIC_SSD tiers. Otherwise the range size must be 24.
         EOT
     }
+
+    precondition {
+      condition     = !startswith(var.filestore_tier, "BASIC") || var.protocol != "NFS_V4_1"
+      error_message = "NFS_V4_1 is not supported on BASIC Filestore tiers."
+    }
   }
 }
--- a/modules/file-system/filestore/variables.tf
+++ b/modules/file-system/filestore/variables.tf
@@ -92,6 +92,7 @@ variable "filestore_tier" {
       "BASIC_SSD",
       "HIGH_SCALE_SSD",
       "ZONAL",
+      "REGIONAL",
       "ENTERPRISE"
     ], var.filestore_tier)
     error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','ZONAL','REGIONAL'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1/Tier."
--- a/modules/file-system/filestore/variables.tf
+++ b/modules/file-system/filestore/variables.tf
@@ -94,7 +94,7 @@ variable "filestore_tier" {
       "ZONAL",
       "ENTERPRISE"
     ], var.filestore_tier)
-    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','ZONAL','REGIONAL'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier."
+    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','ZONAL','REGIONAL'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1/Tier."
   }
 }
 
--- a/modules/file-system/filestore/variables.tf
+++ b/modules/file-system/filestore/variables.tf
@@ -94,11 +94,7 @@ variable "filestore_tier" {
       "ZONAL",
       "ENTERPRISE"
     ], var.filestore_tier)
-    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','HIGH_SCALE_SSD','ZONAL','ENTERPRISE'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier."
-  }
-  validation {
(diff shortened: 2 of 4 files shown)


modules/file-system/filestore/main.tf
```diff
@@ -104,5 +104,10 @@ resource "google_filestore_instance" "filestore_instance" {
         BASIC_HDD or BASIC_SSD tiers. Otherwise the range size must be 24.
         EOT
     }
+
+    precondition {
+      condition     = var.protocol != "NFS_V4_1" || contains(["HIGH_SCALE_SSD", "ZONAL", "REGIONAL", "ENTERPRISE"], var.filestore_tier)
```
tpdownes, line 109: This should fail open rather than closed in case a new tier is created. We do not know the future, and new tiers may exist that support V4.1. Replace this condition instead with what we actually know: NFS v4.1 fails on BASIC tiers.

Rewrite this as a match against tier name beginning with "BASIC" and a similarly wored error message.

modules/file-system/filestore/variables.tf
```diff
@@ -96,10 +96,6 @@ variable "filestore_tier" {
     ], var.filestore_tier)
     error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','HIGH_SCALE_SSD','ZONAL','ENTERPRISE'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier."
```
tpdownes, line 97: Remove reference to the legacy HIGH_SCALE_SSD tier.

https://cloud.google.com/filestore/docs/service-tiers#legacy-tiers

tpdownes (changes requested):
Please create a follow-up task for the next sprint to remove legacy tiers from our documentation but allow them to to continue be explicitly called when desired by the user:

https://cloud.google.com/filestore/docs/service-tiers#should-you-use-legacy

modules/file-system/filestore/main.tf
```diff
@@ -104,5 +104,10 @@ resource "google_filestore_instance" "filestore_instance" {
         BASIC_HDD or BASIC_SSD tiers. Otherwise the range size must be 24.
         EOT
     }
+
+    precondition {
+      condition     = substr(var.filestore_tier, 0, 5) != "BASIC" || var.protocol != "NFS_V4_1"
```
tpdownes, line 109:
```suggestion
      condition     = !startswith(var.filestore_tier, "BASIC") || var.protocol != "NFS_V4_1"
```

tpdownes: 1. Double check my logic, please. :)
2. Bump required_version in versions.tf to `>= 1.3.0`

modules/file-system/filestore/variables.tf
```diff
@@ -90,15 +90,10 @@ variable "filestore_tier" {
     condition = contains([
       "BASIC_HDD",
       "BASIC_SSD",
-      "HIGH_SCALE_SSD",
```
tpdownes, line 93: Please keep HIGH_SCALE_SSD and ENTERPRISE in the `contains` list argument, but leave the error_message as it is. Logic: make the new recommendations obvious, but allow current legacy users to function when they have made an explicit choice.

Leave updated values in blueprint alone, too.

Create a follow-up task to add a `check` block that warns against use of legacy tiers ENTERPRISE and HIGH_SCALE_SSD.

tpdownes (changes requested):

modules/file-system/filestore/variables.tf
```diff
@@ -94,11 +94,7 @@ variable "filestore_tier" {
       "ZONAL",
       "ENTERPRISE"
```
tpdownes, line 95:
```suggestion
      "REGIONAL",
      "ENTERPRISE"
```

modules/file-system/filestore/variables.tf
```diff
@@ -94,11 +94,7 @@ variable "filestore_tier" {
...
     ], var.filestore_tier)
-    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','HIGH_SCALE_SSD','ZONAL','ENTERPRISE'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier."
-  }
-  validation {
-    condition     = !(var.protocol == "NFS_V4_1" && !contains(["HIGH_SCALE_SSD", "ZONAL", "REGIONAL", "ENTERPRISE"], var.filestore_tier))
-    error_message = "NFS_V4_1 is only supported with HIGH_SCALE_SSD, ZONAL, REGIONAL, or ENTERPRISE tiers."
+    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','ZONAL','REGIONAL'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1beta1/Tier."
```
tpdownes, line 97:
```suggestion
    error_message = "Allowed values for filestore_tier are 'BASIC_HDD','BASIC_SSD','ZONAL','REGIONAL'.\nhttps://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/filestore_instance#tier\nhttps://cloud.google.com/filestore/docs/reference/rest/v1/Tier."
```

tpdownes (changes requested):

tpdownes (approved):

PR #3534, opened 2025-01-14
Replace TAS plugin with Kueue

annuay-google:
## What?
- Replace TAS plugin with Kueue for job scheduling
- Deprecate TAS plugin
- Update Jobset based NCCL test to use Kueue TAS in place of TAS plugin

# Why?
Kueue is the officially supported job scheduling utility for A3 Ultra

# Testing
Ran Jobset based NCCL test, verified bandwidth figures

### Submission Checklist


examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
```diff
@@ -169,17 +170,17 @@ deployment_groups:
...
   - id: workload-manager-install
     source: modules/management/kubectl-apply
     use: [a3-ultragpu-cluster]
     settings:
       kueue:
         install: true
         version: v0.10.0
+        config_path: $(vars.kueue_configuration_path)
+        config_template_vars:
+          num_gpus: $(a3-ultragpu-pool.total_gpu_count)
```
ankitkinra, line 182: This is super, so just to confirm if the nodepool size is changed via the toolkit, this config will be updated ?

ankitkinra (changes requested):
Do we need to delete the 2 node nccl test ?

examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
annuay-google: Yes- this number is fetched as an output of the node pool module. Calculated as allocatable_gpu_per_node * static_node_count. This is done so that user doesn't have to pass a redundant input (num_gpus)

annuay-google:
> Do we need to delete the 2 node nccl test ?

Deleting this as well, since we have an n-node test now

annuay-google:
> > Do we need to delete the 2 node nccl test ?
> 
> Deleting this as well, since we have an n-node test now

Deleted

modules/compute/gke-node-pool/outputs.tf
```diff
@@ -50,6 +50,11 @@ output "allocatable_gpu_per_node" {
   value       = local.allocatable_gpu_per_node
 }
 
+output "total_gpu_count" {
```
samskillman, line 53: Wondering if this should be static_gpu_count. Might have implications down the road for configuring Kueue with DWS Flex, since DWS Flex + Kueue could be configured with the total maximum scale.

annuay-google: Updated to static_gpu_count

annuay-google:
> Reference

Addressed

examples/gke-a3-ultragpu/kueue-configuration.yaml.tftpl
```diff
@@ -0,0 +1,59 @@
...
+  name: "a3-ultra"
+spec:
+  namespaceSelector: {} # match all.
+  resourceGroups:
+  - coveredResources: ["nvidia.com/gpu"]
+    flavors:
+    - name: "a3-ultra"
+      resources:
+      - name: "nvidia.com/gpu"
+        nominalQuota: ${num_gpus}
```
mwysokin, line 51: This is something that we should discuss and consider pros and cons of.

With this config users won't be able to schedule non-TAS workloads. The reason is that if a `ResourceFlavor` refers to a topology Kueue expects each Workload that should be mapped to that `ResourceFlavor` to include a topology request. So if you send a Workload without a topology request Kueue won't consider this `ResourceFlavor` and won't admit the Workload and say something like: `Couldn't find ResourceFlavor for this Workload. We're working to improve this experience but that's the situation today. 

So we have 3 options now:

1. We assume that A3Ultra users will always send Workloads with a topology request. Workloads without it won't be admitted.
2. We create 2 separate local queues e.g. multislice-queue and multislice-queue-tas, 2 separate resource flavors TAS and non-TAS and 2 separate cluster queues. That's not ideal because it multiplies the amount of Kubernetes resources and XPK doesn't support multiple kueues only `multislice-queue`. However that's how we run during the validation of A3U with these separate objects.
3. We create 2 resorce flavors TAS and non-TAS pointing to the same nodepool and having the same quota. Pros are that it's simpler, one local queue can be used so XPK will support it and it's generally a good user experience. Cons are: that we don't support borrowing and preemption for TAS workloads yet (it's coming in March) so every cluster queue which has a resource flavor with topology will disable borrowing and preemption for the entire cluster queue. I don't how big of an issue it is for our users. I did some poll internally and it shouldn't be a big issue and the second one is that because there are 2 resource flavors there have to be 2 quotas so either the quota should be divided by 2 (IMO bad and wasteful idea) or the quota should be doubled (full quota for TAS and non-TAS workloads) so both workloads could use the full quota. The downside of this idea is that we haven't tested that scenario thoroughly and there might be race condition that 2 Workloads will be partially admitted at the same time and they'll arrive at a deadlock. The race condition should be highly unlikely because it basically requires 2 workloads to be submitted or admitted at the same time but there still exists this possibility.

Example of scenario 3 would be:

```yaml
kind: ResourceFlavor
apiVersion: kueue.x-k8s.io/v1beta1
metadata:
  name: "a3-ultra-tas"
spec:
  nodeLabels:
    cloud.google.com/gke-nodepool: ${node_pool_name}
  topologyName: "gke-default"
  tolerations:
  - key: "nvidia.com/gpu"
    operator: "Exists"
    effect: NoSchedule
---
kind: ResourceFlavor
apiVersion: kueue.x-k8s.io/v1beta1
metadata:
  name: "a3-ultra"
spec:
  nodeLabels:
    cloud.google.com/gke-nodepool: ${node_pool_name}
```

and then

```yaml
apiVersion: kueue.x-k8s.io/v1beta1
kind: ClusterQueue
metadata:
  name: "a3-ultra"
spec:
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["nvidia.com/gpu"]
    flavors:
    - name: "a3-ultra"
      resources:
      - name: "nvidia.com/gpu"
        nominalQuota: ${num_gpus}
    - name: "a3-ultra-tas"
      resources:
      - name: "nvidia.com/gpu"
        nominalQuota: ${num_gpus}
```

My recommendation is to go with scenario 3 with all its drawbacks since it gets the users the simplest experience. We're working on improving the situation both on the Kueue and XPK sides so the situation is temporary.

mwysokin (commented):
I think we should make an informed decision about how to handle TAS and non-TAS Workloads. I added a comment in the relevant place.

examples/gke-a3-ultragpu/kueue-configuration.yaml.tftpl
annuay-google: 3 seems like an acceptable trade-off, I've updated the Kueue configuration to have 2 resource flavors + updates to cluster queue

annuay-google:
> I think we should make an informed decision about how to handle TAS and non-TAS Workloads. I added a comment in the relevant place.

Addressed

annuay-google:
Re-tested the changes after adding Michal's suggestions, all good

PR #3561, opened 2025-01-19
SlurmGCP. Refactor timestamp parsing

mr0re1:
* stop using "to be deprecated" `datetime.utcnow`;
* fix bug in "Zulu-hack", there is no guarantee that future reservation will always use Zulu-format;
* have just one function to parse timestamps from GCP.


abbas1902 (approved):
LGTM

PR #3563, opened 2025-01-20
Make Partition into dedicated dataclass

mr0re1:
* Improve type-constraints;
* Fixed a bug of using never-defined `partition_feature`.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/base.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/base.py
@@ -1,0 +1,47 @@
+# Copyright 2024 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, List, Dict
+
+from dataclasses import dataclass, field
+
+@dataclass(frozen=True)
+class Partition:
+  name: str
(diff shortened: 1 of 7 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/base.py
```diff
@@ -0,0 +1,47 @@
...
+  nodesets: List[str] = field(default_factory=list)
+  nodesets_dyn: List[str] = field(default_factory=list)
+  nodesets_tpu: List[str] = field(default_factory=list)
+
+  @property
+  def is_tpu(self) -> bool:
+    return len(self.nodesets_tpu) > 0
+  
+  @property
+  def any_dynamic(self) -> bool:
```
abbas1902, line 34: Nit: I'd probably name it something like `has_dynamic`

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
```diff
@@ -180,7 +177,7 @@ def nodeset_dyn_lines(nodeset):
     )
 
 
-def partitionlines(partition, lkp: util.Lookup) -> str:
+def partitionlines(partition: Partition, lkp: util.Lookup) -> str:
```
abbas1902, line 180: Nit: Could we change this to `partition_lines`?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_resume.py
```diff
@@ -74,11 +74,11 @@ def test_group_nodes_bulk(mock_create_placements, mock_tpu):
         "t": TstNodeset(nodeset_name="t"),
       },
       partitions={
-        "p1": TstPartition(
+        "p1": dict(
```
abbas1902, line 77: Why not use the Partition class?

mr0re1: Because it mock raw config that uses Dict, the `Partition` class is not integrated into config yet. It's a follow up action.

PR #3568, opened 2025-01-21
enable queued provisioning on gke nodepool

SwarnaBharathiMantena:
Enable queued provisioning on gke nodepool.
- Add the enable_queued_provisioning boolean variable.
- Update google-beta provider version to 5.21 (associated with queued_provisioning GA)
- Add precondition to ensure placement_policy is not set to COMPACT when enable_queued_provisioning is true.

### Submission Checklist

### DIFF
--- a/modules/compute/gke-node-pool/main.tf
+++ b/modules/compute/gke-node-pool/main.tf
@@ -89,7 +89,7 @@ resource "google_container_node_pool" "node_pool" {
   max_pods_per_node = var.max_pods_per_node
 
   management {
-    auto_repair  = true
+    auto_repair  = var.auto_repair
     auto_upgrade = var.auto_upgrade
   }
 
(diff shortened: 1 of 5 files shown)


modules/compute/gke-node-pool/main.tf
```diff
@@ -107,6 +107,10 @@ resource "google_container_node_pool" "node_pool" {
     }
   }
 
+  queued_provisioning {
+    enabled = var.enable_queued_provisioning
+  }
+
```
annuay-google, line 113: Should this be a dynamic block? Do we even want to pass this field to the provider if we don't want queued provisioning?

modules/compute/gke-node-pool/versions.tf
```diff
@@ -22,7 +22,7 @@ terraform {
     }
     google-beta = {
       source  = "hashicorp/google-beta"
-      version = "> 5"
+      version = "> 5.21"
```
annuay-google, line 25: Can we not bump both providers to most recent terraform version available in toolkit?

modules/compute/gke-node-pool/main.tf
SwarnaBharathiMantena: For the second part of the question, default value is set to false. 
Is there any advantage in defining it as dynamic block? I am using this as reference https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks

PR #3585, opened 2025-01-23
Change cloud rdma installation script to upgrade packages

abbas1902:
Found an issue in the driver installation script where it would do a dnf update indiscriminately (some unrelated package would be uninstallable). We should rather only update the packages that are related to cloud rdma.  Also added conditional driver reloads since its only needed when the kmod package update is updated (and adds significant boot time when done).
### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/install_cloud_rdma_drivers.sh
+++ b/modules/scripts/startup-script/files/install_cloud_rdma_drivers.sh
@@ -19,15 +19,19 @@ OS_VERSION="$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"
 OS_VERSION_MAJOR="$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"//g' -e 's/\..*$//')"
 
 if { [ "${OS_ID}" = "rocky" ] || [ "${OS_ID}" = "rhel" ]; } && { [ "${OS_VERSION_MAJOR}" = "8" ]; }; then
+	KMOD_VERSION="$(dnf list installed | grep "kmod-idpf-irdma" | awk '{print $2}')"
+
 	# Downloading the RDMA packages and installing them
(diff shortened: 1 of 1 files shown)


modules/scripts/startup-script/files/install_cloud_rdma_drivers.sh
```diff
@@ -19,15 +19,19 @@ OS_VERSION="$(awk -F '=' '/VERSION_ID/ {print $2}' /etc/os-release | sed -e 's/"
...
+	sudo dnf upgrade kmod-idpf-irdma rdma-core libibverbs-utils librdmacm-utils infiniband-diags perftest -y
+	NEW_KMOD_VERSION="$(dnf list installed | grep "kmod-idpf-irdma" | awk '{print $2}')"
 
-	# Restarting drivers so they can use RDMA without needing a reboot
-	sudo rmmod idpf
-	sudo rmmod irdma
-	sudo modprobe idpf
+	# Restarting drivers so they can use RDMA without needing a reboot (for new installs or for kmod package updates)
+	if { [ -z "${KMOD_VERSION}" ] || [ "${KMOD_VERSION}" != "${NEW_KMOD_VERSION}" ]; }; then
+		sudo rmmod idpf
```
tpdownes, line 31: To my eye, this sequence of commands doesn't seem consistent. How is `irdma` being reloaded? if it is automatically loaded when you `modprobe idpf`, then I would expect you to need to remove the modules in the opposite order?

tpdownes (commented):
Please address the question. No explicit changes being asked for (yet).

modules/scripts/startup-script/files/install_cloud_rdma_drivers.sh
abbas1902: The modules are independent but the loading of one causes the loading of the other hence the odd ordering.

tpdownes (approved):

PR #3618, opened 2025-01-31
Enhance deploy to output JSON instead of plain-english

nadig-google:
NOTE NOTE NOTE
This a draft Pull Request to solicit early feedback. It has not been tested. This is also not the final implementation.
NOTE NOTE NOTE

This pull request adds support to output JSON instead of the usual "english text output" to "gcluster deploy". It adds a new flag to the deploy command called --json-output which writes a file called cluster_toolkit_output.json with the output messages written JSON format (instead of the usual "english text output" to STDOUT.

This pull request was submitted git commit --no-verify because of failing go staticchecks that prevent git commit from successfully running

What works as of now
- Writes output to JSON file
- Writes unique file per Terraform group to avoid files being overwritten

What does not work as of now:
- Cluster Deployment does not seem to work (I am unsure if the failure is related to my changes)
=============
Related (older and deprecated) PR: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3604


cmd/utils.go
```diff
@@ -82,6 +82,20 @@ func filterYaml(cmd *cobra.Command, args []string, toComplete string) ([]string,
 	return []string{"yaml", "yml"}, cobra.ShellCompDirectiveFilterFileExt
 }
 
+var flagJsonOutput bool
+
+func getJsonOutputBehavior() shell.OutputFormat {
```
mr0re1, line 87: nit `s/getJsonOuptutBehvior/getOuptutFormat`

pkg/shell/terraform.go
```diff
@@ -224,6 +238,31 @@ func promptForApply(tf *tfexec.Terraform, path string, b ApplyBehavior) bool {
...
 }
 
+// This function applies the terraform plan, but generates outputs in JSON format
+// (instead of text)
+func applyPlanJsonOutput(tf *tfexec.Terraform, path string) error {
+	planFileOpt := tfexec.DirOrPlan(path)
+	logging.Info("Running terraform apply on deployment group %s", tf.WorkingDir())
+	// To do: Make file name as a user input
+	// Make the JSON file name unique by having the Terraform group as a substring
+	jsonFilename := "cluster_toolkit_output-" + strings.ReplaceAll(tf.WorkingDir(), "/", ".") + ".json"
```
mr0re1, line 248: This will not work on Windows, not that we claim support, but nonetheless.

cmd/utils.go
```diff
@@ -82,6 +82,20 @@ func filterYaml(cmd *cobra.Command, args []string, toComplete string) ([]string,
...
+	if flagJsonOutput {
+		return shell.JsonOutput
+	}
+	return shell.TextOutput
+}
+
+func addJsonOutputFlag(c *cobra.Command) *cobra.Command {
+	c.Flags().BoolVar(&flagJsonOutput, "json-output", false, "Output errors in JSON format")
```
mr0re1, line 95: AFAIK, It's not  just errors

pkg/shell/terraform.go
```diff
@@ -276,15 +315,25 @@ func applyOrDestroy(tf *tfexec.Terraform, b ApplyBehavior, destroy bool) error {
...
-	if err := applyPlanConsoleOutput(tf, f.Name()); err != nil {
-		return err
+	switch of {
+
+	case JsonOutput:
+		if err := applyPlanJsonOutput(tf, f.Name()); err != nil {
+			return err
+		}
+	case TextOutput: // Text output to the console is also the default choice
+		fallthrough
```
mr0re1, line 325: nit. 
Don't use `fallthrough` ; 
for default , either:
* omit default section
* panic
* set `err`

pkg/shell/terraform.go
mr0re1: + destroy appliaction will override deploy output.

pkg/shell/terraform.go
```diff
@@ -256,14 +295,14 @@ func applyOrDestroy(tf *tfexec.Terraform, b ApplyBehavior, destroy bool) error {
 	// capture Terraform plan in a file
 	f, err := os.CreateTemp("", "plan-)")
 	if err != nil {
+		logging.Info("deploy.go.0000: applyOrDestroy()")
```
mr0re1, line 298: Don't forget to remove

pkg/shell/terraform.go
```diff
@@ -36,6 +36,20 @@ import (
 	"github.com/zclconf/go-cty/cty/gocty"
 )
 
+// OutputFormat determines the format in which the errors are reported.
+// Current supported format are text (default option) and JSON. Future
+// format could be protobuf
+type OutputFormat uint
+
+// 2 output formats are currently supported:
```
mr0re1, line 44: Redundant comment

pkg/shell/terraform.go
mr0re1: The one way to solve this problem would be to pass `OutputSpec` instead of `OuptutFormat` json.

pkg/shell/terraform.go
nadig-google: I picked "panic"

pkg/shell/terraform.go
nadig-google: Removed redundant and updated comment

pkg/shell/terraform.go
nadig-google: Removed

cmd/utils.go
nadig-google: Updated

cmd/utils.go
nadig-google: Updated

pkg/shell/terraform.go
nadig-google: resolved using OS specific change. Let me know if it works

pkg/shell/terraform.go
```diff
@@ -224,6 +238,43 @@ func promptForApply(tf *tfexec.Terraform, path string, b ApplyBehavior) bool {
...
+	case "darwin":
+		fallthrough
+	case "linux":
+		replaceChar = "/"
+	case "windows":
+		replaceChar = "\\"
+	}
+
+	// To do: Make file name as a user input
+	// Make the JSON file name unique by having the Terraform group as a substring
```
mr0re1, line 258: There is no notion of the "group" in this code. `group -> working dir`.

mr0re1: nit `cluster_toolkit_` > `gcluster_`

mr0re1 (approved):

pkg/shell/terraform.go
```diff
@@ -29,13 +29,27 @@ import (
...
+// format could be protobuf
+type OutputFormat uint
+
+// Future option could be ProtoBuf
+const (
+	TextOutput OutputFormat = iota
+	JsonOutput
+)
+
+const supportedOutputFormats = "Text , JSON "
```
mr0re1, line 51: Remove this const.

pkg/shell/terraform.go
```diff
@@ -276,15 +326,24 @@ func applyOrDestroy(tf *tfexec.Terraform, b ApplyBehavior, destroy bool) error {
...
+	case JsonOutput:
+		if err := applyPlanJsonOutput(tf, f.Name()); err != nil {
+			return err
+		}
+	case TextOutput: // Text output to the console is also the default choice
+		if err := applyPlanConsoleOutput(tf, f.Name()); err != nil {
+			return err
+		}
+	default:
+		panic(fmt.Sprintf("Unsupported output format requested. Supported output formats: %s", supportedOutputFormats))
```
mr0re1, line 339: Don't use `Sprintf` here. 
a) yet another change to derail on panic-path;
b) no need for nice output;
c) constant defined somewhere else hurts readability

nadig-google:
Landed this PR instead : https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3654

PR #3639, opened 2025-02-05
Cluster Toolkit - new module for creating Artifact Registries

scott-nag:
### Cluster Toolkit Updates
Introduction of a module to handle Artifact Registry operations: [README](https://github.com/nagconsulting/cluster-toolkit/blob/99ad9cae4fd1a88cc496a72ee0ba64bd5c7a991f/community/modules/container/artifact-registry/README.md)

### OFE Updates
- Re-enabled the latest Terraform (1.4 was being used temporarily due to cluster deletion bug)
- Included an option for Private Service Access to the VPC creation page
- Added views for displaying Artifact Registry repositories that are created alongside clusters, as well as placeholder views that will soon be used to handle container operations
- Updated OFE cluster creation page to include a section for creating Artifact Registries per the TF module
- Fixed an issue with extraction of the Slurm controller location as it had change in the TFState file
- Updated the cluster page to remove the deprecated `enable_smt` field and replace it with `advanced_machine_features.threads_per_core`
- Python update 3.8 -> 3.12

I do have one question about pre-commit/Terraform. [The variables file here](https://github.com/nagconsulting/cluster-toolkit/blob/99ad9cae4fd1a88cc496a72ee0ba64bd5c7a991f/community/modules/container/artifact-registry/variables.tf) contains validation which I have commented out as pre-commit wasn't happy. I was seeing errors like this: `The condition for variable "repo_password" can only refer to the variable itself, using var.repo_password` despite it working when I was testing it locally - is there a different/preferred way that I can re-add these rules again?

Any questions please feel free to give me a shout. Many thanks


scott-nag:
Hello - bumping this PR for visibility as I have just added a number of updates:
- Update: Added a new ‘Containers’ section to OFE. This is primarily used to build containers in ‘standard’ repositories using Cloud Build. Builds are handled by an update to the existing c2 / PubSub functionality
- Update: Added new Ansible role to configure clusters to support these containerised Enroot / Pyxis jobs
- Update: Added option to ‘Enable Private Google Access’ when creating a VPC (required for Artifact Registry / container features)
- Update: Added functionality to create a container-based ‘Application’, and a new section to facilitate these containerised Slurm ‘Jobs’ too
- Update: Simplified the ‘Applications’ form by using Django ‘crispy forms’ package. This means that a lot of the manual Bootstrap HTML/CSS styling can be removed from the front end templates
- Fixed: Ansible role used to configure clusters would sometimes try to retrieve a package unsuccessfully, causing cluster deployment to fail. Included ‘retries’ in the task which resolved the issue
- Fixed: A delay in a freshly created Service Account propagating would occasionally cause a failure when applying roles. Included a mechanism to retry on failure here too
- Fixed: Blueprint generation issue where the ‘compute_startup_scripts’ field in the v6-controller module was deprecated in favour of a similar ‘startup-script’ field in the ‘v6-nodeset’ module.

community/modules/container/artifact-registry/variables.tf
```diff
@@ -0,0 +1,143 @@
...
+  type        = map(string)
+  default     = {}
+}
+
+variable "repo_password" {
+  description = "Optional password/API key. If null, one will be randomly generated."
+  type        = string
+  default     = null
+
+  # validation {
```
mr0re1, line 43: Let's either remove or add description why temporary commented out.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -111,7 +111,7 @@ locals {
     device_name : try(google_compute_disk.controller_disk[0].name, null)
   }
 
-  ghpc_startup_login = var.login_startup_script == null ? [] : [{
+  ghpc_startup_login = [{
```
mr0re1, line 114: Please rebase, should not be needed anymore.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -547,7 +547,7 @@ variable "maintenance_interval" {
 variable "startup_script" {
   description = "Startup script used by VMs in this nodeset"
   type        = string
-  default     = null
+  default     = "# no-op"
```
mr0re1, line 550: Let's rebase, shoudn't be needed anymore

mr0re1 (dismissed):

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -547,7 +547,7 @@ variable "maintenance_interval" {
 variable "startup_script" {
   description = "Startup script used by VMs in this nodeset"
   type        = string
-  default     = "# no-op"
+  default     = null
```
mr0re1, line 550: Why change this? Please revert, unless you needed, in this case lets discuss it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
```diff
@@ -111,7 +111,7 @@ locals {
     device_name : try(google_compute_disk.controller_disk[0].name, null)
   }
 
-  ghpc_startup_login = [{
+  ghpc_startup_login = var.login_startup_script == null ? [] : [{
```
mr0re1, line 114: same

tpdownes: This was removed from PR

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
tpdownes: This was removed from PR

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
tpdownes: Rebased off current develop

community/modules/container/artifact-registry/variables.tf
```diff
@@ -0,0 +1,148 @@
...
+    (e.g., DOCKER_HUB) or null for custom repo.
+  DOC
+  type        = string
+  default     = null
+
+  # To Do: implement validation
+  # validation {
+  #   condition     = ((var.repo_mode != "REMOTE_REPOSITORY" && var.repo_public_repository == null) || (var.repo_mode == "REMOTE_REPOSITORY" && (var.repo_public_repository != null || var.repo_mirror_url != null)))
+  #   error_message = "If repo_mode is REMOTE_REPOSITORY, you must set either repo_public_repository or repo_mirror_url. Otherwise, leave them null."
+  # }
```
tpdownes, line 94: Agreed. As written, this validation block would break for Terraform < 1.9, which breaks our guarantee of Terraform 1.5 compatibility.

The way to do this in Terraform 1.5 is via a lifecycle/precondition block:

https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions

```suggestion
```

tpdownes (changes requested):

community/modules/container/artifact-registry/variables.tf
tpdownes: I am OK if we add the lifecycle/precondition at a later time.

tpdownes: Code addressed this.

tpdownes (approved):
We don't yet have a passing test, but the existing implementation has known problems and manual testing by NAG suggests that the problem may be in our test rather than in the solution itself. We believe customer supportability is best served by merging the solution and doing further evaluation of the test.

RachaelSTamakloe (approved):

PR #3642, opened 2025-02-06
Add startup script to set OFI env vars for Cloud RDMA workloads

abbas1902:
This PR does the following:
- Adds ability for users to enable OFI tunable environment variables in the form of a startup script
- Mounts the lssd attached to h4d-highmem-192-lssd machines

### Submission Checklist

### DIFF
--- a/community/examples/hpc-slurm-h4d.yaml
+++ b/community/examples/hpc-slurm-h4d.yaml
@@ -76,6 +76,11 @@ deployment_groups:
     source: modules/scripts/startup-script
     settings:
       install_cloud_rdma_drivers: true
+      set_ofi_cloud_rdma_tunables: true
+      local_ssd_filesystem:
+        fs_type: ext4
+        mountpoint: /mnt/lssd
+        permissions: "1777"
 
   - id: h4d_nodeset
     source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
@@ -123,8 +128,14 @@ deployment_groups:
       machine_type: n2-standard-4
       enable_login_public_ips: true
 
+  - id: slurm_controller_startup
+    source: modules/scripts/startup-script
+    settings:
+      set_ofi_cloud_rdma_tunables: true
+
   - id: slurm_controller
     source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
     use: [h4d-slurm-net-0, h4d_partition, slurm_login, homefs, appsfs]
     settings:
       enable_controller_public_ips: true
+      controller_startup_script: $(slurm_controller_startup.startup_script)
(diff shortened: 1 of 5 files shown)


modules/scripts/startup-script/main.tf
```diff
@@ -93,6 +93,20 @@ locals {
...
+  ofi_runner = var.set_ofi_cloud_rdma_tunables == "" ? [] : [
+    {
+      type        = "data"
+      destination = "/etc/profile.d/set_ofi_cloud_rdma_tunables.sh"
+      content     = <<-EOT
+        #!/bin/bash
+        export FI_PROVIDER="verbs;ofi_rxm"
+        export FI_OFI_RXM_USE_RNDV_WRITE=1
+        export FI_VERBS_INLINE_SIZE=39
+        export I_MPI_FABRICS="shm:rxm"
```
rohitramu, line 105: The `export` command only makes the variable available to child shells/processes, so it won't be set when the user logs in.

community/examples/hpc-slurm-h4d.yaml
```diff
@@ -76,6 +76,7 @@ deployment_groups:
     source: modules/scripts/startup-script
     settings:
       install_cloud_rdma_drivers: true
+      set_ofi_cloud_rdma_tunables: true
```
rohitramu, line 79: Did we confirm that the env vars won't negatively effect other workloads?  If so, why not put the logic in the RDMA driver installation script?

modules/scripts/startup-script/main.tf
abbas1902: My understanding is since this is placed as a `profile.d` and runs a start up script as long as a user logs in after start up scripts complete it will be there. If they log in prior then they would need to source their .bashrc again

community/examples/hpc-slurm-h4d.yaml
abbas1902: It shouldn't negatively effect other (RDMA) workloads. If someone decides to use a TCP they'd have to override these variables but in any event they'd probably just set this startup script to false. That leads into why I separated them, I didn't want to couple the installation with the env var setting.

modules/scripts/startup-script/main.tf
rohitramu: Ahh I see, missed that it was in `/etc/profile.d`.  Thanks for pointing that out!

community/examples/hpc-slurm-h4d.yaml
rohitramu: That makes sense, thanks for explaining.  Be sure to document that this should be disabled for non-RDMA workloads (probably in the variable description).

rohitramu (approved):

rohitramu (approved):

rohitramu (approved):

PR #3652, opened 2025-02-07
Prevent mixing tz-aware and tz-naive datetime.

mr0re1:
* Add `util.now` to simplify obtaining tz-aware timestamps;
* Make `util.parse_gcp_timestamp` to fallback to UTC;
* Add tests for `Lookup.future_reservation`.


abbas1902 (approved):
LGTM, this should be quite helpful. I think the only remaining spots in the code that has timezone naive datetimes is in load_bq.py. Feel free to switch it to make use of the util method or land as-is (since it only subtracts a timedelta which is works for both naive and tzaware datetimes).

PR #3659, opened 2025-02-10
Switch ofi startup script to not run automatically

abbas1902:
Fixed the boolean logic behind whether the OFI tunable are set for Cloud RDMA (which was introduced in #3642 ). Thereby preventing it from running on all deployments.

Problem: Terraform around running/not running the script had a bug which made it run all the time. This is problematic for MPI workloads that rely on OFI tunables.

Solution: Fix the boolean logic so it accounts for whether `set_ofi_cloud_rdma_tunables` is set to true or false (and defaults to false).

### Submission Checklist

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -53,7 +53,7 @@ HPC deployments on the Google Cloud Platform.`,
 				logging.Fatal("cmd.Help function failed: %s", err)
 			}
 		},
-		Version:     "v1.46.0",
+		Version:     "v1.46.1",
 		Annotations: annotation,
 	}
 )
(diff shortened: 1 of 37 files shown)


modules/scripts/startup-script/variables.tf
```diff
@@ -268,9 +268,3 @@ variable "install_cloud_rdma_drivers" {
   type        = bool
   default     = false
 }
-
-variable "set_ofi_cloud_rdma_tunables" {
```
mr0re1, line 272: Instead of deleting it has to be "deprecated"

mr0re1 (approved):

modules/scripts/startup-script/variables.tf
abbas1902: Added deprecated variable entry

mr0re1 (approved):

mr0re1 (approved):

rohitramu (approved):

mr0re1 (approved):

PR #3672, opened 2025-02-13
Add hyperdisk-ml support for gke-storage

saltysoup:
Hyperdisk-ml is the recommended storage for speeding up load times and inference of larger models eg. DeepSeek-R1, Llama3-405b, which often takes a long time ~20min+

Source: https://cloud.google.com/kubernetes-engine/docs/tutorials/serve-multihost-gpu#speed_up_model_load_times_with_hyperdisk_ml

This PR is to add hyperdisk-ml support by adding this disk type as a valid condition in `variables.tf`, and new templates for storage class and PVC

### DIFF
--- a/modules/file-system/gke-storage/storage-class/hyperdisk-ml-sc.yaml.tftpl
+++ b/modules/file-system/gke-storage/storage-class/hyperdisk-ml-sc.yaml.tftpl
@@ -1,0 +1,24 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: ${name}
+  labels:
+  %{~ for key, val in labels ~}
+    ${key}: ${val}
+provisioner: pd.csi.storage.gke.io
+allowVolumeExpansion: true
+parameters:
+  %{~ endfor ~}
+  type: hyperdisk-ml
+volumeBindingMode: ${volume_binding_mode}
+reclaimPolicy: ${reclaim_policy}
+  %{~ if topology_zones != null ~}
+allowedTopologies:
+- matchLabelExpressions:
+  - key: topology.gke.io/zone
+    values:
+    %{~ for z in topology_zones ~}
+    - ${z}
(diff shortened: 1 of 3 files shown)


modules/file-system/gke-storage/storage-class/hyperdisk-ml-sc.yaml.tftpl
```diff
@@ -0,0 +1,24 @@
...
+kind: StorageClass
+metadata:
+  name: ${name}
+  labels:
+  %{~ for key, val in labels ~}
+    ${key}: ${val}
+provisioner: pd.csi.storage.gke.io
+allowVolumeExpansion: true
+parameters:
+  %{~ endfor ~}
```
wiktorn, line 11: Was this for really meant to be that long?

saltysoup: what do you mean by long? FWIW i used the same template as hyperdisk-extreme (noticed its slightly different to hyperdisk-balanced template).

Hyperdisk-ml doesnt [support customizable IOPS so i took that out](https://cloud.google.com/compute/docs/disks/hyperdisks#feature_summary). I also didnt include custom throughput as it automatically specifies this based on [disk size](https://cloud.google.com/compute/docs/disks/hyperdisks#limits-disk)

wiktorn: I expected something like this:
```suggestion
  %{~ for key, val in labels ~}
    ${key}: ${val}
  %{~ endfor ~}
provisioner: pd.csi.storage.gke.io
allowVolumeExpansion: true
parameters:
```

I think you don't need to repeat `provsioner` and `allowVolmeExpansion`, and most imporatantly, you don't want them missing if no labels are provided. And to me it looks like `hyperdisk-extreme` has the same issue

PR #3691, opened 2025-02-20
Update a3-highgpu image blueprint to use install_nvidia_repo

alyssa-sm:
This PR updates the a3-highgpu blueprint to use a new setting `install_nvidia_repo` which allows for optionally installing the nvidia repository in the image building process. It should be set to false when the image already has the nvidia repository installed.


tpdownes (approved):
Please create a follow-up task to remove the `slurm_version` as a separate PR.

PR #3695, opened 2025-02-21
docs(gke-job-template): example path links

guspan-tanadi:


ighosh98:
@guspan-tanadi are you still working on this?

PR #3752, opened 2025-03-04
Cloud DNS config addition to GKE Cluster module

SwarnaBharathiMantena:
### What?
Cloud DNS is exposed as a DNS provider option in the gke-cluster module. 

### Why?
The Cloud DNS provider replaces the default kube-dns in-cluster provider with Google's managed Cloud DNS service, and is often preferable for larger-node-count clusters.

### Testing
Manually tested provisioning/updates, make tests passed, pre-commit hooks ran fine, and CLA should be on file.

### Submission Checklist


modules/scheduler/gke-cluster/main.tf
```diff
@@ -189,6 +189,13 @@ resource "google_container_cluster" "gke_cluster" {
     }
   }
 
+  dns_config {
```
annuay-google, line 192: Did you test the scenario where this was set to null?

SwarnaBharathiMantena: Do you mean by not passing the input variables? or in specific set to null?

annuay-google (dismissed):
LGTM

modules/scheduler/gke-cluster/main.tf
annuay-google: Either should behave same, but not an issue- the default will ensure no NPE is thrown

SwarnaBharathiMantena: I think it is a good idea to add a validation to make sure the entry for `cluster_dns_scope` and `cluster_dns`, as they accept only three valid values. Thoughts?

SwarnaBharathiMantena: Validations added.

annuay-google (approved):

PR #3770, opened 2025-03-06
Reduce number of startup scripts.

mr0re1:
* Deprecate `compute_startup_script`;
* Remove default no-op scripts, create none instead.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
```diff
@@ -547,15 +547,9 @@ EOD
   default     = 300
 }
 
-variable "compute_startup_script" {
-  description = "Startup script used by the compute VMs."
-  type        = string
-  default     = "# no-op"
-}
-
 variable "compute_startup_scripts_timeout" {
```
harshthakkar01, line 550: so, we provide `startup_script` in nodeset module and timout in `controller` ?

mr0re1: Yep, no change in CX,  but a room for improvements)

harshthakkar01 (approved):

harshthakkar01 (approved):

PR #3775, opened 2025-03-08
Don't take md5 of storage objects to avoid provider bug

mr0re1:
Fixes this type of error that can happen during re-deployments:

```sh
Error: Provider produced inconsistent final plan

When expanding the plan for
module.slurm_controller.module.slurm_files.google_storage_bucket_object.nodeset_startup_scripts["x_startup_sh"]
to include new values learned so far during apply, provider
"registry.terraform.io/hashicorp/google" produced an invalid new value for
.crc32c: was known, but now unknown.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

**Partially** fixed by taking `md5` of source strings instead of resulted objects.

**Follow up:** Address other "pain points" (see https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3775#pullrequestreview-2669023885)

Provider bug: https://github.com/hashicorp/terraform-provider-google/issues/18618

**Tested** by repeated re-deployments with/without changes to python scripts, startup scripts and prologues.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -227,7 +227,7 @@ def _find_tpu_node_action(nodename, state) -> NodeAction:
         return NodeActionUnknown(slurm_state=state, instance_state=inst.status)
 
     return NodeActionUnchanged()
-
+# loloololo
```
wiktorn, line 230:
```suggestion
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
```diff
@@ -205,42 +188,52 @@ resource "google_storage_bucket_object" "devel" {
 # SCRIPTS #
 ###########
 
-resource "google_storage_bucket_object" "controller_startup_scripts" {
-  for_each = {
+locals {
+  ok_chars = "/[^a-zA-Z0-9-_]/"
```
wiktorn, line 192:
```suggestion
  bad_chars = "/[^a-zA-Z0-9-_]/"
```
Actually this regexp matches all the characters you don't want in the filename.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
```diff
@@ -90,25 +88,10 @@ locals {
...
-    login_startup_scripts_md5 = {
-      for o in values(google_storage_bucket_object.login_startup_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    prolog_scripts_md5 = {
-      for o in values(google_storage_bucket_object.prolog_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    epilog_scripts_md5 = {
-      for o in values(google_storage_bucket_object.epilog_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
+    # Extra-files MD5 hashes. Makes config file creation to wait on those files
```
wiktorn, line 91:
```suggestion
    # Extra-files MD5 hashes. Makes config file creation to wait on those files
    slurm_gcp_scripts_md5 = data.archive_file.slurm_gcp_devel_zip.output_md5
```

Data resource output_md5 is known at the plan stage, so this won't make the config.yaml file known only after apply.

wiktorn (commented):
I'd still remove `source` from the:
* google_storage_bucket_object.devel
* google_storage_bucket_object.prolog_scripts
* google_storage_bucket_object.epilog_scripts


Or add `detect_md5hash` with value of md5 hash for those files there.

The last part, that makes `config.yaml` file known after apply (which is a way to trigger the bug) are the `var.nodeset`. In this map lurks [instance template](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf#L87) which changes frequently and its value is known only after apply.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
mr0re1: There is another motivation to include it. Our archive composing is not discriminatory so it will include garbage files (e.g. mypy cache), so it creates unnecessary diff. 
Or may have false-positivie difference in case deployments are performed from different machines.

wiktorn: As long as we are not checking for `slurm_gcp_devel.zip` in `slurmsync.py` I'm fine with removal of this

abbas1902 (approved):

PR #3776, opened 2025-03-08
Stop using `google_compute_image` data source

mr0re1:
Due to https://github.com/hashicorp/terraform-provider-google/issues/21766 stop using `data "google_compute_image"`.

* Following validations has been removed:
  * Check that boot disk size >= image size;
  * Check that image license is aligned with `allow_automatic_updates`.
* Move duplicate `source_image_logic.tf` into `community/modules/internal/slurm-gcp/image_logic`;
* Remove default `instance_image` values from user facing modules.


rohitramu (changes requested):
I think we should wait on this PR until we get a response on https://github.com/hashicorp/terraform-provider-google/issues/21766, for a few reasons:
- The failure mode "disk size is less than image size" is difficult to identify and diagnose in a deployment
- Removing the default image value almost certainly makes the user experience worse - a beginner Cluster Toolkit user may pick a non-Slurm image and be confused as to why nothing's working
- Image deletion seems like a rare problem with an easy workaround (don't delete in-use images)

rohitramu (approved):
Discussed offline.

* I still feel that the failure mode "disk size is less than image size" is difficult to identify and diagnose in a deployment.  We should add this check back when https://github.com/hashicorp/terraform-provider-google/issues/21766 is resolved
* A default image will still be selected if one isn't provided - the specific image just isn't defined in the nodeset or login node settings.  This shouldn't have as big an impact on user experience.

PR #3790, opened 2025-03-11
Stop using `google_compute_image` data source

mr0re1:
**Re-application of reverted**  https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3776

Due to https://github.com/hashicorp/terraform-provider-google/issues/21766 stop using `data "google_compute_image"`.

* Following validations has been removed:
  * Check that boot disk size >= image size;
  * Check that image license is aligned with `allow_automatic_updates`.
* Move duplicate `source_image_logic.tf` into `community/modules/internal/slurm-gcp/image_logic`;
* Remove default `instance_image` values from user facing modules.


community/modules/compute/schedmd-slurm-gcp-v6-nodeset/source_image_logic.tf
```diff
@@ -1,77 +0,0 @@
...
-      EOD
-    }
-    postcondition {
-      condition     = var.disk_size_gb >= self.disk_size_gb
-      error_message = "'disk_size_gb: ${var.disk_size_gb}' is smaller than the image size (${self.disk_size_gb}GB), please increase the blueprint disk size"
-    }
-    postcondition {
-      # Condition needs to check the suffix of the license, as prefix contains an API version which can change.
-      # Example license value: https://www.googleapis.com/compute/v1/projects/cloud-hpc-image-public/global/licenses/hpc-vm-image-feature-disable-auto-updates
-      condition     = var.allow_automatic_updates || anytrue([for license in self.licenses : endswith(license, "/projects/cloud-hpc-image-public/global/licenses/hpc-vm-image-feature-disable-auto-updates")])
```
rohitramu, line 73: Just beware that removing the license check opens us up to likely more support requests - we should probably document that this feature is only available on the HPC image and Slurm Rocky Linux 8 image.

rohitramu (dismissed):

rohitramu:
I've approved, but please run applicable tests before submitting.

tpdownes:
A potentially lighter-weight option that still values the user experience is to move the `data` source inside a `check` block which effectively converts failures to warnings.

https://developer.hashicorp.com/terraform/language/checks#scoped-data-sources

It is unclear to me if `check` blocks run at all on `destroy`. I think the documentation is saying that they do not. The reason it is unclear to me is because `destroy` is just an alias for `apply -destroy` (think of destroy as reverse graph traversal of apply).

PR #3807, opened 2025-03-17
Enable higher performance self-managed NFS server configurations

ndebuhr:
TLDR: This PR enables higher performance self-managed NFS server configurations (community file-system module)

**Boot Disk**

C3 and C4 machine families do not support the [`google_compute_disk`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk) default disk type of `pd-standard` [[1](https://cloud.google.com/compute/docs/general-purpose-machines#c3_disks)][[2](https://cloud.google.com/compute/docs/general-purpose-machines#supported_disk_types_for_c4)][[3](https://github.com/hashicorp/terraform-provider-google/blob/f4f9908e686e965a224982f02a7c3dfefa833010/google/services/compute/resource_compute_disk.go#L756)]. However, the NFS server module doesn't currently allow for an override of the default `pd-standard` boot disk type, so the use of C3 and C4 machines is not possible, despite their importance in Hyperdisk-based optimized NFS setups. 

**NFS Data Disk**

Additionally, most NFS data disk types are currently not supported by the NFS server module: Extreme PD; Hyperdisk Balanced; Hyperdisk Extreme; Hyperdisk ML; and Hyperdisk Throughput. These disk types require throughput and/or IOPS specifications, which are not currently exposed in the module.

**Implementation**

This PR introduces/exposes the boot disk type, provisioned throughput, and provisioned IOPS with null defaults, to avoid breaking changes, configuration drift, and even non-disruptive in-place resource operations.

Null defaults for the provisioned IOPS and throughput values are also ideal from a UX perspective, as default values vary based on storage type and disk size. Additionally, Hyperdisks do not have a consistent "provisioned specification" structure (e.g., Hyperdisk Balanced needs both IOPS and throughput, but Hyperdisk Extreme needs only IOPS). This all considered, an explicit specification (null defaults) seems optimal.

While not explicitly required for the use case, the boot disk size is also exposed, as being locked in to the rather small defaults (usually 10-30 GB) could be limiting for some heavier-weight NFS server implementations.

**Logistics**

* Forked this PR branch from the Toolkit "develop" branch (not main)
* Tested all changes with pre-commit in a local branch
* Confirmed that "make tests" passes all tests
* Unit test coverage looks fine
* Updated all applicable documentation
* Followed Cluster Toolkit Contribution guidelines
   * Contributor License Agreement is on file
* Manually provision-/deprovision-tested the changes across multiple boot disk, data disk, and machine type configurations


SwarnaBharathiMantena (approved):
LGTM

annuay-google:
@SwarnaBharathiMantena can you please work with Neal to merge this PR?

ighosh98 (approved):

ighosh98:
Quick question @ndebuhr Is there some doc that guides customers on what is the ideal value for provisioned IOPS and throughput ? Might make sense to link those.

SwarnaBharathiMantena (approved):

ighosh98 (approved):

PR #3824, opened 2025-03-20
Use MIG for DWS Flex start

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -30,6 +29,7 @@
 )
 from util import lookup
 import tpu
+import mig_flex
```
abbas1902, line 32: Maybe I should get some new glasses but where is this coming from? Is this some package or a python file of some sort?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -41,8 +42,9 @@
     trim_self_link,
     wait_for_operation,
 )
-from util import lookup, NSDict, ReservationDetails
+from util import lookup, ReservationDetails
```
abbas1902, line 45: nit: we can probably just unify this with the other imports on L33-43

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
mr0re1: Forget to `git add`

abbas1902 (approved):
LGTM, going to close this PR and merge it with an extra commit containing some documentation, so this approval is more for the record.

PR #3849, opened 2025-03-26
Relax slurm_cluster_name validation slightly

aaronegolden:
Relaxes the name restrictions slightly, allowing hyphens in the interior and up to 16 characters instead of the previous 10, to improve usability. This makes it possible to at least have names like

cluster-<date> (e.g. cluster-20250326)
slurm-test-0000, slurm-test-0001, etc.


mr0re1:
Discussed offline.

PR #3872, opened 2025-04-01
Add disk size vars for A4

parulbajaj01:
### Submission Checklist

### DIFF
--- a/examples/gke-a4/gke-a4-deployment.yaml
+++ b/examples/gke-a4/gke-a4-deployment.yaml
@@ -45,3 +45,7 @@ vars:
   # In order to target a BLOCK_NAME, extended_reservation can be inputted as
   # <project>/<reservation-name>/reservationBlocks/<reservation-block-name>
   extended_reservation:
+
+  system_node_pool_disk_size_gb: 200
+
+  a4_node_pool_disk_size_gb: 100
--- a/examples/gke-a3-ultragpu/gke-a3-ultragpu-deployment.yaml
+++ b/examples/gke-a3-ultragpu/gke-a3-ultragpu-deployment.yaml
@@ -30,6 +30,6 @@ vars:
   # extended_reservation: RESERVATION_NAME/reservationBlocks/BLOCK_NAME
   extended_reservation: RESERVATION_NAME
   static_node_count: NODE_COUNT
-  system_node_pool_disk_size_gb: # add this
-  a3ultra_node_pool_disk_size_gb: # add this
+  system_node_pool_disk_size_gb: 200
+  a3ultra_node_pool_disk_size_gb: 100
   k8s_service_account_name: # add this


ighosh98 (changes requested):
Default values are already defined in both the blueprints. I don't think these changes are need. 
Please add reasons why this change is required.

parulbajaj01:
> Default values are already defined in both the blueprints. I don't think these changes are need. Please add reasons why this change is required.

It's needed because in the public docs we have mentioned that the user can replace these variables to match the specific values in the deployement.yaml file and since we are adding the disk size vars in the public docs they will be actually needed in the deployment file.

ighosh98:
> > Default values are already defined in both the blueprints. I don't think these changes are need. Please add reasons why this change is required.
> 
> It's needed because in the public docs we have mentioned that the user can replace these variables to match the specific values in the deployement.yaml file and since we are adding the disk size vars in the public docs they will be actually needed in the deployment file.

Thanks. Could you share a link to the docs if they are public?

parulbajaj01:
> > > Default values are already defined in both the blueprints. I don't think these changes are need. Please add reasons why this change is required.
> > 
> > 
> > It's needed because in the public docs we have mentioned that the user can replace these variables to match the specific values in the deployement.yaml file and since we are adding the disk size vars in the public docs they will be actually needed in the deployment file.
> 
> Thanks. Could you share a link to the docs if they are public?

https://cloud.google.com/ai-hypercomputer/docs/create/gke-ai-hypercompute#:~:text=In%20the%20examples/gke%2Da4%2Dhighgpu/gke%2Da4%2Dhighgpu%2Ddeployment.yaml%20file%2C%20replace%20the%20following%20variables%20in%20the%20terraform_backend_defaults%20and%20vars%20sections%20to%20match%20the%20specific%20values%20for%20your%20deployment%3A

examples/gke-a4/gke-a4-deployment.yaml
```diff
@@ -45,3 +45,7 @@ vars:
   # In order to target a BLOCK_NAME, extended_reservation can be inputted as
   # <project>/<reservation-name>/reservationBlocks/<reservation-block-name>
   extended_reservation:
+
+  system_node_pool_disk_size_gb: 200
```
ighosh98, line 49: change to `# add this`

ighosh98 (approved):
general comment: we should come up with a consistent nomenclature/style. For instance adding `#add this` in every blueprint

examples/gke-a4/gke-a4-deployment.yaml
parulbajaj01: done

SwarnaBharathiMantena:
A comment above those variables describing what they are for can help, just like how the other variables have comments.

Issue #3566, opened 2025-01-20
nfs-server module bug: "auto_delete_disk" option not working


scott-nag:
### Describe the bug

[Relevant documentation](https://github.com/GoogleCloudPlatform/cluster-toolkit/tree/main/community/modules/file-system/nfs-server)

Docs state: "NOTE: auto_delete_disk is set to true in this example, which means that running terraform destroy also deletes the disk. To retain the disk after terraform destroy either set this to false or don't include the settings so it defaults to false. Note that with auto_delete_disk: false, you will need to manually delete the disk after destroying a deployment group with nfs-server."

However, the shared volume is deleted regardless of `auto_delete_disk: false` being set in blueprint.

I do notice that the docs state "WARNING: This module has only been tested against the HPC centos7 OS disk image (the default). Using other images may work, but have not been verified." and have tested using an older Centos 7 image with the same result (disk is deleted).

### Steps to reproduce

1. Create a new cluster with a share created by the nfs-server module (example below)
2. Delete the cluster
3. Once deleted the volume will be gone too despite 'auto_delete_disk' being explicitly set to false

### Expected behavior

The volume should remain for future use.

### Actual behavior

The volume is deleted.

### Version (`gcluster --version`)

```gcluster version - not built from official release
Built from 'develop' branch.
Commit info: v1.45.0-25-gf5349cac
Terraform version: 1.9.8
```

### Blueprint

If applicable, attach or paste the blueprint YAML used to produce the bug.

```
---

blueprint_name: hpc-slurm

vars:
  project_id: your-project-here
  deployment_name: hpc-slurm
  region: us-central1
  zone: us-central1-a

deployment_groups:
- group: primary
  modules:
  - id: network
    source: modules/network/vpc

  - id: homefs
    source: community/modules/file-system/nfs-server
    use: [network]
    settings:
      auto_delete_disk: false
      local_mounts: [/home]
      # instance_image:
      #   family: schedmd-slurm-21-08-8-hpc-centos-7
      #   project: schedmd-slurm-public

  - id: debug_nodeset
    source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
    use: [network]
    settings:
      node_count_dynamic_max: 4
      machine_type: n2-standard-2
      enable_placement: false
      allow_automatic_updates: false
      instance_image_custom: false
      instance_image:
        project: schedmd-slurm-public
        family: slurm-gcp-6-8-hpc-rocky-linux-8

[...]

PR #3887, opened 2025-04-05
Remove the broken auto_delete_disk system, and replace it with a working snapshot-based alternative, in the NFS Server module

ndebuhr:
The `auto_delete` parameter (and therefore `auto_delete_disk` variable) within the NFS instance's `boot_disk` block is only effective when the instance implicitly creates the boot disk (i.e., disk properties are defined directly within the instance block using `initialize_params`). Even with `auto_delete=false`, the separate `google_compute_disk` resource goes through it's own destroy lifecycle and the disk is deleted - so ultimately, the variable currently does nothing. 

This PR addresses reproducible issue https://github.com/GoogleCloudPlatform/cluster-toolkit/issues/3566, which was introduced as a regression in https://github.com/GoogleCloudPlatform/cluster-toolkit/commit/bb47067291fefb8b1cfdcac2fce5cd38d17840ef

This PR addresses the `auto_delete` issue, and more broadly expands the data retention options by:

1. Removing the non-functioning auto-delete parameterization
2. Implementing an alternative data protection system, for both the boot disk and the data disk, via `create_snapshot_before_destroy` and `create_boot_snapshot_before_destroy`

Design notes:

1. Retained the existing, heavily-customized boot disk system (i.e., the `null_resource`, custom lifecycle management, and dedicated boot `compute_disk` specification), which blocks the ability to use the boot disk `auto_delete` control.
2. While there's no working functionality to deprecate, the `auto_delete_disk` variable is nevertheless deprecated, rather than simply removed, to inform users of the snapshot-before-destroy alternative (encouraging proactive config changes to avoid accidental data deletion events). Terraform updates to add snapshot-before-destroy are quick and in-place.
3. While the `auto_delete=false` does not impact _deletion behavior_ due to the `google_compute_disk`'s separate destroy lifecycle, it is important to enable the snapshot-before-destroy feature. If the disk is deleted in connection with the VM instance (i.e., `auto_delete=true`), then the snapshot does not trigger. If the disk is deleted during it's own dedicated destroy lifecycle (i.e., `auto_delete=false`), then the snapshot does trigger. So the `auto_delete` has no impact on module deletion behavior, but it needs to be hard-coded as `auto_delete=false` for snapshot enablement.

Manually tested a variety of scenarios with Terraform v1.11.3 and Packer v1.12.0


SwarnaBharathiMantena (approved):
LGTM

tpdownes (changes requested):
@ndebuhr due to a bug in the provider, this would be a breaking change for all Terraform Google provider plugins below 6.14.0. Please modify versions.tf accordingly.

- https://github.com/hashicorp/terraform-provider-google/issues/14707
- https://github.com/hashicorp/terraform-provider-google/commit/a6bcc04f6eea8880ef023185e0073043b14b91b9

My preference would be `moved` block that puts the boot disk management entirely within the google_compute_instance resource but Terraform does not support re-factoring of a resource into an attribute within another resource.

tpdownes (approved):
LGTM. Thank you for the submission and response to change requests!

SwarnaBharathiMantena (approved):

PR #3910, opened 2025-04-10
Use `depends_on` instead of taking md5 of scripts.

mr0re1:
**Motivation:** 
* Reduce exposure to https://github.com/hashicorp/terraform-provider-google/issues/18618;
* Allow for future simpler decomposition of config files.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/main.tf
@@ -40,8 +40,6 @@ resource "random_uuid" "cluster_id" {
 ##################
 
 locals {
-  tp = "${local.bucket_dir}/" # prefix to trim from the bucket path to get a "file name"
-
   config = {
     enable_bigquery_load  = var.enable_bigquery_load
     cloudsql_secret       = var.cloudsql_secret
@@ -89,26 +87,6 @@ locals {
 
     # Providers
     endpoint_versions = var.endpoint_versions
-
-    # Extra-files MD5 hashes
-    # Makes config file creation depend on the files
-    # Allows for informed updates & checks on slurmsync side
-    slurm_gcp_scripts_md5 = google_storage_bucket_object.devel.md5hash,
-    controller_startup_scripts_md5 = {
-      for o in values(google_storage_bucket_object.controller_startup_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    nodeset_startup_scripts_md5 = {
-      for o in values(google_storage_bucket_object.nodeset_startup_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    login_startup_scripts_md5 = {
-      for o in values(google_storage_bucket_object.login_startup_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    prolog_scripts_md5 = {
-      for o in values(google_storage_bucket_object.prolog_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
-    epilog_scripts_md5 = {
-      for o in values(google_storage_bucket_object.epilog_scripts) : trimprefix(o.name, local.tp) => o.md5hash
-    }
   }
 
   x_nodeset         = toset(var.nodeset[*].nodeset_name)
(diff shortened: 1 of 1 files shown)


wiktorn (approved):
This should improve situation significantly, as `google_storage_bucket_object.devel.md5hash` is changing every time the zip file is created, as it store file metadata as well.

The other approach is to create a zip file without metadata, by using following structure:
```hcl
data "local_file" "startup_sh" {
  filename = "${path.module}/files/startup_sh_unlinted"
}

data "local_file" "slurm_gcp_devel" {
  for_each = local.slurm_gcp_devel_zip_files
  filename = "${local.scripts_dir}/${each.key}"
}

data "archive_file" "slurm_gcp_devel_zip" {
  output_path = "${local.build_dir}/${local.slurm_gcp_devel_zip}"
  type        = "zip"
  source {
    content  = data.local_file.startup_sh.content
    filename = "startup.sh"
  }

  # this approach has disadvantage of creating a zip file where all files have date modification of epoch
  dynamic "source" {
    for_each = local.slurm_gcp_devel_zip_files
    content {
      content  = data.local_file.slurm_gcp_devel[source.key].content
      filename = source.key
    }
  }
}
```

I have this in my branch, as startup.sh is not part of devel.zip, but it is used by TPU. This has side-effect of reducing the number of times, when devel.zip is uploaded.

casassg:
neat ty @mr0re1 !

tpdownes (changes requested):
I have used this simplified copy of hpc-slurm.yaml to verify by applying the blueprint once, then changing from n2-standard-2 to n2-standard-4 and re-deploy. In this case it reliably triggers the "inconsistent plan" bug. The PR branch does _not_ appear to fix this problem.

I **think** this particular problem might be manifesting because the config object needs to know the new value of the instance template URL before it is created. It then runs into the problem where the old md5 hash is known but the new md5 hash is not yet known until part way through apply.

```yaml
---

blueprint_name: hpc-slurm

vars:
  deployment_name: hpc-slurm
  region: us-central1
  zone: us-central1-a

# Documentation for each of the modules used below can be found at
# https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md

deployment_groups:
- group: primary
  modules:
  # Source is an embedded module, denoted by "modules/*" without ./, ../, /
  # as a prefix. To refer to a local module, prefix with ./, ../ or /
  - id: network
    source: modules/network/vpc

  # Private Service Access (PSA) requires the compute.networkAdmin role which is
  # included in the Owner role, but not Editor.
  # PSA is a best practice for Filestore instances, but can be optionally
  # removed by deleting the private_service_access module and any references to
  # the module by Filestore modules.
  # https://cloud.google.com/vpc/docs/configure-private-services-access#permissions
  - id: private_service_access
    source: community/modules/network/private-service-access
    use: [network]
  - id: debug_nodeset
    use: [network]
    settings:
      node_count_dynamic_max: 4
      machine_type: n2-standard-2
      allow_automatic_updates: false

  - id: debug_partition
    use:
    - debug_nodeset
    settings:
      partition_name: debug
      is_default: true

  - id: slurm_login
    source: community/modules/scheduler/schedmd-slurm-gcp-v6-login
    use: [network]
    settings:
      machine_type: n2-standard-4
      enable_login_public_ips: true

  - id: slurm_controller
    use:
    - network
    - debug_partition
    - slurm_login
    settings:
```

Further thoughts:

1. It looks like this problem might eventually be solved in this PR:

https://github.com/GoogleCloudPlatform/magic-modules/pull/12032

2. I believe this is probably better/more explicit. We need to add a comment above the `depends_on` field which spells out explicitly why we want the config storage object to be created after these the script objects. Is this to avoid a startup race condition?

3. The comment is suggestive that removal of the md5 hashes will cause problems for slurmsync.py. Maybe in fetch_config_yaml_md5? Please comment.

4. Please also address the comment by @wiktorn

mr0re1:
> It looks like this problem might eventually be solved in this PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/12032

Yes, this is PR for an issue mentioned in description. It's been "almost merged in" for some time.

> I believe this is probably better/more explicit. We need to add a comment above the depends_on field which spells out explicitly why we want the config storage object to be created after these the script objects. Is this to avoid a startup race condition?

Added

> The comment is suggestive that removal of the md5 hashes will cause problems for slurmsync.py. Maybe in fetch_config_yaml_md5? Please comment.

The removed comment is wrong, SlurmGCP doesn't make use of this information.

tpdownes (approved):
Re-tested new changes following example at https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3910#pullrequestreview-2770666610. We do not yet support changes in instance type (most likely due to instance template dependency I noted) but, if I add and remove startup_script settings on the debug nodeset, I can trigger the "inconsistent final plan" error on the develop branch while I cannot trigger it using this branch. LGTM.

alyssa-sm (approved):

Issue #3755, opened 2025-03-05
Slurm controller overwrites existing bucket's IAM policies


siddharthab:
### Describe the bug

In Slurm controller, when using an existing bucket (`create_bucket = false`), the existing bucket's IAM policy bindings are overwritten through resources `google_storage_bucket_iam_binding.viewers` and `google_storage_bucket_iam_binding.legacy_readers`.

### Steps to reproduce

1. Set up a Slurm cluster with the following controller settings:
```
      - id: slurm_controller
        source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
        use:
          ...
        settings:
          ...
          create_bucket: false
          bucket_name: ${vars.bucket_name}
          bucket_dir: gcloud/clusters/slurm
```

### Expected behavior

I would expect the existing bucket's IAM policies to be left untouched, or for the new principals to be added to the existing policies, perhaps through the use of a `google_storage_bucket_iam_member` (instead of `google_storage_bucket_iam_binding`) resource.

### Actual behavior

Terraform configuration overwrites the existing policies to remove all other principals through its use of resources `google_storage_bucket_iam_binding.viewers` and `google_storage_bucket_iam_binding.legacy_readers`.

### Version (`gcluster --version`)

gcluster version v1.46.1
Built from 'main' branch.
Commit info: v1.46.1-0-gc491a4ab2
Terraform version: 1.7.4

### Blueprint

If applicable, attach or paste the blueprint YAML used to produce the bug.

```yaml
---
blueprint_name: slurm

vars:
  deployment_name: slurm
  project_id: moonwalkbio
  region: us-central1
  zone: us-central1-c
  compute_node_machine_type: c2-standard-30
  compute_node_max_num: 30
  service_account_email: default@moonwalkbio.iam.gserviceaccount.com

deployment_groups:
  - group: cellranger

    terraform_backend:
      type: gcs
      configuration:
        bucket: moonwalkbio-terraform
        prefix: state/gcp/projects/moonwalkbio/slurm/cellranger

    modules:
      - id: network
        source: modules/network/pre-existing-vpc
        settings:
          project_id: $(vars.project_id)
          region: $(vars.region)
          network_name: default
          subnetwork_name: default

      - id: workdir
        # We don't have access to parallelstore and is likely not needed.
        # If needed, replace below filestore with commented out parallelstore.
        #source: modules/file-system/parallelstore
        #settings:
        #  local_mount: /shared
[...]

PR #3925, opened 2025-04-11
Use non-authoritative `google_storage_bucket_iam`

mr0re1:
To prevent wiping out pre-existing bindings

https://github.com/GoogleCloudPlatform/cluster-toolkit/issues/3755


tpdownes (approved):
I think this is a good change all on its own. I think it's complicating matters to use both contemporary and legacy Cloud Storage roles. I believe the legacy role is adding only the following permissions:

- storage.buckets.get
- storage.multipartUploads.list

I bet it is only "storage.buckets.get" that is doing any real work. I might consider using the "storage.bucketViewer" contemporary Cloud Storage role (which is in beta).

But I do not want to change 2 things in 1 PR. Please merge this and consider that as a follow-up task.

Reference: https://cloud.google.com/iam/docs/understanding-roles#cloud-storage-roles

PR #3958, opened 2025-04-17
Make login nodes deployable independently of "controller"

mr0re1:
**Motivation:** - allow to provision login group independently of "controller monolith" 

**Changes:**
* Move login nodes infra provisioning into `community/modules/internal/slurm-gcp/login` module;
* Remove login specific fields from stored `config.yaml`;
* Don't provision login artifacts (startup scripts) in `slurm_files` modules;
* Add separate `"${bucket}/login_group_configs/${name}.yaml"` config file for login nodes;
* Add metadata key `slurm_login_group` for VMs to self identify;
* Block login node setup on availability of both general config and group-specific config.

**Breaking change** - requires recreation of login nodes template and instances

**NOTE:** This PR could have been made into non-breaking change by:
* Don't create separate `internal/nodeset` module, but `for_each` resource creation within `controller` module - doesn't break down monolith and makes it harder to reuse in other products;
* Don't add `slurm_login_group` metadata, but rely on VM name parsing - less clean, require moderate amount of refactoring of python-code.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -644,6 +647,18 @@ def _add_nodesets(yamls: List[Any], target: dict):
...
+        lg_cfg = NSDict(lg_yaml)
+        assert lg_cfg.get("group_name"), "group_name is required"
+        lg_name = lg_cfg.group_name
+        assert lg_name not in cfg.login_groups
+        cfg.login_groups[lg_name] = lg_cfg
+
+    if instance_role() == "login":
+        group = instance_login_group()
+        if group not in cfg.login_groups:
+            raise DeffetiveStoredConfigError(f"login group '{group}' does not exist in config")
```
tpdownes, line 661: typo for `Defective`?

mr0re1: It is, but I wouldn't  address it in this PR.

rohitramu (approved):

PR #3976, opened 2025-04-19
Enable specification of system node pool zones in the GKE Cluster module

ndebuhr:
A common pattern with GKE is to use a regional control plane, but nodes in a single zone. This makes most control plane operations a zero-downtime event and offers redundancy, but keeps actual workloads zonally colocated for latency reduction, inter-zone traffic reduction, and simplified storage classes (i.e., no regional disks or volume node affinity complexities). This is especially true for HPC/GPU/TPU node pools, but is also true for "system" components.

At present, this pattern is not possible with the GKE Cluster module system node pool, as the [underlying Terraform module](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool)'s `node_locations` setting, which override's the cluster's default zones, is not exposed.

This PR introduces a variable to optionally define specific system node pool zones. As this is an optional variable with null default, the module's default behavior and existing setups are unaffected.


SwarnaBharathiMantena (approved):
LGTM

Here are terraform links for reference:
1. nodepool: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#node_locations-1
2. cluster: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#node_locations-1

parulbajaj01 (approved):
LGTM

PR #3987, opened 2025-04-23
Remove firewall for Mellanox VPCs

cdunbar13:
This PR removes or prevents firewall rules from being used in a few different VPC modules.  The changes are also reflected in the blueprints that use these VPCs.


modules/network/vpc/main.tf
```diff
@@ -169,10 +169,21 @@ locals {
   }
 }
 
+resource "terraform_data" "network_profile_firewall_validation" {
+  lifecycle {
+    precondition {
+      condition     = !(try(strcontains(var.network_profile, "roce"), false) && length(local.firewall_rules) > 0)
+      error_message = "If var.network_profile contains 'roce', var.firewall_rules must be empty and var.enable_internal_traffic must be false"
```
abbas1902, line 176: nit: Maybe we can use the same language in the [cloud docs](https://cloud.google.com/vpc/docs/rdma-network-profiles#additional_features_that_dont_apply_to_traffic_from_rdma_nics) explaining the RoCE network profile doesn't support Cloud Next Generation Firewall rules.

modules/network/gpu-rdma-vpc/variables.tf
```diff
@@ -104,34 +104,6 @@ variable "delete_default_internet_gateway_routes" {
   default     = false
 }
 
-variable "enable_internal_traffic" {
```
abbas1902, line 107: Correct me if I'm wrong but I think our normal pattern is to deprecate variables? Should we be using that here instead?

cdunbar13: Good point, addressed.

modules/network/vpc/main.tf
cdunbar13: Just added link to explanation.

abbas1902 (dismissed):
LGTM, test failures are unrelated to the changes in this PR.

abbas1902 (approved):

PR #3995, opened 2025-04-24
add non-queue flex-start support in gke

chengcongdu:
### Submission Checklist

### DIFF
--- a/modules/compute/gke-node-pool/variables.tf
+++ b/modules/compute/gke-node-pool/variables.tf
@@ -430,6 +430,23 @@ variable "enable_queued_provisioning" {
   default     = false
 }
 
+variable "enable_flex_start" {
+  description = <<-EOT
+  If true, start the node pool with Flex Start provisioning model.
+  To learn more about flex-start mode, please refer to
+  https://cloud.google.com/kubernetes-engine/docs/how-to/dws-flex-start-training and
+  https://cloud.google.com/kubernetes-engine/docs/how-to/provisioningrequest
+  EOT
+  type        = bool
+  default     = false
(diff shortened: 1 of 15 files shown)


modules/compute/gke-node-pool/variables.tf
```diff
@@ -430,6 +430,23 @@ variable "enable_queued_provisioning" {
...
+  If true, start the node pool with Flex Start provisioning model.
+  To learn more about flex-start mode, please refer to
+  https://cloud.google.com/kubernetes-engine/docs/how-to/dws-flex-start-training and
+  https://cloud.google.com/kubernetes-engine/docs/how-to/provisioningrequest
+  EOT
+  type        = bool
+  default     = false
+}
+
+variable "max_run_duration" {
```
SwarnaBharathiMantena, line 444: Question on `max_run_duration`. Isn't this supposed to be set for the job?
I am confused why this variable is being passed for a nodepool.

My understanding stems from this example: https://screenshot.googleplex.com/8EuWbywUwxiuuzi

chengcongdu: discussed offline and quote from Pawel
```
"There is a difference between DWS queued and non-queued modes. For queued mode max run duration is set per workload on ProvisioningRequest level. For non-queued mode we do not have ProvisioningRequest at all and max run duration is set on the node pool level.  This being said there is a default of one week for mex run duration for a node pool so setting it optional.
The different in config is in enable_queued_provisioning being set to true for queued mode and queued_provisioning not being set at all or being set to false for non-queued mode."
```

PR #4005, opened 2025-04-28
Fixing the kernel upgrade flag for slurm image creation

cdunbar13:
The option of `update_kernel` has never been correct but has not caused issues previously.  This PR makes that fix.


tpdownes (approved):
LGTM

PR #4045, opened 2025-05-02
Managed Lustre Data Import (Hydration)

cdunbar13:
This PR adds the capability to import data from GCS bucket upon creation of the Managed Lustre instance.

The bucket is checked for existence prior to running the import command.  

The import command has an option for a destination folder which was ignored.  This was due to the fact that we're creating a fresh Lustre instance that will not have any data, and the destination folder option requires the specified folder to exist before importing.

Note: This will fail quietly if permissions are not set up correctly.  Without the ability to check specific permissions in Terraform this is difficult to do in an automated way, so users are warned in the documentation.

This was tested (using a modified pfs-managed-lustre-vm.yaml blueprint):
1. Importing data from bucket within the same project
2. Importing data from another project
3. No data import
4. Non-existing bucket

### DIFF
--- a/modules/file-system/managed-lustre/main.tf
+++ b/modules/file-system/managed-lustre/main.tf
@@ -29,11 +29,12 @@ data "google_compute_network_peering" "private_peering" {
 }
 
 locals {
-  server_ip     = split(":", google_lustre_instance.lustre_instance.mount_point)[0]
-  remote_mount  = split(":", google_lustre_instance.lustre_instance.mount_point)[1]
-  fs_type       = "lustre"
-  mount_options = var.mount_options
-  instance_id   = var.name != null ? var.name : "${var.deployment_name}-${random_id.resource_name_suffix.hex}"
+  server_ip        = split(":", google_lustre_instance.lustre_instance.mount_point)[0]
+  remote_mount     = split(":", google_lustre_instance.lustre_instance.mount_point)[1]
+  fs_type          = "lustre"
+  mount_options    = var.mount_options
+  instance_id      = var.name != null ? var.name : "${var.deployment_name}-${random_id.resource_name_suffix.hex}"
+  destination_path = "/"
 
   install_managed_lustre_client_runner = {
     "type"        = "shell"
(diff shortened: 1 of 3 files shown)


modules/file-system/managed-lustre/main.tf
```diff
@@ -47,6 +48,15 @@ locals {
     "args"        = "\"${local.server_ip}\" \"${local.remote_mount}\" \"${var.local_mount}\" \"${local.fs_type}\" \"${local.mount_options}\""
     "destination" = "mount${replace(var.local_mount, "/", "_")}.sh"
   }
+
+  importing_data = try(length(var.import_gcs_bucket_uri) > 0, false)
+  bucket_count   = try(length(data.google_storage_bucket.lustre_import_bucket), 0)
+}
+
+data "google_storage_bucket" "lustre_import_bucket" {
+  count = local.importing_data ? 1 : 0
```
harshthakkar01, line 57: count = <condition> ? 1 : 0

modules/file-system/managed-lustre/main.tf
```diff
@@ -29,11 +29,12 @@ data "google_compute_network_peering" "private_peering" {
...
-  remote_mount  = split(":", google_lustre_instance.lustre_instance.mount_point)[1]
-  fs_type       = "lustre"
-  mount_options = var.mount_options
-  instance_id   = var.name != null ? var.name : "${var.deployment_name}-${random_id.resource_name_suffix.hex}"
+  server_ip        = split(":", google_lustre_instance.lustre_instance.mount_point)[0]
+  remote_mount     = split(":", google_lustre_instance.lustre_instance.mount_point)[1]
+  fs_type          = "lustre"
+  mount_options    = var.mount_options
+  instance_id      = var.name != null ? var.name : "${var.deployment_name}-${random_id.resource_name_suffix.hex}"
+  destination_path = "/"
```
harshthakkar01, line 37: we wont expose custom local mount ?

harshthakkar01, line 37: data will be imported to `local_mount` or root directory ?

cdunbar13: The destination path is under the local_mount.  The `/` is required to specify `<local_mount>/`.  As of right now users shouldn't notice this unless they start digging in the terraform (nothing exposed in variables).

harshthakkar01: okay, so default for local_mount = /shared. and the data will also be imported to /shared directory ?

modules/file-system/managed-lustre/main.tf
cdunbar13: Done

harshthakkar01 (approved):

PR #4057, opened 2025-05-05
Allow deploying cluster without live reservation

vikramvs-gg:
### Submission Checklist


ighosh98 (approved):

modules/compute/gke-node-pool/main.tf
```diff
@@ -298,7 +295,7 @@ resource "google_container_node_pool" "node_pool" {
       condition = (
         (local.input_specific_reservations_count == 0) ||
         (local.input_specific_reservations_count == 1 &&
-          length(local.verified_specific_reservations) > 0 &&
+          (length(local.verified_specific_reservations) > 0 || !var.is_reservation_active) &&
```
SwarnaBharathiMantena, line 298: I think we can narrow down this logic.

If (local.input_specific_reservations_count == 1), (length(local.verified_specific_reservations) can be 1 at max. So, can we make this (length(local.verified_specific_reservations) == 1 ?

Please share your thought on whether it is worth the update.

vikramvs-gg: I checked the code and yes, we allow only 1 reservation, hence it is fine to incorporate the above simplification of condition.We accept list of reservations as per object definition, probably that's why > 0 check was put at first place for verified reservations. But, given that only 1 reservation is supported, it makes sense to simplify the logic for better readability. I've modified the logic as per above and tested the scenarios of cluster provisioning working fine.

SwarnaBharathiMantena (approved):
LGTM

PR #4069, opened 2025-05-06
Better integrate, optimize, and control the gIB NCCL RDMA plugin installer

ndebuhr:
This PR enables faster NCCL RDMA upgrades, a subtle warning for gIB version changes, and integration of the gIB installer code into the `kubectl-apply` module.

Manually tested changes and a variety of scenarios using Terraform v1.11.4.

## Faster NCCL RDMA installation

Previously, when the gIB NCCL plugin version was changed, the rolling update strategy used a `maxUnavailable` of 1 (the default value). This means that each node is updated one-by-one. With image pull times often exceeding 10 seconds, the rollout can leave large clusters in a broken, mixed-NCCL state for hours. Increasing the `maxUnavailable` to 16 reduces this rollout time by over an order of magnitude, while still maintaining the gradual/rolling update characteristics.

## Subtle warning for gIB version changes

An environment with inconsistent NCCL versions can lead to tricky errors. As such, NCCL version changes need to be handled carefully. While the DaemonSet rollout and installer script should fully and cleanly update the environment when the installer image is upgraded, potential unmanaged NCCL environment variables/configurations and other "residuals" could cause issues. This is not so large a concern as to warrant deployment blocking, so a `check` block is added to give a warning when the gIB version changes. Like other `check` blocks in the repo, this warning is wedged between the terraform apply logs and the blueprint outputs after a `gcluster deploy`. I don't believe the warning needs more emphasis than this.

## Integration of the gIB installer code

While largely a prerequisite for the previously-mentioned changes, I believe there's some value in this in it's own right. The integration of the gIB installer code into the `kubectl-apply` module seems like a cleaner and easier UX, while still fully allowing for the old approach of custom `apply_manifests`. The NCCL RDMA installer manifests across the a3-ultra and a4 blueprints are also identical, other than the `gke-accelerator` node affinity, so there's a bit of code deduplication value here.

There's strong precedent for this sort of shift in the `kubectl-apply` module already (Jobset, Kueue,. etc.).

## Additional design notes

1. These changes do not affect default behavior of existing modules (`gib.install` is default `false`) or blueprints (`apply_manifests` works like it always has). 
2. Loosening up the `template_vars` data type (`kubectl-apply` module) was required for the gIB NCCL installer integration, but I believe this is a sensible change in it's own right to enable more advanced templating on top of the manifests. Since the data typing was _relaxed_, rather than changed, this should not introduce any issues.
3. The default accelerator selector is set to the superset of all relevant accelerators (a3-ultra and a4) so that the setup still works if an `accelerator_selector` variable is not supplied. However, an `accelerator_selector` setting is explicitly provided in the a3-ultra and a4 blueprints for clarity.

## Submission Checklist


ighosh98 (changes requested):
The nccl-installer is generally tied to the machine family architecture. I don't think we should be moving it into the manifests folder. For instance we may have a different number of GPUs in a future machine family which can [impact parts of the nccl installer](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/gke-a4/nccl-installer.yaml#L65-L72)

For example: Take a look at this [nccl installer](https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/refs/heads/master/gpudirect-rdma/nccl-rdma-installer-a4x.yaml)
We need to think of a different solution here.

modules/management/kubectl-apply/main.tf
```diff
@@ -222,3 +224,18 @@ module "install_gpu_operator" {
   cleanup_on_fail = true
 
 }
+
+module "install_gib" {
```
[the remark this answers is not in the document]
ndebuhr: No - it's just the single installer manifest.

ndebuhr:
Thanks @ighosh98. Future compatibility around subshapes and cpu architectures is not a lot of parameterization in my opinion - added and tested a couple of additional commits, and they cover the full installer diffs.

I personally think this now represents the right amount of configurability, but with significant code similarities/duplication making this still worth templatizing in the `kubectl-apply` module. We can always push folks to the existing `apply_manifests` for weird one-off cases.

Thoughts?

ighosh98:
> Thanks @ighosh98. Future compatibility around subshapes and cpu architectures is not a lot of parameterization in my opinion - added and tested a couple of additional commits, and they cover the full installer diffs.
> 
> I personally think this now represents the right amount of configurability, but with significant code similarities/duplication making this still worth templatizing in the `kubectl-apply` module. We can always push folks to the existing `apply_manifests` for weird one-off cases.
> 
> Thoughts?

I agree with the idea to templatize the nccl installer. I would prefer that we can define the path to be tied strictly to a machine family and have unique nccl installers per machine family. The two obvious solutions are:
1. We make the apply manifest block a bit more generic and allow users to pass params corresponding to the manifest as well. 
2. We need to have a mechanism to dynamically identify the folder path for a machine family the user wants to deploy. 

Would have to think how to implement one of these two solutions.

ndebuhr:
@ighosh98 Sounds good to me! Revised and retested PR is ready for your review.

modules/management/kubectl-apply/variables.tf
```diff
@@ -105,3 +131,48 @@ variable "nvidia_dra_driver" {
...
+          nodeSelectorTerms = [{
+            matchExpressions = [{
+              key      = "cloud.google.com/gke-gpu",
+              operator = "In",
+              values   = ["true"]
+            }]
+          }]
+        }
+      })
+      accelerator_count = optional(number, 8)
```
ighosh98, line 154: Let's make this a mandatory parameter and not assign a default value

modules/management/kubectl-apply/variables.tf
```diff
@@ -105,3 +131,48 @@ variable "nvidia_dra_driver" {
...
+          nodeSelectorTerms = [{
+            matchExpressions = [{
+              key      = "cloud.google.com/gke-gpu",
+              operator = "In",
+              values   = ["true"]
+            }]
+          }]
+        }
+      },
+      accelerator_count = 8
```
ighosh98, line 169: I don't think we should hard code this.

modules/management/kubectl-apply/variables.tf
```diff
@@ -105,3 +131,48 @@ variable "nvidia_dra_driver" {
...
+              operator = "In",
+              values   = ["true"]
+            }]
+          }]
+        }
+      })
+      accelerator_count = optional(number, 8)
+      }), {
+      image   = "us-docker.pkg.dev/gce-ai-infra/gpudirect-gib/nccl-plugin-gib",
+      version = "v1.0.5",
```
ighosh98, line 157: Let's not hard code this. Let's use config_template_var instead.

ighosh98 (changes requested):
Instead of hard coding values, please use the config_template_var constuct. We have used a similar solution [for wiring Kueue configurations](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/gke-a4/gke-a4.yaml#L240-L242)

Rest looks good to me.

ighosh98:
Sharing code pointers on how Kueue integration implemented the construct:
https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/management/kubectl-apply/main.tf#L81-L94
https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/management/kubectl-apply/main.tf#L43-L56

examples/gke-a4/nccl-installer.yaml.tftpl
```diff
@@ -25,6 +25,8 @@ spec:
       k8s-app: nccl-rdma-installer
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 16
```
ighosh98, line 29: Curious to know why this is 16

ighosh98 (approved):
LGTM
Let's wait for the tests to pass before merging.

ndebuhr:
@ighosh98 Thanks for the comments and call. Made that last change that we discussed - to force specification of the gIB version and accelerator count in the blueprint.

examples/gke-a4/nccl-installer.yaml.tftpl
ndebuhr: Somewhat arbitrary. 1 (the current/default value) is extremely slow. Approximately one order of magnitude seems like a good improvement, but two orders of magnitude might potentially strain networking resources. So, I went for a multiple of 8 (given typical block sizes) that was about one order of magnitude.

ighosh98 (approved):
Will wait for all tests to pass before we merge the PR.

SwarnaBharathiMantena (approved):

PR #4080, opened 2025-05-07
Adding DCGM test to gpu tests for slurm

cdunbar13:
Add test to ensure that the Datacenter Gpu Manager is operational on the Accelerator nodes

Tested on Cloud build with local folder, will run related A* tests.

Tested failures by running the gpu tests on a non accelerator test (slurm-v6-ubuntu) to make sure the return code causes a failure.


tpdownes (approved):
Please observe all tests pass on A3/A4 series VM families

PR #4081, opened 2025-05-07
split kubectl installation into two parts to reduce race conditions

ighosh98:
### Submission Checklist
kubectl installation will start only after cluster creation is complete to reduce the race conditions we have observed.

### DIFF
--- a/modules/management/kubectl-apply/main.tf
+++ b/modules/management/kubectl-apply/main.tf
@@ -41,8 +41,9 @@ data "google_container_cluster" "gke_cluster" {
 data "google_client_config" "default" {}
 
 module "kubectl_apply_manifests" {
-  for_each = local.apply_manifests_map
-  source   = "./kubectl"
+  for_each   = local.apply_manifests_map
+  source     = "./kubectl"
+  depends_on = [var.gke_cluster_exists]
 
   content           = each.value.content
   source_path       = each.value.source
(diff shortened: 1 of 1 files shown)


modules/management/kubectl-apply/main.tf
```diff
@@ -60,6 +61,7 @@ module "install_kueue" {
   source            = "./kubectl"
   source_path       = local.install_kueue ? local.kueue_install_source : null
   server_side_apply = true
+  depends_on        = [var.gke_cluster_exists]
```
mohitchaurasia91, line 64: This looks good, can we also add dependency chaining to perform kubectl apply in a deterministic sequence.

modules/management/kubectl-apply/main.tf
```diff
@@ -41,8 +41,9 @@ data "google_container_cluster" "gke_cluster" {
 data "google_client_config" "default" {}
 
 module "kubectl_apply_manifests" {
-  for_each = local.apply_manifests_map
-  source   = "./kubectl"
+  for_each   = local.apply_manifests_map
+  source     = "./kubectl"
+  depends_on = [var.gke_cluster_exists]
```
SwarnaBharathiMantena, line 46: Where is this var defined? Is it the output of the gke-cluster module?

ighosh98: Yes. Defined in `variables.tf` and passed from gke-cluster module

modules/management/kubectl-apply/main.tf
ighosh98: Is this required? I think we have dependencies established for everything that have interdependencies as of now.

modules/management/kubectl-apply/main.tf
ighosh98: That PR was already merged a while back to fix a different bug

modules/management/kubectl-apply/main.tf
ighosh98: Added dependency of jobset on kueue configure

mohitchaurasia91 (approved):

PR #4091, opened 2025-05-09
Improve API consumption of `_handle_bulk_insert_op`

mr0re1:
* Instead of doing `globalOperations.aggregateList` perform series (usually just 1) of zonal list requests - to stop hitting `HeavyWeightReadRequestsPerMinutePerProject` limits;
* Do not perform server-side filtering on non-indexed fields, instead do it client-side to avoid `ListRequestsFilterCostOverheadPerMinutePerProjectPerRegion` limits;
* Incorporate `createTime` range and `user` (both are indexed) into server-side filtering.

New approach consumes 0 - `filtered_list_cost_overhead`, while old one was effectively churring (on server side) through all operations.

Latency is smaller but not significantly (~2 times improvement) - **no thorough sampling has been done**.

See: https://cloud.google.com/compute/api-quota#api-rate-limits

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Optional, Dict
+from typing import List, Optional, Dict, Any
 import argparse
 from datetime import timedelta
 import shlex
@@ -33,7 +33,6 @@ from util import (
     chunked,
     ensure_execute,
     execute_with_futures,
-    get_insert_operations,
     log_api_request,
     map_with_futures,
     run,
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Optional, Dict
+from typing import List, Optional, Dict, Any
```
tpdownes, line 18: It's not clear to me why you are changing the type hint for the `op` input. Is there an advantage?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -375,6 +374,40 @@ def resume_nodes(nodes: List[str], resume_data: Optional[ResumeData]):
...
+        req = act.list_next(req, result)
+    return ops
+
+
+def _get_failed_instance_inserts(bulk_op: Any, lkp: util.Lookup) -> list[Any]:
+    zones = set() # gather zones that had failed inserts
+    for loc, stat in bulk_op.get("instancesBulkInsertOperationMetadata", {}).get("perLocationStatus", {}).items():
+        pref, zone = loc.split("/", 1)
+        if not pref == "zones":
+            log.error(f"Unexpected location: {loc} in operation {bulk_op['name']}")
```
tpdownes, line 401: What does this key look like for regional bulkInsert requests? Do we have test coverage for that?

tpdownes (commented):
I'm concerned we are missing something in the case when multiple allowed zones are set by the user.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
mr0re1: e.g. : `zones/us-central1-a`

tpdownes (approved):

PR #4092, opened 2025-05-09
Update Open Front End Django dependency and fix OFE integration test

scott-nag:
#4085 django update and fix for ofe test


tpdownes (approved):
Thank you for helping us fix the integration test!

rohitramu (approved):

PR #4138, opened 2025-05-17
Drop CentOS 7 support for Ansible installation

tpdownes:
This PR removes all support and test coverage for installing Ansible into VMs running CentOS 7. CentOS 7 has been EOL since 30 June 2024. The version of Ansible we are using is blocked by the old dependencies available within CentOS 7 so this is a first step in updating Ansible to a more recent version.

Additionally takes the first steps toward a more "fail open" model where new releases of RedHat derivatives may continue to work without modification.

Blocks on #4137.

### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #4139, opened 2025-05-17
Upgrade Ansible to maximum allowed version on oldest supported OS distributions

tpdownes:
This PR implements an in-place upgrade of Ansible after support for CentOS 7 was dropped in #4138. When reviewing our supported operating systems, the minimum Python we must support is Python 3.8 from Ubuntu 20.04. This imposes a set of [version limits](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) which can be summarized as:

- ansible-core 2.13
- ansible 6.x (which depends on ansible-core 2.13)

Ubuntu 20.04 itself will be EOL in 2 weeks on 30 May 2025. The a3-highgpu-8g solution still has dependencies upon Ubuntu 20.04, however, so we cannot immediately drop support for it. Once we transition the A3 High solution to a new OS, our new limit will be imposed by Debian 11 (Python 3.9).

To obtain support for Python 3.8 and above, we must install a newer version of Python in Rocky Linux 8 which includes Python 3.6 in a base installation. Rocky Linux 9 includes Python 3.9 by default, which should suffice, however this PR makes the affirmative choice to update Rocky Linux 8 and 9 to include Python 3.12 _but retain Python 3.6 and 3.9 (respectively) as the default Python 3 interpeter for users_. The reasoning behind this:

- The oldest supported release of Ansible requires Python 3.10 so this is a step we will need to take anyhow on both Rocky Linux 8+9
- Python 3.12 aligns with RedHat Enterprise Linux 10 which was released on 13 May 2025

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/files/install_ansible.sh
+++ b/modules/scripts/startup-script/files/install_ansible.sh
@@ -14,12 +14,12 @@
 # limitations under the License.
 
 set -ex
-REQ_ANSIBLE_VERSION=2.11
-REQ_ANSIBLE_PIP_VERSION=4.10.0
-REQ_PIP_WHEEL_VERSION=0.37.1
-REQ_PIP_SETUPTOOLS_VERSION=59.6.0
-REQ_PIP_MAJOR_VERSION=21
-REQ_PYTHON3_VERSION=6
+REQ_ANSIBLE_VERSION=2.13
+REQ_ANSIBLE_PIP_VERSION=6.7.0
+REQ_PIP_WHEEL_VERSION=0.45.1
+REQ_PIP_SETUPTOOLS_VERSION=75.3.2
+REQ_PIP_MAJOR_VERSION=25
+REQ_PYTHON3_VERSION=8
 
 apt_wait() {
 	while fuser /var/lib/apt/lists/lock >/dev/null 2>&1; do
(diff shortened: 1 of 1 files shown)


modules/scripts/startup-script/files/install_ansible.sh
```diff
@@ -66,14 +66,26 @@ get_python_minor_version() {
...
 	if grep -qi 'ID="rhel"' /etc/os-release; then
-		# Do not set --disablerepo / --enablerepo on RedHat, due to complex repo names
-		# clear array
+		# Do not set --disablerepo / --enablerepo on RedHat, due to
+		# complex repo names; clear array
 		set --
 	fi
-	dnf install "$@" -y python3 python3-pip
-	python_path=$(command -v python3)
+	# On Rocky Linux 9, Python 3.9 is installed by default but this
```
cdunbar13, line 76: This seems like general information (and the second part is duplicated in two places).  Is it worth just putting this at the top?

tpdownes: I would like to eventually simplify this script so that `apt` and `dnf` operations appear in one place. In the interim, I think it's helpful to have the comment wherever the operation appears.

PR #4157, opened 2025-05-20
Auto execute basic Slurm integration test

mr0re1:
Do not require approval for `PR-test-slurm-gcp-v6-simple-job-completion`,
intention is to make it required for PR merging.
This test: doesn't require Filestore, takes ~10 minutes.


tpdownes (approved):
As future work we can consider triggering based on files.

PR #4193, opened 2025-05-26
Slurm: disable public ips

wiktorn:
* disable public IPs for login and controller, as network module provides NAT and private google access, and this is better starting point from security perspective
* provide hints about permissions for other service accounts


mr0re1 (dismissed):

Neelabh94:
Closing since these changes were merged via PR #5950 .

PR #4264, opened 2025-06-10
Add functionality to export slurm config bucket as controller NFS

mr0re1:


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -181,7 +181,7 @@ def create_instances_request(nodes: List[str], placement_group: Optional[str], e
         ),
     )
 
-    if placement_group:
+    if placement_group and excl_job_id is not None:
```
abbas1902: Curious what the motivation is for this, we don't want to enforce "all or nothing" provisioning for non-job associated requests (but still apart of a placement group)?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1066,18 +1071,20 @@ def backoff_delay(start, timeout=None, ratio=None, count: int = 0):
...
     try:
-        resp = get_url(url, headers=HEADERS)
+        resp = requests_lib.get(url, headers=HEADERS)
         resp.raise_for_status()
         return resp.text
-    except RequestException:
-        log.debug(f"metadata not found ({url})")
-        raise Exception(f"failed to get_metadata from {url}")
+    except requests_lib.exceptions.HTTPError:
+        log.error(f"metadata not found ({url})")
```
abbas1902: nit: If it's an HTTPError could we log the response msg/status code as well? Or will that visibility come from elsewhere?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
mr0re1: one needs to check changes before committing , thank you for spotting.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
mr0re1: done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
```diff
@@ -78,74 +94,46 @@ def resolve_network_storage(nodeset=None) -> List[NSMount]:
...
-        server_ip = mount.server_ip.split("@")[0]
-        mount_addr = util.host_lookup(server_ip)
-        return mount_addr == lookup().control_host_addr
-
-    return separate(internal_mount, mounts)
 
+def is_controller_mount(mount) -> bool:
+    # NOTE: Valid Lustre server_ip can take the form of '<IP>@tcp'
+    server_ip = mount.server_ip.split("@")[0]
+    mount_addr = util.host_lookup(server_ip)
```
abbas1902, line 106: nit: Took me a second to realize this is not a mount for the controller but rather a mount that is mounted to the controller (akin to internal mount). So the mounts that you separate in L115 are explicitly the non-controller mounts (a.k.a external mounts in the old code). Thought perhaps the logic had accidentally flipped from the old code but it's all still there. 

I wonder if keeping the name `internal_mount` would make it more obvious. Not really sure.

PR #4275, opened 2025-06-13
Revamp install_cloud_rdma_drivers startup script

abbas1902:
Recently Rocky-8 broke compatibility with its older kernel versions. The RDMA kmod repo had to be updated with new compliant versions and was pushed out however when reloading the kmods this results in a broken setup since the new kmod packages are not compatible with older kernels  (from the HPC VM Image) after reloading idpf/rdma modules.

Options are (1) Update kernel modules + kmod packages so they are both inline with each other, essentially turn the startup script to a one line `dnf update` plus reboot.

(2) Do not update anything for the time being

Since the HPC VM Image is relatively recent we are opting for option 2 for the time being. To be couple with new Slurm images to make use of the new HPC VM image.

### Submission Checklist

### DIFF
--- a/modules/scripts/startup-script/README.md
+++ b/modules/scripts/startup-script/README.md
@@ -326,7 +326,7 @@ No modules.
 | <a name="input_http_proxy"></a> [http\_proxy](#input\_http\_proxy) | Web (http and https) proxy configuration for pip, apt, and yum/dnf and interactive shells | `string` | `""` | no |
 | <a name="input_install_ansible"></a> [install\_ansible](#input\_install\_ansible) | Run Ansible installation script if either set to true or unset and runner of type 'ansible-local' are used. | `bool` | `null` | no |
 | <a name="input_install_cloud_ops_agent"></a> [install\_cloud\_ops\_agent](#input\_install\_cloud\_ops\_agent) | Warning: Consider using `install_stackdriver_agent` for better performance. Run Google Ops Agent installation script if set to true. | `bool` | `false` | no |
-| <a name="input_install_cloud_rdma_drivers"></a> [install\_cloud\_rdma\_drivers](#input\_install\_cloud\_rdma\_drivers) | If true, will install and reload Cloud RDMA drivers. Currently only supported on Rocky Linux 8. | `bool` | `false` | no |
+| <a name="input_install_cloud_rdma_drivers"></a> [install\_cloud\_rdma\_drivers](#input\_install\_cloud\_rdma\_drivers) | If true, will install and reload Cloud RDMA drivers. Currently only supported on Rocky Linux 8. Should not be enabled if using the HPC VM Image. | `bool` | `false` | no |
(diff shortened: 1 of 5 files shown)


nadig-google (approved):
Approving

modules/scripts/startup-script/README.md
```diff
@@ -326,7 +326,7 @@ No modules.
 | <a name="input_http_proxy"></a> [http\_proxy](#input\_http\_proxy) | Web (http and https) proxy configuration for pip, apt, and yum/dnf and interactive shells | `string` | `""` | no |
 | <a name="input_install_ansible"></a> [install\_ansible](#input\_install\_ansible) | Run Ansible installation script if either set to true or unset and runner of type 'ansible-local' are used. | `bool` | `null` | no |
 | <a name="input_install_cloud_ops_agent"></a> [install\_cloud\_ops\_agent](#input\_install\_cloud\_ops\_agent) | Warning: Consider using `install_stackdriver_agent` for better performance. Run Google Ops Agent installation script if set to true. | `bool` | `false` | no |
-| <a name="input_install_cloud_rdma_drivers"></a> [install\_cloud\_rdma\_drivers](#input\_install\_cloud\_rdma\_drivers) | If true, will install and reload Cloud RDMA drivers. Currently only supported on Rocky Linux 8. | `bool` | `false` | no |
+| <a name="input_install_cloud_rdma_drivers"></a> [install\_cloud\_rdma\_drivers](#input\_install\_cloud\_rdma\_drivers) | DEPRECATED: Please do not enable to this startup script as it is no longer necessary. | `bool` | `null` | no |
```
rohitramu: What if the customer is not using the HPC VM image (or one derived from it)?

abbas1902: The original reason for the removal of this script is we essentially cannot protect against incompatibility between kernel/irdma packages if we indiscriminately update the packages all the time. ABI breakages are rare but we'd ideally not want to run the risk of it happening again if possible.

rohitramu: By removing this functionality, it would mean Cluster Toolkit no longer supports RDMA (only the HPC VM image does).

Isn't the better solution to improve the script?

abbas1902: Correct, it would mean we no longer support automated RDMA installations. We would expect it to come from the image. Always promoting to the latest version of the rdma package irrespective of the kernel and reloading the idpf/rdma modules is considered a "risky" operation. Either we tie the rdma packages to a known compatible version of the kernel (which would require maintenance), or we run the risk of encountering a mis-match which leads to a system breakage.

Alternative is doing a full dnf update (to also bump the kernel version), and rebooting the VMs all together. But that would add significant amounts of setup time.

^ I guess that's better than nothing, but I'll still disable this script by default, user can enable if they so choose too.

rohitramu: Discussed offline: maintaining an RDMA driver installation script is currently not feasible due to frequent breaking changes.  It makes sense to remove this functionality and put the responsibility on the VM image.

PR #4276, opened 2025-06-13
SlurmGCP Resume Improvements 

alyssa-sm:
This PR:
- Increases the Message timeout to give commands more time to complete while the controller is under load
- Implements a wrapper for resume.py so it doesn't block slurmctld
- Relaxes All or Nothing Bulk Insert Policy

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume_wrapper.sh
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume_wrapper.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e -o pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+PYTHON_SCRIPT="${SCRIPT_DIR}/resume.py"
+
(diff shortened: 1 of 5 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume_wrapper.sh
```diff
@@ -0,0 +1,42 @@
...
+	SAFE_DIR="/tmp/slurm_resume_data"
+	mkdir -p "$SAFE_DIR"
+
+	RESUME_FILE="${SAFE_DIR}/resumedata.json"
+	cp "$SLURM_RESUME_FILE" "$RESUME_FILE"
+
+	EXTRA_ARGS+=(--resume-file "$RESUME_FILE")
+fi
+
+"${PYTHON_SCRIPT}" "${ALL_ARGS[@]}" "${EXTRA_ARGS[@]}" &
```
mr0re1: nit. It's not clear why do we need to pass `--resume-file` as an explicit arg. Can we do something like this:

```sh
SLURM_RESUME_FILE="${RESUME_FILE}" "${SCRIPT_DIR}/resume.py" $@ &
disown
```

This way we could avoid any changes in python code.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume_wrapper.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Capture all arguments passed by Slurm (the nodelist).
+ALL_ARGS=("$@")
+
+# This array will hold extra argument for resume.py, like the resume data file.
+EXTRA_ARGS=()
+
+# Handle SLURM_RESUME_FILE if provided
+if [ -n "${SLURM_RESUME_FILE-}" ] && [ -f "$SLURM_RESUME_FILE" ]; then
+	SAFE_DIR="/tmp/slurm_resume_data"
```
mr0re1, line 30: That would effectively override previous `resumedata.json`, lets use unique names for those files.

alyssa-sm: Made them unique.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume_wrapper.sh
alyssa-sm: Yes, I changed it to reflect this.

PR #4287, opened 2025-06-17
Adds prevalidation for missing roles

vikramvs-gg:
### Submission Checklist

### DIFF
--- a/modules/internal/iam-validator/main.tf
+++ b/modules/internal/iam-validator/main.tf
@@ -1,0 +1,121 @@
+# Copyright 2025 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Data source that runs the permission check script in-line.
+data "external" "gcp_permission_check" {
+  program = [
+    "/bin/bash",
+    "-c",
+    # The script is updated to output a clean list of missing permissions on failure.
+    <<-EOT
+      set -e
+      if [ "$#" -ne 2 ]; then
+        echo "Usage: $0 <project_id> <permissions_comma_separated>" >&2; exit 1
+      fi
+
+      PROJECT_ID="$1"
+      PERMISSIONS_TO_CHECK="$2"
+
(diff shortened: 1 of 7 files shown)


modules/scheduler/gke-cluster/main.tf
```diff
@@ -62,6 +62,19 @@ locals {
...
+    "container.clusters.create",
+    "container.clusters.delete",
+    "compute.instances.create",
+    "compute.instances.delete",
+    "storage.buckets.create",
+    "iam.serviceAccounts.create",
+    "iam.serviceAccounts.setIamPolicy",
+    "iam.serviceAccounts.actAs",
+    "serviceusage.services.use"
+  ]
```
SwarnaBharathiMantena, line 77: Are all the roles related to storage and serviceAccounts going to be checked for in the cluster module?

Is there a plan to split the tests to the relevant modules? For example, the storage module will test whether the user has roles to create the storage. In this case, the error will still not show up until the deployment starts working with the storage module.

modules/internal/iam-validator/main.tf
```diff
@@ -0,0 +1,121 @@
...
+    <<-EOT
+      set -e
+      if [ "$#" -ne 2 ]; then
+        echo "Usage: $0 <project_id> <permissions_comma_separated>" >&2; exit 1
+      fi
+
+      PROJECT_ID="$1"
+      PERMISSIONS_TO_CHECK="$2"
+
+      TOKEN=$(gcloud auth print-access-token)
```
SwarnaBharathiMantena, line 30: This token generated is going to be encrypted, correct?

modules/internal/iam-validator/main.tf
```diff
@@ -0,0 +1,121 @@
...
+    ,
+    "permission_checker.sh",
+    var.project_id,
+    join(",", var.required_permissions)
+  ]
+}
+
+locals {
+  # A map to suggest roles for missing permissions.
+  permission_to_role_map = {
```
SwarnaBharathiMantena, line 75: One suggestion here: Instead of hard coding this mapping, we can get the mapping info through terraform API: https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/iam_role

This adds to the complexity. So, doing this as part of another PR works as well.

modules/scheduler/gke-cluster/main.tf
ighosh98: +1 We should make this piece into a generic data structure so that any list of modules can be checked. Something along the lines of a map with `key=moduleName` and `value=list[]` where the list contains all the permissions. Then re-use this generic data structure through the blueprint. Feel free to put together a short design document to flesh out the idea and alternatives to it.

ighosh98 (changes requested):
Let's try to make a generic re-usable data structure instead of hard coding the values. Validations for all modules should exist and be done at the start of the cluster provisioning lifecycle.

PR #4312, opened 2025-06-24
Add Kueue 0.12.2 and make it the new default

mwysokin:
This PR makes the following changes:
- Adds support for Kueue 0.12.2 (added version support + manifest with TAS enabled)
  - Some of the highlights of the 0.12 release are:
    - TAS and DWS (through ProvisioningRequest API) support
    - Added optional garbage collector for deactivated workloads
    - Admission Fair Sharing
    - Node hot-swap for TAS workloads
- Removes support for 0.9 and 0.8 versions of Kueue which are no longer supported and would crash with Topology Aware Scheduling feature gate enabled.

### Submission Checklist

### DIFF
--- a/modules/management/kubectl-apply/manifests/kueue-v0.12.2.yaml
+++ b/modules/management/kubectl-apply/manifests/kueue-v0.12.2.yaml
@@ -1,3 +1,17 @@
+# Copyright 2025 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 apiVersion: v1
 kind: Namespace
 metadata:


ighosh98 (changes requested):
This manifest would lead to deployment failures. Please add these code changes appropriately and deploy a cluster using toolkit and confirm if the TAS tests are passing as expected:
1. https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/modules/management/kubectl-apply/manifests/kueue-v0.11.4.yaml#L12626
2. https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/modules/management/kubectl-apply/manifests/kueue-v0.11.4.yaml#L12688-L12691

Please add integration test changes if required.

mwysokin:
@ighosh98 So I believe the changes you mentioned are there already:

Screenshot 2025-06-30 at 13 54 23

and

Screenshot 2025-06-30 at 13 55 30

Or did I miss something?

ighosh98 (dismissed):

ighosh98 (dismissed):

modules/management/kubectl-apply/variables.tf
```diff
@@ -15,7 +15,7 @@
   */
 
 locals {
-  kueue_supported_versions  = ["v0.11.4", "v0.10.1", "v0.10.0", "v0.9.1", "v0.9.0", "v0.8.1"]
```
ighosh98, line 18: @mwysokin Please remove the kueue_versions we don't want to support

ighosh98: Also, can you explain why we would want to deprecate the older versions?

mwysokin: These versions are so old that not only we no longer support them but also at this point we have no idea what was there. There have been so many bug fixes and changes in the last 6 months that we'd like our customers to be close to the most recent versions.

shubpal07 (commented):
Great work on this upgrade! 

I have a couple of quick questions:
1.  In Kueue v0.12.2, is the `TopologyAwareScheduling` feature considered Beta or GA?
2.  Did you notice any significant change in the CPU/memory footprint with this new version?

Thanks

ighosh98 (dismissed):

ighosh98 (approved):

shubpal07 (approved):

PR #4316, opened 2025-06-25
chore/allow hyphens in partition_name and slurm_cluster_name, increase max length to 20 for slurm_cluster_name

rbekhtaoui:
In our setup, we have a Slurm cluster name that exceeds 9 characters and includes hyphens (-). 
Similarly, some of our partition names also contain hyphens. 

These naming patterns were not passing the existing validation rules in the Terraform module.

**Changes made:**

- Updated the regex validation to allow hyphens (-) in both partition_name and slurm_cluster_name.
- Increased the maximum allowed length for slurm_cluster_name to 20 characters (previously limited to 9).

### DIFF
--- a/community/modules/compute/gke-nodeset/variables.tf
+++ b/community/modules/compute/gke-nodeset/variables.tf
@@ -27,8 +27,8 @@ variable "slurm_cluster_name" {
   description = "Cluster name, used in slurm controller"
 
   validation {
-    condition     = var.slurm_cluster_name != null && can(regex("^[a-z](?:[a-z0-9]{0,9})$", var.slurm_cluster_name))
-    error_message = "Variable 'slurm_cluster_name' must be a match of regex '^[a-z](?:[a-z0-9]{0,9})$'."
+    condition     = var.slurm_cluster_name != null && can(regex("^[a-z]([-a-z0-9]{0,20})$", var.slurm_cluster_name))
+    error_message = "Variable 'slurm_cluster_name' must be a match of regex '^[a-z]([-a-z0-9]{0,20})$'."
   }
 }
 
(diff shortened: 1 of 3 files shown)


Priya-Quad:
@rbekhtaoui Please take some action on this PR, since it is open from very long

community/modules/internal/slurm-gcp/instance_template/variables.tf
```diff
@@ -382,8 +382,8 @@ variable "slurm_cluster_name" {
   description = "Cluster name, used for resource naming."
 
   validation {
-    condition     = can(regex("^[a-z](?:[a-z0-9]{0,9})$", var.slurm_cluster_name))
-    error_message = "Variable 'slurm_cluster_name' must be a match of regex '^[a-z](?:[a-z0-9]{0,9})$'."
+    condition     = can(regex("^[a-z]([-a-z0-9]{0,20})$", var.slurm_cluster_name))
```
saara-tyagi27, line 385: From what I understand there are other modules as well that use the 'slurm_cluster_name' variable, why are we changing regex condition here and not consistently for other modules.

rbekhtaoui: you’re absolutely right.
This had already been fixed locally for some time, and the change is indeed meant to be applied consistently across all modules using slurm_cluster_name.
I initially cherry-picked only one of the required commits by mistake. I’ve now addressed this properly with a new commit to ensure consistency across all modules.
Regarding the GKE nodeset, we are not using it at the moment let me know if you would still prefer that I apply the same change there as well.

saara-tyagi27: Yup, looks good now.
It would be ideal to make the behavior consistent on gke nodeset as well.

rbekhtaoui: Done !
thank you

sarthakag (approved):

PR #4292, opened 2025-06-18
[Bugfix] Applying K8 manifests to GKE clusters via URL

shubpal07:
### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #4352, opened 2025-07-03
[v2][Bugfix] Applying K8s manifests to GKE clusters via URL

shubpal07:
### Fix: Correctly apply Kubernetes manifests from remote URLs

NOTE: We had re-raise this PR due to a regression found in the original PR #4292 . The root cause was identified:

```
Some of the integration tests are likely passing manifest objects that do not have a 
source attribute at all (for example, a manifest that only uses the content attribute).
In these cases, manifest.source is null. The startswith() function cannot accept a null 
value; it requires a string. This causes Terraform to stop with an error.
```

This PR resolves a bug that caused Terraform to fail with an `Invalid count argument` error when applying a Kubernetes manifest from a remote URL. The issue occurred when the `kubectl-apply` module had a dependency on another resource i.e. GKE cluster, being created in the same run.

The root cause was a **Terraform plan/apply lifecycle conflict**. The module was trying to fetch the URL content *after* the cluster was created (in the apply phase), but other parts of the plan depended on that content *before* the apply could start.

This fix refactors the `kubectl-apply` module with the following strategy:
* **The parent "controller" module** now fetches all URL-based manifest content during the `terraform plan` phase.
* It then passes the raw **text content** not the URL to the submodule responsible for the apply.
* **The submodule** has been simplified to no longer handle URLs, completely resolving the timing conflict.

This change allows for the robust and predictable deployment of URL-based manifests.

#### How to Test and Verify

1. **Configure a Blueprint**: Use a blueprint that calls the workload-manager-install module (modules/management/kubectl-apply).

2. **Use a URL Source**: In the settings, disable the internal Kueue installation and provide the Kueue manifest via a URL in the apply_manifests list. This configuration specifically targets the fixed logic.

YAML

 ```
# Example blueprint settings:
settings:
  # Ensure internal install is disabled to avoid conflicts
  # kueue:
  #   install: true

  apply_manifests:
  - source: "https://raw.githubusercontent.com/GoogleCloudPlatform/cluster-toolkit/refs/heads/develop/modules/management/kubectl-apply/manifests/kueue-v0.11.4.yaml"
    # Flags needed for large manifests and pre-existing resources
    server_side_apply: true
    force_conflicts: true 

```
3. **Deploy**: Run the deployment command (e.g., ./gcluster deploy ...).

4. **Expected Result**:
The terraform plan and terraform apply phases should complete successfully without any "Invalid count" or other planning errors.

### CODE REVIEW

**Comment by shubpal07** [comment]
Ran the failing integration tests:
1. PR-test-gke-managed-parallelstore [No longer present] Deleted as a part of deprecating parallelstore
2. PR-test-gke-managed-hyperdisk (Passed)

File: modules/management/kubectl-apply/versions.tf [hunk 1]
```diff
@@ -28,6 +28,10 @@ terraform {
       source  = "hashicorp/helm"
       version = "~> 2.17"
     }
+    http = {
+      source  = "hashicorp/http"
+      version = "~> 3.0"
+    }
   }
```
SwarnaBharathiMantena, line 35: The reason it was removed earlier was to overcome a consistent warning.
Please review this PR: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4190

The usage of the provider is not the same. So, can go ahead with the PR as long as the warning is not repeating at the end of blueprint deployment.

Can you please double-check?

shubpal07: @SwarnaBharathiMantena, it feels like a bad commit while resolving the merge conflict. Will remove it.
Thanks for checking

shubpal07: Just cross-checked it @SwarnaBharathiMantena. The `http` is being used in the `kubectl-apply` module. Also checked the deployment logs after applying manifest using URL. No such warning was found.

SwarnaBharathiMantena (approved):
LGTM

SwarnaBharathiMantena (approved):
LGTM

PR #4372, opened 2025-07-09
Added capacity checks for reservations

PayalJakhar:
### Submission Checklist

### **What this change does**

This change adds an early check (pre-validation) to make sure there's enough space (capacity) in a specific machine reservation *before* we try to create the cluster's nodes.

### Why this is important

Right now, if you ask for more machines than are available in a reservation, the cluster setup process only fails much later, when it tries to actually create those machines. This is a problem because:
* It wastes time and computing power trying to get machines that aren't there.
* You don't find out about the capacity problem right away.

### How it was implemented

I added a new section of code to the google_container_node_pool resource block to implement robust validation for reservation affinity. This involved creating two new precondition blocks. One to ensure a single reservation is specified when required, and another to verify that the requested node count does not exceed the available capacity of that reservation.

### DIFF
--- a/modules/compute/gke-node-pool/main.tf
+++ b/modules/compute/gke-node-pool/main.tf
@@ -67,6 +67,20 @@ data "google_container_cluster" "gke_cluster" {
   location = local.cluster_location
 }
 
+data "google_compute_reservation" "specific" {
+  count   = local.input_specific_reservations_count == 1 ? 1 : 0
+  project = var.project_id
+  zone    = var.zones[0]
+  name    = var.reservation_affinity.specific_reservations[0].name
+}
+
+locals {
+  reservation_available_count = (local.input_specific_reservations_count == 1) ? (
+    try(data.google_compute_reservation.specific[0].specific_reservation[0].count, 0) -
+    try(data.google_compute_reservation.specific[0].specific_reservation[0].in_use_count, 0)
+  ) : 0
+}
+
 resource "google_container_node_pool" "node_pool" {
   provider = google
 
@@ -378,6 +392,14 @@ resource "google_container_node_pool" "node_pool" {
     precondition {
       condition     = var.enable_flex_start == true ? (var.spot == false) : true
       error_message = "Both enable_flex_start and spot consumption option cannot be set to true at the same time."
+    }
+    precondition {
+      condition     = var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION" || (var.static_node_count != null && var.static_node_count <= local.reservation_available_count)
+      error_message = "Requested static_node_count (${coalesce(var.static_node_count, "not set")}) exceeds the available reservation capacity (${local.reservation_available_count})."
+    }
+    precondition {
+      condition     = local.input_specific_reservations_count == 1 || var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION"
+      error_message = "Exactly one reservation must be specified when using SPECIFIC_RESERVATION."
     }
   }
 }
--- a/modules/compute/gke-node-pool/README.md
+++ b/modules/compute/gke-node-pool/README.md
@@ -305,6 +305,7 @@ limitations under the License.
 | [null_resource.enable_tcpxo_in_workload](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [null_resource.install_dependencies](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [google_compute_machine_types.machine_info](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_machine_types) | data source |
+| [google_compute_reservation.specific](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_reservation) | data source |
 | [google_compute_reservation.specific_reservations](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_reservation) | data source |
 | [google_container_cluster.gke_cluster](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/container_cluster) | data source |
 


pkg/validators/reservation.go
```diff
@@ -0,0 +1,33 @@
...
+	used := reservation.SpecificReservation.InUseCount
+
+	var available int = int(count - used)
+
+	return available, nil
+}
+
+
+
+
```
PayalJakhar: nit: removed these lines

pkg/validators/reservation.go
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validators
+
+import(
+	"context"
+	"fmt"
+
```
gokamesh: nit: remove empty line.

pkg/validators/reservation.go
```diff
@@ -0,0 +1,42 @@
...
+package validators
+
+import(
+	"context"
+	"fmt"
+	"google.golang.org/api/compute/v1"
+)
+
+func GetAvailableReservationCount(ctx context.Context,
```
pandeydeeksha16: We don't need to use prefix `Get` in the function names: go/go-style/decisions#getters

pkg/validators/reservation.go
```diff
@@ -0,0 +1,42 @@
...
+	project, zone, reservationName string) (int, error) {
+
+	svc, err := compute.NewService(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create compute service: %w", err)
+	}
+
+	reservation, err := svc.Reservations.Get(project, zone, reservationName).Do()
```
pandeydeeksha16: I assume we have tested this change to verify reservation is not nil by any chance

pkg/validators/reservation.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return 0,fmt.Errorf("failed tp get reservation: %w", err)
```
pandeydeeksha16: nit: Typo tp -> to

pkg/validators/reservation.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	count := reservation.SpecificReservation.Count 
+	used := reservation.SpecificReservation.InUseCount
+
+	var available int = int(count - used)
```
pandeydeeksha16: Can we rewrite it as `available := count- used` ?

pkg/validators/validators.go
```diff
@@ -54,8 +56,45 @@ const (
...
+	}
+	if errs.Any() {
+		return errs
+	}
+
+	projectID := bp.Vars.Get(requiredVars[0]).AsString()
+	zone := bp.Vars.Get(requiredVars[1]).AsString()
+	reservationName := bp.Vars.Get(requiredVars[2]).AsString()
+	
+	var requestedCount int
```
pandeydeeksha16, line 78: Do we need to assign some value to requestedCount? LMK if I am missing something

pkg/validators/reservation.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2022 Google LLC
```
SwarnaBharathiMantena: This needs to be 2025

pkg/validators/validators.go
```diff
@@ -54,8 +56,45 @@ const (
 	testZoneInRegionName              = "test_zone_in_region"
 	testModuleNotUsedName             = "test_module_not_used"
 	testDeploymentVariableNotUsedName = "test_deployment_variable_not_used"
+	testReservationCapacityAvailableName= "test_reservation_capacity_available"
 )
 
+func testReservationCapacityAvailable(bp config.Blueprint, _ config.Dict) error {
```
SwarnaBharathiMantena: Looking at the structuring of the other tests included, this function needs to be in the `reservation.go` file. You can keep some updates to code like line 59 (adding the new test to line 59) and lines 255-260 (condition added).

SwarnaBharathiMantena (commented):
Please review this PR: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4057
This is a feature that lets a user specify a reservation that does not exist when the node count is 0.

Please update this validator to not conflict with this feature.

Also, is there a reason for adding this validator here instead of the module, under [preconditions](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/6e8b14bed84858db704d231c8b2070cd846ffc1e/modules/compute/gke-node-pool/main.tf#L256)?

modules/compute/gke-node-pool/main.tf
```diff
@@ -67,6 +67,20 @@ data "google_container_cluster" "gke_cluster" {
   location = local.cluster_location
 }
 
+data "google_compute_reservation" "specific" {
+  count = local.input_specific_reservations_count == 1 ? 1 : 0
+  project = var.project_id
```
parulbajaj01, line 72: Can you fix the indentation here? 
make sure you have the pre-commit hook installed and it's running when you commit changes

modules/compute/gke-node-pool/main.tf
```diff
@@ -379,6 +393,20 @@ resource "google_container_node_pool" "node_pool" {
       condition     = var.enable_flex_start == true ? (var.spot == false) : true
       error_message = "Both enable_flex_start and spot consumption option cannot be set to true at the same time."
     }
+    precondition {
+      condition     = var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION" || var.static_node_count <= local.reservation_available_count
+      error_message = "Requested static_node_count (${var.static_node_count}) exceeds the available reservation capacity (${local.reservation_available_count})."
+    }
+
+    precondition {
```
parulbajaj01: fix indentation here as well

modules/compute/gke-node-pool/main.tf
```diff
@@ -379,6 +393,18 @@ resource "google_container_node_pool" "node_pool" {
       condition     = var.enable_flex_start == true ? (var.spot == false) : true
       error_message = "Both enable_flex_start and spot consumption option cannot be set to true at the same time."
     }
+    precondition {
+      condition     = var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION" || var.static_node_count <= local.reservation_available_count
+      error_message = "Requested static_node_count (${var.static_node_count}) exceeds the available reservation capacity (${local.reservation_available_count})."
+    }
+    precondition {
+      condition     = var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION" || var.static_node_count == null || var.static_node_count > 0
+      error_message = "static_node_count must be greater than 0 or null when using specific reservations."
```
SwarnaBharathiMantena: Can't a user use auto scaling with specific reservation?

modules/compute/gke-node-pool/main.tf
```diff
@@ -379,6 +393,18 @@ resource "google_container_node_pool" "node_pool" {
...
+    }
+    precondition {
+      condition     = local.input_specific_reservations_count == 1 || var.reservation_affinity.consume_reservation_type != "SPECIFIC_RESERVATION"
+      error_message = "Exactly one reservation must be specified in var.reservation_affinity.values when using SPECIFIC_RESERVATION."
```
SwarnaBharathiMantena: " in var.reservation_affinity.values " can be removed as a user potentially does not understand this variable.

SwarnaBharathiMantena (changes requested):

modules/compute/gke-node-pool/main.tf
SwarnaBharathiMantena: I did not check the `null` condition mentioned. Given the null condition, this works with autoscaling as well.

parulbajaj01 (approved):

PR #4331, opened 2025-06-27
OFE: Slurm monitoring update and other fixes

scott-nag:
Note, this follows on from #4217 and #4288. Ideally look at those two PRs first as these features are included here too.

#### Monitoring Infrastructure
* **Real-time SLURM Queue Monitoring:** Enhanced the C2 daemon so that continuously monitors SLURM job queues and node status, providing live updates to the OFE interface
* **Hybrid Job Status Tracking:** Implemented a dual-track system that combines OFE's internal Application -> Job tracking with real-time SLURM data, ensuring accurate job status reporting even when jobs are submitted externally
* **Automatic Job Discovery:** Enhanced the system to automatically detect and track jobs that were submitted outside of OFE, providing complete visibility into all cluster activity
#### Enhanced Cluster Management
* **Cluster Page now shows queue status:** Added real-time SLURM job counts and status indicators to the cluster list page, showing active jobs, pending jobs, and cluster availability
* **Smart Cluster Protection:** Implemented safeguards that prevent cluster deletion or destruction when active jobs are running, with clear visual indicators showing running job counts
#### Advanced Job Management Features
* **Enhanced Job List Interface:** Completely redesigned the jobs page with advanced filtering, search capabilities, and real-time status updates
* **SLURM Integration Display:** Added detailed SLURM job information including job IDs, status, start/end times, exit codes, and additional states (like CONFIGURING, POWER_UP_NODE) to provide complete job lifecycle visibility
* **Smart Auto-Refresh System:** Implemented intelligent page refresh that only activates when there are active jobs or recent activity, with visual indicators showing refresh status and job counts
#### User Experience Improvements
* **Collapsible Sidebar Navigation:** Added a hamburger menu with collapsible sidebar improving screen real-estate utilisation
* **Enhanced Visual Feedback:** Added colour-coded status badges, loading indicators, and real-time counters that provide immediate visual feedback about cluster and job status
* **Improved Error Handling:** Enhanced job status detection with fallback mechanisms that handle complex SLURM exit codes and job state transitions

PR #4411, opened 2025-07-18
Cluster Toolkit VDI Module

scott-nag:
### VDI Module

1. New module for the toolkit which allows users to set up a containerised [Guacamole](https://guacamole.apache.org/) based VDI. I have written a [readme](https://github.com/nagconsulting/cluster-toolkit/blob/e3e815e7ad64f7989b6e3a7cfa947478c47fec5e/community/modules/scripts/vdi-setup/README.md) file which covers the new functionality that is included.

2. Fix for [install_docker.sh](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/437ffc571978f15c1a364da3ab4dd40be3d67531/modules/scripts/startup-script/files/install_docker.yml) as the convenience script at get.docker.com does not support Rocky image.

_Note: this is **not** related to the other PRs relating to OFE that I currently have open (#4331, #4288, #4217) so could be looked at separately._


bytetwin:
@scott-nag - The PR looks good to me except the changes to install_docker.yml. The changes introduced  are very specific to this PR and is not usable across cluster toolkit in general. Introducing this change needs testing to ensure nothing else breaks.

And adding anything to core modules is restricted to the Google team as it comes under our SLA to support. Can you keep your changes to the `community/modules`.  You can add your own docker installation script if needed

bytetwin (changes requested):
@scott-nag - The PR looks good to me except the changes to install_docker.yml. The changes introduced are very specific to this PR and is not usable across cluster toolkit in general. Introducing this change needs testing to ensure nothing else breaks.

And adding anything to core modules is restricted to the Google team as it comes under our SLA to support. Can you keep your changes to the community/modules. You can add your own docker installation script if needed

Priya-Quad:
@scott-nag Please take action on this PR, since it is open from very long

PR #4451, opened 2025-07-25
namespace for gke-modules

PayalJakhar:
### Submission Checklist

**This PR enhances the GKE modules to allow users to specify a custom Kubernetes namespace for deployed resources.
Modules updated:**

- gke-persistent-volume
- gke-storage

Users can now provide a namespace variable to deploy resources into a desired namespace. If the namespace variable is not specified, the modules will continue to use their default namespace settings.

### DIFF
--- a/modules/file-system/gke-persistent-volume/README.md
+++ b/modules/file-system/gke-persistent-volume/README.md
@@ -143,6 +143,7 @@ No modules.
 |------|------|
 | [kubectl_manifest.pv](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
 | [kubectl_manifest.pvc](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
+| [kubectl_manifest.pvc_namespace](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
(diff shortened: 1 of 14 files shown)


parulbajaj01 (changes requested):
Please fix the pre-commit (update the changes in the readme) and also you can remove the "namespace" keyword from the variable values for all 3 variables.

modules/compute/gke-job-template/variables.tf
```diff
@@ -185,3 +185,9 @@ variable "labels" {
   description = "Labels to add to the GKE job template. Key-value pairs."
   type        = map(string)
 }
+
+variable "namespace" {
+  description = "The Kubernetes namespace where the GKE Job will be deployed."
+  type        = string
+  default     = "gke-job-namespace"
```
SwarnaBharathiMantena: Isn't the default here supposed to be the "default" namespace in Kubernetes?
I think when it was not set earlier, "default" was the namespace.
Please review this in other places too for default value.

parulbajaj01: Yes, it should be default ideally. Payal raised a different PR where the integration tests failed for the "default" value stating "it's not able to delete the default namespace in Kubernetes" when destroying the cluster.

@PayalJakhar I would recommend testing these changes in your local before committing. I assume it will fail for the current values of namespaces as well. Can you first verify and test the changes on local and try to figure out the reason of failure.

PR #4498, opened 2025-08-08
Fix: Slurm scripts incorrectly identifying all accelerators as GPUs

sharabiani:
The Slurm scripts, utils.py more specifically, considers all accelerators in an instance_template as gpu. This causes wrong gres.conf setup if the accelerator is TPU.

This PR filters accelerators type with "nvidia-" prefix for GPUs.

### DIFF
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
@@ -388,6 +388,23 @@ def parse_bucket_uri(uri: str):
     return matches.group("bucket"), matches.group("path")
 
 
+def get_template_gpu(template):
+    """get gpu info from machine type or guest accelerators"""
+    gpu_keyword = "nvidia"
+    gpu = None
+    if template.machine_type.accelerators:
+        tma = template.machine_type.accelerators[0]
+        if gpu_keyword in tma.type.lower():
+            gpu = tma
+    elif template.guestAccelerators:
+        tga = template.guestAccelerators[0]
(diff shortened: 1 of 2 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1984,15 +1984,18 @@ def template_info(self, template_link):
         # del template.metadata
 
         # translate gpus into an easier-to-read format
+        gpu_prefix = "nvidia-"
```
nick-stroud: When type is not specified then CTK uses [this definition](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/26f95d86f36ddace4fedb76f25e6bc5b25c403e7/modules/internal/gpu-definition/main.tf#L48-L74) to populate. This prefix seems compatible with all these types.

But I have also seen `type = "NVIDIA H200 141GB"` used and accepted. 

I think it would be safer to do a lower(type).contains("nvidia")
(or w/e correct syntax is)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
nick-stroud, line 1: For the coming set of changes, ideally we could add unit tests to community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests

As written it might be challenging since the template_info seems to do an api lookup. Could we isolate logic into static function get_template_gpu(template) -> (str, int) and then just test that function with various formats we expect.

PR #4499, opened 2025-08-08
Improvements applied to gke-pv-module

sharabiani:
1- Add custom pv_name, pvc_name support

2- Output pvc-name

3- Bind pv to pvc by defining claimref

### DIFF
--- a/modules/file-system/gke-persistent-volume/main.tf
+++ b/modules/file-system/gke-persistent-volume/main.tf
@@ -32,8 +32,8 @@ locals {
   filestore_share_name = trimprefix(var.network_storage.remote_mount, "/")
   base_name            = local.is_gcs ? var.gcs_bucket_name : local.filestore_name
 
-  pv_name  = "${local.base_name}-pv"
-  pvc_name = "${local.base_name}-pvc"
+  pv_name  = var.pv_name != "" ? var.pv_name : "${local.base_name}-pv"
+  pvc_name = var.pvc_name != "" ? var.pvc_name : "${local.base_name}-pvc"
 
   list_mount_options = split(",", var.network_storage.mount_options)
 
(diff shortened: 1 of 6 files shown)


modules/file-system/gke-persistent-volume/variables.tf
```diff
@@ -66,3 +66,15 @@ variable "namespace" {
   type        = string
   default     = "default"
 }
+
+variable "pv_name" {
+  description = "The name for PV. IF not set, a name will be generated based on the storage name."
+  type        = string
+  default     = ""
```
nick-stroud: nit: here and below
For consistency with gcs_bucket_name, I think it would be better to use default = null here. This would also require changing the conditional check in main.tf from var.pv_name != "" to var.pv_name != null.

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -69,6 +71,9 @@ locals {
       labels        = local.labels
       mount_options = local.is_gcs ? local.list_mount_options : null
       bucket_name   = local.is_gcs ? var.gcs_bucket_name : ""
+      namespace     = var.namespace
+      pvc_name      = local.pvc_name
+      namespace     = var.namespace
```
nick-stroud: namespace is listed twice. Remove 1.

modules/file-system/gke-persistent-volume/outputs.tf
```diff
@@ -24,3 +24,8 @@ output "persistent_volume_claims" {
   }
   depends_on = [kubectl_manifest.pvc]
 }
+
+output "persistent_volume_claim_name" {
```
nick-stroud: I would go with pvc_name to be  consistent with input. Toolkit often relies on consistent naming.

PR #4500, opened 2025-08-08
Fix: empty values error for dynamic resources

sharabiani:
### Submission Checklist

### DIFF
--- a/community/modules/internal/slurm-gcp/internal_instance_template/main.tf
+++ b/community/modules/internal/slurm-gcp/internal_instance_template/main.tf
@@ -120,7 +120,7 @@ resource "google_compute_instance_template" "tpl" {
     nic_type           = local.nic_type
     stack_type         = var.stack_type
     dynamic "access_config" {
-      for_each = var.access_config
+      for_each = coalesce(var.access_config, [])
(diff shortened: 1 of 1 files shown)


community/modules/internal/slurm-gcp/internal_instance_template/main.tf
```diff
@@ -120,7 +120,7 @@ resource "google_compute_instance_template" "tpl" {
     nic_type           = local.nic_type
     stack_type         = var.stack_type
     dynamic "access_config" {
-      for_each = var.access_config
+      for_each = coalesce(var.access_config, [])
```
nick-stroud, line 123: This seems like a little bit of a hack since default is already `[]`. 

I suspect this is because someone upstream is setting variable to null. I believe the correct thing to do here is to set `nullable = false` on the input variable and fix the upstream bad actor. 

Would you be willing to attempt this. If it proves problematic then we could move forward as you propose.

sharabiani: Closing this PR. I couldn't reproduce the error. Seems like it is already fixed.

PR #4501, opened 2025-08-08
Add support for regional Instance Template in Slurm scripts

sharabiani:
GCP API endpoint for global and regional instance_templates are different. Therefore, current logic in the slurm-gcp scripts only supports global instance_templates.

This PR adds support for regional instance_templates. GKE creates regional templates and working with GKE nodes require the support for regional instance_templates.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -399,6 +399,32 @@ def trim_self_link(link: str):
         raise Exception(f"'/' not found, not a self link: '{link}' ")
 
 
+def get_self_link_component(link: str, component_name: str):
```
nick-stroud, line 402: Please write unit test for this in community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests

nick-stroud (approved):

PR #4531, opened 2025-08-15
Handle reservation affinity for spot and dws

samskillman:
This commit fixes a bug introduced in #4372 where deployments would fail for GKE node pools using queued provisioning or Spot VMs.

**Problem:**
When a user enabled `enable_queued_provisioning` or `spot`, the module would fail with an "Iteration over null value" error. This was because the underlying Terraform provider requires
`reservation_affinity.consume_reservation_type` to be `NO_RESERVATION` for these configurations. If the user set this value in their blueprint, Terraform would not apply the modules default `[]` for `specific_reservations`, resulting in a `null` value that caused the plan to fail.

**Solution:**
The `gke-node-pool` module has been updated to handle this logic internally, providing a more robust and user-friendly experience.

1.  **Automatic Override:** The module now uses a local variable (`reservation_affinity_override`) to automatically enforce `consume_reservation_type = "NO_RESERVATION"` and `specific_reservations = []` whenever `enable_queued_provisioning` or `spot` is set to `true`. This ensures the correct configuration is always applied, regardless of the users input.
2.  **Documentation Updates:** The variable descriptions for `enable_queued_provisioning`, `spot`, and `reservation_affinity` have been updated to clearly document this automatic behavior. This makes the modules logic transparent and prevents user confusion.

### Submission Checklist


SwarnaBharathiMantena:
The PR #4372 was reverted with PR #4492. So, these changes are not required at this time.

PR #4537, opened 2025-08-18
Deprecate Debian blueprints from a3 mega gpu

rachit-google:
### Submission Checklist

Debian 12 Slurm Image Blueprint for A3-Mega had been deprecated  and there was a notice that it will be removed on or after August 1st 2025. Hence removing the debian subdirectory which consisted of the deprecated blueprints.


rachit-google:
> Can you add description to the PR

added

bytetwin:
@rachit-google - there is a test failure referring to a deleted file

bytetwin (changes requested):
There is a test failure referring to a deleted file

arpit974 (approved):
please run babysit as well...

PR #3615, opened 2025-01-30
Add cuda-toolkit to a3ultra-jbvms blueprint

RachaelSTamakloe:
This PR adds the installation of [cuda-toolkit-12-8](https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=deb_network) to the [a3ultra-vm.yaml](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/machine-learning/a3-ultragpu-8g/a3ultra-vm.yaml) blueprint. 
### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #4553, opened 2025-08-20
added nvidia-repositories script

rachit-google:
### Submission Checklist

Added NVIDIA repositories following the new script in Guest OS accelerator images for A3 Ultra Slurm blueprint.
See below for example:
https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3615/files

### DIFF
--- a/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
@@ -63,6 +63,13 @@ deployment_groups:
         enabled: true
         world_writable: true
       runners:
+      - type: shell
+        destination: install-cuda-toolkit.sh
+        content: |
+          #!/bin/bash
+          set -e -o pipefail
+          sudo add-nvidia-repositories -y
+          sudo apt install -y cuda-toolkit-12-8
       - type: data
         destination: /etc/apt/preferences.d/block-broken-nvidia-container
         content: |
--- a/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
@@ -63,13 +63,6 @@ deployment_groups:
         enabled: true
         world_writable: true
       runners:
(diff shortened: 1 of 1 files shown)


samskillman (changes requested):
Please run the PR tests on this. Also, that script is duplicating the same things lower in the blueprint, and likely opens the build up to upgrading packages at the wrong time (before they are held). Last, `sudo` is not needed here as they run as root.

examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
```diff
@@ -133,31 +133,22 @@ deployment_groups:
...
+        content: |
+          #!/bin/bash
+          set -e -o pipefail
+          add-nvidia-repositories -y
+          apt install -y cuda-toolkit-12-8
+          apt install -y datacenter-gpu-manager-4-cuda12
       - type: ansible-local
-        destination: install_cuda_dcgm.yml
+        destination: settings_nvidia_dcgm.yml
         content: |
```
samskillman, line 146: Let's change this all so that the only thing this section does is create the nvidia-persistenced override. The add-nvidia-repositories script handles everything else.

samskillman (changes requested):
If we take this approach here, then let's do this across a3u and a4 both slurm and vms.

Please make sure to run all tests for A* slurm clusters.

rachit-google:
> If we take this approach here, then let's do this across a3u and a4 both slurm and vms.
> 
> Please make sure to run all tests for A* slurm clusters.

sure. once i get it working for A3 ultra, i will change it for others as well.

examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml
rachit-google: we also have some handlers associated and after creating the persistence override, we are configuring the override and then notifying to reload systemD. should i removed everything and just have the nvidia-persistence override in place?

samskillman: Keep those other handlers as well.

PR #4560, opened 2025-08-21
Add G4 GKE base blueprints

kadupoornima:


examples/gke-g4/gke-g4.yaml
```diff
@@ -0,0 +1,203 @@
...
+  # The number of nodes to be created.
+  static_node_count:
+
+  # Cidr block containing the IP of the machine calling terraform.
+  # The following line must be updated for this example to work.
+  authorized_cidr:
+
+  system_node_pool_disk_size_gb: 100
+  k8s_service_account_name: workload-identity-k8s-sa
+  accelerator_type: nvidia-rtx-pro-6000
```
annuay-google: Why is this user configurable? Do we support other accelerators?

examples/gke-g4/gke-g4.yaml
```diff
@@ -0,0 +1,203 @@
...
+    settings:
+      machine_type: g4-standard-384
+      auto_upgrade: true
+      zones: [$(vars.zone)]
+      static_node_count: $(vars.static_node_count)
+      guest_accelerator:
+      - type: $(vars.accelerator_type)
+        count: $(vars.accelerator_count)
+        gpu_driver_installation_config:
+          gpu_driver_version: LATEST
```
annuay-google: Try to pin this to a specific version. LATEST would be a bit unreliable

annuay-google:
How was the network connectivity tested?

annuay-google:
> How was the network connectivity tested?

i.e. did you verify all the VMs have the additional NICs attached

examples/gke-g4/gke-g4.yaml
kadupoornima: No we don't.. I had initially used it in multiple places hence created as a variable. Hardcoded it now.

examples/gke-g4/gke-g4.yaml
kadupoornima: Discussed offline, added TODO to move to a fixed version in the future.

examples/gke-g4/gke-g4.yaml
```diff
@@ -0,0 +1,202 @@
...
+  zone:
+
+  # The number of nodes to be created.
+  static_node_count:
+
+  # Cidr block containing the IP of the machine calling terraform.
+  # The following line must be updated for this example to work.
+  authorized_cidr:
+
+  system_node_pool_disk_size_gb: 100
```
bytetwin, line 44: should this be user input or hardcoded ?

examples/gke-g4/gke-g4.yaml
```diff
@@ -0,0 +1,202 @@
...
+  k8s_service_account_name: workload-identity-k8s-sa
+  accelerator_count: 8
```
bytetwin: `accelerator_count` i guess should not be a variable that user can change. G4 architecture defines 8 gpu so lets use that directly

bytetwin (changes requested):
Since this is a new machine type, can you add an integration test.

examples/gke-g4/gke-g4.yaml
kadupoornima: Done, thanks.

examples/gke-g4/gke-g4.yaml
kadupoornima: It is given as a user input for other NPIs, no harm in letting it be customisable.

examples/gke-g4/gke-g4.yaml
```diff
@@ -0,0 +1,199 @@
...
+              start_time: "2025-08-01T00:00:00Z"
+              end_time: "2026-04-10T00:00:00Z"
+              exclusion_scope: NO_MINOR_OR_NODE_UPGRADES
+        outputs: [instructions]
+
+      - id: g4-pool
+        source: modules/compute/gke-node-pool
+        use: [g4-cluster, node_pool_service_account]
+        settings:
+          machine_type: g4-standard-384
```
samskillman: How should we handle different shape g4 here? For smaller shapes there is only 1 NIC, for example.

kadupoornima: As of now, only the shape with 8 GPUs is publicly announced ([link](https://cloud.google.com/blog/products/compute/introducing-g4-vm-with-nvidia-rtx-pro-6000)), so we are going with that. The remaining shapes are tentatively proposed.

We had raised this point with the G4 lead. They did not have a preference for one particular type but mentioned the above point.

samskillman: https://cloud.google.com/compute/docs/gpus#rtx-6000-gpus  has the documented shapes.

kadupoornima: Oh this was added recently!

We can make the machine type and accelerator_count user customisable to support various shapes. Do we want to do that then? We would need to add checks to ensure that only the allowed values are being inputted though.

Using this blueprint, tested that cluster creation is successful for machine shapes having 1 or 2 NICs. So unless having this config of 2 subnets for lesser shapes is a con in any way, the rest of the blueprint should be okay to remain the same.

bytetwin: If the machine shapes are public, we should support them as user input values.

And please add integration test to check for both 1 nic and 2 nics. That test should address the query of having 2 subnets.
Lets not seperate the test into another PR but rather deliver it as a single tested unit

annuay-google (approved):

examples/gke-g4/gke-g4.yaml
kadupoornima: Made the changes. Discussed IT offline.

annuay-google (dismissed):

bytetwin (dismissed):

bytetwin (approved):

PR #4564, opened 2025-08-22
Update Toolkit release to v1.63.0

ankitkumar-quad:
Set release-candidate  to version 1.63.0

### DIFF
--- a/tools/create-release-candidate.sh
+++ b/tools/create-release-candidate.sh
@@ -46,7 +46,7 @@ if ! gh auth status; then
 fi
 
 GITDIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


tools/create-release-candidate.sh
```diff
@@ -46,7 +46,7 @@ if ! gh auth status; then
 fi
 
 GITDIR=$(mktemp -d)
-trap 'rm -rf ${GITDIR}' EXIT
+#trap 'rm -rf ${GITDIR}' EXIT
```
abbas1902: why was this change needed?

ankitkumar-quad: Reverted the change in the new commit

PR #4572, opened 2025-08-26
GKE Managed Lustre integration

vikramvs-gg:
The changes in this PR implement the integration of GKE managed lustre. 

*  An example blueprint ([examples/gke-managed-lustre.yaml](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4572/files#diff-6a7f39cb116fca49952a27e75608931358b7813d8d343f83e87888e573887d40)) is added with the integrated managed lustre module enabled.

* Corresponding changes in terraform modules are added to support managed lustre integration.

* [Integration test](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4572/files#diff-1f6ee60661d614ae0ae4ac737e1b9d1aece8d4a09459ea62e6a431a0bbeaa856) with coverage of functionality for managed lustre support is added.

* Testing details:
1. Provisioned the GKE cluster with example blueprint: `gke-managed-lustre.yaml`. Deployment of cluster was successul.
2. Applied the kubernetes pod manifest with persistent volume claim for the managed lustre storage capacity. Pod was applied succesfully.
3. Ran file I/O operations on the managed lustre volume mount via the applied pod. The file I/O were working as expected.
4. Provisioned a separate cluster using the example blueprint to use the same Managed Lustre volume (Cluster was provisioned on same VPC as the first and lustre instance attributes were passed in the blueprint to use existing instance). The second cluster was also provisioned successfully referring the existing managed lustre instance.
5. Applied manifest in second cluster and did the file I/O operations via pod of this second cluster. The file I/O operations were working as expected.
6. Destroyed the clusters and verified the clean destruction of resources that were provisioned.
7. Integration test have good coverage of functionality. The tests provision cluster using example blueprint, applies the kubernetes manifest using a lightweight busybox image. File mount is validated as it should be managed lustre instance and not just any ephemeral storage directory created. File I/O operations are then performed and assertions are validated for read and write contents from the file.


examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+  project_id:  ## Set GCP Project ID Here
+  deployment_name: gke-managed-lstr
+  region: us-central1
+  zone: us-central1-c
+  # Cidr block containing the IP of the machine calling terraform.
+  # The following line must be updated for this example to work.
+  authorized_cidr: <your-ip-address>/32
+  version_prefix: "1.33."
+  base_network_name: $(vars.deployment_name)
+  lustre_instance_id: lustrein-01
```
parulbajaj01, line 26: can we rename it to something more generic, example "gke-lustre-instance"

parulbajaj01: can we rename it to something more generic, example "gke-lustre-instance"

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - subnetwork_name: $(vars.base_network_name)-subnet
+        ranges:
+        - range_name: pods
+          ip_cidr_range: 10.4.0.0/14
+        - range_name: services
+          ip_cidr_range: 10.0.32.0/20
+
+  # Private Service Access (PSA) requires the compute.networkAdmin role which is
+  # included in the Owner role, but not Editor.
+  # PSA is required for all Parallelstore functionality.
```
parulbajaj01: Typo: Parallelstore -> Managed Lustre

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+blueprint_name: gke-managed-lustre
+vars:
+  project_id:  ## Set GCP Project ID Here
+  deployment_name: gke-managed-lstr
+  region: us-central1
+  zone: us-central1-c
```
bytetwin, line 24: name, region and zone should be left to the user right ?

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+        - "988"
+
+  - id: managed-lustre
+    source: modules/file-system/managed-lustre
+    use: [network, private_service_access]
+    settings:
+      name: $(vars.lustre_instance_id)
+      local_mount: /lustre
+      remote_mount: lustrefs
+      size_gib: 18000
```
bytetwin: can be a var ?

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+      per_unit_storage_throughput: 1000
+
+  - id: node_pool_service_account
+    source: community/modules/project/service-account
+    settings:
+      name: gke-np-sa
```
bytetwin, line 92: Won't these hardcoded names conflict with another deployment ie) customer trying to deploy two gke cluster with managed lustre support

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+    source: modules/scheduler/gke-cluster
+    use: [network, workload_service_account]
+    settings:
+      version_prefix: $(vars.version_prefix)
+      release_channel: RAPID
+      enable_managed_lustre_csi: true # Enable Managed lustre for the cluster
+      configure_workload_identity_sa: true
+      enable_private_endpoint: false  # Allows for access from authorized public IPs
+      master_authorized_networks:
+      - display_name: deployment-machine
```
bytetwin: network display name is deployment-machine ?

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+        start_time: "2025-08-01T00:00:00Z"
+        end_time: "2026-10-30T00:00:00Z"
+        exclusion_scope: NO_MINOR_OR_NODE_UPGRADES
+    outputs: [instructions]
+
+  - id: lustre-pv
+    source: modules/file-system/gke-persistent-volume
+    use: [managed-lustre, gke_cluster]
+    settings:
+      capacity_gb: 18000
```
bytetwin: should be variable ?

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+  - id: sample-pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, node_pool_service_account]
+    settings:
+      name: sample-pool
```
bytetwin: sample-pool ?

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -20,21 +20,28 @@ locals {
 }
 
 locals {
-  is_gcs = (var.gcs_bucket_name != null)
+  is_gcs    = (var.gcs_bucket_name != null)
+  is_lustre = (var.lustre_id != null)
```
bytetwin, line 24: Where is the var lustre_id populated in the blueprint ? The default value is null

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -122,17 +154,17 @@ resource "kubectl_manifest" "pvc_namespace" {
 }
 
 resource "kubectl_manifest" "pv" {
-  yaml_body = local.is_gcs ? local.gcs_pv_contents : local.filestore_pv_contents
+  yaml_body = local.is_gcs ? local.gcs_pv_contents : (local.is_lustre ? local.lustre_pv_contents : local.filestore_pv_contents)
```
bytetwin, line 157: These type of ternary conditional expression (including the previous ones) becomes unmaintanable as we add more types. Instead try a map lookup

```
storage_map = {
  gcs = is_gcs
  lustre = is_lustre
  ...
}

gcs_pv = ...
lustre_pv = ...
filrestore_pv = ...

pv_map = {
  gcs = local.gcs_pv
  lustre = local.lustre_pv
  ...
}
selected_storage = one([
    for key, value in local.storage_map : key if value != null
  ])

pv_yaml = local.pv_map[selected_storage]

```

You can modify this code for references in othe rplaces

modules/file-system/gke-persistent-volume/templates/managed-lustre-pv.yaml.tftpl
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: ${pv_name}
+  labels:
+  %{~ for key, val in labels ~}
+    ${key}: ${val}
+  %{~ endfor ~}
+spec:
+  storageClassName: ""
```
bytetwin, line 11: I think the recommendation was to use the default storage class

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -122,17 +154,17 @@ resource "kubectl_manifest" "pvc_namespace" {
...
 
   lifecycle {
     precondition {
-      condition     = (var.gcs_bucket_name != null) != (var.filestore_id != null)
-      error_message = "Either gcs_bucket_name or filestore_id must be set."
+      condition     = (var.gcs_bucket_name != null) != (var.filestore_id != null) || (var.lustre_id != null)
```
parulbajaj01: This condition is incorrect if we want to use only 1 of the mentioned 3. Can we please update the condition?

modules/network/firewall-rules/main.tf
```diff
@@ -14,25 +14,46 @@
...
-  assert {
-    condition     = data.google_compute_subnetwork.subnetwork.private_ip_google_access
-    error_message = "Private Google Access is disabled for subnetwork '${data.google_compute_subnetwork.subnetwork.name}'. This may cause connectivity issues for instances without external IPs trying to access Google APIs and services."
+locals {
+  # Derived values from data source, null if data source is not used
+  derived_project_id   = local.use_subnetwork_data ? data.google_compute_subnetwork.subnetwork[0].project : null
+  derived_network_name = local.use_subnetwork_data ? data.google_compute_subnetwork.subnetwork[0].network : null
+
+  # Effective values: Use var if provided, otherwise use derived value
+  effective_project_id   = coalesce(var.project_id, local.derived_project_id)
```
bytetwin, line 35: local.derived_project_id is more assertive than var.project_id as project_id provided could be wrong. I guess we should swap the variable positions

bytetwin (changes requested):

parulbajaj01 (changes requested):

examples/gke-managed-lustre.yaml
parulbajaj01: The minimum and maximum sizes for each performance tier are subject to specific constraints, as detailed in the official documentation: https://cloud.google.com/managed-lustre/docs/create-instance#performance-tiers

Ideally it should be a var but it's crucial to document how to set these values clearly in the README or blueprint comments to minimize friction for users.

examples/gke-managed-lustre.yaml
parulbajaj01: can be added in the documentation: https://cloud.google.com/managed-lustre/docs/locations

examples/gke-managed-lustre.yaml
parulbajaj01: To prevent naming conflicts, these service accounts are designed to be prefixed with the deployment name for each deployment.

examples/gke-managed-lustre.yaml
parulbajaj01: Other GKE blueprints have "kubectl-access-network" as the value. We can update the name here as well to maintain consistency.

examples/gke-managed-lustre.yaml
parulbajaj01: Can be renamed to managed-lustre-pool or gke-lustre-pool

examples/gke-managed-lustre.yaml
```diff
@@ -0,0 +1,135 @@
...
+  # included in the Owner role, but not Editor.
+  # PSA is required for all Parallelstore functionality.
+  # https://cloud.google.com/vpc/docs/configure-private-services-access#permissions
+  - id: private_service_access
+    source: community/modules/network/private-service-access
+    use: [network]
+    settings:
+      prefix_length: 24
+
+  # Firewall to allow managed lustre connection
```
shri-zz: The capitalization should be "Managed Lustre" in all the comments.

examples/gke-managed-lustre.yaml
parulbajaj01: Same as above, for now, we can keep this value hardcoded. We can add clear instructions to the README and make it a variable in incremental PR's

modules/file-system/gke-persistent-volume/main.tf
parulbajaj01: It's getting created dynamically and passed from the outputs of managed-lustre module.

modules/file-system/gke-persistent-volume/templates/managed-lustre-pv.yaml.tftpl
parulbajaj01: That recommendation was intended when we dynamically provision the Lustre instance using a Storage class. In our case here, we are doing static provisioning by creating the Persistent volume. Hence, StorageClass can go empty here.

We can explore Dynamic provisioning also in the future.

modules/network/firewall-rules/main.tf
vikramvs-gg: There was an issue encountered in the existing implementation of fetching values from data source: `data.google_compute_subnetwork.subnetwork.name is null`. This happens when the data block tries to read information about a subnetwork that hasn’t been created at the time Terraform is evaluating data source _(race condition)_.
On deeper analysis, it was found that the firewall-rules module doesn't actually need to look up the project and network name using a data source; this information can be passed directly from the blueprint  and the vpc  network module outputs. Hence *we should check the vars first* and then fallback to derived values.

examples/gke-managed-lustre.yaml
vikramvs-gg: Renamed to relevant name.

examples/gke-managed-lustre.yaml
vikramvs-gg: This is done.

examples/gke-managed-lustre.yaml
vikramvs-gg: Added the relevant comments including the mentioned doc for enabling user to fill in the region and zone values correctly. Default values are kept for these attrs for reference

examples/gke-managed-lustre.yaml
vikramvs-gg: Made the size and throughput as vars accepting values from user instead of hardcoding in module settings (Keeping default values as reference). Also added some readme comments and mentioned doc for enabling user fill in values of these lustre specific variables correctly.

examples/gke-managed-lustre.yaml
vikramvs-gg: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4572#discussion_r2307409333

examples/gke-managed-lustre.yaml
vikramvs-gg: Made relevant naming change

modules/file-system/gke-persistent-volume/main.tf
vikramvs-gg: Corrected the condition and tested provisioning.

examples/gke-managed-lustre.yaml
vikramvs-gg: This is corrected.

modules/file-system/gke-persistent-volume/main.tf
vikramvs-gg: This can also be taken as incremental change involving re-factoring of gke-persistent-volume module using better practices. Have recorded this note.

parulbajaj01 (approved):
LGTM

PR #4583, opened 2025-08-29
Add Google Cloud NetApp Volumes support

okrause:
Hello,

here is a PR to add support for Google Cloud NetApp Volume, which provides NFS/SMB shared file-system.

It has two new modules:
- netapp-storage-pool: A capacity container to host volumes
- netapp-volume: Volumes are file-systems

The PR contains code and documentation for the two new modules, an example blueprint and updates to the blueprint readme and network_storage.md.

I am unsure if this needs to be a community contribution or a core one. GCNV is a Google first party service and I am part of the GCNV product team. So I consider it "core" contribution, but in the end it depends on how the CT team defines core vs community. Please advise.

I also need guidance if I need to add tests.

Looking forward to the review process.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -64,6 +64,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [xpk-n2-filestore](#xpk-n2-filestore--) ![community-badge] ![experimental-badge]
   * [gke-h4d](#gke-h4d-) ![core-badge]
   * [gke-g4](#gke-g4-) ![core-badge]
+  * [netapp-volumes.yaml](#netapp-volumesyaml--) ![core-badge]
 * [Blueprint Schema](#blueprint-schema)
 * [Writing an HPC Blueprint](#writing-an-hpc-blueprint)
   * [Blueprint Boilerplate](#blueprint-boilerplate)
@@ -1641,6 +1642,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a H4D node pool, a
 This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, along with networks and service accounts. Information about G4 machines can be found [here](https://cloud.google.com/blog/products/compute/introducing-g4-vm-with-nvidia-rtx-pro-6000). The deployment instructions can be found in the [README](/examples/gke-g4/README.md).
 
 [gke-g4]: ../examples/gke-g4
+
+### [netapp-volumes.yaml] ![core-badge]
+
+This blueprint demonstrates how to provision NFS volumes as shared filesystems for compute VMs, using Google Cloud NetApp Volumes. It can be used as an alternative to FileStore in blueprints.
+
+NetApp Volumes is a first-party Google service that provides NFS and/or SMB shared file-systems to VMs. It offers advanced data management capabilities and highly scalable capacity and performance.
+
+NetApp Volume provides:
+
+* robust support for NFSv3, NFSv4.x and SMB 2.1 and 3.x
+* a [rich feature set][service-levels]
+* scalable [performance](https://cloud.google.com/netapp/volumes/docs/performance/performance-benchmarks)
+* FlexCache: Caching of ONTAP-based volumes to provide high-throughput and low latency read access to compute clusters of on-premises data
+* [Auto-tiering](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering) of unused data to optimse cost
+
+Support for NetApp Volumes is split into two modules.
+
+* **netapp-storage-pool** provisions a [storage pool](https://cloud.google.com/netapp/volumes/docs/configure-and-use/storage-pools/overview). Storage pools are pre-provisioned storage capacity containers which host volumes. A pool also defines fundamental properties of all the volumes within, like the region, the attached network, the [service level][service-levels], CMEK encryption, Active Directory and LDAP settings.
+* **netapp-volume** provisions a [volume](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview) inside an existing storage pool. A volume is a file-system which is shared using NFS or SMB. It provides advanced data management capabilities.
+
+You can provision multiple volumes in a pool. For service levels Standard, Premium and Extreme the throughput capability depends on volume size and service level. Every GiB of provisioned volume space adds 16/64/128 KiBps of throughput capability.
+
+#### Steps to deploy the blueprint
+
+To provision the bluebrint, please run:
+
+```shell
+./gcluster create examples/netapp-volumes.yaml --vars "project_id=${GOOGLE_CLOUD_PROJECT}" --vars region=us-central1 --vars zone=us-central1-a
+./gcluster deploy netapp-volumes
+```
+
+After the blueprint deployed, you can login to the VM created:
+
+```shell
+gcloud compute ssh --zone "us-central1-a" "netapp-volumes-0" --project ${GOOGLE_CLOUD_PROJECT} --tunnel-through-iap
+```
+
+A NetApp Volumes volume is provisioned and mounted to /home in all the provisioned VMs. A home directory for your user is created automatically:
+
+```shell
+pwd
+df -h -t nfs
+```
+
+#### Clean Up
+To destroy all resources associated with creating the GKE cluster, run the following command:
+
+```sh
+./gcluster destroy netapp-volumes
+```
+
+[netapp-storage-pool]: ../netapp-storage-pool/README.md
+[service-levels]: https://cloud.google.com/netapp/volumes/docs/discover/service-levels
+[auto-tiering]: https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering
+[netapp-volumes.yaml]: ../examples/netapp-volumes.yaml
 
 ## Blueprint Schema
 
--- a/examples/netapp-volumes.yaml
+++ b/examples/netapp-volumes.yaml
@@ -1,0 +1,107 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+# This blueprint show how to provision shared file systems with Google Cloud NetApp Volumes.
+# It creates a NetApp storage pool and a volume for use by VM instances.
+# It can be used to build compute clusters on top of it and as a drop-in replacement
+# for Filestore in existing blueprints.
+
+blueprint_name: netapp-volumes
+
+vars:
+  project_id: ## Set GCP Project ID Here ##
+  deployment_name: netapp-volumes
+  region:     ## Set GCP Region Here ##
+  zone:       ## Set GCP Zone Here ##
+  pool_service_level: "EXTREME"  # Options: "STANDARD", "PREMIUM", "EXTREME"
+
+# Documentation for each of the modules used below can be found at
+# https://github.com/GoogleCloudPlatform/cluster-toolkit
+
+deployment_groups:
+- group: primary
+  modules:
+  - id: network
+    source: modules/network/vpc
+
+  # Private Service Access (PSA) requires the compute.networkAdmin role which is
+  # included in the Owner role, but not Editor.
+  # https://cloud.google.com/vpc/docs/configure-private-services-access#permissions
+  - id: private_service_access
+    source: community/modules/network/private-service-access
+    use: [network]
+    settings:
+      prefix_length: 24
+      service_name: "netapp.servicenetworking.goog"
+      deletion_policy: "ABANDON"
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: $(vars.deployment_name)-netapp-pool
+      capacity_gib: 2048
+      service_level: $(vars.pool_service_level)
+      region: $(vars.region)
+      # allow_auto_tiering: true
+
+  - id: homefs
+    source: modules/file-system/netapp-volume
+    use: [netapp_pool]             # Create this pool using the netapp-storage-pool module
+    settings:
+      region: $(vars.region)
+      volume_name: $(vars.deployment_name)-homefs
+      capacity_gib: 2048           # Size up to available capacity in the pool
+      large_capacity: false
+      local_mount: "/home"         # Mount point at client when client uses USE directive
+      # mount_options: "..."       # Use custom mount_options for special use cases. Defaults are sane.
+      protocols: ["NFSV3"]         # List of protocols. ["NFSV3], ["NFSv4] or ["NFSV3, "NFSV4"]
+      unix_permissions: "0777"     # Specify default permissions for root inode owned by root:root
+      # If no export policy is specified, a permissive default policy will be applied, which is:
+      #  allowed_clients = "10.0.0.0/8,172.16.0.0/12,192.168.0.0/16" # RFC1918
+      #  has_root_access = true      # no_root_squash enabled
+      #  access_type = "READ_WRITE"
+      # export_policy_rules:
+      # - allowed_clients: "10.10.20.8,10.10.20.9"
+      #   has_root_access: true      # no_root_squash enabled
+      #   access_type: "READ_WRITE"
+      #   nfsv3: true
+      #   nfsv4: false
+      # - allowed_clients: "10.0.0.0/8"
+      #   has_root_access: false     # no_root_squash disabled
+      #   access_type: "READ_WRITE"
+      #   nfsv3: true
+      #   nfsv4: false
+      # tiering_policy:              # Enable auto-tiering. Requires auto-tiering enabled storage pool
+      #   tier_action: "ENABLED"
+      #   cooling_threshold_days: 31 # tier data blocks which have not been touched for 31 days
+      # description: "Shared volume for EDA job"
+      # labels:
+      #   department: eda
+
+  # Example VMs which use homefs
+  - id: gcnv_ubuntu_instances
+    source: modules/compute/vm-instance
+    use: [network, homefs]
+    settings:
+      instance_count: 1
+      machine_type: n2-standard-2
+
+  - id: wait-for-vms
+    source: community/modules/scripts/wait-for-startup
+    settings:
+      instance_names: $(gcnv_ubuntu_instances.name)
+      timeout: 7200
--- a/modules/file-system/netapp-storage-pool/README.md
+++ b/modules/file-system/netapp-storage-pool/README.md
@@ -1,0 +1,193 @@
+## Description
+
+This module creates a [Google Cloud NetApp Volumes](https://cloud.google.com/netapp/volumes/docs/discover/overview)
+storage pool.
+
+NetApp Volumes is a first-party Google service that provides NFS and/or SMB shared file-systems to VMs. It offers advanced data management capabilities and highly scalable capacity and performance.
+NetApp Volume provides:
+
+- robust support for NFSv3, NFSv4.x and SMB 2.1 and 3.x
+- a [rich feature set][service-levels]
+- scalable [performance](https://cloud.google.com/netapp/volumes/docs/performance/performance-benchmarks)
+- FlexCache: Caching of ONTAP-based volumes to provide high-throughput and low latency read access to compute clusters of on-premises data
+- [Auto-tiering](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering) of unused data to optimse cost
+
+Support for NetApp Volumes is split into two modules.
+
+- **netapp-storage-pool** provisions a [storage pool](https://cloud.google.com/netapp/volumes/docs/configure-and-use/storage-pools/overview). Storage pools are pre-provisioned storage capacity containers which host volumes. A pool also defines fundamental properties of all the volumes within, like the region, the attached network, the [service level][service-levels], CMEK encryption, Active Directory and LDAP settings.
+- **netapp-volume** provisions a [volume](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview) inside an existing storage pool. A volume file-system container which is shared using NFS or SMB. It provides advanced data management capabilities.
+
+For more information on this and other network storage options in the Cluster
+Toolkit, see the extended [Network Storage documentation](../../../docs/network_storage.md).
+
+### NetApp storage pool service levels
+
+The netapp-storage-pool module currently supports the following NetApp Volumes [service levels][service-levels]:
+
+- Standard: 16 KiBps throughput per provisioned KiB of volume capacity.
+- Premium: 64 KiBps throughput per provisioned KiB of volume capacity. Optional [auto-tiering].
+- Extreme: 128 KiBps throughput per provisioned KiB of volume capacity. Optional [auto-tiering].
+
+Check the [service level matrix][service-levels] for additional information on capability differences between service levels. Flex service levels are currently not supported, but you can connect to existing Flex volumes using the [pre-existing-network-storage module][pre-existing].
+
+### On-boarding NetApp Volumes
+NetApp Volumes uses [Private Service Access](https://cloud.google.com/vpc/docs/private-services-access) (PSA) to connect volumes to your network. Before you create a storage pool, make sure to [connect NetApp Volumes to your network](https://cloud.google.com/netapp/volumes/docs/get-started/configure-access/networking).
+
+Example of creating a storage pool using a new network:
+
+```yaml
+deployment_groups:
+- group: primary
+  modules:
+  - id: network
+    source: modules/network/vpc
+    settings:
+      region: $(vars.region)
+
+  - id: private_service_access
+    source: community/modules/network/private-service-access
+    use: [network]
+    settings:
+      prefix_length: 24
+      service_name: "netapp.servicenetworking.goog"
+      deletion_policy: "ABANDON"
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: $(vars.deployment_name)-eda-pool
+      capacity_gib: 20000
+      service_level: "EXTREME"
+      region: $(vars.region)
+```
+
+Example of creating a storage pool using an existing network which was already PSA-peered with NetApp Volume:
+
+```yaml
+deployment_groups:
+ - group: primary
+  modules:
+  - id: network
+    source: modules/network/pre-existing-vpc
+    settings:
+      project_id: $(vars.project_id)
+      region: $(vars.region)
+      network_name: $(vars.network)
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network]
+    settings:
+      pool_name: "eda-pool"
+      capacity_gib: 20000
+      service_level: "EXTREME"
+      region: $(vars.region)
+```
+
+### Storage pool example
+
+The following example shows all available parameters in use:
+
+```yaml
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: "mypool"
+      region: "us-west4"
+      capacity_gib: 2048
+      service_level: "EXTREME"
+      active_directory_policy: "projects/myproject/locations/us-east4/activeDirectories/my-ad"
+      cmek_policy: "projects/myproject/locations/us-east4/kmsConfigs/my-cmek-policy"
+      ldap_enabled: false
+      allow_auto_tiering: false
+      description: "Demo storage pool"
+      labels:
+        owner: bob
+```
+
+### NetApp Volumes quota
+
+Your project must have unused quota for NetApp Volumes in the region you will
+provision the storage pool. This can be found by browsing to the [Quota tab within IAM & Admin](https://console.cloud.google.com/iam-admin/quotas) in the Cloud Console.
+Please note that there are separate quota limits for Standard and Premium/Extreme service levels.
+
+See also NetApp Volumes [default quotas](https://cloud.google.com/netapp/volumes/docs/quotas#netapp-volumes-default-quotas).
+
+[service-levels]: https://cloud.google.com/netapp/volumes/docs/discover/service-levels
+[auto-tiering]: https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering
+[pre-existing]: ../pre-existing-network-storage/README.md
+[matrix]: ../../../docs/network_storage.md#compatibility-matrix
+
+## License
+
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.5.7 |
+| <a name="requirement_google"></a> [google](#requirement\_google) | >= 6.45.0 |
+| <a name="requirement_random"></a> [random](#requirement\_random) | ~> 3.0 |
+
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_google"></a> [google](#provider\_google) | >= 6.45.0 |
+| <a name="provider_random"></a> [random](#provider\_random) | ~> 3.0 |
+
+## Modules
+
+No modules.
+
+## Resources
+
+| Name | Type |
+|------|------|
+| [google_netapp_storage_pool.netapp_storage_pool](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/netapp_storage_pool) | resource |
+| [random_id.resource_name_suffix](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/id) | resource |
+| [google_compute_network_peering.private_peering](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_network_peering) | data source |
+
+## Inputs
+
+| Name | Description | Type | Default | Required |
+|------|-------------|------|---------|:--------:|
+| <a name="input_active_directory_policy"></a> [active\_directory\_policy](#input\_active\_directory\_policy) | The ID of the Active Directory policy to apply to the storage pool in the format:<br/>`projects/<project_id>/locations/<location>/activeDirectoryPolicies/<policy_id>` | `string` | `null` | no |
+| <a name="input_allow_auto_tiering"></a> [allow\_auto\_tiering](#input\_allow\_auto\_tiering) | Whether to allow automatic tiering for the storage pool. | `bool` | `false` | no |
+| <a name="input_capacity_gib"></a> [capacity\_gib](#input\_capacity\_gib) | The capacity of the storage pool in GiB. | `number` | `2048` | no |
+| <a name="input_cmek_policy"></a> [cmek\_policy](#input\_cmek\_policy) | The ID of the Customer Managed Encryption Key (CMEK) policy to apply to the storage pool in the format:<br/>`projects/<project>/locations/<location>/kmsConfigs/<name>` | `string` | `null` | no |
+| <a name="input_deployment_name"></a> [deployment\_name](#input\_deployment\_name) | Name of the deployment, used as name of the NetApp storage pool if no name is specified. | `string` | n/a | yes |
+| <a name="input_description"></a> [description](#input\_description) | A description of the NetApp storage pool. | `string` | `""` | no |
+| <a name="input_labels"></a> [labels](#input\_labels) | Labels to add to the NetApp storage pool. Key-value pairs. | `map(string)` | n/a | yes |
+| <a name="input_ldap_enabled"></a> [ldap\_enabled](#input\_ldap\_enabled) | Whether to enable LDAP for the storage pool. | `bool` | `false` | no |
+| <a name="input_network_id"></a> [network\_id](#input\_network\_id) | The ID of the GCE VPC network to which the NetApp storage pool is connected given in the format:<br/>`projects/<project_id>/global/networks/<network_name>`" | `string` | n/a | yes |
+| <a name="input_network_self_link"></a> [network\_self\_link](#input\_network\_self\_link) | Network self-link the pool will be on, required for checking private service access | `string` | n/a | yes |
+| <a name="input_pool_name"></a> [pool\_name](#input\_pool\_name) | The name of the storage pool. Leave empty to generate name based on deployment name. | `string` | `null` | no |
+| <a name="input_private_vpc_connection_peering"></a> [private\_vpc\_connection\_peering](#input\_private\_vpc\_connection\_peering) | The name of the private VPC connection peering. | `string` | `"sn-netapp-prod"` | no |
+| <a name="input_project_id"></a> [project\_id](#input\_project\_id) | ID of project in which the NetApp storage pool will be created. | `string` | n/a | yes |
+| <a name="input_region"></a> [region](#input\_region) | Location for NetApp storage pool. | `string` | n/a | yes |
+| <a name="input_service_level"></a> [service\_level](#input\_service\_level) | The service level of the storage pool. | `string` | `"PREMIUM"` | no |
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| <a name="output_capacity_gb"></a> [capacity\_gb](#output\_capacity\_gb) | Storage pool capacity in GiB. |
+| <a name="output_netapp_storage_pool_id"></a> [netapp\_storage\_pool\_id](#output\_netapp\_storage\_pool\_id) | An identifier for the resource with format `projects/{{project}}/locations/{{location}}/storagePools/{{name}}` |
+<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/modules/file-system/netapp-storage-pool/metadata.yaml
+++ b/modules/file-system/netapp-storage-pool/metadata.yaml
@@ -1,0 +1,20 @@
+# Copyright 2025 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+
+spec:
+  requirements:
+    services:
+    - netapp.googleapis.com
+    - servicenetworking.googleapis.com
--- a/modules/file-system/netapp-storage-pool/variables.tf
+++ b/modules/file-system/netapp-storage-pool/variables.tf
@@ -1,0 +1,133 @@
+/**
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "project_id" {
+  description = "ID of project in which the NetApp storage pool will be created."
+  type        = string
+}
+
+variable "deployment_name" {
+  description = "Name of the deployment, used as name of the NetApp storage pool if no name is specified."
+  type        = string
+}
+
+variable "region" {
+  description = "Location for NetApp storage pool."
+  type        = string
+}
+
+variable "network_id" {
+  description = <<-EOT
+    The ID of the GCE VPC network to which the NetApp storage pool is connected given in the format:
+    `projects/<project_id>/global/networks/<network_name>`"
+    EOT
+  type        = string
+  validation {
+    condition     = length(split("/", var.network_id)) == 5
+    error_message = "The network id must be provided in the following format: projects/<project_id>/global/networks/<network_name>."
+  }
+}
+
+variable "network_self_link" {
+  description = "Network self-link the pool will be on, required for checking private service access"
+  type        = string
+  nullable    = false
+}
+
+variable "private_vpc_connection_peering" {
+  description = "The name of the private VPC connection peering."
+  type        = string
+  default     = "sn-netapp-prod"
+}
+
+variable "pool_name" {
+  description = "The name of the storage pool. Leave empty to generate name based on deployment name."
+  type        = string
+  default     = null
+}
+
+variable "service_level" {
+  description = "The service level of the storage pool."
+  type        = string
+  default     = "PREMIUM"
+  validation {
+    condition     = contains(["STANDARD", "PREMIUM", "EXTREME"], var.service_level)
+    error_message = "Allowed values for service_level are 'STANDARD', 'PREMIUM', or 'EXTREME'."
+  }
+}
+
+variable "capacity_gib" {
+  description = "The capacity of the storage pool in GiB."
+  type        = number
+  default     = 2048
+  validation {
+    condition     = var.capacity_gib >= 2048
+    error_message = "The minimum capacity for the storage pool is 2048 GiB."
+  }
+}
+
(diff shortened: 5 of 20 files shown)


okrause:
Hello, I am back from vacation. Pushed an update to fix an merge conflict caused my changes in the develop branch.

examples/README.md
```diff
@@ -1497,6 +1498,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
 
 [gke-g4]: ../examples/gke-g4
 
+### [netapp-volumes.yaml] ![core-badge]
```
bytetwin, line 1646: community badge instead of core

examples/README.md
```diff
@@ -1497,6 +1498,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
...
+* a [rich feature set][service-levels]
+* scalable [performance](https://cloud.google.com/netapp/volumes/docs/performance/performance-benchmarks)
+* FlexCache: Caching of ONTAP-based volumes to provide high-throughput and low latency read access to compute clusters of on-premises data
+* [Auto-tiering](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering) of unused data to optimse cost
+
+Support for NetApp Volumes is split into two modules.
+
+* **netapp-storage-pool** provisions a [storage pool](https://cloud.google.com/netapp/volumes/docs/configure-and-use/storage-pools/overview). Storage pools are pre-provisioned storage capacity containers which host volumes. A pool also defines fundamental properties of all the volumes within, like the region, the attached network, the [service level][service-levels], CMEK encryption, Active Directory and LDAP settings.
+* **netapp-volume** provisions a [volume](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview) inside an existing storage pool. A volume file-system container which is shared using NFS or SMB. It provides advanced data management capabilities.
```
bytetwin: Should it be "A volume **is** a file-system ...."

examples/README.md
```diff
@@ -1497,6 +1498,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
...
+#### Steps to deploy the blueprint
+
+To provision the bluebrint, please run:
+
+```shell
+./gcluster create examples/netapp-volumes.yaml --vars "project_id=${GOOGLE_CLOUD_PROJECT}"
+./gcluster deploy netapp-volumes
+```
+
+After the blueprint deployed, you can login the to VM created:
```
bytetwin: login **to the** VM

examples/README.md
```diff
@@ -1497,6 +1498,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
...
+
+A NetApp Volumes volume was provisioned and mounted to /home in all the provisioned VMs. A home directory for your user is created automatically:
+
+```shell
+pwd
+df -h -t nfs
+```
+
+#### Clean Up
+To destroy all resources associated with creating the GKE cluster, from Cloud Shell run the following command:
```
bytetwin: I think we can remove the term "cloud shell" as the destroy command can be run from anywhere as long as the gcloud credentials are set

examples/README.md
```diff
@@ -1497,6 +1498,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
...
+./gcluster deploy netapp-volumes
+```
+
+After the blueprint deployed, you can login the to VM created:
+
+```shell
+gcloud compute ssh --zone "us-central1-a" "netapp-volumes-0" --project ${GOOGLE_CLOUD_PROJECT} --tunnel-through-iap
+```
+
+A NetApp Volumes volume was provisioned and mounted to /home in all the provisioned VMs. A home directory for your user is created automatically:
```
bytetwin: **is** provisioned ?

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,106 @@
...
+blueprint_name: netapp-volumes
+
+vars:
+  project_id: ## Set GCP Project ID Here ##
+  deployment_name: netapp-volumes
+  region: us-central1
+  zone: $(vars.region)-a
+
+# Documentation for each of the modules used below can be found at
+# https://github.com/GoogleCloudPlatform/hpc-toolkit/blob/main/modules/README.md
```
bytetwin: https://github.com/GoogleCloudPlatform/cluster-toolkit

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,106 @@
...
+- group: primary
+  modules:
+  # Source is an embedded module, denoted by "modules/*" without ./, ../, /
+  # as a prefix. To refer to a local module, prefix with ./, ../ or /
+  - id: network
+    source: modules/network/vpc
+
+  # Private Service Access (PSA) requires the compute.networkAdmin role which is
+  # included in the Owner role, but not Editor.
+  # PSA is a best practice for Filestore instances, but can be optionally
```
bytetwin: As per this doc https://cloud.google.com/netapp/volumes/docs/get-started/configure-access/networking, PSA seems to be mandatory requirement for netapp rather than optional.

Can you test the blueprint removing PSA or else remove these commented lines

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,106 @@
...
+      deletion_policy: "ABANDON"
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: "netapp-pool"
+      network_id: $(network.network_id)
+      capacity_gib: 2048
+      service_level: "EXTREME"
```
bytetwin: This should be a  varialbe as service level supports different values

Same applies for capacity_gib too as its being used in the `netapp-volume`

modules/file-system/netapp-storage-pool/README.md
```diff
@@ -0,0 +1,195 @@
...
+- a [rich feature set][service-levels]
+- scalable [performance](https://cloud.google.com/netapp/volumes/docs/performance/performance-benchmarks)
+- FlexCache: Caching of ONTAP-based volumes to provide high-throughput and low latency read access to compute clusters of on-premises data
+- [Auto-tiering](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering) of unused data to optimse cost
+
+Support for NetApp Volumes is split into two modules.
+
+- **netapp-storage-pool** provisions a [storage pool](https://cloud.google.com/netapp/volumes/docs/configure-and-use/storage-pools/overview). Storage pools are pre-provisioned storage capacity containers which host volumes. A pool also defines fundamental properties of all the volumes within, like the region, the attached network, the [service level][service-levels], CMEK encryption, Active Directory and LDAP settings.
+- **netapp-volume** provisions a [volume](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview) inside an existing storage pool. A volume file-system container which is shared using NFS or SMB. It provides advanced data management capabilities.
```
bytetwin, line 18: A volume **is a**  file-system ??

modules/file-system/netapp-storage-pool/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright 2023 "Google LLC"
```
bytetwin: 2025

modules/file-system/netapp-storage-pool/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+ * limitations under the License.
+ */
+
+variable "project_id" {
+  description = "ID of project in which the NetApp storage pool will be created."
+  type        = string
+}
+
+variable "deployment_name" {
+  description = "Name of the HPC deployment, used as name of the NetApp storage pool if no name is specified."
```
bytetwin: Lets remove the word hpc

modules/file-system/netapp-storage-pool/versions.tf
```diff
@@ -0,0 +1,36 @@
...
+      source  = "hashicorp/random"
+      version = "~> 3.0"
+    }
+  }
+  # provider_meta "google" {
+  #   module_name = "blueprints/terraform/hpc-toolkit:netapp-storage-pool/v1.60.0"
+  # }
+  # provider_meta "google-beta" {
+  #   module_name = "blueprints/terraform/hpc-toolkit:netapp-storage-pool/v1.60.0"
+  # }
```
bytetwin: Lets enabled this metadata and change version to v1.67.0

modules/file-system/netapp-volume/README.md
```diff
@@ -0,0 +1,198 @@
...
+## Protocol support
+Since Cluster Toolkit is currently built to provision Linux-based compute clusters, this module  supports NFSv3 and NFSv4.1 only. SMB is blocked.
+
+## Large volumes
+Volumes larger than 15 TiB can be created as [Large Volumes](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview#large-capacity-volumes). Such volumes can grow up to 3PiB and can scale read performance up to 29 GiBps. They provide six IP addresses to the volume. They are exported via the `server_ips` output. When connecting a large volume to a client using the USE directive, the mount script will randomly pick an IP from the list to load-balance multiple clients over the available IPs.
+
+This feature is allow-listed GA. To request allow-listing, see [Large Volumes](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/overview#large-capacity-volumes).
+
+## Auto-tiering support
+For auto-tiering enables storage pools you can enable auto-tiering on the volume. For more information, see [manage auto-tiering](https://cloud.google.com/netapp/volumes/docs/configure-and-use/volumes/manage-auto-tiering).
```
bytetwin: **enabled**

modules/file-system/netapp-volume/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Using existing volumes not created by Cluster Toolkit
+NetApp Volumes volumes are regular NFS exports. You can use the [pre-existing-network-storage] module to integrate them into Cluster Toolkit.
```
bytetwin, line 104: volume is repeated

modules/file-system/netapp-volume/metadata.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 "Google LLC"
```
bytetwin: 2025

modules/file-system/netapp-volume/outputs.tf
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+output "network_storage" {
+  description = "Describes a NetApp Volume volume."
```
bytetwin: repeated volume ?

modules/file-system/netapp-volume/metadata.yaml
bytetwin: 2025

modules/file-system/netapp-volume/variables.tf
```diff
@@ -0,0 +1,142 @@
...
+ * limitations under the License.
+ */
+
+variable "project_id" {
+  description = "ID of project in which the NetApp storage pool will be created."
+  type        = string
+}
+
+variable "netapp_storage_pool_id" {
+  description = "The ID of the NetApp storage pool to use for the volume. If not specified, a new storage pool will be created."
```
bytetwin: is storage pool optional ? I don't see a new storage pool created in netapp-volume/main.tf

modules/file-system/netapp-volume/scripts/install-nfs-client.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+# Copyright 2022 Google LLC
```
bytetwin: 2025

modules/file-system/netapp-volume/scripts/mount.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
```
bytetwin: 2025

modules/file-system/netapp-volume/scripts/mount.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright 2022 Google LLC
```
bytetwin: 2025

modules/file-system/netapp-volume/scripts/mount.yaml
bytetwin, line 1: Where is this file used ?

modules/file-system/netapp-volume/versions.tf
```diff
@@ -0,0 +1,32 @@
...
+
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 7.1"
+    }
+  }
+  provider_meta "google" {
+    module_name = "blueprints/terraform/hpc-toolkit:netapp-volume/v1.60.0"
```
bytetwin: v1.67.0

modules/file-system/netapp-volume/versions.tf
```diff
@@ -0,0 +1,32 @@
...
+  }
+  provider_meta "google-beta" {
+    module_name = "blueprints/terraform/hpc-toolkit:netapp-volume/v1.60.0"
```
bytetwin: v1.67.0

bytetwin (commented):
IMO, netapp-storage-pool as well as netapp-volume should be under community/modules instead of modules. Reason being this module is not directly supported by the toolkit team.

I also think there should be an integration test for the blueprint if we are planning to have netapp as a mainstream storage option in toolkit

okrause:
> I also think there should be an integration test for the blueprint if we are planning to have netapp as a mainstream storage option in toolkit

Can you please share some info on integration test to get me started? I need to understand how they are done and how they work.

examples/README.md
okrause: Thank you for the clarification. That was one of the answers I was looking for. Will move it into community.

okrause:
> IMO, netapp-storage-pool as well as netapp-volume should be under community/modules instead of modules. Reason being this module is not directly supported by the toolkit team.

@bytetwin I understand that reasoning. But I want to understand the role of Filestore here. Is Filestore maintained by the toolkit team? If yes, why not pick up NetApp Volumes as well, since it is a more capable and feature rich native Google NFS and SMB file service?

examples/README.md
okrause: Will change. Thank you.

examples/README.md
okrause: Will change. Thank you.

examples/README.md
okrause: Did copy/paste it from a different example. Agree that it should be removed.

examples/README.md
okrause: Will change. Thank you.

examples/netapp-volumes.yaml
okrause: Copy/pasted from different blueprint. Will fix.

examples/netapp-volumes.yaml
okrause: Oops, copied a different blueprint as template and missed to delete it. Thanks for pointing it out.

examples/netapp-volumes.yaml
okrause: Agreed on service level. On capacity_gib: A pool can hold multiple volumes and using the same variable might users  think it is a 1:1 vs 1:n relationship. Basically sum(size of volumes in pool) <= size(aggregate)Three approaches to use it:
1. Add another volume. This shows how a pool can hold multiple volumes. Size of the two volumes should add up to pool size. Give that, volume sizes won't make sense as variables. Pool size maybe, but also questionable.
2. Make it one variables as you suggested
3. Make it two variables to show the values are not necessarily the same.

Variables for sizes are IMHO not scalable, since users may create multiple volumes in a pool.

modules/file-system/netapp-volume/outputs.tf
okrause: Changing to
"Describes a NetApp Volume**s** volume"

modules/file-system/netapp-volume/variables.tf
okrause: It is not optional. It needs to be inherited down to the volume using the **use** directive. I made the default null to fail if nothing is specified. I am replacing it now with a validation to check if valid pool URN is specified.

modules/file-system/netapp-volume/scripts/install-nfs-client.sh
okrause: This was simply copied from Filestore. But let me change the data.

modules/file-system/netapp-volume/scripts/mount.yaml
okrause: This is unclear to me. But Filestore is shipping it as well, so I decided to add a customized version for Netapp Volumes as well. It is unclear to me how exactly it is used in FileStore, but I thought there will be a reason. I assume is is part of the metadata which is used when VMs are deployed. The name says something about ansible. I assume it is part of the magic which makes CT mount the filesystems in the VMs.

modules/file-system/netapp-volume/versions.tf
okrause: Can you please elaborate what this is and how the version numbers come to be? I see I commented this block out in netapp-storage-pool entirely.
I simply try to adhere to the standards in CT, but this is a part of which I don't understand the purpose and how to use it.

cboneti:
Hi @okrause, please push the small changes first (version differences), year for the header and let's run the basic tests here. After that, I can guide you on how to write integration tests.  We are still debating if this module should be in core or in community.

okrause:
> Hi @okrause, please push the small changes first (version differences), year for the header and let's run the basic tests here. After that, I can guide you on how to write integration tests. We are still debating if this module should be in core or in community.

Thank you for the feedback. I did push some changes and looking forward to resolve open questions with you. On my list of item to clarify are:
- core or community
- integration tests
- What is provider_meta information for? How do I know which version number to use?
- I want to use google>=7.4.0, but currently getting constraint conflicts. How to resolve.

Thank you @cboneti  and @bytetwin for supporting this PR review.

modules/file-system/netapp-volume/versions.tf
cboneti: This comes from the current version of the toolkit in main. This helps us understand the versions of the modules that are currently used in the wild and prioritize support.

modules/file-system/netapp-volume/versions.tf
cboneti: see response in the other comment

cboneti:
@okrause 
We are discussing internally the core vs community, when Hanu is back on Monday, we will converge to a final decision. 

**What is provider_meta information for? How do I know which version number to use?**
 I believe I answered that in the commands. You should use the current version of main. 

**I want to use google>=7.4.0, but currently getting constraint conflicts. How to resolve. **
I am not sure, I haven't tried. I believe the toolkit team might have the answer for that. In the past, I saw this type of problems when the provider made breaking changes that would require significant rework within the toolkit code. This might be the case here but the team will know better (another topic to be discussed on Monday).

As for regression tests, if this module goes to core, then it will be absolutely required. 
Regression tests are defined in tools/cloud-build/daily-tests, PTAL. Unless things changed dramatically there, the documentation is probably not complete. It will also be very hard for you to try things out without the general infrastructure so realistically, it might be a better exercise for us to think of the specific post-deploy tests we should run...  like  in tools/cloud-build/daily-tests/ansible_playbooks/test-validation (see in tools/cloud-build/daily-tests/builds/lustre-slurm.yaml + tools/cloud-build/daily-tests/tests/lustre-slurm.yml an example of how these are put together through ansible).

bytetwin:
@okrause - GCP already has a [TF module](https://registry.terraform.io/modules/GoogleCloudPlatform/netapp-volumes/google/latest) for Netapp that creates both storage pool and volume. 
Any reason direct google provider resources are being used instead of the module?

Using the module simplifies the support required as then it falls under dedicated team to support when there are customer issues. This is key when we move the module to core

okrause:
@bytetwin The reasons why I (as the TF usage expert in the GCNV team) don't use that modules are:

- The ownership of this module is unclear to me. Who is maintaining it? Who keeps it current? I know it is not the GCNV engineers who submit TF changes for new features.
- Completeness: The module is missing many features like backup, replication, quotas, etc. I am not saying these features are all relevant for CT, but IMHO the individual resources are easier to use than the module, unless the user uses simple use cases. As for CT applicability: One example where I immediately see an issue is the absence of https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/netapp_volume#nested_mount_options. This means users who can use the module will not get info on IP and export path of a created volume. That's essential to pass the mount information to VMs CT creates.
- Currency: This module did not receive updates for a long time and is/was missing many new features. The update policy is unknown.

I understand the desire to move as much complexity out of CT. I did not build my changes with that in mind. But even with the latest update of the module (which I was unaware of), I am not considering it as an option to build solid automation on top of it. The module looks to me focused on storage admins, not users who want to build end to end stacks.

okrause:
Pushed a few updates. Mostly documentation, removing parts of the support for multiple-IPs per volume for now (cluster toolkit seems to have to done recent changes in the was volumes get mounted) and version numbers.
Open issues:
- provider version conflicts going to 7.4.0 -> can be ignored, but without 7.4.0 pools will never converge (= config drift in TF). Not a functionality issue.
- adding tests. Will work on this now.
- Adding EDA blueprints, but this will happen in new PRs which build on top of this one.

okrause:
Any updates?

sarthakag:
Hi Oliver, 
I'll be working with you on the integration test build. I'll be able to get back to you by tomorrow on this, thank you!

modules/file-system/netapp-storage-pool/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+
+variable "active_directory_policy" {
+  description = <<-EOT
+    The ID of the Active Directory policy to apply to the storage pool in the format:
+    `projects/<project_id>/locations/<location>/activeDirectoryPolicies/<policy_id>`
+    EOT
+  type        = string
+  default     = null
+  validation {
+    condition     = var.active_directory_policy == null || length(split("/", var.active_directory_policy)) == 6
```
sarthakag: I am getting an error on this validation, can you use the ternary operator?
condition     = var.active_directory_policy == null ? true : length(split("/", var.active_directory_policy)) == 6

sarthakag: Here and L110

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+# Documentation for each of the modules used below can be found at
+# https://github.com/GoogleCloudPlatform/hpc-toolkit
+
+deployment_groups:
+- group: primary
+  modules:
+  # Source is an embedded module, denoted by "modules/*" without ./, ../, /
+  # as a prefix. To refer to a local module, prefix with ./, ../ or /
+  - id: network
+    source: modules/network/vpc
```
sarthakag, line 38: Add name here?
```
settings:
      network_name: $(vars.base_network_name)-net
```

modules/file-system/netapp-storage-pool/variables.tf
okrause: This works with TF 1.13. Are you saying this syntax doesn't work with older TF versions and the ternary will work?

tools/cloud-build/daily-tests/tests/netapp-volumes.yml
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+---
+
+test_name: "netapp-volumes"
+deployment_name: "netapp-volumes-{{ build }}"
+workspace: /workspace
+blueprint_yaml: "{{ workspace }}/examples/netapp-volumes.yaml"
+region: europe-southwest1
+zone: europe-southwest1-a
+network: "{{ test_name }}"
```
sarthakag: network: "{{ deployment_name }}-net"

sarthakag (changes requested):
The test passes with the changes requested.

modules/file-system/netapp-storage-pool/variables.tf
sarthakag: Yeah, I had TF 1.5.7 for which it errored out. I see that you've made the changes already, thanks!

sarthakag (dismissed):
Good to go. Test passes. Thanks @okrause !

modules/file-system/netapp-volume/scripts/mount.sh
```diff
@@ -0,0 +1,66 @@
...
+set -e
+SERVER_IPS=$1
+REMOTE_MOUNT=$2
+LOCAL_MOUNT=$3
+FS_TYPE=$4
+MOUNT_OPTIONS=$5
+
+# accept a list of colon-separated IPs and randomly pick one to enable load balancing
+# In recent changes cluster toolkit doesn't seem to use this file anymore,
+# which makes all mounts use the first IP in the list. Needs to be investigated in future.
```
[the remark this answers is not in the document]
okrause: I added support for randomly selecting an IP for load balancing purposes in my PR. It initially worked, but changes in develop branch on how volumes get mounted rendered it not working. So I am baking out that feature from my PR for now. I plan to add support in a later PR, if this is merged.
This PR doesn't break any existing behaviour and is in line with how other NFS volumes get mounted. The PR I plan to add later need to add that feature in a compatible way.

modules/file-system/netapp-storage-pool/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+    error_message = "Allowed values for service_level are 'STANDARD', 'PREMIUM', or 'EXTREME'."
+  }
+}
+
+variable "capacity_gib" {
+  description = "The capacity of the storage pool in GiB."
+  type        = number
+  default     = 2048
+  validation {
+    condition     = var.capacity_gib >= 2048
```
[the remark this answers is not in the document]
okrause: Exceeding the upper bounds is rarely a problem. I don't like adding a check here, since that would mean we have to touch this verification every time GCNV increases the limits. Extra effort for little gain.

IMHO it is a neglect able risk to have deployments fail because user specify too large volumes. This is also in line with other file systems like Filestore, which don't check upper bounds as well.

modules/file-system/netapp-volume/variables.tf
```diff
@@ -0,0 +1,145 @@
...
+  type        = string
+  default     = null
+}
+
+variable "capacity_gib" {
+  description = "The capacity of the volume in GiB."
+  type        = number
+  default     = 1024
+  validation {
+    condition     = var.capacity_gib >= 100
```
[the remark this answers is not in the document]
okrause: Exceeding the upper bounds is rarely a problem. I don't like adding a check here, since that would mean we have to touch this verification every time GCNV increases the limits. Extra effort for little gain.

IMHO it is a neglect able risk to have deployments fail because user specify too large volumes. This is also in line with other file systems like Filestore, which don't check upper bounds as well.

modules/file-system/netapp-volume/main.tf
```diff
@@ -0,0 +1,92 @@
...
+    for_each = var.export_policy_rules == null ? [] : [0]
+    content {
+      dynamic "rules" {
+        for_each = var.export_policy_rules
+        content {
+          access_type     = rules.value.access_type
+          allowed_clients = rules.value.allowed_clients
+          has_root_access = rules.value.has_root_access
+          nfsv3           = rules.value.nfsv3 == null ? contains(var.protocols, "NFSV3") : rules.value.nfsv3
+          nfsv4           = rules.value.nfsv4 == null ? contains(var.protocols, "NFSV4") : rules.value.nfsv4
```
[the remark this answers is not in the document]
okrause: Good suggestion. Incorporating.

community/modules/network/private-service-access/main.tf
```diff
@@ -46,6 +46,7 @@ resource "google_service_networking_connection" "private_vpc_connection" {
   service                 = var.service_name
   reserved_peering_ranges = [google_compute_global_address.private_ip_alloc.name]
   deletion_policy         = var.deletion_policy
+  update_on_creation_fail = var.deletion_policy == "ABANDON" ? true : null
```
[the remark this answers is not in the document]
okrause: The goal of this change is to maintain backward compatibility. Before this change, `update_on_creation_fail` was absent all the time. So I chose to send NULL in that case, which make it absent in the request payload. 

This is a change which was handled in a different PR and is already part of develop. No change here.

modules/file-system/netapp-storage-pool/variables.tf
okrause: I have multiple valuable conditional checks, which I don't like to remove or work around with this crude approach: https://support.hashicorp.com/hc/en-us/articles/43291233547027-Error-Invalid-reference-in-variable-validation-in-Terraform-versions-prior-to-1-9

I am going to require TF >= 1.9.0 in my modules. This way users will run into an error which tells them they need a newer TF, instead of giving them an `Error: Invalid reference in variable validation` error.

examples/README.md
```diff
@@ -63,6 +63,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [xpk-n2-filestore](#xpk-n2-filestore--) ![community-badge] ![experimental-badge]
   * [gke-h4d](#gke-h4d-) ![core-badge]
   * [gke-g4](#gke-g4-) ![core-badge]
+  * [netapp-volumes.yaml](#netapp-volumesyaml--) ![community-badge]
```
sarthakag: s/community-badge/core-badge?

examples/README.md
```diff
@@ -1631,6 +1632,61 @@ This blueprint uses GKE to provision a Kubernetes cluster and a G4 node pool, al
 
 [gke-g4]: ../examples/gke-g4
 
+### [netapp-volumes.yaml] ![core-badge]
+
+This blueprint demonstrates how to provision NFS volumes as shares filesystems for compute VMs, using Google Cloud NetApp Volumes. It can be used as an  alternative to FileStore in blueprints.
```
sarthakag: s/shares/shared?

tools/cloud-build/daily-tests/tests/netapp-volumes.yml
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+---
+
+test_name: "netapp-volumes"
+deployment_name: "netapp-volumes-{{ build }}"
+workspace: /workspace
+blueprint_yaml: "{{ workspace }}/examples/netapp-volumes.yaml"
+region: us-central1
+zone: us-central1-a
+network: "{{ test_name }}-net"
```
sarthakag: s/test_name/deployment_name?

modules/file-system/netapp-storage-pool/versions.tf
```diff
@@ -0,0 +1,39 @@
...
+  provider_meta "google" {
+    module_name = "blueprints/terraform/hpc-toolkit:netapp-storage-pool/v1.70.0"
+  }
+  provider_meta "google-beta" {
+    module_name = "blueprints/terraform/hpc-toolkit:netapp-storage-pool/v1.70.0"
+  }
+
+  # Require Terraform version 1.9.0 or higher, since that added support
+  # for variable conditions. See https://support.hashicorp.com/hc/en-us/articles/43291233547027-Error-Invalid-reference-in-variable-validation-in-Terraform-versions-prior-to-1-9
+  required_version = ">= 1.9.0"
```
cboneti: I would not go above the currently supported version which is 1.5.7. This is *very* important. Please make the appropriate fixes to work on 1.5.7

cboneti (changes requested):
Hi, thanks for this. Could you please fix the code so it works on terraform 1.5.7?

sarthakag:
Hi @okrause, can you please resolve the open conversations which have been fixed?

modules/file-system/netapp-storage-pool/versions.tf
okrause: Fixed. Now require 1.5.7 min version.

tools/cloud-build/daily-tests/tests/netapp-volumes.yml
okrause: Fixed

examples/README.md
okrause: Fixed. Thanks for the catch.

okrause:
> Hi @okrause, can you please resolve the open conversations which have been fixed?

Done.

bytetwin:
@okrause Can you please squash all commits into one. Will do one final review on the squash and will be enable the merge after that

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+vars:
+  project_id: ## Set GCP Project ID Here ##
+  deployment_name: netapp-volumes
+  region: us-central1
+  zone: us-central1-a
+  pool_service_level: "EXTREME"  # Options: "STANDARD", "PREMIUM", "EXTREME"
+
+# Documentation for each of the modules used below can be found at
+# https://github.com/GoogleCloudPlatform/hpc-toolkit
```
bytetwin: https://github.com/GoogleCloudPlatform/cluster-toolkit

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+# It can be used to build compute clusters on top of it and as a drop-in replacement
+# for Filestore in existing blueprints.
+
+blueprint_name: netapp-volumes
+
+vars:
+  project_id: ## Set GCP Project ID Here ##
+  deployment_name: netapp-volumes
+  region: us-central1
+  zone: us-central1-a
```
bytetwin: region and zone should be left open like project_id unless netapp volumes is supported only in us-central1

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+  pool_service_level: "EXTREME"  # Options: "STANDARD", "PREMIUM", "EXTREME"
+
+# Documentation for each of the modules used below can be found at
+# https://github.com/GoogleCloudPlatform/hpc-toolkit
+
+deployment_groups:
+- group: primary
+  modules:
+  # Source is an embedded module, denoted by "modules/*" without ./, ../, /
+  # as a prefix. To refer to a local module, prefix with ./, ../ or /
```
bytetwin: can remove the above comment

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+    settings:
+      prefix_length: 24
+      service_name: "netapp.servicenetworking.goog"
+      deletion_policy: "ABANDON"
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: "netapp-pool"
```
bytetwin: Shouldn't pool_name be unique ? Otherwise it will cause conflict with a similar deployment. You can derive the pool name using deployment name like `$(vars.deployment_name)-netapp-pool`

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+      network_id: $(network.network_id)
```
bytetwin: This is not needed as the `use: network` will internally get translated to network_id requirement by gcluster

examples/netapp-volumes.yaml
```diff
@@ -0,0 +1,110 @@
...
+      service_level: $(vars.pool_service_level)
+      region: $(vars.region)
+      # allow_auto_tiering: true
+
+  - id: homefs
+    source: modules/file-system/netapp-volume
+    use: [netapp_pool]             # Create this pool using the netapp-storage-pool module
+    settings:
+      region: $(vars.region)
+      volume_name: "homefs"
```
bytetwin: this too should be unique within a storage pool i believe, if user tries to spin up multiple volumes within a pool.

You can define this in the var section and leave it open with a comment or you can prefix deployment name

modules/file-system/netapp-storage-pool/README.md
bytetwin: Can you please fix this. The phrase `A volume file-system container` is missing `is` for the sentence to complete

modules/file-system/netapp-storage-pool/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Standard: 16 KiBps throughput per provisioned KiB of volume capacity.
+- Premium: 64 KiBps throughput per provisioned KiB of volume capacity. Optional [auto-tiering].
+- Extreme: 128 KiBps throughput per provisioned KiB of volume capacity. Optional [auto-tiering].
+
+Check the [service level matrix][service-levels] for additional information on capability differences between service levels. Flex service levels are currently not supported, but you can connect to existing Flex volumes using the [pre-existing-network-storage module][pre-existing].
+
+### On-boarding NetApp Volumes
+NetApp Volumes uses [Private Service Access](https://cloud.google.com/vpc/docs/private-services-access) (PSA) to connect volumes to your network. Before you create a storage pool, make sure to [connect NetApp Volumes to your network](https://cloud.google.com/netapp/volumes/docs/get-started/configure-access/networking).
+
+Example of creating a storage pool using a new network:
```
bytetwin, line 36: please fix the example based on the comments raised in `examples/netapp-volumes.yaml`

modules/file-system/netapp-storage-pool/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+}
+
+variable "private_vpc_connection_peering" {
+  description = "The name of the private VPC connection peering."
+  type        = string
+  default     = "sn-netapp-prod"
+}
+
+variable "pool_name" {
+  description = "The name of the storage pool. Leave empty to use generates name based on deployment name."
```
bytetwin: Typo fix needed as below. 
"Leave empty to generate name based on deployment name"

modules/file-system/netapp-volume/README.md
```diff
@@ -0,0 +1,201 @@
...
+- id: homefs
+  source: modules/file-system/pre-existing-network-storage
+  settings:
+    server_ip: ## Set server IP here ##
+    remote_mount: nfsshare
+    local_mount: /home
+    fs_type: nfs
+```
+
+This creates a resource in Cluster Toolkit which references the specified NFS export, which will be mounted at `/home` by clients whuch USE it.
```
bytetwin: :s/whuch/which/

:s/USE/use/

modules/file-system/netapp-volume/outputs.tf
```diff
@@ -0,0 +1,66 @@
...
+  Runner to mount the file-system using an ansible playbook. The startup-script
+  module will automatically handle installation of ansible.
+  - id: example-startup-script
+    source: modules/scripts/startup-script
+    settings:
+      runners:
+      - $(your-fs-id.mount_runner)
+  ...
+  EOT
+  value       = local.mount_runner
```
bytetwin, line 50: Is this mount option supposed to be used in vm startup script ? I didn't see this in the example.

modules/file-system/netapp-volume/scripts/install-nfs-client.sh
```diff
@@ -0,0 +1,37 @@
...
+		else
+			echo "Unsupported version of centos/RHEL/Rocky"
+			return 1
+		fi
+		yum install --disablerepo="*" --enablerepo=${enable_repo} -y nfs-utils
+	elif [ -f /etc/debian_version ] || grep -qi ubuntu /etc/lsb-release || grep -qi ubuntu /etc/os-release; then
+		apt-get update --allow-releaseinfo-change-origin --allow-releaseinfo-change-label
+		apt-get -y install nfs-common
+	else
+		echo 'Unsuported distribution'
```
bytetwin: :s/Unsuported/Unsupported/

modules/file-system/netapp-volume/variables.tf
bytetwin: Please change the description as it states new pool will be created if not provided

modules/file-system/netapp-volume/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+variable "labels" {
+  description = "Labels to add to the NetApp volume. Key-value pairs."
+  type        = map(string)
+}
+
+variable "local_mount" {
+  description = "Mountpoint for this volume. Note: If set to the same as the `name`, it will trigger a known Slurm bug ([troubleshooting](../../../docs/slurm-troubleshooting.md))."
```
bytetwin: Can you add a condition to check if the mount != `name` to avoid the slurm bug

modules/file-system/netapp-volume/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+  description = <<-EOT
+    If true, the volume will be created with large capacity.
+    Large capacity volumes have 6 IP addresses and a minimal size of 15 TiB.
+    EOT
+  type        = bool
+  default     = false
+}
+
+variable "unix_permissions" {
+  description = "UNIX permissions for root inode the volume."
```
bytetwin: missing `in` - root inode in the volume

bytetwin (changes requested):

modules/file-system/netapp-volume/outputs.tf
okrause: This is code I took from filestore module to emulate how filestore plugs into ClusterToolkit. It is unclear to me how CT uses it, but I provide it for compatibility.
On that topic: I actually need to provide 6 IPs for large volumes or FlexCache and have the clients use one of them randomly for load balancing. I did implement that in the mount.sh script, but it looks like CT meanwhile changed the way it mounts clients, which made my approach not working. I want to pick up this issue after this PR is merged in a smaller follow-on PR. But need to investigate how CT mounts volumes provided by USE directive now.

modules/file-system/netapp-volume/scripts/install-nfs-client.sh
okrause: This is a copy from other modules. I am happy to fix it, but you may want to have a look into [duplicate-diff.py](tools/duplicate-diff.py). 
My questions:
1. should I add my `install-nfs-client.sh` and `mount.sh` to this file?
2. If I fix the type in my `install-nfs-client.sh`, who is going to fix it in the other modules?

modules/file-system/netapp-volume/variables.tf
okrause: I copied that from FileStore. First I thought it is a SLURM bug, but investigated now and it seems to be a FileStore-related problem. Using the same name for volume and mount works fine in GCNV. Just tested it with SLURM. I am removing the text about it.

examples/netapp-volumes.yaml
okrause: done

examples/netapp-volumes.yaml
okrause: done. Also updates examples/README.md to reflect the requirement to specify these variables.

examples/netapp-volumes.yaml
okrause: done

examples/netapp-volumes.yaml
okrause: done

modules/file-system/netapp-storage-pool/README.md
okrause: done

modules/file-system/netapp-volume/scripts/install-nfs-client.sh
okrause: I fixed the typo in my version of the file and leave it open if my files should be subject to duplicate checking, which would require fixes in all the other modules.

modules/file-system/netapp-volume/outputs.tf
bytetwin: The way these work is install*-client.sh is required to install the necessary software packages/drivers for the specific storage type and then mount.sh to mount the remote storage to the local directory in the vm.

[startup_from_network_storage.tf](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/compute/vm-instance/startup_from_network_storage.tf) is where the these scripts are called to perform the action. 

Your install-nfs-client.sh is installing only the nfs related package but the netapp volume also supports SMB. You need to install the respective drivers for smb too (and mount.sh should cater to using this driver for mount) though I also see a note in one of the readmes as

```
## Protocol support
Since Cluster Toolkit is currently built to provision Linux-based compute clusters, this module supports NFSv3 and NFSv4.1 only. SMB is blocked.
```
If above is true, then you may want to remove the SMB options in other readmes

For your 2nd question, volume mount varies by orchestrator type - slurm, gke and vm. The VM mounts as described in above approach.

GKE uses CSI drivers. For slurm, you can mount the same way as a vm - take a look at this [example](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml#L434-L453)

modules/file-system/netapp-storage-pool/README.md
```diff
@@ -0,0 +1,193 @@
...
+    settings:
+      project_id: $(vars.project_id)
+      region: $(vars.region)
+      network_name: $(vars.network)
+
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network]
+    settings:
+      pool_name: "eda-pool"
```
bytetwin, line 82: nit: this pool name also should be derived to avoid conflict

modules/file-system/netapp-storage-pool/README.md
```diff
@@ -0,0 +1,193 @@
...
+### Storage pool example
+
+The following example shows all available parameters in use:
+
+```yaml
+  - id: netapp_pool
+    source: modules/file-system/netapp-storage-pool
+    use: [network, private_service_access]
+    settings:
+      pool_name: "mypool"
```
bytetwin, line 97: nit: this pool name also should be derived to avoid conflict

modules/file-system/netapp-volume/README.md
```diff
@@ -0,0 +1,201 @@
...
+The following examples show the use of netapp-volume. They builds on top of an storage pool which can be provisioned using the [netapp-storage-pool][netapp-storage-pool] module.
+
+### Example with minimal parameters
+
+```yaml
+  - id: home_volume
+    source: modules/file-system/netapp-volume
+    use: [netapp_pool]  # Create this pool using the netapp-storage-pool module
+    settings:
+      volume_name: "eda-home"
```
bytetwin, line 45: nit: this volume name also should be derived to avoid conflict

modules/file-system/netapp-volume/README.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+### Example with all parameters
+
+```yaml
+  - id: shared_volume
+    source: modules/file-system/netapp-volume
+    use: [netapp_pool]              # Create this pool using the netapp-storage-pool module
+    settings:
+      volume_name: "eda-shared"
```
bytetwin, line 60: nit: this volume name also should be derived to avoid conflict

modules/file-system/netapp-storage-pool/README.md
okrause: This blueprint is mostly an example on how to create pools and volumes and how to use them in a VM. It serves as example on how to integrate GCNV into more complex workload blueprints.
Since data is usually not ephemeral, I suggest that volumes and pools use well-known names, which the user cares about. I like to stick to the existing, fixed names for this blueprint.

modules/file-system/netapp-storage-pool/README.md
okrause: This blueprint is mostly an example on how to create pools and volumes and how to use them in a VM. It serves as example on how to integrate GCNV into more complex workload blueprints.
Since data is usually not ephemeral, I suggest that volumes and pools use well-known names, which the user cares about. I like to stick to the existing, fixed names for this blueprint.

modules/file-system/netapp-volume/README.md
okrause: This blueprint is mostly an example on how to create pools and volumes and how to use them in a VM. It serves as example on how to integrate GCNV into more complex workload blueprints.
Since data is usually not ephemeral, I suggest that volumes and pools use well-known names, which the user cares about. I like to stick to the existing, fixed names for this blueprint.

modules/file-system/netapp-volume/README.md
okrause: This blueprint is mostly an example on how to create pools and volumes and how to use them in a VM. It serves as example on how to integrate GCNV into more complex workload blueprints.
Since data is usually not ephemeral, I suggest that volumes and pools use well-known names, which the user cares about. I like to stick to the existing, fixed names for this blueprint.

okrause:
@bytetwin Thank you for the review. All items reported are nit issues. I want to get back to the question: What is missing to merge this PR. We need to come to a close. Please advise.

modules/file-system/netapp-volume/outputs.tf
okrause: I am adding the volume to SLURM using the USE directive. This works fine. A few weeks ago, I was able to use a tweaked mount.sh to pass multiple IPs and mount round robin. But that stopped working. I investigated and it turned out that mount.sh was not being used at all anymore. I plan to investigate how the mounts are happening now and how i can insert doing round-robin mount of multiple NFS server IPs, but for now my goal is to get this work here finished and then built an addition on top of it.
Thank you for the information you provided!

bytetwin: Yes, install and mount for the supported storage types are directly done in slurm not in cluster toolkit code.

I would suggest you to try this [approach](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/machine-learning/a3-ultragpu-8g/a3ultra-slurm-blueprint.yaml#L434-L453)

tools/cloud-build/daily-tests/tests/netapp-volumes.yml
```diff
@@ -0,0 +1,28 @@
...
+blueprint_yaml: "{{ workspace }}/examples/netapp-volumes.yaml"
+region: us-central1
+zone: us-central1-a
+network: "{{ deployment_name }}-net"
+remote_node: "{{ deployment_name }}-0"
+post_deploy_tests:
+- test-validation/test-mounts.yml
+custom_vars:
+  mounts:
+  - /home
```
bytetwin, line 28: @okrause Can you add this snippet after the custom_vars(at the same level) for the test to pass. This is to cater for the providing the region and zone value to the blueprint


```
cli_deployment_vars:
  region: "{{ region }}"
  zone: "{{ zone }}"
```

okrause: Thank you for the support. I did what you said. Hope I got everything right.

okrause: I saw my commit picked up an unrelated file(s). Working on it.

okrause: I cleaned up the history. Looks fine now.

sarthakag (approved):

modules/file-system/netapp-volume/variables.tf
okrause: done

PR #4601, opened 2025-09-02
Adding unkillable_step_timeout (UnkillableStepTimeout) to the H4D blueprint.

Neelabh94:
It has been observed that for large jobs on 32 node+ clusters, once a job is cancelled, we often breach the default UnkillableStepTimeout of 300 seconds. This results in the affected nodes being marked as "DRAIN" rendering them unusable for any future jobs without explicit restart operations.

With this change we are providing an additional parameter within the blueprint to alter this value based on the workload and usage. The change was tested on a 32 node cluster that ran Intel MPI benchmark. All the nodes successfully returned to "IDLE" state after the job was cancelled while they were being marked "DRAIN" earlier.

An earlier PR added SlurmdTimeout to the blueprint to mitigate this issue but intensive testing revealed that to be an incorrect approach, hence removing the slurmd_timeout(SlurmdTimeout) parameter from the blueprint to avoid unnecessary clutter. Users can still specify slurmd_timeout in blueprints since it is still accepted as a valid blueprint param.

### Submission Checklist


samskillman (approved):
LGTM

PR #4562, opened 2025-08-22
Add Slurm default_auth_key

ACW101:
### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #4607, opened 2025-09-04
Add slurm-gke blueprint

ACW101:
### Submission Checklist


ACW101:
This PR include using the existing NFS server on the controller to distribute slurm key to nodeset running on GKE. If this is accepted, we can close #4562 as it's no longer needed.

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+vars:
+  project_id: managed-slurm-gke-dev
+  deployment_name: kevinmcw2
+  region: asia-southeast1
+  zone: asia-southeast1-c
+  authorized_cidr: 0.0.0.0/0 # <your-ip-address>/32
+  gcp_public_cidrs_access_enabled: false
+  gke_nodeset_replicas: 2
+  slinky_image: us-east1-docker.pkg.dev/managed-slurm-gke-dev/default/slurmd:ubuntu24-24.11-vanilla-pyxis
```
pawloch00: users are not able to pull this image. This should not be included here

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+      enable_login_public_ips: true
+
+  - id: slurm_controller
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
+    use:
+    - network
+    - slurm_login
+    - debug_partition
+    - compute_partition
+    - homefs
```
pawloch00, line 187: do we need homefs on controller machine?

ACW101: This came from the [hpc-slurm blueprint](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/hpc-slurm.yaml#L118) which this bluepoint is based on. Unless there is a reason we don't want it attached to controller, I think it's a good idea to keep them consistent across blueprints.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: Removed.

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+
+blueprint_name: slurm-gke
+
+vars:
+  project_id: managed-slurm-gke-dev
```
nick-stroud: Remove project name. For your development flow you can pass in as a cli arg or put in a deployments file.

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+  deployment_name: kevinmcw2
```
nick-stroud: pick something more representative for a public example.

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+  region: asia-southeast1
+  zone: asia-southeast1-c
+  authorized_cidr: 0.0.0.0/0 # <your-ip-address>/32
+  gcp_public_cidrs_access_enabled: false
+  gke_nodeset_replicas: 2
+  slinky_image: # your slinky image
```
nick-stroud: Where will user get this image? Can we point to the one published by SchedMD or do we need to bake in additional things?

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,201 @@
...
+      bandwidth_tier: gvnic_enabled
+      allow_automatic_updates: false
+
+  - id: compute_partition
+    source: community/modules/compute/schedmd-slurm-gcp-v6-partition
+    use:
+    - compute_nodeset
+    settings:
+      partition_name: compute
+      suspend_time: -1
```
nick-stroud: I think we can scrap this nodeset/partition from the example. I don't think it is adding anything and just makes example more convoluted.

community/modules/compute/gke-nodeset/README.md
nick-stroud, line 1: Please add to flat list of modules found in modules/README.md

Same for the gke-partition module.

community/modules/compute/gke-nodeset/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+variable "image" {
+  description = "The image for slurm daemon"
+  type        = string
+  default     = null
+
+  validation {
+    condition     = var.image != null
+    error_message = "Variable 'image' must be provided."
+  }
+}
```
nick-stroud, line 44: This is more concise and achieves the same goal:
```
variable "image" {
  description = "The image for slurm daemon"
  type        = string
  nullable    = false
}
```

community/modules/compute/gke-nodeset/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+
+variable "node_pool_names" {
+  description = "If set to true. The node group VMs will have a random public IP assigned to it. Ignored if access_config is set."
+  type        = list(string)
+  nullable    = false
+}
+
+variable "node_count" {
```
nick-stroud: community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md uses [node_count_static](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md#input_node_count_static) and [node_count_dynamic_max](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md#input_node_count_dynamic_max).

Maybe we should keep common here?

community/modules/compute/gke-nodeset/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+  description = "Name of the TPU accelerator (cloud.google.com/gke-tpu-accelerator annotation). Required when has_tpu=true"
+  type        = string
+  default     = null
+}
+
+variable "tpu_topology" {
+  description = "TPU topology. Required when has_tpu=true"
+  type        = string
+  default     = null
+}
```
nick-stroud: For tpu properties, are their any inter-dependencies. Ex. If one is set to A then another must be set to X? Or if one is set another must also be set.

If so we may want to make the tpu fields a compound object and do some validation on the input.

community/modules/compute/gke-nodeset/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+  description = "Path directory within `bucket_name` for Slurm cluster file storage."
+  type        = string
+  nullable    = false
+}
+
+variable "slurm_bucket" {
+  description = "GCS Bucket of Slurm cluster file storage."
+  type        = any
+  nullable    = true
+}
```
nick-stroud, line 84: Do we need to take slurm_bucket_name and slurm_bucket.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: We have a modified version of [slinky containers](https://github.com/SlinkyProject/containers/tree/main). I'll sync with you offline about the license implication of including this in CTK.

community/modules/compute/gke-nodeset/variables.tf
ACW101: Yes, I believe we can derive `tpu_chips_per_node` based on `tpu_accelerator` and `tpu_topology`. There is a table of supported topology for each TPU generation like https://cloud.google.com/tpu/docs/v6e. We can do some validation on the the (tpu_accelerator, tpu_topology) combination user provided. Should I do this in the same PR or do this in a separate one so I can update [schedmd-slurm-gcp-v6-nodeset-tpu](https://github.com/ACW101/cluster-toolkit/tree/blueprint/community/modules/compute/schedmd-slurm-gcp-v6-nodeset-tpu) as well?

community/modules/compute/gke-nodeset/variables.tf
ACW101: Removed `slurm_bucket_name`.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: done.

community/modules/compute/gke-nodeset/README.md
ACW101: done.

community/modules/compute/gke-nodeset/variables.tf
ACW101: done.

community/modules/compute/gke-nodeset/variables.tf
ACW101: done.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: done.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: done.

community/modules/compute/gke-nodeset/main.tf
```diff
@@ -0,0 +1,71 @@
...
+
+data "google_storage_bucket" "this" {
+  name = var.slurm_bucket[0].name
+
+  depends_on = [var.slurm_bucket]
+}
+
+### Slurm NodeSet
+locals {
+  nodeset = {
```
nick-stroud, line 51: Remind me, what is the marker that we use to know this is a GKE nodeset. I would think we would have some field in here like `slinky_nodeset=true`.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
```diff
@@ -0,0 +1,206 @@
...
+        image: ${image}
+        imagePullPolicy: IfNotPresent
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/bash
+              - -c
+              - scontrol update nodename=$NODE_NAME state=down reason=preStop &&
+                scontrol delete nodename=$NODE_NAME;
```
nick-stroud: IIUC this runs when the pods are stopping. Are these actually running as part of our solution. I suspect we don't actually want this. Suspend process would be config is updated (for static nodes) or slurm decides to power off (autoscaling nodes). Slurm patches slinky and handles move to power off idle state.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
```diff
@@ -0,0 +1,206 @@
...
+          claimName: slurm-key-pvc
+      - emptyDir:
+          medium: Memory
+        name: slurm-spool
+      - emptyDir:
+          medium: Memory
+        name: slurm-log
+      - name: home-pvc
+        persistentVolumeClaim:
+          claimName: ${home_pvc}
```
nick-stroud, line 199: Eventually we will want mounts to be dynamic and scalable. Think: I have a home filesystem and an /apps and a /scratch that are all separate shared filesystems.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
```diff
@@ -0,0 +1,206 @@
...
+          -print0 | xargs -0r chown -v \"$${SLURM_USER}:$${SLURM_USER}\"\n\tfind \"$${SLURM_DIR}\"
+          -type f -regextype posix-extended -regex \"^.*/(pro|epi)log-.*$\" -print0
+          | xargs -0r chmod -v 755\n\n\t# Inject secrets into certain config files\n\tlocal
+          dbd_conf=\"slurmdbd.conf\"\n\tif [[ -f \"$${SLURM_MOUNT}/$${dbd_conf}\" ]];
+          then\n\t\techo \"Injecting secrets from environment into: $${dbd_conf}\"\n\t\trm
+          -f \"$${SLURM_DIR}/$${dbd_conf}\"\n\t\tenvsubst <\"$${SLURM_MOUNT}/$${dbd_conf}\"
+          >\"$${SLURM_DIR}/$${dbd_conf}\"\n\t\tchown -v \"$${SLURM_USER}:$${SLURM_USER}\"
+          \"$${SLURM_DIR}/$${dbd_conf}\"\n\t\tchmod -v 600 \"$${SLURM_DIR}/$${dbd_conf}\"\n\tfi\n\n\t#
+          Display Slurm directory files\n\tls -lAF \"$${SLURM_DIR}\"\n}\n\nfunction
+          main() {\n\tinit::common\n\tinit::slurm\n}\nmain\n"
```
nick-stroud, line 139: Is there any way to make the format of this string more human readable. I think it would be more maintainable over time.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
```diff
@@ -0,0 +1,206 @@
...
+        value: present
+      volumes:
+      - emptyDir:
+          medium: Memory
+        name: etc-slurm
+      - emptyDir: {}
+        name: run
+      - name: slurm-config
+        persistentVolumeClaim:
+          claimName: slurm-key-pvc
```
nick-stroud: nit: in GCE version, we would unmount this volume after retrieving the key. I suspect this stays mounted. I think that is ok for this example.

community/modules/compute/gke-partition/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+  nullable    = false
+}
+
+variable "slurm_bucket_dir" {
+  description = "Path directory within `bucket_name` for Slurm cluster file storage."
+  type        = string
+  nullable    = false
+}
+
+variable "slurm_bucket" {
```
nick-stroud, line 39: Again, I suspect we don't need both slurm_bucket and slurm_bucket_name

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
nick-stroud: I think we can address this in future PR but if we plan to only support home. Which is maybe fine for POC example then we should validate that this is the only network_storage on the cluster.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
pawloch00: This is leftover from slinky default nodeset. We may remove it.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: For my understanding, what are usually stored in the home? If we store assets that will requires for running slurm worload, we might need to think about attaching separate storage for different user in the future and in that case GCS might be a more scalable solution. But I agree this can be in future PR.

community/modules/compute/gke-nodeset/main.tf
ACW101: Did some digging, looks like it determine type based on the instance template metadata.

https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py#L1705-L1706

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: Yeah this volume cannot be unmounted for the duration of pod execution. If we want to change this behavior we will need to mount it without csi.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: removed.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: I tried to reformat this inline but couldn't get the terraform templating to work correctly. One option is to pull this out into a separate template file to manage and pass it in as a template var. Wdyt?

community/modules/compute/gke-partition/variables.tf
ACW101: missed this, corrected.

community/examples/slurm-gke/slurm-gke.yaml
nick-stroud: I believe for this PR we are just pointing to the os slinky image.

community/modules/compute/gke-nodeset/variables.tf
nick-stroud: I think maybe as part of recent update we removed all of the templates except nodeset-general, or maybe it was always that way. Either way, 

The following code is fairly misleading:
```
  template_type = contains(local.machine_type_templates, var.machine_type) ? var.machine_type : (var.has_gpu ? "gpu-general" : var.has_tpu ? "tpu-general" : "general")
```

If our first draft only works for non-accelerator types we should not accept all the other variables just to create downstream error. 

I recommend we remove unsupported variables and update code path to just support single template file. Before doing so I would save a branch somewhere in your fork that has the current version so that we can easily bring it back when we are ready to support all the machine types.

community/modules/compute/gke-nodeset/main.tf
nick-stroud: I am realizing the full chain here and having some pause. 

We create a gke-node-pool -> this implicitly creates an instance template to back the node pool -> we look up that instance template in a data block based on naming -> we pass the instance template to slurm -> slurm looks up the instance template -> slurm decides if it is gke based on labels. 

This feels convoluted to me. We know it is a slurm nodeset so why not just pass that information in the config.yaml. I also think this would make debugging easier as the behavior could be known from looking at the config. I will bring up discussion offline.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
nick-stroud: For the sake of argument, lets assume that we give the users the ability to pick any storage system they want (lustre, filestore, gcs) and then can mount it under any file path they choose. And we allow them to provide a list of such storages. This is more or less how our existing slurm cluster implementation works now.

If you look at the [outputs from gke-persistent-volumes](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/acbf29f73e70716c6e81a5400d61092b4f956f0c/modules/file-system/gke-persistent-volume/outputs.tf#L17C9-L26) you can see that the `persistent_volume_claims` output includes a mount path and mount options. 

I think that instead of taking the claim name we should take the `persistent_volume_claims` object and we should use the path specified there instead of hard coding /home.

In the future we would want to expand support from a single claim to a list of claims.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
nick-stroud: Did you try using the `- |` syntax to define a multiline string? I think this would allow you to get rid of the escape characters.

If not that it seems like the least we could do would be to put each new line on a new line.

I would also be open to passing in as a var if that could improve things. Feel free to propose something and we can iterate.

community/modules/compute/gke-nodeset/variables.tf
ACW101: Agreed, I removed the unused variable for now and keep the minimal required ones for generic machine type.

community/modules/compute/gke-nodeset/main.tf
ACW101: Added a new attribute `gke-nodepool` to nodeset config and used it to determine if the node is gke or not, see 94474564600f312adfd1766138bb3393604080ba.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: See commit a61adbbab05dc2f19ec33dfff2af8fbdb9d1d67e.

I see your point. Instead of passing around PVCs, I propose following the same convention as the existing GCE nodeset to [pass the output network_storage of filestore](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf#L560-L570) and create PV/PVC within the gke nodeset module. 

In addition, I have encountered some terraform error on destroy path related to the order of destruction of PVC, namespace and nodeset resources. To have more control over k8s objection creation and deletion, I might need to have all k8s resources in a `helm` chart in the next revision and passing in network storage here will also help with that migration.

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
ACW101: `- |` was what I tried. Also I learned these are third-party code that came from [slurm operator helm chart](https://github.com/SlinkyProject/slurm-operator/blob/a7ea2de176cf1755a895564c66ee98113c86f424/internal/builder/scripts/initconf.sh) so we probably won't be changing it. In the followup PR when I switch from `kubectl_apply` to `helm_release` we will be able to stop relying on terrafom template and I'll be able to make it more readible.

community/modules/compute/gke-nodeset/variables.tf
```diff
@@ -0,0 +1,108 @@
...
+  type        = list(string)
+  nullable    = false
+}
+
+variable "network_storage" {
+  description = "An array of network attached storage mounts to be configured on nodes."
+  type = object({
+    server_ip             = string,
+    remote_mount          = string,
+    local_mount           = string,
```
nick-stroud, line 97: For long term interface stability I think this should take a list.

Then since the current code only uses a list of size 1 and hardcodes it to home, we should add the following validations:

1. list is of size 1
2. local_mount = /home

community/modules/compute/gke-nodeset/templates/nodeset-general.yaml.tftpl
nick-stroud: I think this is a reasonable solution, see my follow up: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4607#discussion_r2366764123

community/modules/compute/gke-nodeset/main.tf
nick-stroud: This looks better to me.

community/examples/slurm-gke/slurm-gke.yaml
nick-stroud, line 1: With so much code in this PR I think it will be important to have an integration test to make sure we keep it from going stale. 

I would like to see integration test as a separate PR and then we can submit them at relatively the same time.

community/modules/compute/gke-nodeset/variables.tf
ACW101: Done in 7f742e37a827be63ccf7758d721eefd80a889811

community/examples/slurm-gke/slurm-gke.yaml
ACW101: I ended up including it in the same PR 6872248507dbd7653731ef43040d073151ebe94e

tools/cloud-build/daily-tests/tests/slurm-gke.yml
```diff
@@ -0,0 +1,42 @@
+# Copyright 2024 Google LLC
```
pawloch00, line 1: Can we have simple task ex. srun -p gke hostname -s included as part of this test?

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,175 @@
+# Copyright 2025 Google LLC
```
pawloch00, line 1: I created a cluster based on this blueprint. In the final message I see:

```
To test Slurm functionality, connect to the controller to use Slurm client commands:
  kubectl exec -it statefulsets/slurm-controller \
    --namespace=slurm \
    -- bash --login
    ```
 Is this some leftover from slinky? Because I don't recall possibility to connect to controller with this command

File: tools/cloud-build/daily-tests/tests/slurm-gke.yml (hunk 18 above)
Comment (ACW101) [reply tools/cloud-build/daily-tests/tests/slurm-gke.yml:1]: Currently this test already run tests defined in [test-partitions.yml](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-mounts.yml), which include command like `srun mount`. Do you think we still need to add this test?

File: community/examples/slurm-gke/slurm-gke.yaml (hunk 19 above)
Comment (ACW101) [reply community/examples/slurm-gke/slurm-gke.yaml:1]: Good catch, removed irrelevant instruction from slinky in ce743604a7b5adbe036deaea175e160e18167e0f

File: tools/cloud-build/daily-tests/tests/slurm-gke.yml (hunk 18 above)
Comment (pawloch00) [reply tools/cloud-build/daily-tests/tests/slurm-gke.yml:1]: No need for this in such case. Thanks for reassuring

Comment (pawloch00) [inline tools/cloud-build/daily-tests/tests/slurm-gke.yml:1]: This test does not checks if node stays idle for the whole time. We need gke nodes to not be powered up/down - stay idle/comp for the whole time. We either need this to be checked here, or add UT covering this somewhere(probably in slurmsync tests?)

**Review by nick-stroud — APPROVED** [review]

File: community/examples/slurm-gke/slurm-gke.yaml [hunk 20]
```diff
@@ -0,0 +1,174 @@
...
+---
+
+blueprint_name: slurm-gke
+
+vars:
+  project_id: ## Set GCP Project ID Here ##
+  deployment_name: slurm-gke
+  region: us-west4
+  zone: us-west4-c
+  authorized_cidr: 0.0.0.0/0 # <your-ip-address>/32
```
samskillman: Can we follow the pattern in the other GKE blueprints? https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/examples/gke-a4/gke-a4.yaml

No default authorized_cidr, regions, zone.

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,174 @@
...
+  gcp_public_cidrs_access_enabled: false
```
samskillman: This is the default, any reason to put it here?

samskillman (dismissed):
Minor comments to bring closer to other GKE blueprints, otherwise LGTM

samskillman (dismissed):

community/examples/slurm-gke/slurm-gke.yaml
ACW101: Removed.

community/examples/slurm-gke/slurm-gke.yaml
ACW101: Updated in ee07bb06ca8c47b3061a10a1bf08330798a41289.

samskillman (approved):

nick-stroud (approved):

community/examples/slurm-gke/slurm-gke.yaml
```diff
@@ -0,0 +1,192 @@
...
+blueprint_name: slurm-gke
+
+vars:
+  # The following variables should be over-written in the deployment.yaml file.
+  # Your GCP Project ID
+  project_id: ## Set GCP Project ID Here ##
+
+  # This should be unique across all of your Cluster
+  # Toolkit Deployments.
+  deployment_name: slurm-gke
```
abbas1902: just fyi getting this error when deploying the blueprint as-is:
```

  on main.tf line 213, in module "slurm_controller":
 213:   slurm_cluster_name           = var.deployment_name
    ├────────────────
    │ var.slurm_cluster_name is "slurm-gke"

Variable 'slurm_cluster_name' must be a match of regex
'^[a-z](?:[a-z0-9]{0,9})$'.
```

ACW101: Since this is minor and PR already approved, I'll relax the validation a bit in a separate PR to allow `-`.

abbas1902: It may be easier to pick a different default name, I think that validation has been around for a long time (so may be tricky to evaluate unintended side effects)

ACW101: Fixed.

PR #4618, opened 2025-09-07
Refactoring in gke persistent module

vikramvs-gg:
The changes include the following refactoring in gke-persistent-volume module:
1. Map Lookups for Templates and Template Variables instead of nested ternary operators used across the file.
2. Simplified Content Rendering: pv_content and pvc_content are now generated with single templatefile calls, using the maps to get the correct path and variables.

This overalls improves the readability and maintainability of the module as the storage types are added to this module in future.


modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -20,117 +20,107 @@ locals {
...
+    pv_name   = local.pv_name
+    pvc_name  = local.pvc_name
+    labels    = local.labels
+    capacity  = "${var.capacity_gb}Gi"
+    namespace = var.namespace
+  }
+  pvc_template_vars = {
+    gcs       = local.common_pvc_vars
+    lustre    = local.common_pvc_vars
+    filestore = local.common_pvc_vars
```
bytetwin: If the pvc_vars is the same for all storage types, why is this map needed ?

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -20,117 +20,107 @@ locals {
...
+  }
 
-  gcs_pv_contents = templatefile(
-    "${path.module}/templates/gcs-pv.yaml.tftpl",
-    {
+  # Variables for PV templates (specific to each storage type)
+  pv_template_vars = {
```
bytetwin, line 78: This template vars has many repeated values like pv_name, capacity, labels, namespace etc.

Defining them as
```
 common_pv_vars = {
    pv_name = local.pv_name
    labels = local.labels
.....
}
```

And use it as below

```
pv_template_vars = {
   gcs = merge(local.common_pv_vars, 
                        {
                              mount_options = var.gcs_bucket_name != null ? split(",", var.network_storage.mount_options) : []
                        }
.... for other storage types
}
```

This will be more cleaner

modules/file-system/gke-persistent-volume/outputs.tf
```diff
@@ -15,17 +15,22 @@
...
-  depends_on = [kubectl_manifest.pvc]
 }
 
 output "pvc_name" {
-  description = "The name for k8s PVC created by this module."
+  description = "The name of the Kubernetes PVC created by this module."
   value       = local.pvc_name
 }
+
+output "pvc_namespace" {
```
bytetwin: Why is this output needed as the pvc is in the same namespace as input

modules/file-system/gke-persistent-volume/main.tf
vikramvs-gg: Removed this map and used local.common_pvc_vars instead as all the attrs are same.

modules/file-system/gke-persistent-volume/main.tf
vikramvs-gg: Used common pv_vars and merged storage specific attrs in storage pv_vars as suggested.

modules/file-system/gke-persistent-volume/outputs.tf
vikramvs-gg: This is not needed, was added for just debugging.

bytetwin (approved):

PR #4620, opened 2025-09-08
Added support for managed luster in a3-high

rachit-google:
Added support for managed luster in a3-high

### DIFF
--- a/examples/machine-learning/a3-highgpu-8g/ml-slurm-a3-1-image.yaml
+++ b/examples/machine-learning/a3-highgpu-8g/ml-slurm-a3-1-image.yaml
@@ -102,7 +102,7 @@ deployment_groups:
           set -e -o pipefail
           ansible-galaxy role install googlecloudplatform.google_cloud_ops_agents
           ansible-pull \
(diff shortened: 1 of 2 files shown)


examples/machine-learning/a3-highgpu-8g/ml-slurm-a3-1-image.yaml
```diff
@@ -102,7 +102,7 @@ deployment_groups:
           set -e -o pipefail
           ansible-galaxy role install googlecloudplatform.google_cloud_ops_agents
           ansible-pull \
-              -U https://github.com/GoogleCloudPlatform/slurm-gcp -C 6.9.1 \
+              -U https://github.com/GoogleCloudPlatform/slurm-gcp -C 6.10.5 \
```
arpit974: this change might already be taken care by a PR from laveen

arpit974 (commented):
similar to other PR:
we should also add "install_managed_lustre": false,"
in below block
- type: data
        destination: /var/tmp/slurm_vars.json
        content: |

PR #4644, opened 2025-09-12
Update H4D blueprint: disable automatic updates, provide image info, and delete duplicate filestore

Neelabh94:
The PR makes the following changes to the H4D blueprint:

1> Disable automatic updates to prevent unattended updates leading to unstable clusters.

2> Explicitly mention the image used for nodeset, controller and login nodes within the blueprint instead of relying solely on the default in CTK.

3> Removes duplicate file store "appfs". There's no benefit of having two separate file store instances in the base blueprint. There are probably use cases of using separate file stores for specific apps but is not a general practice and hence needs to be removed.

### Submission Checklist

### DIFF
--- a/examples/hpc-slurm-h4d.yaml
+++ b/examples/hpc-slurm-h4d.yaml
@@ -19,6 +19,9 @@ vars:
   region: us-central1
   zone: us-central1-a
   rdma_net_range: 192.168.128.0/18
+  instance_image:
+    family: slurm-gcp-6-11-hpc-rocky-linux-8
+    project: schedmd-slurm-public
 
   #Provisioning model, select one, lack of selection will rely on on-demand capacity
   h4d_dws_flex_enabled: false
@@ -61,15 +64,6 @@ deployment_groups:
       filestore_share_name: homeshare
       local_mount: /home
 
-  - id: appsfs
-    source: modules/file-system/filestore
-    use: [h4d-slurm-net-0]
-    settings:
-      filestore_tier: BASIC_SSD
-      size_gb: 2560
-      filestore_share_name: appsshare
-      local_mount: /apps
-
   - id: h4d_startup
     source: modules/scripts/startup-script
     settings:
@@ -83,6 +77,9 @@ deployment_groups:
     source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
     use: [h4d_startup, h4d-slurm-net-0]
     settings:
+      # Unattended upgrades are disabled to prevent automatic daily updates which may lead to potential instability.
+      # For more details, see https://cloud.google.com/compute/docs/instances/create-hpc-vm#disable_automatic_updates
+      allow_automatic_updates: false
       bandwidth_tier: gvnic_enabled
       machine_type: h4d-highmem-192-lssd
       node_count_static: 2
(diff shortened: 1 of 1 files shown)


examples/hpc-slurm-h4d.yaml
```diff
@@ -62,15 +65,6 @@ deployment_groups:
       filestore_share_name: homeshare
       local_mount: /home
 
-  - id: appsfs
```
bytetwin, line 64: What was the rationale behind this additional file store and why is this removed ? Can you add details to the PR description.

Also has this updated blueprint been tested - it will fail because appsfs is a dependency to controller module

examples/hpc-slurm-h4d.yaml
```diff
@@ -19,6 +19,9 @@ vars:
   region: us-central1
   zone: us-central1-a
   rdma_net_range: 192.168.128.0/18
+  slurm_image:
```
bytetwin: This is a default image specified in the nodeset and login/controller nodes - https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf#L77-L93

Why is it needed here ?

bytetwin (dismissed):

examples/hpc-slurm-h4d.yaml
Neelabh94: Removed the dependency from controller.

examples/hpc-slurm-h4d.yaml
Neelabh94: The idea is to simply make it more explicit for the users to know what image they are using and provide flexibility for easy edits of the image

samskillman: Instead of a new variable, use instance_image, then below it gets inherited automatically.

examples/hpc-slurm-h4d.yaml
```diff
@@ -84,11 +78,15 @@ deployment_groups:
...
     settings:
+      # Unattended upgrades are disabled to prevent automatic daily updates which may lead to potential instability.
+      # For more details, see https://cloud.google.com/compute/docs/instances/create-hpc-vm#disable_automatic_updates
+      allow_automatic_updates: false
       bandwidth_tier: gvnic_enabled
       machine_type: h4d-highmem-192-lssd
       node_count_static: 2
       node_count_dynamic_max: 0
       enable_placement: false
+      instance_image: $(vars.slurm_image)
```
samskillman:
```suggestion
```

examples/hpc-slurm-h4d.yaml
```diff
@@ -132,11 +130,13 @@ deployment_groups:
     settings:
       machine_type: n2-standard-4
       enable_login_public_ips: true
+      instance_image: $(vars.slurm_image)
```
samskillman:
```suggestion
```

examples/hpc-slurm-h4d.yaml
```diff
@@ -132,11 +130,13 @@ deployment_groups:
...
 
   - id: slurm_controller
     source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
-    use: [h4d-slurm-net-0, h4d_partition, slurm_login, homefs, appsfs]
+    use: [h4d-slurm-net-0, h4d_partition, slurm_login, homefs]
     settings:
       enable_controller_public_ips: true
+      instance_image: $(vars.slurm_image)
```
samskillman:
```suggestion
```

examples/hpc-slurm-h4d.yaml
samskillman:
```suggestion
  instance_image:
```

samskillman (changes requested):
Left some inline suggestions

arpit974 (approved):

PR #4657, opened 2025-09-17
Add nvidia-imex-* to list of held packages

nick-stroud:
I opted to add them everywhere for consistency (when really only needed on A4X).

Tested: confirmed that holding on image that does not have package installed does not fail.


samskillman (approved):

RachaelSTamakloe (commented):
I believe `examples/machine-learning/a3-megagpu-8g/a3mega-slurm-gcsfuse-lssd-blueprint.yaml` needs to be updated as well.

PR #4682, opened 2025-09-22
Enable Private Nodes by default in GKE Node Pool

kadupoornima:
Disable public ip allocation to gpu node pool.

This is a potential security concern highlighted by the networking team. Fixed it by changing the default value of the enable_private_nodes variable in gke-node-pool module to true.


shubpal07 (approved):
LGTM

shubpal07 (approved):

PR #4699, opened 2025-09-25
feat: Add GKE Inference Gateway support

SinaChavoshi:
This PR introduces support for the GKE Inference Gateway by adding a new feature flag to the `gke-cluster` module.
Key changes:

- Added a new boolean variable, `enable_inference_gateway`, to the `gke-cluster` module. When set to `true`, this flag:
   - Enables the `HttpLoadBalancing` add-on in the GKE cluster, a prerequisite for the Gateway API.
   - Deploys the necessary Inference Gateway Custom Resource Definitions (CRDs) directly from the official Kubernetes SIGs repository.

 - Created a new example blueprint, `gke-a3-highgpu-inference-gateway.yaml`, to demonstrate how to enable this feature. This blueprint also includes the required `REGIONAL_MANAGED_PROXY` 
      subnet.
 - Updated the `examples/README.md` to include documentation for the new blueprint, guiding users on how to deploy a sample workload after the cluster is provisioned.

### Submission Checklist


samskillman:
Hi @SinaChavoshi  - would you mind rebasing your changes on top of the current upstream `develop` branch, and make the target of this PR to go to the `develop` branch as well? That follows our development pattern. Thanks!

SinaChavoshi:
> Hi @SinaChavoshi - would you mind rebasing your changes on top of the current upstream `develop` branch, and make the target of this PR to go to the `develop` branch as well? That follows our development pattern. Thanks!

Done.

examples/gke-a3-highgpu-inference-gateway.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - monitoring.viewer
+      - stackdriver.resourceMetadata.writer
+      - storage.objectAdmin
+      - artifactregistry.reader
+
+  - id: gpunets
+    source: modules/network/multivpc
+    settings:
+      network_name_prefix: $(vars.deployment_name)-gpunet
+      global_ip_address_range: 192.169.0.0/16
```
samskillman: I think you grabbed this range from a different example where there is a known bug. I think we want to use `192.168.0.0/16` here instead, which is the actual private ip range.

examples/gke-a3-highgpu-inference-gateway.yaml
```diff
@@ -0,0 +1,128 @@
...
+        gvnic_start_index: 1
+      enable_inference_gateway: true
+    outputs: [instructions]
+
+  - id: a3_highgpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, gpunets, node_pool_service_account]
+    settings:
+      machine_type: a3-highgpu-8g
+      static_node_count: 2
```
samskillman: Is it valuable here to make machine_type a variable at the top? What would stop us from using the same structure here for G4, for example? 

Let's definitely make static_node_count a variable above. Any reason why it should be static? Is there a nice way to configure auto-scaling? For a3 does that require matching to a reservation in most cases? If so, we should include the consumption models in this blueprint.

examples/gke-a3-highgpu-inference-gateway.yaml
```diff
@@ -0,0 +1,128 @@
...
+      zones: [$(vars.zone)]
+    outputs: [instructions]
+
+  - id: cpu_pool
+    source: modules/compute/gke-node-pool
+    use: [gke_cluster, node_pool_service_account]
+    settings:
+      name: "default-pool"
+      machine_type: "e2-standard-2"
+      static_node_count: 1
```
samskillman: Typically I see auto-scaling for the default pool so that we don't end up hitting issues for non-gpu workloads running out of resources. Any thoughts here?

examples/gke-a3-highgpu-inference-gateway.yaml
```diff
@@ -0,0 +1,128 @@
...
+      zones: [$(vars.zone)]
+
+  - id: workload_component_install
```
samskillman: Are Kueue/Jobset used in inference gateway work?

samskillman (changes requested):
Mostly it looks good, thank you for this contribution! I've added a few suggestions that we should discuss/fix up before merging.

examples/gke-a3-highgpu-inference-gateway.yaml
SinaChavoshi: Done.

examples/gke-a3-highgpu-inference-gateway.yaml
SinaChavoshi: Thanks for the suggestion! Making machine_type a variable is a great idea to help us limit the number of blueprints we need to maintain. To keep the scope of this CL down, I'll create an issue to track this as a follow-up feature.

Regarding static_node_count, Done. I've moved it to a variable at the top. I'm specifically focusing this blue print to match the vllm sample here https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway. 

Finally regarding reservation, yes that's a great point. I have added a corresponding section for reservation and Spot VM configuration.

examples/gke-a3-highgpu-inference-gateway.yaml
SinaChavoshi: Completely agree, updated to use auto scaling.

examples/gke-a3-highgpu-inference-gateway.yaml
SinaChavoshi: Done, removed both.

examples/gke-a3-highgpu-inference-gateway.yaml
SinaChavoshi: created #4717 to track machine type parametrization.

SinaChavoshi: regarding reservation ... I will comment the reservation out as if the user don't have a reservation on the project they will get the following error if they choose ANY_RESERVATION

```
Error: Iteration over null value

  on modules/embedded/modules/compute/gke-node-pool/reservation_definitions.tf line 21, in locals:
  21:   input_reservation_names    = [for r in try(var.reservation_affinity.specific_reservations, []) : split("/", r.name)[0]]
    ├────────────────
    │ var.reservation_affinity.specific_reservations is null

A null value cannot be used as the collection in a 'for' expression.

```

cboneti (approved):
lgtm, thanks

cboneti (approved):
Approved, pending passing all relevant tests.

samskillman (approved):

examples/README.md
```diff
@@ -50,6 +50,7 @@ md_toc github examples/README.md | sed -e "s/\s-\s/ * /"
   * [gke-a3-ultragpu.yaml](#gke-a3-ultragpuyaml-) ![core-badge]
   * [gke-a3-megagpu](#gke-a3-megagpuyaml-) ![core-badge]
   * [gke-a3-highgpu](#gke-a3-highgpuyaml-) ![core-badge]
+  * [gke-a3-highgpu-inference-gateway.yaml](#gke-a3-highgpu-inference-gatewayyaml--) ![core-badge]
```
raushan2016, line 53: needs to be `-` instead of `--`

SinaChavoshi: Good catch! thank you! addressed under PR #4839

PR #4741, opened 2025-10-09
Adding GKE sample for running nvidia-bug-report

raushan2016:
### Changes
- Adding examples on how to run `nvidia-bug-report` when using COS OS image in GKE as a pod.
- Renaming from being GCE specific to GCP level so that both GCE and GKE examples is available for COS OS image.

### Why
- Parity with existing example of running `nvidia-bug-report` on GCE VM with GKE node using COS OS image
- Running `nvidia-bug-report` can be challenging as it will require some kernel modules as well as SSH into the GKE node. This approach uses K8s native to run it as K8s Pod using the same image which was previous used for running as docker container for GCE node using COS Image.

### Submission Checklist


community/cos-nvidia-bug-report/bug-report-pod.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - effect: NoSchedule
+    key: nvidia.com/gpu
+    operator: Exists
+  containers:
+  - name: bug-reporter-container
+    resources:
+      limits:
+        nvidia.com/gpu: "8"
+      requests:
+        nvidia.com/gpu: "8"
```
bytetwin: This hardcoding will prevent the pod being scheduled on nodes with < 8 gpus. Lets make this an arg for user to fill or clarify in readme

raushan2016: Updated the doc and made it a user field

PR #4796, opened 2025-10-28
Add NCCL integration test for A3 Mega GPUs 

saara-tyagi27:
This PR introduces a new integration test to validate the performance of the NVIDIA NCCL test on A3 Mega GPU instances.

The test performs the following steps:

1. Clones the NCCL tests repository and builds the tests.
2. Runs the all_reduce_perf and all_gather_perf tests to measure the average bus bandwidth.
3. Asserts that the average bus bandwidth is above a certain threshold.

This test will help to ensure that NCCL is performing as expected on A3 Mega GPU instances.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-nccl.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-nccl.yml
@@ -14,7 +14,6 @@
 
 ---
 - name: Run prerequisite NCCL scripts
-  # Use the 'shell' module to execute complex shell commands
   shell: |
     set -x -e
     cd {{ ansible_user_dir }}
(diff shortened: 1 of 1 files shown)


tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-nccl.yml
```diff
@@ -0,0 +1,97 @@
...
+
+- name: Extract average bus bandwidth
+  shell: "echo '{{ nccl_output.stdout }}' | awk -F':' '{print $NF}'"
+  register: avg_bus_bandwidth
+  changed_when: false
+  when: nccl_output.stdout is defined and nccl_output.stdout != ""
+
+- name: Ensure average bus bandwidth is sufficient
+  fail:
+    msg: "Average bus bandwidth is {{ avg_bus_bandwidth.stdout | trim }} GB/s, which is below the threshold of 30 GB/s."
```
LAVEEN, line 96: If the average bandwidth for a3 mega slurm is more than 100 GB/s , why are we setting this to 30 GB/s?

saara-tyagi27: I checked in gke nccl tests and the threshold value was set to 30. So i used the same for slurm.

PR #4834, opened 2025-11-11
Adding A4X Base Support to JBVM

LAVEEN:
This PR adds support for the gpu_topology attribute within the placement_policy for VM instances. This is necessary to support the upcoming A4x GPUs, which require a specific GPU topology to be defined for optimal performance.


modules/compute/vm-instance/variables.tf
```diff
@@ -321,18 +321,18 @@ variable "placement_policy" {
     condition     = var.placement_policy == null ? true : try(keys(var.placement_policy), null) != null
     error_message = <<-EOT
     The var.placement_policy should be either unset/null or be a map/object with
-    fields: vm_count (number), availability_domain_count (number), collocation (string), max_distance (number).
+    fields: vm_count (number), availability_domain_count (number), collocation (string), max_distance (number), gpu_topology (string).
```
bytetwin: This comment is for line 318 type = any.

Terraform supports optional since 1.5 version. So we should make use of that and change the type optional(map(....))

modules/compute/vm-instance/variables.tf
```diff
@@ -314,25 +314,30 @@ variable "placement_policy" {
...
-  type    = any # It's a workaround of lack of `optional` in Terraform 1.2
+  type = object({
+    vm_count                = optional(number)
+    availability_domain_count = optional(number)
+    collocation             = optional(string)
+    max_distance            = optional(number)
+    gpu_topology            = optional(string)
+  })  
   default = null
   validation {
```
bytetwin: With the new type check, the validation blocks are not needed. The type checking will ensure the placement_policy is either null or a map of provided keys with each value of the specified type.

Please ensure that this type validation is working on terraform version 1.5.7 (the ctk supported version). You can remove the valdiation blocks after  testing to ensure that wrong values are flagged by the type checker

kvenkatachala333 (approved):

PR #4836, opened 2025-11-12
Add DWS Flex Start feature for G4 VMs

kadupoornima:
This change adds an example blueprint showing how to use the DWS Flex Start feature for G4 VMs.

This change has been tested by successfully deploying the cluster and running sample jobs to verify dynamic node creation and deletion.

About Flex-start VMs: https://cloud.google.com/compute/docs/instances/about-flex-start-vms

We have also refactored the gke-consumption-options folder into individual NPIs (initially was only for A3U).


kadupoornima:
Closed after discussion. Will publish these blueprints internally.

PR #4845, opened 2025-11-13
Modify the wait-for-startup-script to fix test failures

agrawalkhushi18:
1. A key log line, `Finished google-startup-scripts.service - Google Compute Engine Startup Scripts`, was present in our Ansible test ([wait-for-startup-script.yml](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/tools/cloud-build/daily-tests/ansible_playbooks/tasks/wait-for-startup-script.yml))  but was missing in the wait-for-startup-status.sh script. This mismatch was causing the ansible-vm test failure. This line is now added to the .sh script, resolving the inconsistency.                                      
                                                                                                                                                                
2. The script is also refactored to more reliably detect the true status of a startup script. It now used grep to more precisely extract the numeric exit code from specific log messages. Earlier the sed, was brittle and could include non-numeric lines in STATUS, leading to  incorrect comparisons.
Crucially, it prioritizes explicit failure messages from our script before checking for generic "Finished" messages from the GCE service. This prevents the script from incorrectly reporting success when the user's script has actually failed which was the case earlier when I tested.

### Submission Checklist

### DIFF
--- a/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
+++ b/community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
@@ -58,11 +58,12 @@ fetch_cmd="gcloud compute instances get-serial-port-output ${INSTANCE_NAME} --po
 # finishes on the new guest agent
 FINISH_LINE="startup-script exit status"
 # Match string for failures on the new guest agent
-FINISH_LINE_ERR="Script.*failed with error:"
+FINISH_LINE_ERR="Script \"startup-script\" failed with error:"
 
 # NEW: Accept also these finish lines as success.
 STARTUP_SCRIPT_SUCCEEDED_LINE="google-startup-scripts.service: Succeeded."
 STARTUP_SCRIPT_FINISHED_LINE="Finished Google Compute Engine Startup Scripts."
+STARTUP_SCRIPT_SERVICE_FINISHED_LINE="Finished google-startup-scripts.service - Google Compute Engine Startup Scripts."
 
 NON_FATAL_ERRORS=(
 	"Internal error"
@@ -72,7 +73,7 @@ until [[ now -gt deadline ]]; do
 	ser_log=$(
 		set -o pipefail
 		${fetch_cmd} 2>"${error_file}" |
-			c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}\|${STARTUP_SCRIPT_SUCCEEDED_LINE}\|${STARTUP_SCRIPT_FINISHED_LINE}"
+			c1grep "${FINISH_LINE}\|${FINISH_LINE_ERR}\|${STARTUP_SCRIPT_SUCCEEDED_LINE}\|${STARTUP_SCRIPT_FINISHED_LINE}\|${STARTUP_SCRIPT_SERVICE_FINISHED_LINE}"
 	) || {
 		err=$(cat "${error_file}")
 		echo "$err"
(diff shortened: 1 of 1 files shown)


community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -95,31 +96,44 @@ until [[ now -gt deadline ]]; do
...
-	echo "startup-script finished successfully"
-elif [[ "${STATUS}" == 1 ]]; then
+
+# --- Prioritize explicit failure from the script itself ---
+if [[ "${LAST_EXIT_STATUS}" == 1 ]]; then
 	echo "startup-script finished with errors, ${INSPECT_OUTPUT_TEXT}"
 	echo "${fetch_cmd}"
+	exit 1
+# --- Then explicit success from the script itself ---
+elif [[ "${LAST_EXIT_STATUS}" == 0 ]]; then
```
bytetwin, line 114: how can `LAST_EXIT_STATUS` be 0 ? Your grep command is looking for the failure log message

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
```diff
@@ -95,31 +96,44 @@ until [[ now -gt deadline ]]; do
 done
 
 # This line checks for an exit code - the assumption is that there is a number
-# at the end of the line and it is an exit code
-STATUS=$(sed -r 's/.*([0-9]+)\s*$/\1/' <<<"${ser_log}" | uniq)
+# at the end of the line and it is an exit code.
+# Modified to correctly extract the last numeric exit status from the relevant log line.
+LAST_EXIT_STATUS=$(echo "${ser_log}" | grep -oP "(?<=Script \"startup-script\" failed with error: exit status )[0-9]+" | tail -n 1)
```
bytetwin, line 100: Does it always print `Script "startup-script"` failed ? in that case why do we have greedy match regex here - https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/4845/files#diff-1377115189997b6be62a5f9785554d9e996c36db632bbd7395d96c19b65623a5R61

Can you reverify this

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
agrawalkhushi18: `LAST_EXIT_STATUS` can be 0 because it first tries to extract from explicit failure messages. If no failure is found, it falls back to a more  generic log line : `startup-script exit status`, which can indeed report 0 for successful script execution.

community/modules/scripts/wait-for-startup/scripts/wait-for-startup-status.sh
agrawalkhushi18: As per my understanding, the modules/scripts/startup-script module, specifically its main.tf file, is responsible for assembling and injecting the startup-script into a Google Cloud VM's metadata. This script is then executed by the Google Compute Engine guest agent. The phrase "startup-script" in the log is a consistent identifier for the metadata key that we control. So, I think it always will print `Script "startup-script" failed with error:`. Should we replace that greedy match with the exact words to make it consistent then?

bytetwin: I think we should.

bytetwin (approved):

PR #4872, opened 2025-11-23
removing deletion protection on filestore instances to force delete them in test project

bytetwin:
Filestore instances have deletion protection on that leave them as orphaned resources in the test project. Force deleting them

### DIFF
--- a/tools/clean-filestore-limit.sh
+++ b/tools/clean-filestore-limit.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2022 Google LLC
+# Copyright 2025 Google LLC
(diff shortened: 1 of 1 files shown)


tools/clean-filestore-limit.sh
```diff
@@ -18,6 +18,7 @@
 
 set -e -o pipefail
 
+# shellcheck disable=SC2329
```
bytetwin, line 21: Shellcheck can't determine functions usage when its [invoked via trap](https://www.shellcheck.net/wiki/SC2329)

LAVEEN (approved):

PR #4919, opened 2025-12-02
feat(validations): Add generalized early regex validation

saara-tyagi27:
Description:

This PR adds a generic, module-scoped validation framework so modules can declare input validation rules in their metadata.yaml. Metadata validators run against settings defined in a module’s settings block (not global blueprint vars) and validate the resolved value when the setting is present. A runtime-only Blueprint.YamlCtx is used so validators can detect whether a setting was explicitly provided and skip optional rules when it wasn’t. 

  Key Changes:

   * Generalized `RegexValidator`: Instead of hardcoding validation logic for specific variables, this new validator consumes a regex pattern and a list of variables from the module's metadata and applies
     the validation dynamically. This allows us to easily add new regex-based validations without changing the core Go code.
   * New Validations: As part of this change, I have added regex validations for the following variables:
       * slurm_cluster_name
       * partition_name
       * name (for gke-node-pool and resource-policy)
   * Added a runtime-only Blueprint.YamlCtx that records source-file positions for blueprint paths; validators use bp.YamlCtx.Pos(path) to detect whether a module.setting was explicitly provided in the user’s YAML so optional rules can be skipped when the key wasn't present (even if a value would be available via defaults or variable inheritance).
 
Testing:

All validations have been tested by introducing invalid values into the appropriate blueprint files and confirming that the gcluster deploy command fails with the expected error message.

### DIFF
--- a/community/modules/compute/notebook/metadata.yaml
+++ b/community/modules/compute/notebook/metadata.yaml
@@ -18,11 +18,3 @@ spec:
     services:
     - notebooks.googleapis.com
     - storage.googleapis.com
-ghpc:
-  validators:
-  - validator: regex
-    inputs:
-      vars: [deployment_name]
-      pattern: "^[a-z](?:[a-z0-9]{0,6})$"
-      scope: module
-    error_message: "'deployment_name' must contain only lowercase letters and numbers, start with a letter, and be 1-7 characters long."
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -291,6 +291,9 @@ type Blueprint struct {
 	path string
 	// records of intentions to stage file (populated by ghpc_stage function)
 	stagedFiles map[string]string
+	// YamlCtx holds parsed YAML positions so validators can tell if a module setting
+	// was explicitly present in the user's source (runtime-only, not serialized).
+	YamlCtx *YamlCtx `yaml:"-"`
 }
 
 func (bp *Blueprint) Clone() Blueprint {
@@ -467,6 +470,9 @@ func NewBlueprint(path string) (Blueprint, *YamlCtx, error) {
 		return Blueprint{}, &ctx, err
 	}
 	bp.path = absPath
+	// Attach parsed YAML context to the Blueprint so validators can determine
+	// whether a module.setting path was explicitly present in the user's YAML.
+	bp.YamlCtx = &ctx
 	return bp, &ctx, nil
 }
 
--- a/pkg/validators/metadata_validators_test.go
+++ b/pkg/validators/metadata_validators_test.go
@@ -45,7 +45,7 @@ func TestRegexValidator(t *testing.T) {
 	validator := RegexValidator{}
 
 	t.Run("fails_on_invalid_name", func(t *testing.T) {
-		// rule targeting module.setting "name"
+		// rule targeting module.setting "name" via vars:
 		rule := modulereader.ValidationRule{
 			Validator:    "regex",
 			ErrorMessage: "'name' must be lowercase and start with a letter.",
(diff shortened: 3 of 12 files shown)


cmd/deploy.go
```diff
@@ -160,3 +162,34 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
 	}
 	return shell.ExportOutputs(tf, artifactsDir, applyBehavior, outputFormat)
 }
+
```
cboneti: I am surprised we have to add this to deploy.go, whereas all other validators worked without this dependency, right?

pkg/validators/basic_validators.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2023 Google LLC
```
cboneti: This being a new file, I would recommend against naming it basic_validators.  The word basic doesn't mean much in this context. Are these all the metadata related validators?

cboneti: ah, also copyright 2023 -> 2025?

community/modules/compute/notebook/metadata.yaml
```diff
@@ -18,3 +18,10 @@ spec:
     services:
     - notebooks.googleapis.com
     - storage.googleapis.com
+ghpc:
+  validations:
```
cboneti: you should align this with the blueprint language. Make sure you have the same terms whenever possible

pkg/config/config.go
cboneti, line 230: Here, we define validators that are defined in the blueprint... I recommend aligning these to the validators in the modules.  Example: here we use "validator", etc.. perhaps you could use a very similar syntax? 
https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/526b97efb9699786065b486df41ce9a5f746c527/docs/blueprint-validation.md for more

pkg/validators/basic_validators.go
saara-tyagi27: Yes, these are all metadata related validators. What do u think about the name "metadata_validators.go"

cmd/deploy.go
saara-tyagi27: Last I tried incorporating the logic in create but it didn't seem to work so added it to deploy. But makes much more sense for validations to run as part of a create process.
Fixed this, the validations now run in the create stage.

community/modules/compute/notebook/metadata.yaml
saara-tyagi27: I have aligned it as much possible to the blueprint language. 
Had to leave in "error message" section as that part is required and separate for diff variables.

pkg/config/config.go
saara-tyagi27: Thankyou for the suggestion, this makes sense and I have now aligned the module Validator to this one as much as possible.

cboneti (commented):
I left one comment there. Thanks for this!!

pkg/validators/validators.go
```diff
@@ -115,9 +116,46 @@ func Execute(bp config.Blueprint) error {
...
+
+				if err := validator.Validate(bp, mod, rule, group, j); err != nil {
+					// The validator is responsible for creating a BpError with the correct path.
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
saara-tyagi27: I observed that during the current implementation of validations that happen later also give errors 1 at a time so i went ahead with that.

pkg/validators/validators.go
```diff
@@ -229,3 +267,8 @@ func validators(bp config.Blueprint) []config.Validator {
 	}
 	return vs
 }
+
+// Validator is the interface that all validation patterns must implement.
+type Validators interface {
+	Validate(bp config.Blueprint, mod config.Module, rule modulereader.ValidationRule, group config.Group, modIdx int) error
+}
```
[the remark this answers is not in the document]
saara-tyagi27: Done.

pkg/validators/metadata_validators_test.go
```diff
@@ -0,0 +1,64 @@
...
+	err := validator.Validate(bp, bp.Groups[0].Modules[0], rule, bp.Groups[0], 0)
+	if err == nil {
+		t.Errorf("Expected validation error, but got nil")
+	}
+
+	expectedError := "deployment_groups[0].modules[0].settings.name: 'name' must be lowercase and start with a letter."
+	if err.Error() != expectedError {
+		t.Errorf("Expected error message '%s', but got '%s'", expectedError, err.Error())
+	}
+}
```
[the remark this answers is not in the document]
saara-tyagi27: Done.

PR #4941, opened 2025-12-04
fix(slurm): power down node instead of mark as down

### LINKED ISSUE #4940: Nodes permanently marked DOWN on resume failure (e.g., stockouts) instead of returning to pool
### Describe the bug

When the Slurm GCP controller's `resume.py` script encounters a GCP API error during instance creation (specifically `bulkInsert`), such as `ZONE_RESOURCE_POOL_EXHAUSTED` (stockout), it calls `down_nodes_notify_jobs`. This function explicitly executes `scontrol update nodename=... state=down`.

This behavior permanently marks the node as `DOWN` in Slurm, requiring manual administrator intervention (`scontrol update node=... state=resume`) to make the node available again. This defeats the purpose of an autoscaling cluster during transient errors like stockouts, as Slurm cannot retry provisioning the node or allocate a different node for the pending job once the node is marked `DOWN`.

### Steps to reproduce

1. Configure a Slurm cluster using the `schedmd-slurm-gcp-v6-controller` module.
2. Submit a batch job that requests nodes in a region/zone currently experiencing high demand or stockouts (to trigger `ZONE_RESOURCE_POOL_EXHAUSTED`).
3. Observe the behavior of the `resume.py` script and the node state in `slurmctld` when the API call fails.

### Expected behavior

If provisioning fails due to a cloud provider error (especially a capacity issue), the node should be marked as `POWER_DOWN`. This would allow Slurm's power management logic to put the node back into the idle pool (`IDLE~`), enabling the scheduler to retry provisioning later or select a different node for the job automatically.

### Actual behavior

The node is marked as `DOWN` with the reason set to the GCP API error message. The node remains in this state indefinitely until an admin manually fixes it.

### Version (`gcluster --version`)

v1.70.0

### Blueprint

NA

### Expanded Blueprint

NA

### Output and logs

```text
[2025-12-04T15:11:25.071] update_node: node smallslurm-a21-0 reason set to: GCP Error: ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS: The zone 'projects/sq-cash-ml-slurm-prod/zones/us-central1-a' does not have enough resources available to fulfill the request.  '(resource type:compute)'.
[2025-12-04T15:11:25.071] Killing JobId=16571 on failed node smallslurm-a21-0
[2025-12-04T15:11:25.072] update_node: node smallslurm-a21-0 state set to DOWN
...
[2025-12-04T15:18:57.028] node smallslurm-a21-0 not resumed by ResumeTimeout(600), setting DOWN and POWERED_DOWN

```

### Execution environment

- OS: Linux (Slurm Controller)
- Shell: bash
- go version: N/A

### Additional context

The issue is located in resume.py in the down_nodes_notify_jobs function. 
Current code:
```python
run(f"{lookup().scontrol} update nodename={nodelist} state=down reason={reason_quoted}", check=False)
```
Proposed fix is to change state to POWER_DOWN.

casassg:
## Summary
Fixes an issue where nodes are permanently marked as `DOWN` when the `resume.py` script encounters a GCP API error (such as `ZONE_RESOURCE_POOL_EXHAUSTED` / stockouts) during bulk insert operations.
## Description
Currently, when `resume.py` handles a failed bulk insert operation, it calls `down_nodes_notify_jobs`, which executes:
`scontrol update nodename=... state=down ...`
This forces the node into a manual failure state, requiring administrator intervention to resume. This behavior is detrimental during transient cloud errors (like stockouts) because it prevents Slurm from retrying the allocation on other available nodes or retrying the same node later.
This PR changes the state update from `state=down` to `state=POWER_DOWN`. This allows Slurm to return the node to a power-save state (`IDLE~`), keeping it eligible for future scheduling attempts without manual intervention.
## Changes
- Modified `down_nodes_notify_jobs` in `resume.py` to use `state=POWER_DOWN` instead of `state=down`.

Fixes #4940


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -469,8 +469,8 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
         run(f"{lookup().scontrol} notify {job.job_id} {reason_quoted}", check=False)
 
     nodelist = util.to_hostlist(nodes)
-    log.error(f"Marking nodes {nodelist} as DOWN, reason: {reason}")
-    run(f"{lookup().scontrol} update nodename={nodelist} state=down reason={reason_quoted}", check=False)
+    log.error(f"Marking nodes {nodelist} as POWER_DOWN, reason: {reason}")
```
bytetwin: Thanks @casassg for the contribution. While there is a merit in this idea, there are few concerns in this implementation

1. This method is called not only called for bulkinsert failures. Making it power_down in other cases is not suitable.
2. bulkinsert failures can be for various reasons like quota, permissions, stockout etc . Applying a blanket power_down state may result in an infinite loop where slurm continues to power up this node for jobs but gcloud continues to deny the request

casassg: Mmm fair enough. I can try to reduce power down to stock out

casassg: lmk what do you think @bytetwin

casassg: any way we can get this checked or CI running?

sarthakag: Approved the CI workflows.

sarthakag: Hi @casassg
What do you mean by stock out state?

I don't think we can simply add the power_down state to fix transient errors - there has to be a limit to the retries.

PR #4973, opened 2025-12-10
Make waiting for kueue installation configurable, and wait for kueue in the G4 GKE blueprint

kadupoornima:
Made the `wait` parameter in kueue installation configurable. Updated the GKE G4 blueprint to explicitly set `wait: true` for Kueue installation, allowing the deployment process to pause until Kueue is fully ready.

GKE G4 daily test was failing when attempting to submit a Kubernetes job, specifically encountering an `Internal Server Error (HTTP 500)` caused by a broken admission webhook. The test had succeeded after 1 retry attempt.

This PR fixes the probable root cause.

Manual cluster creation and deletion were successfully tested with this change.

### DIFF
--- a/tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a4x.yml
+++ b/tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a4x.yml
@@ -116,16 +116,16 @@
   debug:
     msg: "{{nccl_test_logs.stdout}}"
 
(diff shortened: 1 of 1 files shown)


SwarnaBharathiMantena (approved):
I think, ideally we should keep the actual terraform module default same on our end too.

The default is True on the actual terraform module - https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#wait-1

Due to some testing results earlier, we set the default to **False**. In case we see value in it, we can reset it in future.

PR #5015, opened 2025-12-18
Fix the missing filestore_csi driver setup in the H4D blueprint

SwarnaBharathiMantena:
The filestore_csi driver setup in the H4D blueprint is missing. Fixed by setting `enable_filestore_csi` to `true`.


SwarnaBharathiMantena:
PR-test-gke-h4d integration test passed.

PR #5039, opened 2025-12-29
Add automated GCP resource cleanup script and Cloud Build pipeline

simrankaurb:
This PR adds a new cleanup script designed to identify and remove stale GCP resources while strictly protecting active or explicitly exempted infrastructure. The solution is split into two primary components:

1.  **`tools/cleanup.sh`**: A Bash script that orchestrates the discovery and deletion of various GCP resources (Compute Engine, GKE, TPU, VPC, etc.) based on age and protection rules.
2.  **`tools/cloud-build/project-cleanup.yaml`**: A Cloud Build configuration to automate the execution of the cleanup script with built-in retry logic and environment variable management.

#### **Key Features**
*   **Hierarchical Resource Protection**: The script doesn't just check for names; it proactively identifies "protected" resources. For example, if a GKE cluster is marked for preservation, the script automatically identifies and protects its underlying node pools, Managed Instance Groups (MIGs), specific VM instances, disks, and even the VPC networks they inhabit.
*   **Flexible Exclusions**: Supports two methods for whitelisting resources:
    *   **Global Exclusion File**: A text file (stored in GCS) containing names of resources to ignore.
    *   **Label-based Exemption**: Resources with the label `cleanup-exemption-date` followed by a valid `YYYY-MM-DD` date will be skipped until that date has passed.
*   **Multi-Phase Cleanup**: Resources are deleted in an order that respects dependencies (e.g., clusters and VMs are deleted before the networks and subnets they use).
*   **Safety Mechanisms**: 
    *   **Dry Run Mode**: Allows to see what would be deleted without taking action.
    *   **Time Cutoffs**: Defaults to deleting resources older than 2 hours, with a longer 60-day window for VM images to preserve base environment states.

---

### **What this Pull Request is doing**
This PR establishes a "garbage collection" system for the project's GCP environment. It solves the problem of resource leakage (and subsequent costs) by:
1.  **Scanning** the project for nearly all major resource types (GKE, Compute, TPU, Filestore, Images, VPCs, IAM bindings).
2.  **Filtering** out anything that is too new (less than 2 hours old), explicitly listed in an exclusion file, or labeled for protection.
3.  **Mapping Dependencies** to ensure that if a core resource (like a VM) is being kept, its required infrastructure (like its IP address or network) is also kept, even if that infrastructure wasn't explicitly protected.
4.  **Executing Deletion** via `gcloud` commands in a specific order to avoid dependency errors.
5.  **Automating** the entire process via Cloud Build, which can be triggered on a schedule or manually to keep the project clean.

### Submission Checklist

### DIFF
--- a/tools/cleanup.sh
+++ b/tools/cleanup.sh
@@ -35,12 +35,18 @@ log() {
 check_dependencies() {
 	log "INFO" "Checking for required command-line tools..."
 	local dependencies=("gcloud" "awk" "grep" "sort" "date" "sed" "basename")
+	local missing_deps=()
+
 	for cmd in "${dependencies[@]}"; do
 		if ! command -v "$cmd" &>/dev/null; then
-			log "ERROR" "Missing required dependency: $cmd. Please install it via cloud build and try again."
-			exit 1 # Dependencies are critical, we must exit immediately here.
+			missing_deps+=("$cmd")
 		fi
 	done
+
+	if [ ${#missing_deps[@]} -ne 0 ]; then
+		log "ERROR" "Missing required dependencies: ${missing_deps[*]}. Please install them and try again."
+		exit 1 # Dependencies are critical, we must exit after reporting all missing ones.
+	fi
 	log "INFO" "All dependencies are satisfied."
 }
 
@@ -53,11 +59,9 @@ load_exclusions() {
 		local line="$1"
 		local trimmed_line
 		trimmed_line=$(echo "$line" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-		if [[ -n "$trimmed_line" ]] && [[ "$trimmed_line" != \#* ]]; then
-			if [[ -z "${EXCLUSION_MAP[${trimmed_line}]:-}" ]]; then
-				EXCLUSION_MAP["${trimmed_line}"]=1
-				((line_count++))
-			fi
+		if [[ -n "$trimmed_line" ]] && [[ "$trimmed_line" != \#* ]] && [[ -z "${EXCLUSION_MAP[${trimmed_line}]:-}" ]]; then
+			EXCLUSION_MAP["${trimmed_line}"]=1
+			((line_count++))
 		fi
 	}
 
@@ -579,7 +583,7 @@ process_addresses() {
 			[[ -z "$name" ]] && continue
 			if ! is_excluded "$name" "${labels_str:-}"; then
 				if [[ "$status" == "IN_USE" ]]; then
-					log "DEBUG" "Skipping IN_USE Regional Address $name ($region)."
+					log "SKIP" "Skipping IN_USE Regional Address $name ($region)."
 					continue
 				fi
 				execute_delete "Regional Address" "$name" "($region)" \
@@ -600,7 +604,7 @@ process_addresses() {
 			[[ -z "$name" ]] && continue
 			if ! is_excluded "$name" "${labels_str:-}"; then
 				if [[ "$status" == "IN_USE" ]]; then
-					log "DEBUG" "Skipping IN_USE Global Address $name."
+					log "SKIP" "Skipping IN_USE Global Address $name."
 					continue
 				fi
 				execute_delete "Global Address" "$name" "(Global)" \
@@ -762,7 +766,7 @@ process_filestore() {
 
 		if [[ -n "${network_uri}" && -n "${PROTECTED_NETWORK_URIS[${full_network_uri}]:-}" ]]; then
 			if [[ -z "${EXCLUSION_MAP[${name}]:-}" ]]; then
-				log "INFO" "SKIP : Filestore $name ($location) on PROTECTED network $network_basename (URI: $full_network_uri)"
+				log "SKIP" "Filestore $name ($location) on PROTECTED network $network_basename (URI: $full_network_uri)"
 				EXCLUSION_MAP["${name}"]=1
 			fi
 			continue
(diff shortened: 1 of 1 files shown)


tools/cleanup.sh
```diff
@@ -0,0 +1,910 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
simrankaurb: We have this in cloud build already and shellscript picks it from there so, not needed.

tools/cleanup.sh
```diff
@@ -0,0 +1,936 @@
...
+						log "ERROR" "Failed to remove binding for $member in $role"
+						((ERROR_COUNT++)) || true
+					fi
+				fi
+				((count++)) || true
+			fi
+		done
+	done <<<"$policy_data"
+	log "INFO" "Finished processing IAM deleted members. $count bindings actioned."
+}
```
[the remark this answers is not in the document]
simrankaurb: The script exclusively removes bindings for members prefixed with deleted:serviceAccount:. These bindings are already pointing to non-existent service accounts and have no impact on current permissions or functionality. A race condition might, at worst, cause the script to fail to remove one of these obsolete bindings, which can be picked up on a subsequent run. It's unlikely to corrupt or remove active, valid bindings.

tools/cleanup.sh
```diff
@@ -0,0 +1,971 @@
...
+		log "ERROR" "Cannot access GCS exclusion file: $EXCLUSION_FILE. Please check the path and permissions."
+		exit 1
+	fi
+	while IFS= read -r line || [[ -n "$line" ]]; do
+		process_line "$line"
+	done < <(gcloud storage cat "$EXCLUSION_FILE")
+
+	if [[ ${#EXCLUSION_MAP[@]} -eq 0 ]]; then
+		log "ERROR" "No valid exclusion entries loaded from $EXCLUSION_FILE. Exiting to prevent accidental deletion."
+		exit 1
```
Neelabh94, line 83: We can use different exit codes throughout the script to differentiate between re-triable and non re-triable errors. This can be done in a Follow-up PR

tools/cleanup.sh
```diff
@@ -0,0 +1,971 @@
...
+	fi
+
+	if [[ -n "$labels_str" ]]; then
+		IFS=';' read -ra LABEL_PAIRS <<<"$labels_str"
+		for PAIR in "${LABEL_PAIRS[@]}"; do
+			local KEY VAL
+			KEY="${PAIR%%=*}"
+			VAL="${PAIR#*=}"
+			if [[ "$KEY" == "cleanup-exemption-date" ]]; then
+				if [[ "$VAL" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]]; then
```
Neelabh94: The if block can be rationalized to reduce the complexity. Something like this can be used.
// Combined check for YYYY-MM-DD format and logical date validity
          if [[ "$VAL" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]] && \
             exp_seconds=$(date -d "$VAL + 1 day" -u +%s 2>/dev/null); then
            
            current_seconds=$(date -u +%s)
            if [[ "$exp_seconds" -gt "$current_seconds" ]]; then
              log "SKIP" "$resource_name (Label: cleanup-exemption-date=$VAL, valid)"
              return 0 # Excluded
            else
              log "INFO" "$resource_name (Label: cleanup-exemption-date=$VAL expired)"
              return 1 # Not excluded
            fi
          else
            log "WARNING" "$resource_name (Label: cleanup-exemption-date invalid format or value: $VAL)"
            return 1 # Not excluded
          fi

tools/cleanup.sh
```diff
@@ -0,0 +1,971 @@
...
+		trimmed_line=$(echo "$line" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+		if [[ -n "$trimmed_line" ]] && [[ "$trimmed_line" != \#* ]]; then
+			if [[ -z "${EXCLUSION_MAP[${trimmed_line}]:-}" ]]; then
+				EXCLUSION_MAP["${trimmed_line}"]=1
+				((line_count++))
+			fi
+		fi
+	}
+
+	log "INFO" "Exclusion file is a GCS path. Streaming content..."
```
Neelabh94: This log is slightly confusing. should be reworded to read something along the lines "the Exclusion file must be a GCS path, verifying.."

tools/cleanup.sh
```diff
@@ -0,0 +1,971 @@
...
+	if ! gcloud storage ls "$EXCLUSION_FILE" >/dev/null 2>&1; then
+		log "ERROR" "Cannot access GCS exclusion file: $EXCLUSION_FILE. Please check the path and permissions."
+		exit 1
+	fi
+	while IFS= read -r line || [[ -n "$line" ]]; do
+		process_line "$line"
+	done < <(gcloud storage cat "$EXCLUSION_FILE")
+
+	if [[ ${#EXCLUSION_MAP[@]} -eq 0 ]]; then
+		log "ERROR" "No valid exclusion entries loaded from $EXCLUSION_FILE. Exiting to prevent accidental deletion."
```
Neelabh94: We are exiting if no exclusion entries are found. This assumption that there will 'always' be some exclusion file, needs clarification.

tools/cleanup.sh
```diff
@@ -0,0 +1,1001 @@
...
+
+		if ! is_excluded "$name" "${labels_str:-}"; then
+			log "INFO" "Processing Filestore instance for potential deletion: $name in $location"
+
+			if [[ "$DRY_RUN" == "true" ]]; then
+				log "DRY-RUN" "Would disable deletion protection on Filestore: $name ($location)"
+				log "DRY-RUN" "Would delete Filestore: $name ($location)"
+				((count++)) || true
+			else
+				log "EXECUTE" "Attempting to disable deletion protection on Filestore: $name ($location)"
```
Neelabh94, line 789: nit: This can be INFO.  Same at other places where it is used.

tools/cleanup.sh
simrankaurb: Done, thanks!

tools/cleanup.sh
simrankaurb: Reworded log and added comment for better clarity, thanks!

tools/cleanup.sh
simrankaurb: Thanks for this! Added comment with explanation.

tools/cleanup.sh
simrankaurb: Thanks for this! That's a good suggestion. I'll take this up as a follow up PR.

tools/cleanup.sh
simrankaurb: When reviewing logs, especially after an incident or during debugging, filtering for lines with EXECUTE allows for quick identification of all state-changing actions taken by the script. Mixing these with general INFO messages would make this analysis more difficult.

PR #5056, opened 2026-01-02
Support Tag Key and Tag Value creation on Cluster Toolkit [via Tag module]

SwarnaBharathiMantena:
### Description
The tag module creates a tag key and the provided tag values.
If the key already exists, then the tag values passed are associated with the existing tag key.

The module creates of two resources; [google_tags_tag_key](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_key) and [google_tags_tag_value](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_value).

### Manual Testing
1. _Scenario 1_: Create a new tag key with two values.
    - Tag key: `environment`
    - Tag values: `dev`, `staging`
--> In this case, a new tag key gets created with tag values provided.

2. _Scenario 2_: Existing tag key with inconsistent (old and new) tag values.
    - Tag key: `environment`
    - Tag values: `dev`, `prod`
--> In this case, a new key is not be created as the key already exists. The values are appended so a new tag value `prod` is created and the existing values are not overwritten. As `dev` already exists, it throws an error similar to:
```
Error: Error waiting to create TagValue: Error waiting for Creating TagValue: Error code 6, message: <eye3 title='ALREADY_EXISTS'/> generic::ALREADY_EXISTS: A TagValue with short name 'dev' already exists under parent 'tagKeys/xxxxxxxxxxx7691'
```

3. _Scenario 3_: Existing tag key with inconsistent (old and new) tag values.
    - Tag key: `environment_2`
    - Tag values: `dev`, `prod`
    - Specific purpose and purpose_data:
    ```
          tag_key_purpose_data:
            network: "hpc-toolkit-gsc/8506222224444488888"
            # "https://www.googleapis.com/compute/alpha/projects/hpc-toolkit-gsc/global/networks/8506647392441480133" #"hpc-toolkit-gsc/gke-h4d-rdma-net"
    ```
    --> In this case, a new tag key gets created with tag values provided.

### Follow-up
- An integration test validating the tag module will be added as part of the H4D blueprint update.

### DIFF
--- a/modules/management/tag/main.tf
+++ b/modules/management/tag/main.tf
@@ -48,6 +48,17 @@ resource "google_tags_tag_key" "key" {
   description  = var.tag_key_description
   purpose      = var.tag_key_purpose
   purpose_data = var.tag_key_purpose_data
+
+  lifecycle {
+    precondition {
+      condition = (
+        var.tag_key_purpose == "GCE_FIREWALL" ?
+        (var.tag_key_purpose_data != null ? contains(keys(var.tag_key_purpose_data), "network") : false) :
+        true
+      )
+      error_message = "When tag_key_purpose is set to 'GCE_FIREWALL', tag_key_purpose_data must be provided and must contain a 'network' key."
+    }
+  }
 }
 
 # Create values using the resolved ID
--- a/modules/management/tag/variables.tf
+++ b/modules/management/tag/variables.tf
@@ -67,15 +67,6 @@ variable "tag_key_purpose_data" {
   description = "Purpose data cannot be changed once set. Purpose data corresponds to the policy system that the tag is intended for. For example, the GCE_FIREWALL purpose expects data in the following map format: network = \"<project-id>/<network-id>\" (or) Network URI (or) selfLinkWithId."
   type        = map(string)
   default     = null
-
-  validation {
-    condition = (
-      var.tag_key_purpose == "GCE_FIREWALL" ?
-      (var.tag_key_purpose_data != null ? contains(keys(var.tag_key_purpose_data), "network") : false) :
-      true
-    )
-    error_message = "When tag_key_purpose is set to 'GCE_FIREWALL', tag_key_purpose_data must be provided and must contain a 'network' key."
-  }
 }
 
 variable "tag_values" {


modules/management/tag/main.tf
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+# Use the external data source
+data "external" "check_tag_key" {
```
bytetwin: Curious why the data source wont work here and a gcloud solution is used

https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/tags_tag_key

modules/management/tag/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+variable "tag_key_parent" {
+  description = "The resource name of the new TagKey's parent. Must be of the form organizations/{org_id} or projects/{project_id_or_number}."
+  type        = string
```
bytetwin, line 19: can you add a validation for this as it expects a specific regex

modules/management/tag/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+}
+
+variable "tag_key_description" {
+  description = "User-assigned description of the TagKey. Must not exceed 256 characters."
+  type        = string
+  default     = ""
+}
+
+variable "tag_key_purpose" {
```
bytetwin, line 54: validation as it can only accept 2 values

modules/management/tag/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+  description = "A purpose cannot be changed once set. A purpose denotes that this Tag is intended for use in policies of a specific policy engine, and will involve that policy engine in management operations involving this Tag. Possible values are: GCE_FIREWALL, DATA_GOVERNANCE."
+  type        = string
+  default     = null
+}
+
+variable "tag_key_purpose_data" {
+  description = "Purpose data cannot be changed once set. Purpose data corresponds to the policy system that the tag is intended for. For example, the GCE_FIREWALL purpose expects data in the following map format: network = \"<project-id>/<network-id>\" (or) Network URI (or) selfLinkWithId."
```
bytetwin, line 67: Likewise validation especially for GCE_FIREWALL purpose

modules/management/tag/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+  type        = map(string)
+  default     = null
+}
+
+variable "tag_value" {
```
bytetwin: should it be tag_values to better align with list object type ?

modules/management/tag/README.md
```diff
@@ -0,0 +1,95 @@
+## Description
+The tag module creates a tag key and the provided tag values.
+
+If the key already exists, then the tag values passed are associated with the existing tag key.
+
+The module creates of two resources; [google_tags_tag_key](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_key) and [google_tags_tag_value](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_value).
```
bytetwin: nit: `The module creates two resources`

modules/management/tag/main.tf
SwarnaBharathiMantena: If the key does not exist when using data source, the deployment fails. TF cannot re-route it to create a new tag key at that stage.
So, I had to update the code to use gcloud command so that now we can create a new tag key based on the gcloud command result.

modules/management/tag/README.md
```diff
@@ -3,7 +3,7 @@ The tag module creates a tag key and the provided tag values.
 
 If the key already exists, then the tag values passed are associated with the existing tag key.
 
-The module creates of two resources; [google_tags_tag_key](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_key) and [google_tags_tag_value](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_value).
+The module creates two resources; [google_tags_tag_key](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_key) and [google_tags_tag_value](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_value).
 
 ### Example
```
bytetwin, line 8: the example needs an update on `tag_values` rename

modules/management/tag/main.tf
SwarnaBharathiMantena: Found a way around to use TF code for the `create if not exists` logic.

PR #5062, opened 2026-01-05
Add precondition checks to disallow setting conflicting consumption options

kadupoornima:
Problem: The blueprint validator did not prevent a user from configuring a GKE node pool with a few conflicting consumption options. While these combinations are syntactically valid in the blueprint, it results in a `badRequest` error from the GKE API during terraform apply. This lead to wasted time deploying and debugging, as the failure occurs late in the terraform apply process. 

Solution : In this PR, we added precondition checks to prevent conflicting combinations during the blueprint check/generate phase itself for the following:

- Both `spot: true` and `enable_queued_provisioning: true` options cannot be set at the same time. 
- If Spot consumption option is enabled (`spot: true`), then reservation should not be set.

Tested these changes locally.


SwarnaBharathiMantena (approved):
LGTM

PR #5063, opened 2026-01-05
feat(validations): Add AllowedEnumValidator for module-scoped enum checks

saara-tyagi27:
##  Description
This PR extends the module-scoped validation framework by introducing an AllowedEnumValidator. This allows modules to enforce that input settings match a predefined list of permitted values directly from metadata.yaml. This effectively moves logic previously handled by Terraform validation blocks to an earlier stage in the HPC Toolkit deployment lifecycle.

### Key Changes
- **New AllowedEnumValidator:** Implements a generic validator that checks if a module setting is within a defined list of allowed values. 

### Functionality:
- Case Sensitivity: Defaults to true to mirror Terraform's contains() behavior, but can be toggled via case_sensitive in the metadata.
- Null Handling: Supports an allow_null flag to allow optional settings to bypass the check if not provided in the blueprint.

## Example Usage
The **_allowed_enum_** validator enables module authors to define a restricted list of permitted values for any variable. For example, a networking module can now enforce that **network_routing_mode is strictly GLOBAL or REGIONAL**.

### DIFF
--- a/pkg/validators/metadata_validators.go
+++ b/pkg/validators/metadata_validators.go
@@ -141,12 +141,14 @@ func (v *AllowedEnumValidator) Validate(
 	group config.Group,
 	modIdx int) error {
 
+	modPath := config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source
+
 	// 1. Parse Metadata Inputs (flags)
 	caseSensitive, err := parseBoolInput(rule.Inputs, "case_sensitive", true)
 	if err != nil {
 		return config.BpError{
 			Err:  fmt.Errorf("validation rule for module %q: %v", mod.ID, err),
-			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
+			Path: modPath,
 		}
 	}
 
@@ -154,7 +156,7 @@ func (v *AllowedEnumValidator) Validate(
 	if err != nil {
 		return config.BpError{
 			Err:  fmt.Errorf("validation rule for module %q: %v", mod.ID, err),
-			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
+			Path: modPath,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/validators/metadata_validators.go
```diff
@@ -75,3 +76,103 @@ func (r *RegexValidator) Validate(
...
+			Err:  fmt.Errorf("validation rule for module %q: %v", mod.ID, err),
+			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
+		}
+	}
+
+	caseSensitive, _ := rule.Inputs["case_sensitive"].(bool)
+	if _, ok := rule.Inputs["case_sensitive"]; !ok {
+		caseSensitive = true
+	}
+	allowNull, _ := rule.Inputs["allow_null"].(bool)
```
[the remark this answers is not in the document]
saara-tyagi27: Done.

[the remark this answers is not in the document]
saara-tyagi27: Added robust boolean parsing. Done.

pkg/validators/metadata_validators.go
```diff
@@ -75,3 +76,117 @@ func (r *RegexValidator) Validate(
...
+			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
+		}
+	}
+
+	// 2. Normalize the 'allowed' list
+	allowedRaw, ok := rule.Inputs["allowed"]
+	if !ok {
+		return config.BpError{
+			Err:  fmt.Errorf("validation rule for module %q is missing an 'allowed' list", mod.ID),
+			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
```
arpit974: can this be defined as a separate variable and used at various places.

pkg/validators/metadata_validators_test.go
```diff
@@ -140,3 +140,143 @@ func TestRegexValidator(t *testing.T) {
...
+		BlueprintName: "test-bp",
+		Groups: []config.Group{
+			{
+				Name: "primary",
+				Modules: []config.Module{
+					{
+						ID:     "test-module",
+						Source: "test/module",
+						Settings: config.NewDict(map[string]cty.Value{
+							"network_routing_mode": cty.StringVal("GLOBAL"),
```
arpit974, line 155: please add a test for regional as well.

pkg/validators/metadata_validators.go
saara-tyagi27: Makes sense. Done.

pkg/validators/metadata_validators_test.go
saara-tyagi27: Added test for regional as well.

PR #5078, opened 2026-01-08
Update slurm topology test assertion params, and fix precommit errors with pytest and mypy

SwarnaBharathiMantena:
Update slurm topology test assertion params, and fix precommit errors with pytest and mypy.

#### 1. `AssertionError: The two sets did not match.`
The previous `self.assert_equal(r_rack_set, s_rack_set, "The two sets did not match.")` code fails sometimes as expected because the lists might not be ordered the same way and return `true` each time.
This is fixed by updating the type during the assertion step to a set of frozenset (because set of set is not possible given set should have hashable entries).

#### 2. Issues running `pytest-check` and `mypy-check` 
The `pytest-check` and `mypy-check` tests failed due to missing dependent packages. Updating `language: system` to `language: system` ensures that a separate venv is created for the test in specific.

#### 3. Additional mypy-check failure
Mypy fails on **options because it sees a generic Dict[str, str]. The ClientOptions class has very specific internal requirements (some fields must be specific types like Credentials or Callables). By explicitly naming the arguments (e.g., api_endpoint=...), Mypy can verify that the value you are passing matches the expected type for that specific parameter.

### DIFF
--- a/tools/python-integration-tests/slurm_topology.py
+++ b/tools/python-integration-tests/slurm_topology.py
@@ -39,7 +39,7 @@ class SlurmTopologyTest(test.SlurmTest):
         r_rack_set = [set(v) for v in r_rack.values()]
         s_rack_set = [set(v) for v in s_rack.values()]
 
-        self.assert_equal(sorted(r_rack_set), sorted(s_rack_set), "The two sets did not match.")
+        self.assert_equal({frozenset(s) for s in r_rack_set}, {frozenset(s) for s in s_rack_set}, "The two sets did not match.")
 
     def get_slurm_topology(self):
         stdin, stdout, stderr = self.ssh_client.exec_command("scontrol show topo")


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -336,12 +336,15 @@ def create_client_options(api: ApiEndpoint) -> ClientOptions:
...
-    co = ClientOptions(**options)
+    
+    # Explicitly pass arguments to the constructor
+    api_endpoint = f"https://{api.value}.{ud}/{ver}/" if ver else None
+    universe_domain_val = ud if (ud and ud != DEFAULT_UNIVERSE_DOMAIN) else None
+
+    co = ClientOptions(
+        api_endpoint=api_endpoint,
+        universe_domain=universe_domain_val
+    )
```
cboneti: I might be wrong here, but this has the side effect that you might be setting universe_domain=None, versus  not specifying it at all, is there any way for us to validate if it would change anything?

cboneti (approved):
LGTM

PR #5085, opened 2026-01-08
upgrading slurm-gcp version to 6.10.6.

arpit974:
upgrading slurm-gcp version to 6.10.6.

### DIFF
--- a/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-gcsfuse-lssd-blueprint.yaml
+++ b/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-gcsfuse-lssd-blueprint.yaml
@@ -30,7 +30,7 @@ vars:
   a3mega_maintenance_interval: ""
   enable_placement: false
   local_mount_homefs: /home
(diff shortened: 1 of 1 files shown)


community/examples/hpc-build-slurm-image.yaml
```diff
@@ -23,7 +23,7 @@ vars:
   image_build_machine_type: n2d-standard-16
   build_from_image_family: hpc-rocky-linux-8
   build_from_image_project: cloud-hpc-image-public
-  build_from_git_ref: 6.9.1
+  build_from_git_ref: 6.10.6
```
LAVEEN, line 26: I might have missed this ,but may i know why are we upgrading to 6.10.6 and not 6.10.7 as a3high is at 6.10.7 ? or in other way why a3h is at 6.10.7 and  not at 6.10.6?

arpit974: all bp in repo is at 6.10.6 barring a3h - completely rewritten blueprint, just trying to keep repo consistent, can have another WI to move all to 6.10.7

PR #5114, opened 2026-01-16
Fix transient aperture device error in A3M slurm yaml

agrawalkhushi18:
This PR addresses transient "Aperture devices not found" errors on A3-Mega nodes.
The hardware that manages these devices boots at the same time as the compute node. If the node finishes booting first, it looks for the devices before they are ready, leading to a "not found" error.

A move to udev rules was made to automate mounting, but the system's "readiness check" was too rigid—it would fail instantly if the device was even a second late. 

These changes improves the reliability of Aperture device initialization:
- Implements Polling in ExecCondition: Replaces the static directory check with a 60-second polling loop (30×2s) to wait for device readiness.

### DIFF
--- a/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
@@ -452,11 +452,6 @@ deployment_groups:
             "data-root": "$(vars.localssd_mountpoint)/docker"
           }
       runners:
-      - type: shell
-        destination: enable_fabric_manager.sh
-        content: |
-          #!/bin/bash
-          systemctl enable nvidia-fabricmanager
       - type: ansible-local
         destination: slurm_aperture.yml
         content: |


examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -469,7 +474,7 @@ deployment_groups:
                 mode: 0o644
                 content: |
                   [Service]
-                  ExecCondition=/usr/bin/test -d /dev/aperture_devices/
+                  ExecCondition="/usr/bin/bash -c 'for i in \$(seq 30); do if [ -n \"\$(ls -A /dev/aperture_devices 2>/dev/null)\" ]; then exit 0; fi; sleep 2; done; exit 1'"
```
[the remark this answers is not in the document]
bytetwin: This is a good comment. Please address this fix

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -452,6 +452,11 @@ deployment_groups:
             "data-root": "$(vars.localssd_mountpoint)/docker"
           }
       runners:
+      - type: shell
+        destination: enable_fabric_manager.sh
+        content: |
+          #!/bin/bash
+          systemctl enable nvidia-fabricmanager
```
bytetwin: Why is this needed ? I don't think fabricmanager starts even with this.

agrawalkhushi18: While the polling loop effectively handles the boot-time race condition mentioned in the description, this explicitly ensures that the Fabric Manager is deterministically managed by the OS rather than relying on implicit image-level triggers. We have previously encountered customer issues where A3-Mega nodes were non-functional because the `nvidia-fabricmanager.service` was in a failed state or not started correctly after a restart. So, I thought this can increase the reliability.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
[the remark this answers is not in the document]
agrawalkhushi18: Done.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
agrawalkhushi18: The script does not explicitly enable the nvidia-fabricmanager service when checked by running the status command. Therefore removing the script. Apologies for the confusion.

PR #5115, opened 2026-01-16
feat(validations): Add early range based pattern validation

AdarshK15:
### **Summary**

This PR introduces the range validator to the module-scoped validation framework and applies it to several core and community modules. By moving numeric and length validation into metadata.yaml files, these constraints are caught earlier in the deployment lifecycle.

### **Key Changes**

**New RangeValidator:** Implements a generic validator to enforce numeric boundaries or length constraints on module settings.

**Functionality:**
&nbsp;&nbsp;&nbsp;&nbsp;- **Numeric Bounds:** Supports min and max checks for integer values.
&nbsp;&nbsp;&nbsp;&nbsp;- **Length Handling:** Includes a length_check flag to validate the size of a list; instead of the raw value.

**Changes:**
Added range validation for the following modules:

&nbsp;&nbsp;&nbsp;&nbsp;- **_compute/gke-nodeset_**: Validates count of entries in network_storage and filestore_id.
&nbsp;&nbsp;&nbsp;&nbsp;- **_compute/mig_**: Ensures at least one version is provided.
&nbsp;&nbsp;&nbsp;&nbsp;- **_scheduler/htcondor-access-point_**: Limits the number of zones.
&nbsp;&nbsp;&nbsp;&nbsp;- **_scheduler/schedmd-slurm-gcp-v6-controller_**: Limits the number of static IPs to a maximum of 1.
&nbsp;&nbsp;&nbsp;&nbsp;- **_compute/gke-node-pool_**: Enforces valid upgrade settings count ranges.
&nbsp;&nbsp;&nbsp;&nbsp;- **_internal/slurm-gcp/nodeset_tpu_**: Enforces valid node count ranges.
&nbsp;&nbsp;&nbsp;&nbsp;- **_network/multivpc_**: Enforces a network count between 2 and 8.
&nbsp;&nbsp;&nbsp;&nbsp;- **_file-system/filestore_**: Enforces minimum capacity requirements.

Added regex validation for the following modules:
&nbsp;&nbsp;&nbsp;&nbsp;- **_database/slurm-cloudsql-federation_**: Validates network_id to ensure the correct path format.
&nbsp;&nbsp;&nbsp;&nbsp;- **_modules/file-system/filestore_**: Validates network_id to ensure the correct path format.
&nbsp;&nbsp;&nbsp;&nbsp;- **_modules/file-system/parallelstore_**: Validates network_id to ensure the correct path format.
&nbsp;&nbsp;&nbsp;&nbsp;- **_modules/scheduler/gke-cluster_**: Validates network_id to ensure the correct path format.
&nbsp;&nbsp;&nbsp;&nbsp;- **_database/slurm-cloudsql-federation_**: Validates network_id to ensure the correct path format.
&nbsp;&nbsp;&nbsp;&nbsp;- **_compute/notebook_**: Validates mount_runner.args to ensure 5 elements.

### **Testing**
Added unit tests covering numeric ranges, list of ints, and also list length validation scenarios.

### **Documentation** 
Updated docs/blueprint-validation.md to include usage instructions and examples for the new validator.


pkg/validators/metadata_validators.go
```diff
@@ -192,3 +192,137 @@ func (v *AllowedEnumValidator) Validate(
...
+				}
+			}
+			if len(values) == 1 {
+				if values[0].Type() != cty.String {
+					return config.BpError{
+						Err:  fmt.Errorf("range validator with 'delimiter' expects a string value, but got %s", values[0].Type().FriendlyName()),
+						Path: path,
+					}
+				}
+				stringVal := values[0].AsString()
```
kvenkatachala333: When length_check is true and a delimiter is provided, the code calls values[0].AsString() without checking if the value is null or unknown.

Add a guard similar to the numeric check (lines 252-254), If values[0] is null or unknown, the validator should either skip it or return a clear error instead of risking a panic.

pkg/validators/metadata_validators.go
```diff
@@ -192,3 +192,137 @@ func (v *AllowedEnumValidator) Validate(
...
+				segments := strings.Split(stringVal, *delimiter)
```
kvenkatachala333: If a user provides an empty string ("") for a variable like mount_runner.args, strings.Split returns a slice of length 1 (containing one empty string).

Verify if the intention is for an empty string to count as 1 segment or 0 segments. If 0 is preferred, the logic should handle the empty string case explicitly.

pkg/validators/metadata_validators.go
```diff
@@ -192,3 +192,137 @@ func (v *AllowedEnumValidator) Validate(
...
+			continue
+		}
+		if val.Type() == cty.Number {
+			f, _ := val.AsBigFloat().Float64()
+			if f != float64(int64(f)) {
+				return config.BpError{
+					Err:  fmt.Errorf("range validator only supports integer numbers, not %v", f),
+					Path: path,
+				}
+			}
```
kvenkatachala333, line 241: The implementation explicitly restricts validation to integers only. If there are use cases for range validation on floating-point numbers (e.g., fractional GBs or cost thresholds), this would need to be expanded. Currently, it is strictly for counts and integer values.

community/modules/internal/slurm-gcp/nodeset_tpu/metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+---
+
+ghpc:
+  validators:
+  - validator: range
+    inputs:
+      vars: [node_count_static]
+      min: 0
+      length_check: false
+    error_message: "The 'node_count_static' value must be at least 0."
```
LAVEEN, line 22: should this error message be greater than or equal to 0 or the given message looks appropriate . Please check on this which one should be best over here.

modules/file-system/filestore/metadata.yaml
```diff
@@ -20,6 +20,12 @@ spec:
 
 ghpc:
   validators:
+  - validator: range
+    inputs:
+      vars: [size_gb]
+      min: 1024
+      length_check: false
+    error_message: "The 'size_gb' value must be at least 1024 GiB. See https://cloud.google.com/filestore/docs/service-tiers for details."
```
LAVEEN, line 27: do we have max validation for same ?

pkg/validators/metadata_validators.go
```diff
@@ -192,3 +192,137 @@ func (v *AllowedEnumValidator) Validate(
...
+		return r.checkBounds(len(values), min, max, customErrMsg, path)
+	}
+
+	for _, val := range values {
+		if val.IsNull() || !val.IsKnown() {
+			continue
+		}
+		if val.Type() == cty.Number {
+			f, _ := val.AsBigFloat().Float64()
+			if f != float64(int64(f)) {
```
LAVEEN, line 236: should we simply this using math.Trunc() ?

pkg/validators/metadata_validators.go
AdarshK15: removed support for delimiter, now this issue doesn't occur.

pkg/validators/metadata_validators.go
AdarshK15: done.

community/modules/internal/slurm-gcp/nodeset_tpu/metadata.yaml
AdarshK15: I added 'must be at least 0' to be consistent with existing error messages.

pkg/validators/metadata_validators.go
AdarshK15: Yes, Currently there are no validations for floating point numbers, adding this support now would add overhead.
We can certainly add floating-point support if specific floating point use cases are identified in the future.

modules/file-system/filestore/metadata.yaml
AdarshK15: I verified the code and I couldn't find a max value defined for size_gb, only the 1024 GiB minimum.

pkg/validators/metadata_validators.go
AdarshK15: The float64(int64(f)) check ensures the input is a whole number. If user gives a floating point number then, math.Trunc() could lead to data loss or overflows when converting to int for the bound checks.

PR #5129, opened 2026-01-21
Migrate configure_kueue from gavinbunney to helm

agrawalkhushi18:
This PR migrates the `configure_kueue`  module from the community `gavinbunney/kubectl` provider to the official `hashicorp/helm provider`. This is part of the work to remove the third‑party kubectl-apply dependency and standardize on the official Helm provider for chart-based installs and upgrades.
The module now deploys the Kueue chart via helm_release (helm provider) instead of applying raw manifests through the kubectl provider.

The changes have been tested in local and the builds were successful.

### Submission Checklist

### DIFF
--- a/modules/management/kubectl-apply/main.tf
+++ b/modules/management/kubectl-apply/main.tf
@@ -19,6 +19,15 @@ locals {
   cluster_name     = local.cluster_id_parts[5]
   cluster_location = local.cluster_id_parts[3]
   project_id       = var.project_id != null ? var.project_id : local.cluster_id_parts[1]
+  kueue_config_content = (
+    var.kueue.config_path != null && var.kueue.config_path != "" ?
+    (
+      endswith(var.kueue.config_path, ".tftpl") || length(try(var.kueue.config_template_vars, {})) > 0 ?
+      templatefile(var.kueue.config_path, try(var.kueue.config_template_vars, {})) :
+      file(var.kueue.config_path)
+    ) : ""
+  )
+  configure_kueue = local.install_kueue && try(var.kueue.config_path, "") != ""
 
   # 1. First, Identify manifests that are explicitly enabled.
   enabled_manifests = {
@@ -46,7 +55,9 @@ locals {
       server_side_apply = manifest.server_side_apply
       wait_for_rollout  = manifest.wait_for_rollout
     }
-  })
+    }
+
+  )
 
   install_kueue             = try(var.kueue.install, false)
   install_jobset            = try(var.jobset.install, false)
@@ -106,7 +117,7 @@ module "install_kueue" {
 
 module "configure_kueue" {
   source           = "./helm_install"
-  count            = local.install_kueue ? ((var.kueue.config_path == null ? "" : var.kueue.config_path) != "" ? 1 : 0) : 0
+  count            = local.configure_kueue ? 1 : 0
   release_name     = "kueue-config"
   chart_name       = "${path.module}/raw-config-chart"
   chart_version    = "0.1.0"
@@ -116,16 +127,7 @@ module "configure_kueue" {
 
   values_yaml = [
     yamlencode({
-      manifests = [
-        for doc in split("\n---", (
-          (var.kueue.config_path == null ? "" : var.kueue.config_path) != ""
-          ? (length(var.kueue.config_template_vars == null ? {} : var.kueue.config_template_vars) > 0 || length(regexall("\\.tftpl$", var.kueue.config_path == null ? "" : var.kueue.config_path)) > 0
-            ? templatefile(var.kueue.config_path, var.kueue.config_template_vars == null ? {} : var.kueue.config_template_vars)
-          : file(var.kueue.config_path))
-          : ""
-        )) : trimspace(doc)
-        if length(trimspace(doc)) > 0
-      ]
+      manifests = [for doc in split("\n---", local.kueue_config_content) : trimspace(doc) if length(trimspace(doc)) > 0]
     })
   ]
 


modules/management/kubectl-apply/main.tf
```diff
@@ -102,18 +102,38 @@ module "install_kueue" {
   depends_on = [var.gke_cluster_exists]
 }
 
+resource "time_sleep" "wait_for_webhook" {
+  create_duration = "120s"
+  depends_on      = [module.install_kueue]
+}
+
```
shubpal07, line 117: Do we need `time_sleep` even after using wait = true? Ideally it shouldn't be.

modules/management/kubectl-apply/main.tf
```diff
@@ -102,18 +102,38 @@ module "install_kueue" {
...
 module "configure_kueue" {
-  source        = "./kubectl"
-  source_path   = local.install_kueue ? try(var.kueue.config_path, "") : null
-  template_vars = local.install_kueue ? try(var.kueue.config_template_vars, null) : null
-  depends_on    = [module.install_kueue]
+  source           = "./helm_install"
+  count            = local.install_kueue ? ((var.kueue.config_path == null ? "" : var.kueue.config_path) != "" ? 1 : 0) : 0
```
shubpal07: nit: may help in code readability
```
locals {
  # ... existing locals ...
  configure_kueue = local.install_kueue && try(var.kueue.config_path, "") != ""
}

module "configure_kueue" {
  # ...
  count = local.configure_kueue ? 1 : 0
  # ...
}
```

modules/management/kubectl-apply/main.tf
```diff
@@ -102,18 +102,38 @@ module "install_kueue" {
...
+          (var.kueue.config_path == null ? "" : var.kueue.config_path) != ""
+          ? (length(var.kueue.config_template_vars == null ? {} : var.kueue.config_template_vars) > 0 || length(regexall("\\.tftpl$", var.kueue.config_path == null ? "" : var.kueue.config_path)) > 0
+            ? templatefile(var.kueue.config_path, var.kueue.config_template_vars == null ? {} : var.kueue.config_template_vars)
+          : file(var.kueue.config_path))
+          : ""
+        )) : trimspace(doc)
+        if length(trimspace(doc)) > 0
+      ]
+    })
+  ]
```
shubpal07: We may think here to make the logic less denser. We may move the content reading logic into a `local` value.
```
locals {
    kueue_config_content = (
        var.kueue.config_path != null && var.kueue.config_path != "" ? 
        (
            endswith(var.kueue.config_path, ".tftpl") || length(try(var.kueue.config_template_vars, {})) > 0 ?
            templatefile(var.kueue.config_path, try(var.kueue.config_template_vars, {})) :
            file(var.kueue.config_path)
        ) : ""
    )
}

```
resource/module block
```
manifests = [for doc in split("\n---", local.kueue_config_content) : trimspace(doc) if length(trimspace(doc)) > 0]

```

modules/management/kubectl-apply/main.tf
```diff
@@ -102,18 +102,38 @@ module "install_kueue" {
...
+  create_duration = "120s"
+  depends_on      = [module.install_kueue]
+}
+
 module "configure_kueue" {
-  source        = "./kubectl"
-  source_path   = local.install_kueue ? try(var.kueue.config_path, "") : null
-  template_vars = local.install_kueue ? try(var.kueue.config_template_vars, null) : null
-  depends_on    = [module.install_kueue]
+  source           = "./helm_install"
```
shubpal07, line 119: Using custom helm charts can also help us in migrating `kubectl_apply_manifests` to helm

tools/cloud-build/daily-tests/ansible_playbooks/tasks/rescue_gcluster_failure.yml
```diff
@@ -37,7 +37,7 @@
     register: gcluster_destroy
     changed_when: gcluster_destroy.changed
     run_once: true
-    ansible.builtin.command: ./gcluster destroy {{ deployment_name }} --auto-approve
+    ansible.builtin.command: ./gcluster destroy {{ deployment_name }} --robust --auto-approve
```
shubpal07: FYI: This will make robust destroy default across all integration tests.

modules/management/kubectl-apply/main.tf
shubpal07: Can we test removing time_sleep and just using wait - true ?

agrawalkhushi18: Yes, thanks for suggesting that. Setting wait=true along with added dependency of system_node_pool_id solves the problem. We do not need this time sleep. The changes are addressed in another [PR #5147](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/5147).

tools/cloud-build/daily-tests/ansible_playbooks/tasks/rescue_gcluster_failure.yml
agrawalkhushi18: This flag has been removed now as it was added to solve the destruction deadlock. The exact workaround for the issue has been resolved in PR #5147 .

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Done, thanks!

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Thanks for the suggestion. Done.

modules/management/kubectl-apply/main.tf
```diff
@@ -19,6 +19,15 @@ locals {
   cluster_name     = local.cluster_id_parts[5]
   cluster_location = local.cluster_id_parts[3]
   project_id       = var.project_id != null ? var.project_id : local.cluster_id_parts[1]
+  kueue_config_content = (
+    var.kueue.config_path != null && var.kueue.config_path != "" ?
+    (
+      endswith(var.kueue.config_path, ".tftpl") || length(try(var.kueue.config_template_vars, {})) > 0 ?
```
SwarnaBharathiMantena, line 25: I think `var.kueue.config_path` need not end with ".tftpl". It can be a yaml with / without any config_template_vars as well.

SwarnaBharathiMantena: I see that the other file formats like `yaml` are covered in the logic. So, please ignore the above comment.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Yes, agreed. This can be taken  up as a follow up task after configure_kueue is migrated to helm. I have created a separate ticket to handle the migration of kubectl_apply_manifests.

SwarnaBharathiMantena (approved):

modules/management/kubectl-apply/main.tf
shubpal07: Thanks

modules/management/kubectl-apply/main.tf
```diff
@@ -19,6 +19,15 @@ locals {
...
+      templatefile(var.kueue.config_path, try(var.kueue.config_template_vars, {})) :
+      file(var.kueue.config_path)
+    ) : ""
+  )
+  configure_kueue = local.install_kueue && try(var.kueue.config_path, "") != ""
```
shubpal07, line 30: NIT: 
Nested ternaries may be difficult to understand the logic. To  improve **readability** and **maintainability**, we may consider to break  the logic down into distinct boolean states (`has_config`, `is_template`) before calculating the final content.

```tf
locals {
  # 1. Extract inputs
  kueue_path = try(var.kueue.config_path, "")
  kueue_vars = try(var.kueue.config_template_vars, {})

  # 2. Define the conditions
  has_kueue_config = local.kueue_path != null && local.kueue_path != ""

  # It is a template if it has the extension OR if variables are provided
  is_kueue_template = local.has_kueue_config && (
    endswith(local.kueue_path, ".tftpl") || length(local.kueue_vars) > 0
  )

  # 3. Calculate the content 
  kueue_config_content = local.has_kueue_config ? (
    local.is_kueue_template ? 
      templatefile(local.kueue_path, local.kueue_vars) : 
      file(local.kueue_path)
  ) : ""
}

```

modules/management/kubectl-apply/main.tf
```diff
@@ -105,17 +116,23 @@ module "install_kueue" {
...
+  namespace        = "kueue-system"
+  create_namespace = true
+  wait             = false # Configuration resources (Queues) usually don't need wait
 
-  server_side_apply = true
-  wait_for_rollout  = true
+  values_yaml = [
+    yamlencode({
+      manifests = [for doc in split("\n---", local.kueue_config_content) : trimspace(doc) if length(trimspace(doc)) > 0]
+    })
```
shubpal07, line 131: The logic `split("\n---", ...)` works for standard Linux-formatted files, but it introduces a fragility we should guard against:
1. **Windows Line Endings**: If the content comes from a file saved with CRLF (\r\n), the split on \n--- will fail to find the delimiter, treating the whole file as one invalid manifest.

Suggestion: 
```sh
locals {
  # Normalize CRLF to LF to ensure the split works across platforms
  normalized_content = replace(local.kueue_config_content, "\r\n", "\n")
}

manifests = [
  for doc in split("\n---", local.normalized_content) : 
  trimspace(doc) 
  if length(trimspace(doc)) > 0
]

```

2. **Embedded Separators**: The logic `split("\n---", ...)` relies on a simple string search, which is "context-unaware." It cannot distinguish between a YAML document separator and the string `\n---` appearing inside a multi-line text block ex:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: script-config
  namespace: kueue-system
data:
  # A shell script that uses '---' as a visual separator
  setup.sh: |
    #!/bin/bash
    echo "Starting configuration..."
    
    # Terraform will split the file here!
    ---
    
    echo "Configuration complete."

```
Terraform splits this into two parts.

Part 1: A valid-looking ConfigMap that abruptly ends at ---.
Part 2: A text blob starting with echo "Configuration complete."
Result: The deployment will fails because Part 2 is not a valid Kubernetes manifest (it has no apiVersion or kind)

**Suggestion**: Since terraform's built-in parser `yamldecode` function might [fail on multi-document files](https://github.com/hashicorp/terraform/issues/29729) (files with multiple --- sections),  a more reliable solution is to use a **regular expression** that is **strict** about where the separator appears. Ex:

```tf
locals {
  # 1. Normalize line endings to \n (solves the Windows issue)
  # 2. Ensure the file ends with a newline to make regex matching consistent
  normalized_content = "${replace(local.kueue_config_content, "\r\n", "\n")}\n"

  # Regex Explanation:
  # \n---     Match a newline followed by three dashes
  # [ \t]*    Allow optional trailing spaces or tabs (valid YAML)
  # \n        Ensure it is followed immediately by a newline (the start of the next doc)
  #
  # This avoids matching "---" inside a line or "---" followed by text on the same line.
  manifest_separator_regex = "/\\n---[ \\t]*\\n/"
}

manifests = [
  for doc in split(local.manifest_separator_regex, local.normalized_content) : 
  trimspace(doc) 
  if length(trimspace(doc)) > 0
]

```

modules/management/kubectl-apply/main.tf
```diff
@@ -105,17 +116,23 @@ module "install_kueue" {
...
+  source           = "./helm_install"
+  count            = local.configure_kueue ? 1 : 0
+  release_name     = "kueue-config"
+  chart_name       = "${path.module}/raw-config-chart"
+  chart_version    = "0.1.0"
+  namespace        = "kueue-system"
+  create_namespace = true
+  wait             = false # Configuration resources (Queues) usually don't need wait
 
-  server_side_apply = true
```
shubpal07, line 113: In the previous `kubectl` implementation, SSA(`server_side_apply`) was often necessary to bypass the 256KB annotation size limit for large CRDs like Kueue's.

We may want to confirm that Helm's release storage fully mitigates the size limit issue for these manifests? I assume yes, but it would be good to verify we won't hit "Resource too large" errors when applying the larger Kueue CRDs/custom CRDs (in future) without SSA.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Thanks for the suggestion. I will take it up as a follow up item to this PR.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Thanks for the suggestion. I will take a look into this in a follow up PR.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: That's a great observation. Unlike the `kubectl` provider, which uses Client-Side Apply and stores the entire manifest in a `kubectl.kubernetes.io/last-applied-configuration `annotation on the resource itself (hitting the 256KB limit), Helm stores the entire release state in a Kubernetes Secret. Standard Kubernetes Secrets have a size limit of 1MB. By moving the manifest data from individual resource annotations to a centralized Helm Secret, we get a much larger buffer (1MB vs 256KB) for complex resources like Kueue's CRDs.

PR #5136, opened 2026-01-27
Update validate_configs.sh logic to accommodate deployment yamls

SwarnaBharathiMantena:
### Problem
All the blueprints which have deployment yaml files are excluded from validation at this time. This is because the `validate_configs.sh` script only supports a complete blueprint.

### Fixes / Changes
* `validate_configs.sh` - significant refactoring of the validation logic to support verifying deployment YAMLs and to use the new mocking script. With the current code, lots of blueprints are excluded from validation due to this reason. Rename the CONFIGS variable to TARGET_BLUEPRINTS to clarify what the variable defines.
* `get_mock_vars.py` - introduced dynamic variable mocking to automatically generate required input variables for blueprints, preventing validation failures due to missing inputs.
* `hpc-slinky.yaml` - Minor update to outputs entry identified using this script. (which was not identified earlier because the path was excluded.)


tools/validate_configs/get_mock_vars.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    def load_yaml(path: str) -> Dict[str, Any]:
+        if not path:
+            return {}
+        try:
+            with open(path, 'r') as f:
+                data = yaml.safe_load(f) or {}
+                return data.get('vars', {})
+        except Exception as e:
+            print(f"Error reading {path}: {e}", file=sys.stderr)
```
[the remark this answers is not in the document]
bytetwin: Please address this

tools/validate_configs/get_mock_vars.py
```diff
@@ -0,0 +1,95 @@
...
+            final_mock_vars[var_name] = predefined_vars[var_name]
+            continue
+
+        # Priority 2: Heuristic defaults for missing/empty values
+        # (Your original logic for determining if a mock value is needed)
+        needs_mock = (
+            var_value is None or 
+            var_value == "" or 
+            (isinstance(var_value, str) and var_value.isupper() and 
+             any(x in var_value for x in ["SIZE", "COUNT", "NAME", "ID", "REGION", "ZONE", "PROJECT", "BUCKET"]))
```
kadupoornima, line 69: case sensitive ok?

tools/validate_configs/validate_configs.sh
```diff
@@ -41,9 +48,23 @@ run_test() {
...
-	${GHPC_PATH} create "${BP_PATH}" -l ERROR \
-		--skip-validators="${VALIDATORS_TO_SKIP}" "${deployment_args[@]}" \
-		--vars="project_id=${PROJECT},deployment_name=${DEPLOYMENT}" >/dev/null ||
+
+	# Predefine common variables
+	PREDEFINED_VARS="project_id=${PROJECT},deployment_name=${DEPLOYMENT},region=us-central1,zone=us-central1-a,zones=['us-central1-a'],authorized_cidr=0.0.0.0/0"
+	PREDEFINED_VARS+=",health_check_schedule='0 0 * * 0',number_of_vms=2,static_node_count=2,slurm_cluster_name=testname,a3_partition_name=testname"
+
+	# Generate all other mock variables dynamically
+	# We use python to parse the blueprint and fill in the rest
```
kadupoornima, line 63: If this is a new dependency, users running validation locally might encounter an error if they haven't installed it? If so we should add in the automatic downloads list.

tools/validate_configs/validate_configs.sh
```diff
@@ -96,56 +115,89 @@ run_test() {
...
+	dir=$(dirname "$example")
+	base=$(basename "$example" .yaml)
+
+	# Handle cases like a3mega-slurm-gcsfuse-lssd-blueprint.yaml sharing a3mega-slurm-deployment.yaml
+	# or blueprints named without '-blueprint' suffix.
+	prefix=${base%-blueprint}
+	deployment="$dir/${prefix}-deployment.yaml"
+
+	if [ ! -f "$deployment" ]; then
+		# Try to find any deployment file in the same directory if the specific one doesn't exist
```
kadupoornima, line 177: If we have a blueprint `cluster-v2.yaml` and two deployment files `cluster-deployment.yaml` and `cluster-v2-deployment.yaml`, we end up selecting `cluster-deployment.yaml` instead of `cluster-v2-deployment.yaml`?
nit: can sort possible_deployments by descending length or checking for an exact match before falling back to prefix matching.

kadupoornima (changes requested):

tools/validate_configs/validate_configs.sh
SwarnaBharathiMantena: Added it to the makefile

tools/validate_configs/get_mock_vars.py
SwarnaBharathiMantena: Skipped making the uppercase check case-insensitive. This check is a specific heuristic to identify placeholder strings (like PROJECT_ID or BUCKET_NAME). Making it case-insensitive would cause the script to overwrite valid, user-provided values with generic mocks, breaking the validation of correctly configured blueprints.

SwarnaBharathiMantena:
Closing this ticket as this effort was deprioritized.

PR #5147, opened 2026-01-28
Update wait flag and resolving helm_release deadlock destruction error

agrawalkhushi18:
This PR addresses transient kueue-webhook build failures caused by asynchronous Helm installs (wait=false) by enabling `wait=true`. To prevent the resulting destruction-time race condition where node pools are deleted before Helm resources, it introduces an explicit dependency on the system_node_pool to ensure a clean, sequential teardown.

Changes: 

- gke-cluster/outputs.tf: Added system_node_pool_id output to expose the node pool's unique identifier for dependency tracking.

- kubectl-apply/variables.tf: Introduced system_node_pool_id variable to receive the dependency anchor from the cluster module.

- kubectl-apply/main.tf: Forced wait = true for Kueue installation and passed the system_node_pool_id into the install_kueue module.

- helm_install/main.tf: Implemented a tf_dependency_anchor using a Helm set block to force Terraform to honor sequential resource ordering during destruction.

- Updated GKE blueprints to remove explicit setting of the wait flag

To clean up the orphaned resources created through the ansible post-deploy test, a cleanup step to remove these resources has been added for helm and terraform to handle the gcluster destroy gracefully in the build environment.


SwarnaBharathiMantena:
Added `release-breaking-changes` label as the default value of `wait` is being updated.

modules/scheduler/gke-cluster/outputs.tf
```diff
@@ -102,3 +102,8 @@ output "gke_version" {
   description = "GKE cluster's version."
   value       = google_container_cluster.gke_cluster.master_version
 }
+
+output "system_node_pool_id" {
+  description = "The ID of the system node pool."
+  value       = one(google_container_node_pool.system_node_pools[*].id)
+}
```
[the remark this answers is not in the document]
agrawalkhushi18: done

modules/management/kubectl-apply/main.tf
```diff
@@ -87,7 +87,7 @@ module "kubectl_apply_manifests" {
 module "install_kueue" {
   source           = "./helm_install"
   count            = local.install_kueue ? 1 : 0
-  wait             = try(var.kueue.wait, false)
+  wait             = true
```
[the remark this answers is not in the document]
agrawalkhushi18: done

modules/management/kubectl-apply/variables.tf
```diff
@@ -191,3 +190,9 @@ variable "gib" {
     }
   }
 }
+
+variable "system_node_pool_id" {
```
bytetwin, line 194: If this is used only to derive the dependency id, why not retain the same name instead of this and then translating in mian.tf. But that would mean, you have modify every blueprint to pass the translation explicitly.

modules/management/kubectl-apply/helm_install/variables.tf
```diff
@@ -210,3 +210,9 @@ variable "postrender" {
   })
   default = null # Disabled by default
 }
+
+variable "deployment_dependency_ids" {
```
bytetwin: Name can be better. How about just dependencies ?

agrawalkhushi18: Sure, we can keep it that way.

modules/management/kubectl-apply/variables.tf
SwarnaBharathiMantena: I am inclined towards keeping it as `system_node_pool_id` as it will make it seamless with addition of cluster module id to the `use` section in kubectl-apply module.
Also because this variable is only used for kueue installation at this time and only the system nodepool has been identified as a dependency.

modules/management/kubectl-apply/variables.tf
```diff
@@ -117,7 +117,6 @@ variable "kueue" {
   description = "Install and configure [Kueue](https://kueue.sigs.k8s.io/docs/overview/) workload scheduler. If `wait` is true, the installation will wait for resources to be ready. A configuration yaml/template file can be provided with config_path to be applied right after kueue installation. If a template file provided, its variables can be set to config_template_vars."
```
SwarnaBharathiMantena: If `wait` is being removed, please update the description here.

modules/management/kubectl-apply/helm_install/variables.tf
agrawalkhushi18: Done.

modules/management/kubectl-apply/variables.tf
agrawalkhushi18: Thanks for pointing that out. Updated the description.

modules/management/kubectl-apply/variables.tf
agrawalkhushi18: Since the system node pool is the only dependency currently identified for resolving the Kueue deadlock, I think we can proceed with this specific name to keep the blueprint configuration simple. If we encounter customer requirements for additional dependencies in the future, we can refactor the code to be more generic at that time.

PR #5149, opened 2026-01-29
Adding terraform trigger for cleanup script and spacing out onspot tests

simrankaurb:
This PR introduces a daily Cloud Build trigger for resource cleanup and adjusts the scheduling of on-spot integration tests to prevent overlapping resource contention

### DIFF
--- a/tools/cloud-build/provision/daily-cleanup.tf
+++ b/tools/cloud-build/provision/daily-cleanup.tf
@@ -63,3 +63,29 @@ module "daily_project_cleanup_slurm_schedule" {
   schedule    = "0 0 * * MON-FRI"
   retry_count = 4
 }
+
+resource "google_cloudbuild_trigger" "daily_project_cleanup" {
+  name        = "DAILY-project-cleanup"
+  description = "A resource cleanup script to run periodically"
+  tags        = [local.notify_chat_tag]
+
+  git_file_source {
+    path      = "tools/cloud-build/project-cleanup.yaml"
+    revision  = local.ref_develop
+    uri       = var.repo_uri
+    repo_type = "GITHUB"
+  }
+
+  source_to_build {
+    uri       = var.repo_uri
+    ref       = local.ref_develop
+    repo_type = "GITHUB"
+  }
+}
+
+module "daily_project_cleanup_schedule" {
+  source      = "./trigger-schedule"
+  trigger     = google_cloudbuild_trigger.daily_project_cleanup
+  schedule    = "0 22 * * *"
+  retry_count = 4
+}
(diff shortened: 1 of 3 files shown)


tools/cloud-build/provision/daily-cleanup.tf
```diff
@@ -63,3 +63,29 @@ module "daily_project_cleanup_slurm_schedule" {
...
+    path      = "tools/cloud-build/project-cleanup.yaml"
+    revision  = local.ref_develop
+    uri       = var.repo_uri
+    repo_type = "GITHUB"
+  }
+
+  source_to_build {
+    uri       = var.repo_uri
+    ref       = local.ref_develop
+    repo_type = "GITHUB"
```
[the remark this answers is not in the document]
simrankaurb: Keeping them here for consistency with existing code, will take up another task to remove redundant code across all files. Thanks.

tools/cloud-build/provision/daily-cleanup.tf
```diff
@@ -63,3 +63,29 @@ module "daily_project_cleanup_slurm_schedule" {
...
+  }
+}
+
+module "daily_project_cleanup_schedule" {
+  source      = "./trigger-schedule"
+  trigger     = google_cloudbuild_trigger.daily_project_cleanup
+  schedule    = "0 22 * * *"
+  retry_count = 4
```
LAVEEN, line 90: Can we please check on this? Why 4 retries are needed for cleanup to run successfully?
One first,second and third attempt what are leftover resources that doesn't gets cleaned up .

tools/cloud-build/provision/list_tests.py
```diff
@@ -54,9 +54,12 @@
     # (set_of_tests, min_distance)
     ((
         "ml-a4-highgpu-slurm",
-        "ml-a4-highgpu-onspot-slurm",
```
LAVEEN, line 57: if this two runs parallelly ? doesn't this creates a resource collision with current architecture that we have?
        "ml-a4-highgpu-slurm",
        "ml-a4-highgpu-onspot-slurm"

tools/cloud-build/provision/daily-cleanup.tf
simrankaurb: This is to keep it consistent with other cleanup scripts in this file: https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/tools/cloud-build/provision/daily-cleanup.tf.

tools/cloud-build/provision/list_tests.py
simrankaurb: I have confirmed from the past few runs that they are running in different regions so resource conflict should not be a problem. Rather, for ml-a4-highgpu-onspot-slurm and gke-a4-onspot tests, we might face collisions and I have spaced them out.

PR #5160, opened 2026-01-30
feat(validations): Add early conditional validation

AdarshK15:
### **Summary**
  This PR introduces the conditional validator to the module-scoped validation framework. This validator allows module settings to be validated based on the state or value of another "trigger" setting and ensures that constraints are caught earlier in the deployment lifecycle.

###   Key Changes
  New `ConditionalValidator`: Implements a logic-based validator that enforces that a dependent variable is set or matches a specific value only when a trigger variable condition is met. This is useful for cross-variable dependencies (e.g., if feature X is enabled, setting Y is required).

###   Functionality:
   * If **trigger_value** or **dependent_value** is omitted, the validator checks if the variable is simply "set" (non-null, true bool, positive integer, and non-empty list/tuple/map).
   * If a value is provided, it must match exactly. This also supports matching against null to check if a variable is explicitly omitted.
   * Optional Handling: Respects the optional flag to skip validation if the trigger variable is not found in the blueprint.

  ### Testing
  Added comprehensive unit tests covering:
   * Trigger scenarios: checking both "is set" logic and "exact match" logic.
   * Dependent enforcement: validating that variables are either set or match expected values when conditions are met.
   * Verification of error messages to ensure they correctly report mismatched values.

  ### Documentation
  Updated docs/blueprint-validation.md to include usage instructions and examples for the new conditional validator.

### DIFF
--- a/modules/compute/gke-node-pool/main.tf
+++ b/modules/compute/gke-node-pool/main.tf
@@ -392,7 +392,7 @@ resource "google_container_node_pool" "node_pool" {
     }
     precondition {
       condition     = var.enable_flex_start == true ? (var.reservation_affinity.consume_reservation_type == "NO_RESERVATION") : true
-      error_message = "enable_flex_start only works with reservation_affinity consume_reservation_type NO_RESERVATION."
+      error_message = "enable_flex_start requires reservation_affinity.consume_reservation_type to be 'NO_RESERVATION'."
     }
     precondition {
       condition     = !(var.enable_flex_start && var.spot)


modules/compute/gke-node-pool/metadata.yaml
```diff
@@ -35,3 +35,38 @@ ghpc:
       vars: [local.upgrade_settings.max_unavailable]
       min: 0
     error_message: "'max_unavailable' value must be at least 0."
+  - validator: conditional
+    inputs:
+      trigger: enable_flex_start
+      trigger_value: true
+      dependent: reservation_affinity.consume_reservation_type
+      dependent_value: "NO_RESERVATION"
+    error_message: "enable_flex_start only works with reservation_affinity consume_reservation_type NO_RESERVATION."
```
LAVEEN: what is this error message indicates? Not able to follow completely.

AdarshK15: updated the error message.

PR #5166, opened 2026-02-02
feat(validations): Add early required validation

AdarshK15:
### Summary
  This PR introduces the required validator to the module-scoped validation framework. It allows enforcing the presence of mandatory settings or explicitly forbid specific settings (e.g., for deprecation) directly within metadata.yaml. By moving these checks into the metadata layer, these constraints are validated earlier in the deployment lifecycle.

###   Key Changes
  New **RequiredValidator**: Implements a validator to enforce the presence or absence of module settings.

  **Functionality:**
   * Required (Default): Validates that all variables in the provided list are "set".
   * Forbidden: When configured with forbidden: true, validates that none of the specified variables are "set". This is
     useful for preventing the use of deprecated variables or enforcing "must not be set" constraints.
   * Active: A variable is considered "set" if it is a non-empty string, a non-zero number, a true boolean, or a
     non-empty collection (list/map).

  ### Testing
  Added unit tests in pkg/validators/metadata_validators_test.go covering:
   * Pass/Fail scenarios for missing required variables.
   * Pass/Fail scenarios for present forbidden variables.

  ### Documentation
  Updated docs/blueprint-validation.md to include usage instructions and YAML examples for both enforcing required
  variables and forbidding unwanted ones.

### DIFF
--- a/pkg/validators/metadata_validators.go
+++ b/pkg/validators/metadata_validators.go
@@ -333,7 +333,7 @@ func (r *RequiredValidator) Validate(
 	rule modulereader.ValidationRule,
 	group config.Group,
 	modIdx int) error {
-  
+
 	var unsetVarNames []string
 	var setVarNames []string
 
@@ -380,7 +380,7 @@ func (r *RequiredValidator) Validate(
 			msg = fmt.Sprintf("%s: %s", rule.ErrorMessage, msg)
 		}
 		return config.BpError{Err: fmt.Errorf("%s", msg), Path: modPath}
-  }
+	}
 	return nil
 }
 


docs/blueprint-validation.md
```diff
@@ -150,6 +150,35 @@ ghpc:
     error_message: "The 'versions' list must contain at least one version."
 ```
 
+### Required Validator
+The `required` validator ensures that a specific set of variables is either present or absent (forbidden). This is useful for enforcing mandatory inputs or preventing conflicting configurations.
```
arpit974: nit: might need to rephrase this.

pkg/validators/metadata_validators.go
```diff
@@ -295,3 +295,64 @@ func (r *RangeValidator) Validate(
...
+		return nil
+	}
+	modPath := config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source
+
+	varsList, ok := parseStringList(rule.Inputs["vars"])
+	if !ok {
+		return config.BpError{Err: fmt.Errorf("validation rule for module %q is missing 'vars'", mod.ID), Path: modPath}
+	}
+	for _, varName := range varsList {
+		values, _, _ := getModuleSettingValues(bp, group, modIdx, mod, varName)
```
arpit974, line 355: why are we not getting error here.?

docs/blueprint-validation.md
AdarshK15: Updated description.

pkg/validators/metadata_validators.go
```diff
@@ -322,3 +322,64 @@ func (e *ExclusiveValidator) Validate(
...
+	}
+	if len(unsetVarNames) > 0 {
+		msg := fmt.Sprintf("missing required settings: %s", strings.Join(unsetVarNames, ", "))
+		if rule.ErrorMessage != "" {
+			msg = fmt.Sprintf("%s: %s", rule.ErrorMessage, msg)
+		}
+		return config.BpError{Err: fmt.Errorf("%s", msg), Path: modPath}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
AdarshK15: I considered using IterateRuleTargets, but it skips calling the handler when a variable is missing from the blueprint. Since the core purpose of this validator is to identify these missing variables, I used a manual implementation to ensure the handler catches missing or null values and aggregates them into a single error message.

pkg/validators/metadata_validators.go
AdarshK15: We aggregate the set and unset variables first because the validation logic differs depending on 'deprecated' flag. At this point, we donot have context on if variable is required or not so If we catch error here, we skip calling handler for missing or null variables.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/metadata.yaml
```diff
@@ -36,3 +36,8 @@ ghpc:
     inputs:
       vars: [instance_image.name, instance_image.family]
     error_message: "In 'instance_image', at most one of the \"family\" or \"name\" fields can be set."
+  - validator: required
+    inputs:
+      vars: [enable_smt, disable_smt]
+      deprecated: true
+    error_message: "DEPRECATED: Use `advanced_machine_features.threads_per_core` instead."
```
Neelabh94: Deprecated usually refers to functional but discouraged features. Since we are erroring out here, we should use clear messaging that is not supported

arpit974 (approved):

PR #5220, opened 2026-02-10
Install Kueue in TPU base bps of v6e and 7x

SwarnaBharathiMantena:
Install Kueue in TPU base bps of v6e and 7x. At this time, the Kueue is installed in the advanced blueprints. We want to install Kueue on all out blueprints.


shubpal07 (approved):
LGTM

PR #5229, opened 2026-02-12
Module deprecation warning system

vikramvs-gg:
**feat: Add Module Deprecation Warning System**
This pull request introduces validations in gcluster engine to automatically warn users when they are using a deprecated module in a blueprint. 

**Why is it needed**
This change provides a clear, in-tool mechanism to notify users about a module's deprecation status, allowing for smoother migrations and empowering users to plan migrations proactively.

  **How It Works**
  The system operates through metadata-driven checks integrated into the toolkit's core blueprint validation process:

   1. Module Metadata:
       * A ghpc block in each module's metadata.yaml can now include a deprecation_date (in YYYY-MM-DD format) and an alternative_module suggestion.
       
   2. Automated Checks:
       * During blueprint processing, the tool checks each module's deprecation metadata.
       * If a deprecation_date is present and valid, a warning is issued, indicating if the module's deprecation date is in the future (planned removal) or in the past (no longer supported).
       * If the deprecation_date is malformed (e.g., "tomorrow"), a fatal error is immediately triggered, stopping the blueprint validation to prevent deployment with invalid configurations.
       * If no deprecation_date is specified, the module is considered non-deprecated, and no warning is issued.

**Testing**
- Validated scenarios of warnings and errors for deprecated and non deprecated modules
- Unit tests for go code changes running successfully.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -137,14 +137,17 @@ func validateMaybeDie(bp config.Blueprint, ctx config.YamlCtx) {
 	}
 	logging.Error("%s", renderError(err, ctx))
 
-	var errorMsg strings.Builder
-	errorMsg.WriteString("One or more blueprint validators has failed. See messages above for suggested actions.\n")
-	errorMsg.WriteString("General troubleshooting guidance and instructions for configuring alidators are shown below:\n\n")
-	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-troubleshooting\n")
-	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-validation\n\n")
-	errorMsg.WriteString("Validators can be silenced or treated as warnings or errors:\n\n")
-	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-validation-levels\n")
-	logging.Error("%s", errorMsg.String())
+	const errorMsg = `One or more blueprint validators has failed. See messages above for suggested
+actions. General troubleshooting guidance and instructions for configuring validators are shown below.
+
+- https://goo.gle/hpc-toolkit-troubleshooting
+- https://goo.gle/hpc-toolkit-validation
+
+Validators can be silenced or treated as warnings or errors:
+
+- https://goo.gle/hpc-toolkit-validation-levels
+`
+	logging.Error("%s", errorMsg)
 
 	switch bp.ValidationLevel {
 	case config.ValidationWarning:
--- a/pkg/config/deprecation_validate.go
+++ b/pkg/config/deprecation_validate.go
@@ -41,8 +41,9 @@ func validateDeprecation(modID ModuleID, info modulereader.ModuleInfo) error {
 	var msgBuilder strings.Builder
 	if currentTime.Before(deprecationDate) {
 		// Phase 1: Announcement & Warning Period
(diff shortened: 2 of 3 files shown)


cmd/create.go
```diff
@@ -137,30 +137,25 @@ func validateMaybeDie(bp config.Blueprint, ctx config.YamlCtx) {
...
-	logging.Error("- https://goo.gle/hpc-toolkit-validation-levels")
-	logging.Error("")
+	var errorMsg strings.Builder
+	errorMsg.WriteString("One or more blueprint validators has failed. See messages above for suggested actions.\n")
+	errorMsg.WriteString("General troubleshooting guidance and instructions for configuring alidators are shown below:\n\n")
+	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-troubleshooting\n")
+	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-validation\n\n")
+	errorMsg.WriteString("Validators can be silenced or treated as warnings or errors:\n\n")
+	errorMsg.WriteString("- https://goo.gle/hpc-toolkit-validation-levels\n")
+	logging.Error("%s", errorMsg.String())
```
bytetwin: Instead of string builder, why not a multi line string ? This string is static

pkg/config/deprecation_validate.go
```diff
@@ -0,0 +1,55 @@
...
+
+	currentTime := time.Now()
+	alternativeModule := info.Metadata.Ghpc.AlternativeModule
+
+	var msgBuilder strings.Builder
+	if currentTime.Before(deprecationDate) {
+		// Phase 1: Announcement & Warning Period
+		msgBuilder.WriteString(fmt.Sprintf("The module %s will be deprecated on %s. Module will be removed on this date.", modID, deprecationDate.Format("2006-01-02")))
+		msgBuilder.WriteString("\nNo new features will be added to the module. Bug fixes will be avoided, unless absolutely critical. No new blueprints should use this module.")
+	} else {
```
bytetwin, line 47:
```
msgBuilder.WriteString(fmt.Sprintf(`The module %s will be deprecated on %s. Module will be removed on this date.
No new features will be added to the module. Bug fixes will be avoided, unless absolutely critical. No new blueprints should use this module.`, 
    modID, deprecationDate.Format("2006-01-02")))
```

Use multi line strings

pkg/logging/logging.go
```diff
@@ -49,13 +50,19 @@ func Info(f string, a ...any) {
 	infolog.Printf("%s: %s", formatTs(), msg)
 }
 
-// Error prints info to stderr but does not end the program
+// Warn prints message to stderr but does not end the program
+func Warn(f string, a ...any) {
+	msg := fmt.Sprintf(f, a...)
+	errorlog.Printf("%s: %s: %s", formatTs(), WarningColor.Sprint("WARNING"), msg)
```
bytetwin: only the string WARNING would have different color, not the msg ?

bytetwin (changes requested):

cmd/create.go
vikramvs-gg: Used multiline string since it is a constant

pkg/config/deprecation_validate.go
vikramvs-gg: Fixed it

pkg/logging/logging.go
vikramvs-gg: Makes sense to have the message as well with warning color to be more visible. Fixed.

swarnabm (approved):

PR #5252, opened 2026-02-17
Support future reservation in name check validator

saara-tyagi27:
### Summary

This pr adds support for future reservation checking in test_reservation_exists validator by making an api call to future reservation.

### DIFF
--- a/pkg/validators/cloud.go
+++ b/pkg/validators/cloud.go
@@ -254,53 +254,74 @@ func testZoneInRegion(bp config.Blueprint, inputs config.Dict) error {
 	return TestZoneInRegion(m["project_id"], m["zone"], m["region"])
 }
 
-// findReservationInOtherZones searches for a reservation by name across all zones
-// in the project, checking both standard and future reservation pools.
-func findReservationInOtherZones(ctx context.Context, s *compute.Service, projectID string, name string) ([]string, error) {
+// Helper interface to treat Standard and Future reservations generically
+type zoneResource interface {
+	GetName() string
+	GetZone() string
+}
+
+// Wrapper for compute.Reservation
+type stdRes struct{ *compute.Reservation }
+
+func (r stdRes) GetName() string { return r.Name }
+func (r stdRes) GetZone() string { return r.Zone }
+
+// Wrapper for compute.FutureReservation
+type futRes struct{ *compute.FutureReservation }
+
+func (r futRes) GetName() string { return r.Name }
+func (r futRes) GetZone() string { return r.Zone }
+
+func extractZonesFromItems[T any](items map[string]T, name string, extractor func(T) []zoneResource) []string {
 	foundInZones := []string{}
-	var err, fErr error
+	for _, scopedList := range items {
+		for _, res := range extractor(scopedList) {
+			if res.GetName() == name {
+				parts := strings.Split(res.GetZone(), "/")
+				foundInZones = append(foundInZones, parts[len(parts)-1])
+			}
+		}
+	}
+	return foundInZones
+}
 
+func findReservationInOtherZones(ctx context.Context, s *compute.Service, projectID string, name string) ([]string, error) {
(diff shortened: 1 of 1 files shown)


pkg/validators/cloud.go
```diff
@@ -296,24 +296,54 @@ func testMachineTypeInZoneAvailability(bp config.Blueprint, inputs config.Dict)
...
+	// If we found results in either pool, we return them and ignore errors from the other
+	// (e.g. if one API is disabled but the other works).
+	if len(foundInZones) > 0 {
+		return foundInZones, nil
+	}
+
+	// If both failed and we found nothing, return the last encountered error.
+	if err != nil && fErr != nil {
+		return nil, lastErr
+	}
```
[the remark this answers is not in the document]
saara-tyagi27: Done.

pkg/validators/cloud.go
```diff
@@ -328,21 +358,28 @@ func TestReservationExists(ctx context.Context, reservationProjectID string, zon
...
+		return nil
+	}
+
+	// 3. Access Check: If both failed, check for metadata blindness (403/400).
+	// We handle this because users might be allowed to CONSUME but not DESCRIBE a shared reservation.
 	if msg, isSoft := getSoftWarningMessage(err, "test_reservation_exists", reservationProjectID, "Compute Engine API", "compute.reservations.get"); isSoft {
 		fmt.Println(msg)
-		return nil // Skip and continue
+		return nil
 	}
```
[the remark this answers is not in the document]
saara-tyagi27: Done.

pkg/validators/cloud.go
```diff
@@ -254,24 +254,48 @@ func testZoneInRegion(bp config.Blueprint, inputs config.Dict) error {
...
+	if err == nil {
+		for _, scopedList := range aggList.Items {
+			for _, res := range scopedList.Reservations {
+				if res.Name == name {
+					parts := strings.Split(res.Zone, "/")
+					foundInZones = append(foundInZones, parts[len(parts)-1])
+				}
+			}
+		}
 	}
```
Neelabh94, line 303: Can we do an early exit here? If the reservation is found when searching in standard reservation pool, do we need to continue to search for it in the Future reservation pool too?
If not, we can have a check here like below:
if len(foundInZones) > 0 {
		return foundInZones, nil
	}
to prevent resource intensive AggregatedList calls on Future reservations. Wdyt?

saara-tyagi27: I suggest keeping both searches because Standard and Future reservations are in two different lists. Reservations for hardware like A4 or B200 often show up in the Future pool. Searching both ensures we give the user a complete "hint" if they pick the wrong zone.

Do you think it's worth the extra API call for better accuracy?

saara-tyagi27: Added early exit as the current functionality is an additional check and not a deployment breaking step.

pkg/validators/cloud.go
```diff
@@ -254,25 +254,74 @@ func testZoneInRegion(bp config.Blueprint, inputs config.Dict) error {
...
+			}
+			return res
+		})
+		if len(found) > 0 {
+			return found, nil
+		}
+	}
+
+	// If both failed and we found nothing, return the errors
+	if err != nil || fErr != nil {
```
Neelabh94, line 320: nit:  We can enhance the UX here by using the below method appropriately.

var errs []error
errs = append(errs, fmt.Errorf("standard reservations: %w", err)) # in the standard check's if-else
errs = append(errs, fmt.Errorf("future reservations: %w", fErr)) # in the future reservation check's if-else

if len(errs) > 0 {  # at line 320
 return nil, fmt.Errorf("failed to list reservations: %w", errors.Join(errs...))
}

saara-tyagi27: Thank you for the suggestion, I will address this in a follow up pr.

PR #5253, opened 2026-02-17
fix: use relative path for slurm_files build_dir

mtibben:
The `slurm_files` module defines `build_dir = abspath("${path.module}/build")`, which stores a machine-specific absolute path in terraform state via `google_storage_bucket_object.source` (for the `devel` and `devel_compute` resources).

When a different engineer runs `terraform plan` from their machine, the absolute path differs, causing terraform to detect a spurious change and plan an unnecessary replacement of the GCS objects — even though the actual file content is identical.

Removing `abspath()` makes `build_dir` use the relative path from `path.module`, which is consistent across machines and eliminates the spurious diffs.

This change is limited to `build_dir`, which is only used in two places: as the `output_path` for the `archive_file` data sources that generate the devel zip files. Both `archive_file` and `google_storage_bucket_object` work correctly with relative paths — terraform resolves them from the root module directory, which is the standard working directory for all terraform operations.

The other `abspath()` calls in this file (`scripts_dir`, `etc_dir`, and the hybrid-mode variables) are unaffected. Those paths are used in `file()` and `templatefile()` calls which read content inline — the paths themselves are never stored in terraform state, so they don't cause cross-machine drift.


SwarnaBharathiMantena (approved):
LGTM

Tests ran successfully. The failing ones here are known issues on develop.

PR #5282, opened 2026-02-25
Migrate kubectl_apply_manifest module to helm

agrawalkhushi18:
This PR migrates the `kubectl_apply_manifest` module from the `gavinbunney/kubectl` provider to the official `hashicorp/helm` provider.

### Key Changes

**1. Re-implementation of `kubectl_apply_manifests`**
- Replaced the underlying module logic to use `hashicorp/helm` via a local `raw-config-chart`.

**2. Enhanced Manifest Processing in `locals`**
- Relocated and consolidated all file-reading, templating, and directory-scanning logic from the submodule into the main module's `locals` block.
- The Helm provider requires pure YAML string inputs. The logic pre-processes all input types (URLs, `.tftpl` templates, raw files, and directories) into a single rendered string before passing it to Helm.

**3. Server-Side Apply (SSA) Deprecation**
- Removed `server_side_apply` support as it is no longer necessary. Few blueprints have been modified accordingly.
- Helm stores release state in GZIP-compressed Secrets (limit ~1MB+), effectively bypassing the 256KB constraint of the `kubectl.kubernetes.io/last-applied-configuration` annotation that necessitated SSA for large CRDs.
- Modified the kubectl-apply README.md according to the helm provider settings.

**4. Release Naming Stability**
- Introduced a `random_id` resource to generate a unique 4-byte suffix for each Helm release (e.g., `manifest-apply-ceab0dfc-0`).
- Added `hashicorp/random` to `versions.tf` to support conflict-free release naming.
- This prevents namespace collisions when multiple modules (e.g., `gke-cluster`, `gke-node-pool`, and user workloads) instantiate `kubectl-apply` simultaneously within the same blueprint. The 4-byte length was chosen to guarantee uniqueness while staying comfortably within Kubernetes' 53-character name limit.

**5. Atomic Operations**
- Enabled `atomic = true` by default on the Helm release.
- Ensures that failed manifest applications (e.g., a typo or quota error) automatically trigger a rollback, preventing the cluster from being left in a "half-applied" zombie state.

NOTE: 
- The extra step to re-apply nccl-config and nri-device-injector manifest is removed from the test file because the gpu_direct.tf has been updated to fetch the latest manifests instead of an older commit hash which was causing the discrepancy in nccl tests.
- The changes to the outputs.tf of slinky module resolves a deployment race condition where `gke-nodeset` Custom Resources were being applied before the required slinky Helm CRDs finished installing. By anchoring explicit `depends_on` constraints to the slinky module outputs, we force the Terraform DAG to correctly serialize both the creation and destruction phases of the deployment.

### Submission Checklist

### DIFF
--- a/community/examples/slurm-gke/slurm-gke.yaml
+++ b/community/examples/slurm-gke/slurm-gke.yaml
@@ -121,6 +121,15 @@ deployment_groups:
       machine_type: c2-standard-16
       zones: [$(vars.zone)]
 
+  - id: gke_ns_manifest
+    source: modules/management/kubectl-apply
+    use: [gke_cluster]
+    settings:
+      apply_manifests:
+      - source: $(ghpc_stage("./files/slurm-namespace.yaml.tftpl"))
+        template_vars:
+          namespace: $(vars.slurm_namespace)
+
   - id: slinky
     source: community/modules/scheduler/slinky
     use:


modules/management/kubectl-apply/main.tf
```diff
@@ -41,23 +41,28 @@ locals {
...
-      template_vars     = manifest.template_vars
-      server_side_apply = manifest.server_side_apply
-      wait_for_rollout  = manifest.wait_for_rollout
-    }
+      content = (
+        contains(keys(local.url_manifests), tostring(index)) ? data.http.manifest_from_url[tostring(index)].body :
+        try(manifest.source, "") != "" ? (
+          endswith(manifest.source, "/") ? (
+            join("\n---\n", [
+              for f in fileset(manifest.source, "*") : (
```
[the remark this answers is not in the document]
agrawalkhushi18: The fileset now uses the more specific glob pattern {*.yaml,*.yml,*.tftpl} to ensure only likely manifest and template files are processed from directories, making it more robust.

modules/management/kubectl-apply/main.tf
```diff
@@ -80,20 +85,30 @@ data "google_container_cluster" "gke_cluster" {
...
 
-  content           = each.value.content
-  source_path       = each.value.source
-  template_vars     = each.value.template_vars
-  server_side_apply = each.value.server_side_apply
-  wait_for_rollout  = each.value.wait_for_rollout
+  release_name  = "manifest-apply-${random_id.release_suffix.hex}-${each.key}"
+  chart_name    = "${path.module}/raw-config-chart"
+  chart_version = "0.1.0"
+  namespace     = "default"
```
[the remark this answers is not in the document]
agrawalkhushi18: By including namespace = manifest.namespace in the processed_apply_manifests_map  and passing it to the module call, we have removed the hardcoded "default" value.
Assuming manifest.namespace defaults to null in your variables.tf, the helm_release will now naturally fall back to the provider's default context, mimicking the previous kubectl experience.

modules/compute/gke-node-pool/gpu_direct.tf
```diff
@@ -26,9 +26,9 @@ locals {
...
       gpu_direct_manifests = [
-        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/fee883360a660f71ba07478db95d5c1325322f77/gpudirect-tcpx/nccl-tcpx-installer.yaml",      # nccl_plugin v3.1.9 for tcpx
-        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/fee883360a660f71ba07478db95d5c1325322f77/gpudirect-tcpx/nccl-config.yaml",              # nccl_configmap
-        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/fee883360a660f71ba07478db95d5c1325322f77/nri_device_injector/nri-device-injector.yaml", # nri_plugin
+        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/gpudirect-tcpx/nccl-tcpx-installer.yaml",      # nccl_plugin v3.1.9 for tcpx
+        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/gpudirect-tcpx/nccl-config.yaml",              # nccl_configmap
+        "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nri_device_injector/nri-device-injector.yaml", # nri_plugin
```
shubpal07: We know that these manifests point to an old commit, but directly basing them to the `master` version poses an inherent risk of upstream breaking changes. Some options worth considering for:
1. We have [PR](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/5246) up (which will remove these direct manifest injections from gpu_direct).

2. We might also consider to use the last stable and tested commit hash of `container-engine-accelerators`

modules/management/kubectl-apply/README.md
```diff
@@ -15,9 +15,9 @@ This module simplifies the following functionality:
     * A single URL to a manifest file. Ex.: `https://github.com/.../myrepo/manifest.yaml`.
 
     > **Note:** Applying from a URL has important limitations. Please review the [Considerations & Callouts for Applying from URLs](#applying-manifests-from-urls-considerations--callouts) section below.
-    * A single local YAML manifest file (`.yaml`). Ex.: `./manifest.yaml`.
+    * A single local YAML manifest file (`.yaml` or `.yml`). Ex.: `./manifest.yaml`.
     * A template file (`.tftpl`) to generate a manifest. Ex.: `./template.yaml.tftpl`. You can pass the variables to format the template file in `template_vars`.
-    * A directory containing multiple YAML or template files. Ex: `./manifests/`. You can pass the variables to format the template files in `template_vars`.
+    * A directory containing multiple YAML or template files. Ex: `./manifests/` or `./manifests`. The module correctly identifies directories even if the trailing slash is omitted. For security and stability, the module only processes files with `.yaml`, `.yml`, or `.tftpl` extensions. Other files in the directory (like `README.md` etc. ) are automatically ignored.
```
shubpal07, line 20: It will be a good idea if we can provide some documentation/guidance here to users regarding `Helm-based Manifest Application`.
Some points: Helm Resource ownerships, Large manifests and CRDs, Release suffixes, etc

modules/management/kubectl-apply/main.tf
```diff
@@ -80,20 +94,30 @@ data "google_container_cluster" "gke_cluster" {
...
+  wait          = each.value.wait_for_rollout
+  timeout       = 1200
+  values_yaml = [
+    yamlencode({
+      manifests = [
+        for doc in split("\n---", each.value.content) : trimspace(doc)
+        if length(trimspace(doc)) > 0
+      ]
+    })
+  ]
```
shubpal07, line 155: The way we parse multi-document YAMLs uses `split("\n---", content)`. This is highly **error-prone** because `\n---` is frequently used inside valid Kubernetes objects (like embedded shell scripts or YAML blocks defined as strings in `ConfigMaps`). Splitting at the Terraform level will shred these manifests and cause deployment failures.

Since the `raw-config-chart` uses a range loop in its template, we don't actually need to split the string in Terraform. We can simply pass the entire unbroken block of YAML text as a single element in the manifests list, and Kubernetes will natively parse the stream correctly.
By keeping the string whole in Terraform, we push the responsibility of slicing up the YAML documents to the native [Kubernetes decoders](https://kubernetes.io/docs/concepts/workloads/management/#organizing-resource-configurations)

```hcl
  values_yaml = [
    yamlencode({
      # Pass the entire unbroken string to Helm. Helm will parse inner '---' natively.
      manifests = length(trimspace(each.value.content)) > 0 ? [each.value.content] : []
    })
  ]
```
*(Note: I noticed this exact same issue exists in for module "configure_kueue". It would be awesome if we could apply this same fix to `kueue_config_content` Refer [CP](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/management/kubectl-apply/main.tf#L129-L133))*

modules/management/kubectl-apply/README.md
```diff
@@ -165,17 +165,7 @@ To ensure a reliable deployment, you must manually enforce the correct order of
 
 4. **Run the deployment command again.** Since the CRDs are now guaranteed to exist in the cluster, this second apply will succeed reliably.
 
-#### **2. Large Manifests (CRDs)**
-
-* **Issue:** Applying very large manifests can fail with a `metadata.annotations: Too long` error.
-* **Solution:** Enable Server-Side Apply by setting `server_side_apply: true` for the manifest entry.
-
-#### **3. Conflicts on Re-application**
```
shubpal07, line 173: Is there any guidance on Helm based conflicts which can arise due to resource ownerships and re-deployment of resources? If any, we might need to point them here (with resolution steps).

tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a3-high.yml
```diff
@@ -29,16 +29,16 @@
...
-- name: Apply NCCL Config and NRI Device Injector
-  delegate_to: localhost
-  ansible.builtin.shell: |
-    # Apply the NCCL configuration required by the all-gather scripts
-    kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/gpudirect-tcpx/nccl-config.yaml
+# - name: Apply NCCL Config and NRI Device Injector
+#   delegate_to: localhost
+#   ansible.builtin.shell: |
+#     # Apply the NCCL configuration required by the all-gather scripts
+#     kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/gpudirect-tcpx/nccl-config.yaml
```
shubpal07: NIT: commented code

tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a3-high.yml
```diff
@@ -29,16 +29,16 @@
...
 
-    # Apply the NRI Device Injector manifest
-    kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nri_device_injector/nri-device-injector.yaml
-  register: infra_setup
-  changed_when: "'created' in infra_setup.stdout or 'configured' in infra_setup.stdout"
+#     # Apply the NRI Device Injector manifest
+#     kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nri_device_injector/nri-device-injector.yaml
+#   register: infra_setup
+#   changed_when: "'created' in infra_setup.stdout or 'configured' in infra_setup.stdout"
```
shubpal07: NIT: commented code

modules/management/kubectl-apply/main.tf
```diff
@@ -41,23 +41,37 @@ locals {
...
+            ])
+            ) : (
+            endswith(manifest.source, ".tftpl") || length(try(manifest.template_vars, {})) > 0 ?
+            templatefile(manifest.source, try(manifest.template_vars, {})) :
+            file(manifest.source)
+          )
+        ) : coalesce(manifest.content, "")
+      )
+      wait_for_rollout = manifest.wait_for_rollout
+      namespace        = manifest.namespace
```
shubpal07, line 109: The logic blocks in locals for processing manifest.source have become quite complex and nested. 

Could we abstract this into intermediate maps to make the source resolution explicitly clear? Here is a order we can refer:

```
# 1. First, Identify manifests that are explicitly enabled.
# 2. Identify URL-based manifests
# 3. Identify directory-based manifests
# 4. Identify single-file manifests
# 5. Rebuild the map by populating the 'content' field for all manifests
```

tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a3-high.yml
agrawalkhushi18: Removed the redundant step from test file.

tools/cloud-build/daily-tests/ansible_playbooks/test-validation/test-gke-a3-high.yml
agrawalkhushi18: Removed the redundant step from test file.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Great catch. I have updated the logic to pass the entire unbroken block of YAML as a single element in the manifests list in both the modules.

modules/management/kubectl-apply/variables.tf
```diff
@@ -102,12 +102,12 @@ variable "cluster_id" {
 variable "apply_manifests" {
   description = "A list of manifests to apply to GKE cluster using kubectl. For more details see [kubectl module's inputs](kubectl/README.md).\n NOTE: The `enable` input acts as a FF to apply a manifest or not. By default it is always set to `true`. "
```
SwarnaBharathiMantena: Is it good to update the description here to mention that Helm is being used. Also, we need to update the reference to the kubectl module stated here I think. Instead add reference to helm_install if applicable.

modules/management/kubectl-apply/variables.tf
```diff
@@ -102,12 +102,12 @@ variable "cluster_id" {
...
-    source            = optional(string, null)
-    template_vars     = optional(map(any), null)
-    server_side_apply = optional(bool, false)
-    wait_for_rollout  = optional(bool, true)
+    enable           = optional(bool, true)
+    content          = optional(string, null)
+    source           = optional(string, null)
+    template_vars    = optional(map(any), null)
+    wait_for_rollout = optional(bool, true)
+    namespace        = optional(string, null)
```
SwarnaBharathiMantena, line 110: What does "null" namespace map to? Is it "default" namespace?

modules/compute/gke-node-pool/gpu_direct.tf
SwarnaBharathiMantena: Yes, reference to `master` can cause issues. Please update this to latest commit.

modules/management/kubectl-apply/main.tf
```diff
@@ -80,20 +94,28 @@ data "google_container_cluster" "gke_cluster" {
...
-  content           = each.value.content
-  source_path       = each.value.source
-  template_vars     = each.value.template_vars
-  server_side_apply = each.value.server_side_apply
-  wait_for_rollout  = each.value.wait_for_rollout
-
-  providers = {
-    kubectl = kubectl
-  }
+  release_name  = "manifest-apply-${random_id.release_suffix.hex}-${each.key}"
```
SwarnaBharathiMantena, line 143: I think making it stable (which is not working in our favor right now) or adding timestamp instead might be more useful from a user perspective. If the user wants to work with `helm rollback` or `helm history`, then the random suffix can cause confusion on which release is the latest. We can also consider exposing a release version suffix if that is a better approach.

modules/management/kubectl-apply/README.md
SwarnaBharathiMantena: The readme has reference of server_side_apply in 5 locations. Please update those.

modules/management/kubectl-apply/variables.tf
agrawalkhushi18: By setting the default to null, we are enabling contextual inheritance. In Terraform, passing null to the namespace attribute of a helm_release tells the provider to use its own default configuration. This exactly mimics the behavior of the previous kubectl implementation while giving advanced users the choice to be explicit when they want to co-locate Helm metadata with resources in a custom namespace.

modules/management/kubectl-apply/variables.tf
agrawalkhushi18: Good catch. I have updated the description.

modules/management/kubectl-apply/README.md
agrawalkhushi18: Updated the references. Thanks!

modules/compute/gke-node-pool/gpu_direct.tf
agrawalkhushi18: I have updated the link to the latest commit for this PR. Revamping of gke-a3-high PR would handle the injection of these manifests by adding local manifests in the sub-directory instead of applying it through URLs in gpu_direct.tf. Thanks Shubham for incorporating this change.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Thanks for the suggestion, Shubham! I have refactored the locals block into the multi-step mapping. While the logic for identifying explicitly enabled and URL-based manifests was already present, I have now added the directory-based classification to simplify it a bit.

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Thanks for the suggestion. I'll look into using module names as a suffix for manifest releases to make them easier to identify. I'll move this to a separate ticket so we can take it up after the current changes.

modules/management/kubectl-apply/README.md
agrawalkhushi18: I have updated the README to include a section on Helm-based Manifest Application.
Thanks

modules/management/kubectl-apply/README.md
agrawalkhushi18: Helm uses specific labels and annotations to track and manage the lifecycle of Kubernetes resources. If we encounter a persistent conflict where a leftover unmanaged cluster resource blocks your deployment, we must manually delete the conflicting resource  from the cluster (`kubectl delete <kind> <name>`) and then re-run your `gcluster deploy`. This has been mentioned as a point in the `Re-deployment Conflicts` section.

PR #5352, opened 2026-03-12
feat: Cluster Toolkit Slurm HA Controllers

andybubu:
## 1. Slurm HA Controller Implementation

The HA controller implementation builds upon the standard controller setup but introduces critical changes to support High Availability via a Primary/Backup model and shared state.

### Core Modifications
-   **setup.py**:
    -   **Setup Order**: `setup_network_storage()` is now called *before* `setup_key()` and `setup_jwt_key()`. This is essential because HA relies on shared network storage (Filestore/NFS) to synchronize keys between controllers.
    -   **Key Synchronization**: `setup_key()` was refactored to handle `is_primary` logic.
        -   **Primary**: Generates keys and copies them to the shared state directory (`slurmdirs.state`) i.e `/var/log/spool/slurm`.
        -   **Backup**: Waits for keys to appear in the shared directory and configures `slurmctlcld`.
    -   **Controller Setup**: refactored `setup_controller()` to handle `is_primary` logic. It configures the instances as `primary` and `backup` controllers base on the instance respective roles.
    -   **Munge Keys**: `setup_key()` now also syncs via shared state, similar to jwt keys.

-   **conf.py**:
    -   Updates to `install_slurm_conf` to generate `SlurmctldHost` entries for receiving HA heartbeat messages.
    -   Updates to `install_slurmdbd_conf` to support `DbdBackupHost`.

- **controller.tf**:
   - Updated to add backup_controller instance template,  instance and related resources blocks provided `enable_backup_controller` is true.
 

## 2. Backward Compatibility

The implementation preserves backward compatibility for single-controller deployments:
-   **Guarded by Variable**: The HA functionality is controlled by the `enable_backup_controller` Terraform variable, which defaults to `false`.
-   **Default Behavior**: When `enable_backup_controller` is false, the module behaves exactly as the standard single-controller deployment:
    -   `slurm.conf` is generated with a single controller host.
    -   Key synchronization logic is skipped in favor of local generation.
    -   No backup controller resources are created or configured.

## Submission Checklist

## Test
1. Slurm HA deployed
Screenshot 2026-03-07 at 12 51 41

2. Stop slurmctld on primary controller
image

3. Observe logs on primary controller:
image

4. Observe logs on backup controller:
image

5. shutdown primary controller
image

6. Observe logs on backup controller:
image

8. Check cluster status on backup controller:
    [Note jobs still running and also successful new job submission from backup controller]
image

10. Restart primary controller:
image

11. Observe cluster status:
image

12. Jobs completed successfully:
image

### DIFF
--- a/community/examples/hpc-slurm-ha.yaml
+++ b/community/examples/hpc-slurm-ha.yaml
@@ -1,0 +1,200 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+blueprint_name: hpcslurmha
+
+vars:
+  project_id: # Set Project ID
+  deployment_name: hpcslurmha
+  region: us-central1
+  zone: us-central1-a
+
+terraform_backend_defaults:
+  type: gcs
+  configuration:
+    bucket: # Set Bucket Name Here
+
+deployment_groups:
+- group: primary
+  modules:
+  # Service Accounts
+  - id: controller_sa
+    source: modules/project/service-account
+    kind: terraform
+    settings:
+      deployment_name: $(vars.deployment_name)
+      project_id: $(vars.project_id)
+      name: controller-sa
+      description: Controller Service Account
+      display_name: Controller SA
+      project_roles:
+      - compute.instanceAdmin.v1
+      - iam.serviceAccountUser
+      - storage.objectViewer
+      - logging.logWriter
+      - monitoring.metricWriter
+
+  - id: login_sa
+    source: modules/project/service-account
+    kind: terraform
+    settings:
+      deployment_name: $(vars.deployment_name)
+      project_id: $(vars.project_id)
+      name: login-sa
+      description: Login Node Service Account
+      display_name: Login SA
+      project_roles:
+      - logging.logWriter
+      - monitoring.metricWriter
+      - storage.objectViewer
+
+  - id: compute_sa
+    source: modules/project/service-account
+    kind: terraform
+    settings:
+      deployment_name: $(vars.deployment_name)
+      project_id: $(vars.project_id)
+      name: compute-sa
+      description: Compute Node Service Account
+      display_name: Compute SA
+      project_roles:
+      - logging.logWriter
+      - monitoring.metricWriter
+      - storage.objectViewer
+
+  # Network
+  - id: network
+    source: modules/network/vpc
+    settings:
+      network_name: $(vars.deployment_name)-net
+      subnetworks:
+      - subnet_name: $(vars.deployment_name)-primary-subnet
+        subnet_region: $(vars.region)
+        subnet_ip: 10.80.0.0/24
+
+  # Filestore (Shared /home)
+  - id: home_fs
+    source: modules/file-system/filestore
+    settings:
+      network_id: $(network.network_id)
+      size_gb: 1024
+      zone: $(vars.zone)
+
+  # Debug Logic (Define Login Node properties, does not create VM directly)
+  - id: debug_nodes
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-login
+    settings:
+      name_prefix: debug
+      num_instances: 1
+      machine_type: n1-standard-2
+      service_account_email: $(login_sa.service_account_email)
+      service_account_scopes: ["https://www.googleapis.com/auth/cloud-platform"]
+      enable_oslogin: true
+      subnetwork_self_link: $(network.subnetwork_self_link)
+      enable_login_public_ips: false # NO EXTERNAL IP
+
+  # Nodeset (Compute)
+  - id: compute_nodes
+    source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
+    settings:
+      name: a2-nodes
+      node_count_dynamic_max: 4
+      node_count_static: 0
+      machine_type: n1-standard-2
+      service_account_email: $(compute_sa.service_account_email)
+      service_account_scopes: ["https://www.googleapis.com/auth/cloud-platform"]
+      enable_oslogin: true
+      subnetwork_self_link: $(network.subnetwork_self_link)
+      enable_placement: false
+      network_storage:
+      - server_ip: $(home_fs.network_storage.server_ip)
+        remote_mount: $(home_fs.network_storage.remote_mount)
+        local_mount: /home
+        fs_type: nfs
+        mount_options: _netdev,hard,intr
+
+  # Partition
+  - id: debug_partition
+    source: community/modules/compute/schedmd-slurm-gcp-v6-partition
+    use: [compute_nodes]
+    settings:
+      partition_name: debug
+      is_default: true
+
+  # Filestore (Dedicated State for Controller)
+  - id: controller_fs
+    source: modules/file-system/filestore
+    settings:
+      network_id: $(network.network_id)
+      size_gb: 1024
+      zone: $(vars.zone)
+
+  # Cloud SQL (External DB)
+  - id: slurm_db
+    source: community/modules/database/slurm-cloudsql-federation
+    settings:
+      deployment_name: $(vars.deployment_name)
+      project_id: $(vars.project_id)
+      region: $(vars.region)
+      tier: db-f1-micro
+      sql_instance_name: $(vars.deployment_name)-db
+      network_id: $(network.network_id)
+      use_psc_connection: true
+      subnetwork_self_link: $(network.subnetwork_self_link)
+
+  # Controller (Custom HA Module)
+  - id: slurm_controller
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
+    use: [debug_partition]
+    settings:
+      enable_backup_controller: true
+      slurm_cluster_name: hpcslurmha
+      backup_zone: us-central1-f
+      # Disable PD state disk, rely on /var/spool/slurm filestore share
+      controller_state_disk: null
+      cloudsql:
+        server_ip: $(slurm_db.cloudsql.server_ip)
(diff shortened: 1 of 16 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -490,21 +546,26 @@ def setup_controller():
...
     pass
 
-
 def setup_login():
     """run login node setup"""
     log.info("Setting up login")
 
     lkp = lookup()
-    slurmctld_host = f"{lkp.control_host}"
+    primary_slurmctld_host = f"{lkp.control_host}"
```
arpit974: (inside def setup_login()), and (inside def setup_compute()).
The multiple config-server option string splits on space boundaries under systemd argument parsing on login and compute nodes, causing startup failure. Also, using name-only variables removes the parenthesized address IP fallback format host(IP) that baseline develop uses, making configless nodes entirely dependent on dynamic cloud DNS to boot. Let's clean-format this as a single outer-quoted comma-separated list and restore the parenthesized IP fallback.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
```diff
@@ -327,6 +329,59 @@ def install_slurm_conf(lkp: util.Lookup) -> None:
...
+                log.info(f"Successfully resolved backup_controller hostname '{backup_name}' to {backup_ip}")
+                break
+            except Exception as e:
+                if i < retries - 1:
+                    log.warning(f"Failed to resolve backup_controller hostname '{backup_name}' (Attempt {i + 1}/{retries}): {e}. Retrying in {snooze} seconds...")
+                    time.sleep(snooze)
+                else:
+                    log.error(f"Failed to resolve backup_controller hostname '{backup_name}' after {retries} attempts: {e}")
+                    raise
+
```
arpit974, line 364: The primary controller's startup script blocks inside a 5-minute dynamic DNS polling loop waiting for the backup VM to register. When static IPs are defined in Terraform, we should serialize the static backup IP to GCS config.yaml to instantly bypass this DNS resolution retry loop, keeping the primary VM boot time under 10 seconds.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/slurm_files.tf
arpit974, line 142: Having compute nodes mount keys locally from the primary VM's disk creates a single point of failure that halts autoscaling when the primary is offline. Additionally, because key mounts are hard-mounted, any transient storage dropouts during boot will freeze the VM setup thread in an uninterruptible kernel hang. Let's add dynamic locals to auto-discover the Filestore state spool /var/spool/slurm mount, automatically route munge/slurm keys through it using resilient NFS soft mounts, and turn off default primary VM NFS exports during HA to prevent filesystem freezes on failovers.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -514,7 +575,10 @@ def setup_login():
     if not lkp.cfg.enable_slurm_auth:
       run("systemctl restart munge", timeout=30)
     run("systemctl enable sackd", timeout=30)
-    run("systemctl restart sackd", timeout=30)
+    try:
+        run("systemctl restart sackd", timeout=30)
+    except Exception:
```
arpit974, line 580: Ignoring auth daemon (sackd / munge) startup failures inside broad try-except blocks masks broken VM setups while reporting setup completed successfully. Let's remove the try-except wrapper around daemon restarts so setup script failures register normally, letting GCE health checks cleanly capture unhealthy node provisioning.

andybubu: Wrapping restart in a try-except block tolerates harmless, transient startup races.
Leaving status check unwrapped guarantees that any actual unhealthy state is strictly caught and reported, letting GCE health checks cleanly capture true failures.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
andybubu: This follows the SchedMD [docs](https://slurm.schedmd.com/configless_slurm.html#SETUP) recommendation for details sackd --conf-server options which takes "$host[:$port]" and not "$host(ip)[:$port]"

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -541,12 +605,18 @@ def setup_compute():
 
     lkp = lookup()
     util.chown_slurm(dirs.scripts / "config.yaml", mode=0o600)
-    slurmctld_host = f"{lkp.control_host}"
+    primary_slurmctld_host = f"{lkp.control_host}"
     if lkp.control_addr:
         slurmctld_host = f"{lkp.control_host}({lkp.control_addr})"
```
arpit974: seems like typo: slurmctld_host should be primary_slurmctld_host. Otherwise, the IP fallback is ignored and the primary host configuration remains un-updated.

arpit974 (commented):
During parallel startup, client nodes (login and compute nodes) mount the storage volume and immediately try to copy munge.key/slurm.key from it. However, because the primary controller takes a few minutes to boot and generate these keys, the copy command immediately throws a FileNotFoundError and crashes the node setup.

We should add a wait/retry loop (e.g., polling every 10 seconds up to 300 seconds) to wait for the keys to appear in the NFS share before executing the copy.

andybubu:
> During parallel startup, client nodes (login and compute nodes) mount the storage volume and immediately try to copy munge.key/slurm.key from it. However, because the primary controller takes a few minutes to boot and generate these keys, the copy command immediately throws a FileNotFoundError and crashes the node setup.
> 
> We should add a wait/retry loop (e.g., polling every 10 seconds up to 300 seconds) to wait for the keys to appear in the NFS share before executing the copy.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -541,12 +597,15 @@ def setup_compute():
...
     lkp = lookup()
     util.chown_slurm(dirs.scripts / "config.yaml", mode=0o600)
-    slurmctld_host = f"{lkp.control_host}"
-    if lkp.control_addr:
-        slurmctld_host = f"{lkp.control_host}({lkp.control_addr})"
-    slurmd_options = [
-        f'--conf-server="{slurmctld_host}:{lkp.control_host_port}"',
-    ]
+    primary_control_host = lkp.control_host or lkp.control_addr
+    backup_control_host = lkp.cfg.get("slurm_backup_controller_name") or lkp.cfg.get("slurm_backup_controller_ip")
```
arpit974: Using lkp.control_host or lkp.control_addr will always evaluate to the hostname (since lkp.control_host is always defined and truthy), which ignores the IP fallback entirely.

To preserve the baseline develop behavior where clients can connect directly via IP in private/DNS-less networks during boot, we should restore the host(IP) parenthesized syntax

andybubu: Firstly, host(IP) parenthesized syntax is not valid syntax for configless setup of slurmd and sackd services as per SchedMD [documentation](https://slurm.schedmd.com/configless_slurm.html). 

To address this issue, we  can swapped the evaluation order of the hostnames and IP fallback variables so that the IP address `lkp.control_addr / slurm_backup_controller_ip` takes precedence when defined, falling back to the hostnames `lkp.control_host / slurm_backup_controller_name` if the IPs are absent.

arpit974 (approved):

community/examples/hpc-slurm-ha.yaml
LAVEEN, line 1: can we please add new test for this new blueprint?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -231,35 +231,79 @@ def setup_key(lkp: util.Lookup) -> None:
...
+                    util.chown_slurm(shared_key, mode=0o400)
+                else:
+                    util.chown_slurm(shared_key, mode=0o600)
+            except Exception as e:
+                log.warning(f"Failed to sync key to shared storage: {e}")
+
     else:
-        if dst.exists():
-            log.info("key already exists. Skipping key generation.")
+        # Wait for key to appear in shared storage and copy to local destination
```
LAVEEN, line 277: i think this would timeout in case if someone deploy the blueprint without filestore, does the terraform covers this validation ?

LAVEEN: if not please  add one.

community/examples/hpc-slurm-ha.yaml
arpit974: since the blueprint is in community folder, test can be added as a separate exercise and not part of this PR.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
arpit974: we do have a condition in main.tf, if no share storage and enable_backup_controller flag is true,

deployment will fail with error:

2026-06-08T08:54:32Z: Error: exit status 1

  on main.tf line 144, in module "slurm_controller":
 144:   enable_backup_controller     = true
    ├────────────────
    │ var.controller_state_disk is null
    │ var.enable_backup_controller is true
    │ var.network_storage is empty list of object

When enable_backup_controller is true, controller_state_disk must be set to
null and network_storage must be provided (non-empty) to share state.

This was checked by the validation rule at
../_modules/embedded/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf:94,3-13.


which is the intent.

andybubu: ➕

LAVEEN (approved):

PR #5420, opened 2026-03-28
NAP support on GKE Clusters (gke-cluster module)

SwarnaBharathiMantena:
**GKE Node Auto-Provisioning (NAP) Support**: Added support for GKE Node Auto-Provisioning (NAP) in the gke-cluster module, enabling dynamic node pool provisioning based on workload requirements.

**Terraform Configuration Updates**: Updated Terraform variables and main configuration to include cluster_autoscaling settings, allowing for dynamic resource limits and auto-provisioning defaults.

**Go Configuration Logic**: Implemented Go-based logic to parse machine types and inject internal variables for accelerator configuration, ensuring proper NAP setup.

**Testing**: Added comprehensive unit tests to validate the new autoscaling configuration logic and accelerator extraction.


pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+		origMt = mod.Settings.Get("machine_type")
+	}
+	mod.Settings = mod.Settings.With("machine_type", cty.StringVal(machineType))
+
+	// Defensively restore the original machine_type setting in all circumstances
+	defer func() {
+		if hasMt {
+			mod.Settings = mod.Settings.With("machine_type", origMt)
+		}
+	}()
```
shubpal07: If `hasMt` was false originally (i.e., `machine_type` was not set in `mod.Settings`), the defer function does nothing. This may leave the temporary `machine_type` set in `mod.Settings` after the function returns. 

This side effect may affect subsequent expansion steps that expect `machine_type` to be unset. 

The alternative we may think of is if `hasMt` is false, the deferred function can remove `machine_type` from `mod.Settings` or set it to a state that represents "unset" if removal is not supported by the cty wrapper.


```suggestion
	var origMt cty.Value
	hasMt := mod.Settings.Has("machine_type")
	if hasMt {
		origMt = mod.Settings.Get("machine_type")
	}
	mod.Settings = mod.Settings.With("machine_type", cty.StringVal(machineType))

	// Defensively restore the original machine_type setting in all circumstances
	defer func() {
		if hasMt {
			mod.Settings = mod.Settings.With("machine_type", origMt)
		} else {
			m := mod.Settings.Items()
			delete(m, "machine_type")
			mod.Settings = NewDict(m)
		}
	}()
```

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+func processAutoscalingLimit(resVal cty.Value, bp Blueprint, mod *Module, accCache map[string]cachedAccInfo) (cty.Value, bool, error) {
+	if !resVal.Type().IsObjectType() {
+		return cty.Value{}, false, nil
+	}
+	resMap := resVal.AsValueMap()
+
+	mtVal, ok := resMap["autoprovisioning_machine_type"]
+	if !ok || mtVal.IsNull() || !mtVal.IsKnown() {
+		return cty.Value{}, false, nil
+	}
```
shubpal07, line 112: If a user knows the specific GKE resource type string (e.g., `tpu-v6e-slice` or `nvidia-h100-80gb`) and specifies it directly in `autoprovisioning_resource_type` while omitting `autoprovisioning_machine_type`, the above check will fail. 

The function returns `false`, and the calling loop omits this limit entirely from the final list sent to Terraform. This may result into valid user configurations to be silently dropped if they don't use the machine-type-based resolution path. 
We may think check if `autoprovisioning_resource_type` is already provided. If it is, we may skip the resolution logic and pass the limit through.

shubpal07 (changes requested):

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+			return 0, fmt.Errorf("autoprovisioning_max_count must be greater than 0 for machine type %s, got %d", machineType, maxCount)
+		}
+		if maxCount%acceleratorsPerVM != 0 {
+			return 0, fmt.Errorf("autoprovisioning_max_count (%d) for machine type %s must be a multiple of its native capacity (%d)", maxCount, machineType, acceleratorsPerVM)
+		}
+		return maxCount, nil
+	}
+	// Default to a large number (e.g., 1000) to allow Node Auto-Provisioning
+	// to scale beyond a single VM capacity.
+	return 1000, nil
```
kadupoornima: We are silently setting the upper limit as 1000 here. This could cause unexpected ceilings if users try to scale a massive cluster.

Instead of defining this here, we could define this default directly in variables.tf, like `autoprovisioning_max_count = optional(number, 1000)`. This would ensure that the default is self-documenting in Terraform.

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if err := json.Unmarshal([]byte(cfgJson), &data); err != nil {
+		return 0, "", fmt.Errorf("failed to unmarshal machine configurations: %w", err)
+	}
+
+	if acc, ok := data.GPUs[machineType]; ok {
+		return acc.Count, acc.Type, nil
+	} else if acc, ok := data.TPUs[machineType]; ok {
+		// Map TPU machine type to GKE accelerator identifier
+		tpuAccMap := map[string]string{
```
kadupoornima: Can extract `tpuAccMap` to a package-level variable instead of instantiating on every call.

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+		if mcVal.Type() == cty.Number {
+			f, _ := mcVal.AsBigFloat().Float64()
+			return int(f), true
+		}
+	}
+	return 1, false
+}
+
+func validateAndExtractTotalAccelerators(maxCount int, maxCountPassed bool, acceleratorsPerVM int, machineType string) (int, error) {
+	if maxCountPassed {
```
kadupoornima: We could move this check near the function call itself instead of passing it as a param and checking.

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func extractMaxCount(resMap map[string]cty.Value) (int, bool) {
+	if mcVal, ok := resMap["autoprovisioning_max_count"]; ok && !mcVal.IsNull() && mcVal.IsKnown() {
+		if mcVal.Type() == cty.Number {
+			f, _ := mcVal.AsBigFloat().Float64()
+			return int(f), true
+		}
+	}
+	return 1, false
```
kadupoornima: If the return value is false, does the count matter? Should we keep it as 0, false?

kadupoornima (changes requested):

pkg/config/autoscaling.go
SwarnaBharathiMantena: Updated.

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	if maxCount%acceleratorsPerVM != 0 {
+		return 0, fmt.Errorf("autoprovisioning_max_count (%d) for machine type %s must be a multiple of its native capacity (%d)", maxCount, machineType, acceleratorsPerVM)
+	}
+	return maxCount, nil
+}
+
+func getAcceleratorCountAndType(machineType string, bp Blueprint, mod *Module) (int, string, error) {
+	var origMt cty.Value
+	hasMt := mod.Settings.Has("machine_type")
```
scaliby: can we copy `Module` instead of mutating it?

pkg/config/autoscaling.go
SwarnaBharathiMantena: I have refactored getAcceleratorCountAndType to use a copy of the module instead of mutating it, addressing @scaliby's comment. This removes the complex defer logic and makes the code much cleaner!

pkg/config/autoscaling.go
SwarnaBharathiMantena: Thanks for suggesting, updated the code.

scaliby (approved):
LGTM!

SwarnaBharathiMantena:
The failing tests are either capacity issues or known issues the team is working on.

SwarnaBharathiMantena:
The failing tests are either capacity issues or known issues the team is working on.

pkg/config/machine_mappings.json
```diff
@@ -0,0 +1,61 @@
...
+    "v4": "tpu-v4-podslice",
+    "l4": "nvidia-l4",
+    "rtx": "nvidia-rtx-pro-6000"
+  },
+  "accelerator_shorthand_map": {
+    "l4-1": "g2-standard-12",
+    "l4-2": "g2-standard-24",
+    "l4-4": "g2-standard-48",
+    "l4-8": "g2-standard-96",
+    "rtx-6000-1": "g4-standard-48",
```
kadupoornima, line 32: What about fractional GPUs? ex: `g4-standard-24`

pkg/config/machine_mappings.json
```diff
@@ -0,0 +1,61 @@
...
+    "h100-80gb-8": "a3-highgpu-8g",
+    "h100-mega-80gb-8": "a3-megagpu-8g",
+    "h200-141gb-8": "a3-ultragpu-8g",
+    "b200-8": "a4-highgpu-8g",
+    "gb200-4": "a4x-highgpu-4g",
+    "v4-8": "ct4p-hightpu-4t",
+    "v5p-1": "ct5p-hightpu-1t",
+    "v5p-2": "ct5p-hightpu-2t",
+    "v5p-4": "ct5p-hightpu-4t",
+    "v5e-1": "ct5lp-hightpu-1t",
```
Neelabh94: the common name for this is actually v5litepod

pkg/config/machine_mappings.json
```diff
@@ -0,0 +1,61 @@
+{
+  "machine_family_to_label_map": {
```
Neelabh94, line 2: these maps are pretty similar to the ones already defined in hardware.go? Can we re-use them here?

SwarnaBharathiMantena: Yes, that is the plan for a followup PR. Multiple files have it. My plan is to consolidate and maintain a single `machine_mappings.json`. This will help us reference to this json from any other go file and can serve as a single source of truth wrt terraform modules as well. `json` files are easy to consume in the terraform code.

pkg/config/machine_mappings.json
```diff
@@ -0,0 +1,61 @@
...
+    "ct6e-standard": "tpu-v6e-slice",
+    "ct5lp-hightpu": "tpu-v5e-slice",
+    "ct5p-hightpu": "tpu-v5p-slice",
+    "ct4p-hightpu": "tpu-v4-podslice",
+    "v6e": "tpu-v6e-slice",
+    "v5e": "tpu-v5e-slice",
+    "v5p": "tpu-v5p-slice",
+    "v4": "tpu-v4-podslice",
+    "l4": "nvidia-l4",
+    "rtx": "nvidia-rtx-pro-6000"
```
kadupoornima: Can remove this line. `"g4-standard": "nvidia-rtx-pro-6000",` already exists.

pkg/config/machine_mappings.json
SwarnaBharathiMantena: Added new entries to support these machine types.

modules/scheduler/gke-cluster/README.md
```diff
@@ -149,6 +149,7 @@ limitations under the License.
...
.../>  the precise `autoprovisioning_resource_type` (e.g. `nvidia-h100-80gb`) expected by Terraform.<br/><br/>  WARNING: Enabling autoscaling defaults to effectively unlimited scaling (1,000,000 CPU cores and 10,000,000 GB of memory) unless specific limits are passed in. This may result in large unexpected billing charges if a workload misconfiguration occurs. | <pre>object({<br/>    limits = list(object({<br/>      autoprovisioning_machine_type  = optional(string)<br/>      autoprovisioning_resource_type = optional(string)<br/>      autoprovisioning_max_count     = optional(number, 1000)<br/>    }))<br/>    service_account_email         = optional(string, "")<br/>    oauth_scopes                  = optional(list(string), ["https://www.googleapis.com/auth/cloud-platform"])<br/>    autoprovisioning_disk_size_gb = optional(number, 100)<br/>    autoprovisioning_disk_type    = optional(string, "pd-balanced")<br/>    autoprovisioning_auto_upgrade = optional(bool, true)<br/>    autoprovisioning_auto_repair  = optional(bool, true)<br/>    autoprovisioning_cpu_max      = optional(number, 1000000)<br/>    autoprovisioning_memory_max   = optional(number, 10000000)<br/>  })</pre> | `null` | no |
```
kadupoornima: nit: Makes readme a little bulky

https://screenshot.googleplex.com/FMyzwGqUMVXLbQ8

pkg/config/autoscaling.go
```diff
@@ -0,0 +1,187 @@
...
+// Update that file when adding support for new machine families.
+
+//go:embed machine_mappings.json
+var machineMappingsJSON []byte
+
+// ExpandClusterAutoscaling intercepts the cluster_autoscaling variable,
+// parses machine_type in Go, and injects internal variables for maximum chips and accelerator type.
+type cachedAccInfo struct {
+	count int
+	t     string
```
kadupoornima: nit: can give a more explanatory name.

kadupoornima (dismissed):

modules/scheduler/gke-cluster/README.md
SwarnaBharathiMantena: Having this information captured here lets a user know which variables are supported. The other option is to describe the variable in the README separately. Might be an extra step for a user who wants to know more. I will discuss with the team and proceed on this on a followup PR.

modules/internal/tpu-definition/main.tf
```diff
@@ -38,9 +42,10 @@ locals {
 
   # Project shared JSON into the expected format for tpu_chip_count_map (machine_type -> count)
   tpu_chip_count_map = {
-    for k, v in try(local.accelerators_json.tpus, {}) : k => v.count
+    for k, v in local.tpu_accelerators : k => v.count
   }
 
+
```
Neelabh94, line 48: nit: new line not needed here.

SwarnaBharathiMantena:
Test failures:

Reservation does not exist
1. [PR-test-gke-a3-highgpu](https://pantheon.corp.google.com/cloud-build/builds/dc557244-20c4-442f-b764-a409425603f1?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod)
2. [PR-test-gke-h4d](https://pantheon.corp.google.com/cloud-build/builds/aada2000-33a8-479c-824f-d29916b28369?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod)

Capacity issues
3. [PR-test-gke-a4-onspot](https://pantheon.corp.google.com/cloud-build/builds;region=global/a4504ae6-89ac-4b88-9dc1-2b20ed7c5188?e=13803378&mods=monitoring_api_prod&project=hpc-toolkit-dev)

Known issues
4. [PR-test-gke-g4-onspot](https://pantheon.corp.google.com/cloud-build/builds/e6a88c0c-aa8f-43a2-8bb3-41be7a82fab5?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod)
5. [PR-test-gke-tpu-7x](https://pantheon.corp.google.com/cloud-build/builds/3d93743e-f2e9-4524-97fd-acdf54626077?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod)
6. [PR-test-gke-tpu-v6e-flex](https://pantheon.corp.google.com/cloud-build/builds/1c8cc024-e929-4a22-bade-79389b515d41?project=hpc-toolkit-dev&e=13803378&mods=monitoring_api_prod)
7. [PR-test-gke-a3-highgpu-onspot](https://pantheon.corp.google.com/cloud-build/builds;region=global/3285d1fd-9dce-4bfd-9679-6a6340e419b1?e=13803378&mods=monitoring_api_prod&project=hpc-toolkit-dev)

modules/scheduler/gke-cluster/main.tf
```diff
@@ -65,6 +65,37 @@ locals {
...
+  nap_cpu_max    = local.autoscaling_enabled ? local.autoscaling_config.autoprovisioning_cpu_max : null
+  nap_memory_max = local.autoscaling_enabled ? local.autoscaling_config.autoprovisioning_memory_max : null
+
+  user_provided_resource_types = local.has_autoscaling_limits ? [for limit in local.autoscaling_config.limits : limit.autoprovisioning_resource_type] : []
+
+  add_default_cpu    = local.autoscaling_enabled && !contains(local.user_provided_resource_types, "cpu")
+  add_default_memory = local.autoscaling_enabled && !contains(local.user_provided_resource_types, "memory")
+
+  machine_mappings = jsondecode(var.machine_mappings_json)
+}
```
Neelabh94, line 97: can we redefine the whole block like this:
```
locals {
  # Consolidate all NAP logic into one object
  # If autoscaling is disabled, this whole object becomes null
  nap = var.cluster_autoscaling == null ? null : {
    config          = var.cluster_autoscaling
    service_account = var.cluster_autoscaling.service_account_email != "" ? var.cluster_autoscaling.service_account_email : local.sa_email
    
    # Derived properties only relevant when NAP is enabled
    user_resource_types = [for limit in var.cluster_autoscaling.limits : limit.autoprovisioning_resource_type]
    has_limits          = length(var.cluster_autoscaling.limits) > 0
    
    # Logic to determine if default CPU/Memory limits should be injected
    add_default_cpu     = !contains([for limit in var.cluster_autoscaling.limits : limit.autoprovisioning_resource_type], "cpu")
    add_default_memory  = !contains([for limit in var.cluster_autoscaling.limits : limit.autoprovisioning_resource_type], "memory")
  }

  machine_mappings = jsondecode(var.machine_mappings_json)
}
```

Consequently:
In your google_container_cluster resource, the dynamic block becomes much more intuitive:

```
dynamic "cluster_autoscaling" {
  for_each = local.nap != null ? [1] : []
  content {
    enabled             = true
    autoscaling_profile = var.autoscaling_profile

    dynamic "resource_limits" {
      for_each = concat(
        local.nap.add_default_cpu ? [{ type = "cpu", min = 1, max = local.nap.config.autoprovisioning_cpu_max }] : [],
        local.nap.add_default_memory ? [{ type = "memory", min = 1, max = local.nap.config.autoprovisioning_memory_max }] : [],
        # ... additional limits ...
      )
      # ... rest of the content ...
    }
  }
}
```

Neelabh94 (approved):
added some nit comments on improvements. They can be done as a follow up PR.

LGTM!

kadupoornima (approved):

modules/scheduler/gke-cluster/main.tf
SwarnaBharathiMantena: Thank you, noted. I will raise a new PR for these changes.

shubpal07 (approved):

PR #5422, opened 2026-03-29
feat: Add explicit opt-in Quota Availability Validator

kvenkatachala333:
This PR introduces a Quota Availability Validator to the Cluster Toolkit as an explicit, opt-in feature. It enables a "fail fast" mechanism by verifying resource capacity before deployment, without causing regressions or unexpected latency for existing blueprints.

Key Features

- Explicit Opt-in Design: Disabled by default. Only executes when explicitly listed in your blueprint YAML.
- Zero Regression for Existing Blueprints: No unexpected latency (1–3s) or permission blockers (e.g., compute.projects.get) for users who don't need real-time quota checks.
- Real-Time API Capability: Integrates with the Compute Engine Quotas API to aggregate and check module requirements.

How to Enable in Blueprint

```
validators:
- validator: test_quota_availability
  inputs:
    project_id: $(vars.project_id)
    region: $(vars.region) 
```    
 
Resource Coverage

- Compute (CPUs): Family-specific metrics (e.g., C3_CPUS, H100_CPUS).
- GPUs: Maps accelerator types (A100, H100, L4, etc.) to regional/global metrics (including GPUS_ALL_REGIONS).
- Storage: PD-Standard, SSD, Balanced, Extreme, and Hyperdisk Balanced (including IOPS/Throughput).
- Specialty services: Filestore capacity and TPU core requirements.
- Networks: NETWORKS and SUBNETWORKS global quotas.

Resiliency & Performance

- Exponential Backoff: Up to 5 retries for rate limits (429) or transient errors.
- Regional Caching: In-memory caching for projects and regions to minimize duplicate API calls.
- Unit Tests: Dedicated tests in pkg/validators/quota_test.go with mock GCP clients.

### DIFF
--- a/pkg/validators/quota.go
+++ b/pkg/validators/quota.go
@@ -1,0 +1,788 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validators
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"regexp"
+
+	"strconv"
+	"strings"
+	"sync"
+
+	"hpc-toolkit/pkg/config"
+	"hpc-toolkit/pkg/logging"
+
+	"math"
+	"time"
+
+	"github.com/zclconf/go-cty/cty"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/option"
+)
+
+const (
+	testQuotaAvailabilityName = "test_quota_availability"
+	localSSDSizeGB            = 375.0
+)
+
+var (
+	ErrUnknownValue        = errors.New("value is unknown")
+	machineTypeFamilyRegex = regexp.MustCompile(`^([a-z][0-9]+[a-z]?)-`)
+)
+
+type QuotaClient interface {
+	GetRegion(projectID, region string) (*compute.Region, error)
+	GetProject(projectID string) (*compute.Project, error)
+	GetMachineType(projectID, zone, machineType string) (*compute.MachineType, error)
+}
+
+type GCPQuotaClient struct {
+	svc          *compute.Service
+	regions      sync.Map
+	projects     sync.Map
+	machineTypes sync.Map
+}
+
+func NewGCPQuotaClient(ctx context.Context, projectID string) (*GCPQuotaClient, error) {
+	opts := []option.ClientOption{}
+	if projectID != "" {
+		opts = append(opts, option.WithQuotaProject(projectID))
+	}
+
+	svc, err := compute.NewService(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create compute service: %w", err)
+	}
+
+	return &GCPQuotaClient{svc: svc}, nil
+}
+
+func retryCall[T any](op func() (T, error)) (T, error) {
+	var res T
+	var err error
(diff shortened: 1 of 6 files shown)


pkg/validators/quota.go
```diff
@@ -0,0 +1,788 @@
...
+
+	var metricMap = map[string]string{
+		"h100": "NVIDIA_H100_GPUS",
+		"l4":   "NVIDIA_L4_GPUS",
+		"t4":   "NVIDIA_T4_GPUS",
+		"v100": "NVIDIA_V100_GPUS",
+		"p100": "NVIDIA_P100_GPUS",
+		"p4":   "NVIDIA_P4_GPUS",
+		"k80":  "NVIDIA_K80_GPUS",
+	}
```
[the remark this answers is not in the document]
kvenkatachala333: Handled

pkg/validators/quota.go
```diff
@@ -0,0 +1,788 @@
...
+		cores, errC := strconv.ParseFloat(coresStr, 64)
+		if errC == nil {
+			metric := fmt.Sprintf("%s_TPUS", strings.ToUpper(ver))
+
+			isTpuPreemptible := false
+			if settings.Has("preemptible") {
+				v, _ := evalBool(bp, settings.Get("preemptible"))
+				if v {
+					isTpuPreemptible = true
+				}
```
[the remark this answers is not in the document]
kvenkatachala333: Handled

kvenkatachala333:
gke-a3-highgpu, ml-a3-highgpu-slurm, gke-h4d >> Reservations have been removed from dev project, hence skipped
slurm-gcp-v6-rocky8, slurm6-tpu, gke-a4x >> couldn't find zone to deploy and daily tests have been disabled for the same

Hence skipped these 6 tests, rest all are successful

pkg/validators/quota.go
```diff
@@ -0,0 +1,791 @@
...
+
+const (
+	testQuotaAvailabilityName = "test_quota_availability"
+	localSSDSizeGB            = 375.0
+)
+
+var (
+	ErrUnknownValue        = errors.New("value is unknown")
+	machineTypeFamilyRegex = regexp.MustCompile(`^([a-z][0-9]+[a-z]?)-`)
+	acceleratorMetricMap   = map[string]string{
```
LAVEEN, line 47: why is this harded , why it doesnt cover all the families?

kvenkatachala333: explicitly map these common families because GCP Quota metric names are specific strings defined by the API and don't always follow a formula (for example, the A100 has a specific metric for its 80GB variant as seen in the special case above)

For any GPU family not listed in this map, the function actually falls through to a fallback mechanism at the bottom (lines 394–402). It automatically sanitizes the input and constructs the metric name in the standard NVIDIA_<FAMILY>_GPUS format. This ensures that future or less common GPU families are covered without requiring constant updates to the hardcoded map

pkg/validators/quota.go
```diff
@@ -0,0 +1,791 @@
...
+		"h100": "NVIDIA_H100_GPUS",
+		"l4":   "NVIDIA_L4_GPUS",
+		"t4":   "NVIDIA_T4_GPUS",
+		"v100": "NVIDIA_V100_GPUS",
+		"p100": "NVIDIA_P100_GPUS",
+		"p4":   "NVIDIA_P4_GPUS",
+		"k80":  "NVIDIA_K80_GPUS",
+	}
```
AdarshK15, line 55: Suggestion: Can we use compute API to fetch these dynamically instead of static content

saara-tyagi27 (approved):

pkg/validators/quota.go
kvenkatachala333: Thanks for the suggestion! However, the Compute API (like AcceleratorTypes) does not actually expose the associated Quota Metric strings (e.g., NVIDIA_L4_GPUS).

Used this static map to translate the Accelerator Type to the Quota Metric Name without requiring additional API latency or extra IAM permissions (like compute.acceleratorTypes.list).

For any new or unmapped GPUs, the function already has a string-transformation fallback built-in to handle them dynamically.

AdarshK15 (approved):

LAVEEN (approved):
LGTM.

PR #5452, opened 2026-04-03
Ensure fully qualified URLs for reservation subblocks

scaliby:
Fixes URI formatting when targeting reservation blocks/subblocks in the `gke-node-pool` module. 

When users provide a suffix in the `specific_reservations[0].name` field (e.g., `my-res/reservationBlocks/my-block`), the module previously dropped the suffix for inactive reservations, and failed to include the required `zone` in the URI for active ones. 

This PR updates `active_reservation_values` and `default_reservation_values` to properly construct the fully qualified `projects/{project}/zones/{zone}/reservations/{name}/reservationBlocks/{block}` URI whenever a suffix is detected.

### DIFF
--- a/modules/compute/gke-node-pool/reservation_definitions.tf
+++ b/modules/compute/gke-node-pool/reservation_definitions.tf
@@ -32,7 +32,7 @@ data "google_compute_reservation" "specific_reservations" {
     {} :
     {
       for pair in flatten([
-        for zone in try(var.zones, []) : [
+        for zone in(var.zones != null ? var.zones : []) : [
           for i, reservation_name in try(local.input_reservation_names, []) : {
             key : "${local.input_reservation_projects[i]}/${zone}/${reservation_name}"
             zone : zone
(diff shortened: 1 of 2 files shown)


modules/compute/gke-node-pool/reservation_definitions.tf
```diff
@@ -32,7 +32,7 @@ data "google_compute_reservation" "specific_reservations" {
     {} :
     {
       for pair in flatten([
-        for zone in try(var.zones, []) : [
+        for zone in(var.zones != null ? var.zones : []) : [
```
jamOne-, line 35: Is it different from the `try` that was here before?

jamOne- (approved):
Did you test it?

scaliby:
> Did you test it?

I did and it seems to be working.

modules/compute/gke-node-pool/reservation_definitions.tf
scaliby: Yes, the new check explicitly converts a `null` value into an empty list to prevent iteration errors, whereas the previous `try()` function allowed `null` to pass through and break the loop.

SwarnaBharathiMantena (approved):
LGTM!

The failing tests are not related. Two of the failing tests are spot consumption tests which are not related. One of them is due to expired reservation and two others failed due to nodepool getting created in ERROR state.

Neelabh94 (approved):

PR #5476, opened 2026-04-09
Integrate storage profile in GCSFuse

parulbajaj01:
This PR integrates GCS Fuse Storage Profiles directly into the GKE persistent volume modules. We are moving away from the older method of manually configuring GCS mounts on the node host (via pre-existing-network-storage) and switching to using Kubernetes-native StorageClasses like gcsfusecsi-training.

**What changed?**

**Module Updates**: Added the `gcsfuse_storage_class_name` setting into the gke-persistent-volume module. It automatically maps inputs into the PV/PVC specs cleanly.

**IAM Security**: Instead of trying to force Terraform to create global custom roles (which needs administrative permissions we don't want to grant automation runners), both the `gke.gcsfuse.profileUser` role creation and service account bindings are kept manual. I've added straightforward instructions in the module's README.

**Pre-flight Checks**: Adde Go validations that check if permissions are set up correctly before provisioning to avoid messy deployment failures.

**Testing**
Everything builds successfully locally.

If a user updates their blueprint and runs `gcluster deploy -w` on an existing cluster which earlier had the mount options config, the only thing in this case that the user might need to do is deleting the PV's and PVC's manually because the first time, the PVCs were created with an empty storageClassName: "" (or defaulting to standard). Now, the update will try to change that field to `gcsfusecsi-checkpointing` and `gcsfusecsi-training`. Kubernetes rejects the patch because it does not allow changing the storage class of an existing, bound PVC.
To apply the new storage profiles, user will need to delete the old PVCs and PVs so that Terraform can recreate them from scratch with the new settings. Since the data is stored externally in GCS buckets, deleting the Kubernetes volume objects will not delete the actual data in GCS. 

### Submission Checklist

### DIFF
--- a/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
+++ b/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
@@ -33,7 +33,7 @@ vars:
   system_node_pool_disk_size_gb: 200
   a3ultra_node_pool_disk_size_gb: 100
   accelerator_type: nvidia-h200-141gb
-  version_prefix: "1.33."
+  version_prefix: "1.35."
 
   # # To enable Managed-Lustre please uncomment this section and fill out the settings.
   # # Additionally, please uncomment the private_service_access, lustre_firewall_rule, managed-lustre and lustre-pv modules.
@@ -303,45 +303,23 @@ deployment_groups:
       k8s_service_account_name: workload-identity-k8s-sa
     outputs: [instructions]
 
-  # Create a remote mount of training_bucket using
-  # mount options optimized for reading training data.
-  # Based on Source of truth https://github.com/GoogleCloudPlatform/gcsfuse/blob/d1373b665b7f60e98856d2181f1193396ef16427/samples/gke-csi-yaml/gpu/training-pv.yaml#L15
-  # Some of the options might be available only on latest GKE version, please check the cluster version to meet the required version https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-perf
-  - id: gcs-training
-    source: modules/file-system/pre-existing-network-storage
-    settings:
-      remote_mount: $(training_bucket.gcs_bucket_name)
-      local_mount: /training-data
-      fs_type: gcsfuse
-      mount_options: "implicit-dirs,metadata-cache:ttl-secs:-1,metadata-cache:stat-cache-max-size-mb:-1,metadata-cache:type-cache-max-size-mb:-1,file-cache:max-size-mb:-1,file-cache:cache-file-for-range-read:true"
-
-  # Create a remote mount of checkpoint_bucket using mount
-  # options optimized for writing and reading checkpoint data.
-  # Based on Source of truth https://github.com/GoogleCloudPlatform/gcsfuse/blob/d1373b665b7f60e98856d2181f1193396ef16427/samples/gke-csi-yaml/gpu/checkpointing-pv.yaml#L15
-  # Some of the options might be available only on latest GKE version, please check the cluster version to meet the required version https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-perf
-  - id: gcs-checkpointing
-    source: modules/file-system/pre-existing-network-storage
-    settings:
-      remote_mount: $(checkpoint_bucket.gcs_bucket_name)
-      local_mount: /checkpoint-data
-      fs_type: gcsfuse
-      mount_options: "implicit-dirs,metadata-cache:ttl-secs:-1,metadata-cache:stat-cache-max-size-mb:-1,metadata-cache:type-cache-max-size-mb:-1,file-cache:max-size-mb:-1,file-cache:cache-file-for-range-read:true,file-cache:enable-parallel-downloads:true,rename-dir-limit=200000"
-
-  # Persistent Volume for training data
+  # Persistent Volume for training data using GCS Fuse Storage Profile
   - id: training-pv
     source: modules/file-system/gke-persistent-volume
-    use: [gcs-training, a3-ultragpu-cluster]
+    use: [training_bucket, a3-ultragpu-cluster]
     settings:
       gcs_bucket_name: $(training_bucket.gcs_bucket_name)
       capacity_gib: 1000000
+      gcsfuse_storage_class_name: gcsfusecsi-training
 
-  # Persistent Volume for checkpoint data
+  # Persistent Volume for checkpoint data using GCS Fuse Storage Profile
   - id: checkpointing-pv
     source: modules/file-system/gke-persistent-volume
-    use: [gcs-checkpointing, a3-ultragpu-cluster]
+    use: [checkpoint_bucket, a3-ultragpu-cluster]
     settings:
       gcs_bucket_name: $(checkpoint_bucket.gcs_bucket_name)
       capacity_gib: 1000000
+      gcsfuse_storage_class_name: gcsfusecsi-checkpointing
 
   # This is an example job that will install and run an `fio`
   # benchmark against the training and checkpointing buckets.
--- a/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
+++ b/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
@@ -303,6 +303,9 @@ deployment_groups:
       k8s_service_account_name: workload-identity-k8s-sa
     outputs: [instructions]
 
+  # When utilizing GCS Fuse Storage Profiles, you must ensure the GKE Service Agent
+  # possesses the 'gke.gcsfuse.profileUser' custom IAM role assigned out-of-band.
+  # Please refer to 'modules/file-system/gke-persistent-volume/README.md' for instructions.
   # Persistent Volume for training data using GCS Fuse Storage Profile
   - id: training-pv
     source: modules/file-system/gke-persistent-volume
--- a/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
+++ b/examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
@@ -308,7 +308,6 @@ deployment_groups:
     source: modules/file-system/gke-persistent-volume
     use: [training_bucket, a3-ultragpu-cluster]
     settings:
-      gcs_bucket_name: $(training_bucket.gcs_bucket_name)
       capacity_gib: 1000000
       gcsfuse_storage_class_name: gcsfusecsi-training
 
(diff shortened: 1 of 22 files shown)


examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
```diff
@@ -303,45 +303,23 @@ deployment_groups:
...
-  # Persistent Volume for training data
+  # Persistent Volume for training data using GCS Fuse Storage Profile
   - id: training-pv
     source: modules/file-system/gke-persistent-volume
-    use: [gcs-training, a3-ultragpu-cluster]
+    use: [training_bucket, a3-ultragpu-cluster]
     settings:
       gcs_bucket_name: $(training_bucket.gcs_bucket_name)
       capacity_gib: 1000000
+      gcsfuse_storage_class_name: gcsfusecsi-training
```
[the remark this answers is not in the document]
parulbajaj01: The training_bucket also outputs network_storage with the same structure.

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -77,10 +78,11 @@ locals {
   # Variables for PV templates, merging common vars with type-specific ones.
   pv_template_vars = {
     gcs = merge(local.common_pv_vars, {
-      mount_options = var.gcs_bucket_name != null ? split(",", var.network_storage.mount_options) : []
-      bucket_name   = var.gcs_bucket_name
-      namespace     = var.namespace
-      pvc_name      = local.pvc_name
+      mount_options      = var.gcs_bucket_name != null ? [for opt in split(",", var.network_storage.mount_options) : opt if !contains(["defaults", "_netdev", "implicit_dirs"], opt)] : []
```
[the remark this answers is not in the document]
gargnitingoogle: +1, `implcit_dirs` should be removed only if storage_class_name is set to one of the AI/ML storage class names.

gargnitingoogle: I am not clear why we were adding the `defaults` and `_netdev` options earlier, but not anymore, these have nothing to do with storage-class-name/storage-profiles.

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -153,3 +159,10 @@ resource "kubectl_manifest" "pvc" {
   yaml_body  = local.pvc_content
   depends_on = [kubectl_manifest.pv, kubectl_manifest.pvc_namespace]
 }
+
+resource "google_project_iam_member" "gcsfuse_agent_binding" {
+  count   = var.gcsfuse_storage_class_name != "" ? 1 : 0
+  project = split("/", var.cluster_id)[1]
```
gargnitingoogle: You could avoid the duplication of logic by using a local

```tf
locals {
  project_id = split("/", var.cluster_id)[1]
}
project = local.project_id
role = "projects/${local.project_id}/roles/gke.gcsfuse.profileUser"
```

gargnitingoogle (commented):
Minor comments other than whatever Gemini code assistant has commented. Otherwise LGTM. We should also confirm by testing these changes before going ahead with full-on review.

modules/file-system/gke-persistent-volume/main.tf
gargnitingoogle: Is this to do with the removal of the `gcs-training`, `gcs-checkpointing` resources ?

parulbajaj01: They are by default getting added to the mount options for gcsfuse - https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/file-system/cloud-storage-bucket/variables.tf#L43-L47

gargnitingoogle: @parulbajaj01  does that mean they were not needed earlier also ?

parulbajaj01: No they were being used earlier. But now since we are using storage profiles which is automatically managing them, so we don't need to specify them.

parulbajaj01: I’ve confirmed that while the `pre-existing-network-storage` module originally set mount_options to "defaults,_netdev", our blueprints were providing custom values that completely overwrote these defaults. Consequently, the original "defaults,_netdev" settings were never actually in effect.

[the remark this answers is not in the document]
parulbajaj01: done

examples/gke-a3-ultragpu/gke-a3-ultragpu.yaml
```diff
@@ -303,45 +303,23 @@ deployment_groups:
...
-      mount_options: "implicit-dirs,metadata-cache:ttl-secs:-1,metadata-cache:stat-cache-max-size-mb:-1,metadata-cache:type-cache-max-size-mb:-1,file-cache:max-size-mb:-1,file-cache:cache-file-for-range-read:true,file-cache:enable-parallel-downloads:true,rename-dir-limit=200000"
-
-  # Persistent Volume for training data
+  # Persistent Volume for training data using GCS Fuse Storage Profile
   - id: training-pv
     source: modules/file-system/gke-persistent-volume
-    use: [gcs-training, a3-ultragpu-cluster]
+    use: [training_bucket, a3-ultragpu-cluster]
     settings:
       gcs_bucket_name: $(training_bucket.gcs_bucket_name)
```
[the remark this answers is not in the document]
parulbajaj01: Fixed

modules/file-system/gke-persistent-volume/main.tf
parulbajaj01: Refactored.

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -153,3 +159,10 @@ resource "kubectl_manifest" "pvc" {
   yaml_body  = local.pvc_content
   depends_on = [kubectl_manifest.pv, kubectl_manifest.pvc_namespace]
 }
+
+resource "google_project_iam_member" "gcsfuse_agent_binding" {
+  count   = var.gcsfuse_storage_class_name != "" ? 1 : 0
```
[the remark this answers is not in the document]
parulbajaj01: Done

modules/file-system/gke-persistent-volume/main.tf
```diff
@@ -77,10 +78,11 @@ locals {
   # Variables for PV templates, merging common vars with type-specific ones.
   pv_template_vars = {
     gcs = merge(local.common_pv_vars, {
-      mount_options = var.gcs_bucket_name != null ? split(",", var.network_storage.mount_options) : []
-      bucket_name   = var.gcs_bucket_name
-      namespace     = var.namespace
-      pvc_name      = local.pvc_name
+      mount_options      = var.gcs_bucket_name != null ? [for opt in split(",", var.network_storage.mount_options) : opt if !contains(["defaults", "_netdev"], opt) && !(opt == "implicit_dirs" && var.gcsfuse_storage_class_name != "" && var.gcsfuse_storage_class_name != null)] : []
```
vikramvs-gg: The current line combines multiple conditions, a loop, and a fallback in a single expression which is complex to read. Lets make it more readable using `locals`. Maybe something like the below:
```
locals {
  # 1. Check if a GCS Fuse storage profile is active
  has_gcsfuse_storage_profile = var.gcsfuse_storage_class_name != "" && var.gcsfuse_storage_class_name != null

  # 2. Extract mount options as a list (fallback to empty list if not GCS)
  gcs_mount_opts_raw = var.gcs_bucket_name != null ? split(",", var.network_storage.mount_options) : []

  # 3. Apply filters to clean up the options
  gcs_mount_options = [
    for opt in local.gcs_mount_opts_raw : opt
    if !contains(["defaults", "_netdev"], opt) &&                       # Filter out defaults
       !(opt == "implicit_dirs" && local.has_gcsfuse_storage_profile)   # Filter implicit_dirs if profile is active
  ]
}
```

Then, in pv_template_vars map, we can simply refer gcs_mount_options:
```
  pv_template_vars = {
    gcs = merge(local.common_pv_vars, {
      mount_options      = local.gcs_mount_options
      bucket_name        = var.gcs_bucket_name
      namespace          = var.namespace
      pvc_name           = local.pvc_name
      storage_class_name = var.gcsfuse_storage_class_name
    })
    # ... lustre and filestore sections
  }
```

pkg/validators/cloud.go
```diff
@@ -495,3 +498,86 @@ func testMachineTypeInZoneAvailability(bp config.Blueprint, inputs config.Dict)
...
+	if err != nil {
+		logging.Error("WARNING: Could not fetch IAM policy for %s: %v. Skipping IAM binding check.", projectID, err)
+		return
+	}
+
+	agentMember := fmt.Sprintf("serviceAccount:service-%d@container-engine-robot.iam.gserviceaccount.com", project.ProjectNumber)
+	roleGranted := false
+
+	for _, binding := range policy.Bindings {
+		if binding.Role == roleName || binding.Role == fmt.Sprintf("projects/%s/roles/gke.gcsfuse.profileUser", projectID) {
```
vikramvs-gg: The roleName is defined as same value in the `testGCSFuseIAMRoleExistsCheck` function from where this `checkGCSFuseIAMBinding` is being called, hence this OR condition seems to be redundant.

pkg/validators/cloud.go
```diff
@@ -495,3 +498,86 @@ func testMachineTypeInZoneAvailability(bp config.Blueprint, inputs config.Dict)
...
+		return handleClientError(err)
+	}
+
+	roleName := fmt.Sprintf("projects/%s/roles/gke.gcsfuse.profileUser", projectID)
+	_, err = s.Projects.Roles.Get(roleName).Do()
+	if err != nil {
+		var herr *googleapi.Error
+		if errors.As(err, &herr) && herr.Code == 404 {
+			return config.HintError{
+				Err: fmt.Errorf("custom IAM role 'gke.gcsfuse.profileUser' was not found in project %s", projectID),
```
vikramvs-gg: nit: `gke.gcsfuse.profileUser` is used at multiple places in the function. We can define a const for string.

pkg/validators/cloud.go
```diff
@@ -495,3 +498,86 @@ func testMachineTypeInZoneAvailability(bp config.Blueprint, inputs config.Dict)
 func testDiskTypeInZoneAvailability(bp config.Blueprint, inputs config.Dict) error {
 	return testResourceInZoneAvailability(bp, inputs, "test_disk_type_in_zone", "disk_type", "disk type", validateDiskTypeInZone)
 }
+
+// testGCSFuseIAMRoleExists checks if the custom IAM role 'gke.gcsfuse.profileUser' exists.
```
vikramvs-gg: There are some hard failures (like the 404 role not found) and there are soft failures (like permission), we can add comment over the function describing high level logic and why some failures are soft while others not would improve maintainability of code.

modules/file-system/gke-persistent-volume/main.tf
parulbajaj01: done

vikramvs-gg (approved):

PR #5484, opened 2026-04-13
Update base image driver to 580 in a3m and a4x blueprint

saara-tyagi27:
### Summary

This pull request migrates the A3 Mega and A4x high GPU infrastructures to Ubuntu 24.04 and NVIDIA driver 580.

### Key changes

- **Base Image Upgrade:** Updated the base image for A3 Mega GPU configurations from Ubuntu 22.04 to Ubuntu 24.04.
- **NVIDIA Driver and Toolkit Updates:** Upgraded NVIDIA components to version 580 and updated the CUDA toolkit installation to version 13.0.
- **Compatibility Fixes:** Added shell scripts to handle Enroot environment requirements, patched the DMABUF import helper for kernel 6.17 compatibility, and implemented a mechanism to signal early completion of the packer process.

### DIFF
--- a/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
@@ -35,7 +35,7 @@ vars:
   enable_login_public_ips: true
   enable_controller_public_ips: true
   localssd_mountpoint: /mnt/localssd
-  build_slurm_from_git_ref: 6.10.10
+  build_slurm_from_git_ref: 6.12.1
   #Provisioning models (set to true or fill in reservation name, pick only one)
   a3mega_reservation_name: ""  # supply reservation name
   a3mega_dws_flex_enabled: false
--- a/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
@@ -26,7 +26,7 @@ vars:
   sys_net_range: 172.16.0.0/16
   base_image:
     project: ubuntu-os-accelerator-images
-    image: ubuntu-accelerator-2204-amd64-with-nvidia-570-v20260210
+    image: ubuntu-accelerator-2404-amd64-with-nvidia-580-v20260225
   a3mega_partition_name: a3mega
   local_mount_homefs: /home
   instance_image:
@@ -83,6 +83,12 @@ deployment_groups:
         enabled: true
         world_writable: true
       runners:
+      - type: data
+        destination: /etc/apt/preferences.d/block-broken-nvidia-container
+        content: |
+          Package: nvidia-container-toolkit nvidia-container-toolkit-base libnvidia-container-tools libnvidia-container1
+          Pin: version 1.17.7-1
+          Pin-Priority: 100
 
       # The following holds NVIDIA software that was already installed on the
       # accelerator base image to be the same driver version. This reduces the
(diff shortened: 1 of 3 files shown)


examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -153,46 +159,24 @@ deployment_groups:
...
-              nvidia-container-toolkit \
-              datacenter-gpu-manager-4-cuda12=1:4.5.2-1 \
-              datacenter-gpu-manager-4-dev=1:4.5.2-1 \
-              datacenter-gpu-manager-4-core=1:4.5.2-1
+           wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/cuda-keyring_1.1-1_all.deb -O /tmp/cuda-keyring.deb
+           dpkg -i /tmp/cuda-keyring.deb
+           apt update -y
+           apt install -y \
+               cuda-toolkit-13-0 \
+               nvidia-container-toolkit \
```
LAVEEN, line 172: doesn't this pin dcgmi version ?

saara-tyagi27: This pins the installation to the CUDA 13 branch (so it pulls the latest DCGM version compatible with CUDA 13), but it removes the specific version pin (like 4.5.2-1) we had previously.

Dcgm version pinning will be taken up in upcoming prs.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -35,7 +35,7 @@ vars:
   enable_login_public_ips: true
   enable_controller_public_ips: true
   localssd_mountpoint: /mnt/localssd
-  build_slurm_from_git_ref: 6.12.1
+  build_slurm_from_git_ref: 6.10.10
```
LAVEEN: with the image update, why there is a downgrade of slurm version?

saara-tyagi27: Thankyou for pointing this out, this change was left during testing. Reverted it.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -263,31 +262,44 @@ deployment_groups:
                 state: stopped
                 enabled: false
 
-      - type: ansible-local
-        destination: install_dmabuf.yml
+      - type: shell
```
samskillman: Why are we switching this to a shell script?

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -153,46 +159,24 @@ deployment_groups:
               -i localhost, --limit localhost --connection=local \
               -e @/var/tmp/slurm_vars.json \
               ansible/playbook.yml
-      - type: ansible-local
```
samskillman, line 169: Please confirm that this can be removed and that a gve-dkms package exists already in the image.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -341,6 +353,14 @@ deployment_groups:
           apt-get install --assume-yes google-cloud-cli
           # Clean up the bash executable hash for subsequent steps using gsutil
           hash -r
+      - type: shell
```
samskillman: Is this because of the hostname issues? Let's use the cloud-config setting instead to write / set up hostname with something like this in the builder vm metadata

user-data: |
  #cloud-config
  prefer_fqdn_over_hostname: false
  preserve_hostname: true
  create_hostname_file: true

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
saara-tyagi27: I confirmed the gve driver is built-in (intree: Y) on both images. I wonder if the install step was originally there because older kernels needed it for missing features.

Since kernel 6.17 likely (assuming here) has these features built-in and the old 1.4.3 package fails to compile on this new kernel, I propose removing the step. Does that sound correct, or was there a specific reason for version 1.4.3 I'm missing?

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
saara-tyagi27: The script manually sets the status to 'done' to prevent Packer from hanging and timing out at the end of the build. We did encounter a secondary bug where the FQDN broke the gcloud call inside it, so I will incorporate your cloud-config suggestion to handle hostnames robustly.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
saara-tyagi27: I confirmed the gve packages are present.

saara-tyagi27:
A4x test cannot be run due to reservation not available. To handle that the testing has been done manually.

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
```diff
@@ -385,6 +452,7 @@ deployment_groups:
       metadata:
         user-data: |
           #cloud-config
+          create_hostname_file: true
```
LAVEEN, line 477: please check if we need this?

L:224 line covers this i believe

LAVEEN (dismissed):
LGTM

examples/machine-learning/a3-megagpu-8g/a3mega-slurm-blueprint.yaml
saara-tyagi27: Good catch, Removed the duplication on line 224.

LAVEEN (approved):

PR #5493, opened 2026-04-14
Support Compute Endpoint Override for Slurm image building and cluster deployment

AdarshK15:
### Summary
This PR enables Cluster Toolkit to provide custom compute endpoint overrides for Slurm image building and cluster deployment. 

**Packer Upgrade:** I upgraded the googlecompute Packer plugin to ~> 1.2.5, as older versions don't support the custom_endpoints variable.
**Gcloud override & Compute endpoint Variables:** I added gcloud_path_override and compute_endpoint_version as input variables to both the Packer and startup-script modules. This lets us pass these values directly from the blueprint.
**Startup Script Wrapper:** I added a gcloud wrapper in the startup script to set the custom compute endpoint. This ensures that when ansible-pull runs the slurm-gcp playbook to install Slurm, all underlying gcloud commands correctly use the provided override.
**Customizable Licenses:** I changed image_licenses field into a variable, to allow it to be overridden as the default production license URL may not be available.
**Explicit Region:** Introduced an explicit `region` variable for the `googlecompute` builder. Without this change, the logic truncated the zone name to infer the region name, which failed for zones that did not follow the `region-[a-z]` naming convention.

### Testing
I verified these changes by building and deploying a4high, a4x, and g4 clusters. I also verified Rocky Linux 8 image building and cluster creation using the hpc-build-slurm-image.yaml blueprint.

### DIFF
--- a/modules/packer/custom-image/README.md
+++ b/modules/packer/custom-image/README.md
@@ -257,16 +257,19 @@ No resources.
 ## Inputs
 
 | Name | Description | Type | Default | Required |
-| ---- | ----------- | ---- | ------- | :------: |
+|------|-------------|------|---------|:--------:|
 | <a name="input_accelerator_count"></a> [accelerator\_count](#input\_accelerator\_count) | Number of accelerator cards to attach to the VM; not necessary for families that always include GPUs (A2). | `number` | `null` | no |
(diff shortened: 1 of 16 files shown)


LAVEEN (approved):
LGTM

PR #5500, opened 2026-04-14
feat(kueue): add declarative super-slicing operator injection and resource limit overrides

jamOne-:
### What
- **Super-Slicing Operator Injection:** Added declarative Terraform configuration to seamlessly provision a Kueue cluster ready for super-slicing workloads (matching the `xpk` setup) via the "pathways approach". 
- **Dynamic CRD Linking:** Refactored deeply nested logic inside `modules/management/kubectl-apply/main.tf` to elegantly parse, extract, and merge `admissionChecks` and `resourceGroups` dynamically across multiple `ClusterQueue` definitions of the same name.
- **Resource and Replica Configs:** Introduced customizable replica counts (`controller_replicas`) and resource limits/requests (`controller_cpu`, `controller_memory`) for both the Kueue and Jobset controllers. 
- **XPK Parity Defaults:** If `enable_slice_controller` is set to true and no manual overrides are provided, both controllers will automatically scale and size to match standard `xpk` benchmarks (Kueue: 3 Replicas, 16 CPU / 64Gi | Jobset: 4 CPU / 16Gi).
- **CRD Buffer:** Implemented a 15-second `time_sleep` buffer (`wait_for_kueue_crd`) immediately after installing Kueue to ensure the Kubernetes API server has fully registered the new Custom Resource Definitions before attempting to create instances of them.

### Why
- To support configuring super-slicing seamlessly and robustly inside the cluster-toolkit without relying on `local-exec` bash patching. 
- The 15-second CRD registration wait resolves a known race condition flakiness where the API server would reject instances of newly installed CRDs (like `Topology` and `ClusterQueue`) if the Helm installation completed extremely quickly.

### Submission Checklist
- [x] Test all changes with pre-commit in a local branch
- [x] Add or modify unit tests to cover code changes *(N/A - HCL declarative logic)*
- [x] Follow Cluster Toolkit Contribution guidelines


jamOne-:
@SwarnaBharathiMantena please run the tests for this PR 🙏

modules/management/kubectl-apply/variables.tf
```diff
@@ -120,12 +120,16 @@ variable "apply_manifests" {
 
 
 variable "kueue" {
-  description = "Install and configure [Kueue](https://kueue.sigs.k8s.io/docs/overview/) workload scheduler. A configuration yaml/template file can be provided with config_path to be applied right after kueue installation. If a template file provided, its variables can be set to config_template_vars."
+  description = "Install and configure [Kueue](https://kueue.sigs.k8s.io/docs/overview/) workload scheduler. A configuration yaml/template file can be provided with config_path to be applied right after kueue installation. If a template file provided, its variables can be set to config_template_vars. Additional fields: enable_slice_controller sets up super-slicing; controller_replicas, controller_cpu, and controller_memory allow overriding manager resources."
   type = object({
     install                  = optional(bool, false)
     version                  = optional(string, "0.13.3")
```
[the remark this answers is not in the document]
jamOne-: I think that's ignorable, isn't it? Super slicing will have a special blueprint anyway

modules/management/kubectl-apply/kueue/super-slicing.yaml.tftpl
```diff
@@ -0,0 +1,24 @@
+apiVersion: kueue.x-k8s.io/v1beta1
+kind: Topology
+metadata:
+  name: ${try(super_slice_topology_name, "super-slice-topology")}
```
SwarnaBharathiMantena, line 4: Do we expect the topology to change for a new TPU family? Is the `nodeLabel: cloud.google.com/gke-tpu-partition-4x4x4-id` TPU 7x specific?

modules/management/kubectl-apply/main.tf
```diff
@@ -19,8 +19,36 @@ locals {
...
+  install_gib               = try(var.gib.install, false)
+  install_asapd_lite        = try(var.asapd_lite.install, false)
+
+  enable_slice_controller      = var.kueue.enable_slice_controller
+  kueue_super_slicing_cpu      = "16"
+  kueue_super_slicing_memory   = "64Gi"
+  kueue_super_slicing_replicas = 3
+  jobset_super_slicing_cpu     = "4"
+  jobset_super_slicing_memory  = "16Gi"
+  kueue_cpu                    = var.kueue.controller_cpu != null ? var.kueue.controller_cpu : (local.enable_slice_controller ? local.kueue_super_slicing_cpu : null)
```
SwarnaBharathiMantena, line 37: This logic is going to intake `var.kueue.controller_cpu` value even when the value is less than `16` which is the `kueue_super_slicing_cpu` value.

Is that expected? Is it better to add a validation instead and let the user know?

This applies to all the superslicing related config/variables here.

modules/management/kubectl-apply/main.tf
```diff
@@ -47,16 +75,26 @@ locals {
   cluster_queues        = [for doc in local.parsed_kueue_docs : doc if try(doc.kind, "") == "ClusterQueue"]
   other_docs            = [for doc in local.parsed_kueue_docs : doc if try(doc.kind, "") != "ClusterQueue"]
   merged_cluster_queues = { for cq in local.cluster_queues : cq.metadata.name => cq... }
+  merged_cluster_queues_with_lists = {
```
SwarnaBharathiMantena: There are too many intermediate variables on this logic.

Can we please simplify this?

One suggestion:
```
locals {
  # 1. Combine splitting and decoding into one step
  kueue_parsed = [for doc in split("\n---", local.kueue_config_content) : yamldecode(doc) if trimspace(doc) != ""]

  # 2. Group ClusterQueues by name directly from the parsed list
  cluster_queue_groups = { for doc in local.kueue_parsed : doc.metadata.name => doc... if try(doc.kind, "") == "ClusterQueue" }

  # 3. Merge ClusterQueues by inheriting cqs[0] and overwriting only the spec
  final_cluster_queues = [
    for name, cqs in local.cluster_queue_groups : merge(cqs[0], {
      spec = merge(
        try(cqs[0].spec, {}),
        { resourceGroups = flatten([for cq in cqs : try(cq.spec.resourceGroups, [])]) },
        # Aggregated admissionChecks logic
        length(distinct(compact(flatten([for cq in cqs : try(cq.spec.admissionChecks, [])])))) > 0 ? {
          admissionChecks = distinct(compact(flatten([for cq in cqs : try(cq.spec.admissionChecks, [])])))
        } : {}
      )
    })
  ]

  # 4. Construct final manifest list
  final_kueue_manifests = concat(
    [for doc in local.kueue_parsed : yamlencode(doc) if try(doc.kind, "") != "ClusterQueue"],
    [for doc in local.final_cluster_queues : yamlencode(doc)]
  )
}
```

SwarnaBharathiMantena: This edits lines 73-101.

modules/management/kubectl-apply/main.tf
```diff
@@ -182,15 +231,47 @@ module "install_kueue" {
...
 
   dependencies = var.system_node_pool_id != null ? [var.system_node_pool_id] : []
 
   depends_on = [var.gke_cluster_exists]
 }
 
+# Sleep for 15 seconds after Kueue installation to allow the Kubernetes API server
+# enough time to fully register the newly created Custom Resource Definitions (CRDs)
+# before we attempt to create instances of them (e.g. ClusterQueue, Topology).
+resource "time_sleep" "wait_for_kueue_crd" {
```
SwarnaBharathiMantena: @agrawalkhushi18 can you please review this code. Does helm usage overcome this requirement?

SwarnaBharathiMantena (changes requested):

modules/management/kubectl-apply/main.tf
agrawalkhushi18: Ideally, the dependency of `configure_kueue` on `install_kueue` as per the previous code should work as intended allowing the CRDs to register before the CRs get created since we have set `wait=true` in install_kueue.

modules/management/kubectl-apply/kueue/super-slicing.yaml.tftpl
jamOne-: I assume it won't change for the new TPU family that supports super-slicing. This is more of a super-slicing specific label, as it only supports joining 4x4x4 cubes.

modules/management/kubectl-apply/main.tf
jamOne-: My reasoning was that if a user wants to use a different value, we should allow that.

Is it doable (and a good practice) to print a warning maybe? E.g. "Kueue controller CPU is lower than the recommended default for super-slicing."?

modules/management/kubectl-apply/main.tf
jamOne-: OK, I'll revert this change.

modules/management/kubectl-apply/main.tf
jamOne-: Done, thanks.

jamOne-:
I'll add a few more validations to the PR. Moving to draft.

PR #5508, opened 2026-04-16
Update Toolkit release to v1.89.0

siddhartha-quad:
Set release-candidate to version 1.89.0


siddhartha-quad:
raised for testing purpose

PR #5519, opened 2026-04-17
[Telemetry] Add support to collect the Billing Account ID metric

kadupoornima:
This pull request introduces the **capability to capture and report the GCP billing account ID associated with a project** within the Cluster Toolkit Telemetry system. This change involves integrating the Google Cloud Billing API, updating the telemetry metadata structure, and ensuring the new data point is correctly extracted and included in event reporting.

Changes:

* **Dependency Update**: Added the `cloud.google.com/go/billing` library to the project dependencies.
* **Telemetry Enhancement**: Implemented functionality to fetch and include the GCP billing account ID in telemetry data.
* **Refactoring**: Updated the blueprint key retrieval logic and added unit tests for the new billing account ID extraction.

**KINDLY NOTE THAT TELEMETRY DATA IS NOT BEING COLLECTED YET.**


pkg/telemetry/collector.go
```diff
@@ -163,6 +166,18 @@ func getOSVersion() string {
 	}
 }
 
+func getBillingAccountId(bp config.Blueprint) string {
+	projectID := getKeyFromBlueprint("project_id", bp)
+	if projectID != "" {
+		ctx := context.Background()
+		billingAccount := getProjectBillingAccount(ctx, projectID)
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/telemetry/collector.go
SwarnaBharathiMantena, line 51: `BuildConcordEvent` uses a mutex (c.mu) to protect the map during reads. Can we ensure that the  `CollectMetrics` uses it during writes? At this time, this could cause a data race if telemetry collection ever runs concurrently with event building. To avoid this, we can wrap the map assignments in CollectMetrics with c.mu.Lock() and c.mu.Unlock().

pkg/telemetry/collector_util.go
```diff
@@ -17,13 +17,18 @@ package telemetry
...
 	"hpc-toolkit/pkg/config"
 	"os"
 	"os/exec"
 	"regexp"
+
 	"strings"
 	"time"
 
+	billing "cloud.google.com/go/billing/apiv1"
+	"cloud.google.com/go/billing/apiv1/billingpb"
```
SwarnaBharathiMantena, line 27: Styling: Re-arrange this so that the Google Go Style Guide is followed, by grouping standard library imports separate from external third-party libraries. Currently, the new imports are interspersed with existing ones like `fmt` and `strings`.

pkg/telemetry/collector_util.go
```diff
@@ -82,14 +87,30 @@ func ifModulesMatchPatterns(modulesList []string, patterns []string) string {
...
+		v, _ := val.Unmark()
+		if !v.IsNull() && v.Type() == cty.String {
+			return v.AsString()
+		}
+	}
+	return ""
+}
+
+// getProjectBillingAccount fetches the billing account associated with a given GCP project in the format "billingAccounts/{billing_account_id}". If billing is disabled for the project, this will return an empty string.
+var getProjectBillingAccount = func(ctx context.Context, projectID string) string {
```
SwarnaBharathiMantena, line 100: Let us ensure that returning an empty string on all errors is consistent with how other metrics (like Region or Machine Type) should behave if their keys are missing or malformed.
(This might already be the case. Please double-check.)

SwarnaBharathiMantena (commented):
Should this PR also include an update to the README file of Telemetry?

pkg/telemetry/collector_util.go
kadupoornima: Done.

pkg/telemetry/collector_util.go
kadupoornima: Yes, an empty string is returned, similar to the other metrics.

pkg/telemetry/collector.go
kadupoornima: We already use the mutex in CollectMetrics method also.

kadupoornima:
> Should this PR also include an update to the README file of Telemetry?

Not required, this is just another metric.

SwarnaBharathiMantena (approved):
LGTM!

PR #5559, opened 2026-04-28
adding new global static ip module.

arpit974:
Adding a new module for provisioning Google Cloud Global Static IP addresses.

This module allows for the allocation of named global static IPs using Terraform's for_each pattern. It provides a standard way to manage static IP resources required by Global Load Balancers across different environments.

### Submission Checklist

### DIFF
--- a/modules/network/global-static-ip/main.tf
+++ b/modules/network/global-static-ip/main.tf
@@ -1,0 +1,19 @@
+# Copyright 2026 "Google LLC"
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "google_compute_global_address" "ips" {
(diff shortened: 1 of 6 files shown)


modules/network/global-static-ip/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+variable "project_id" {
+  description = "GCP Project ID"
+  type        = string
+}
+variable "ip_names" {
+  description = "List of global static IP names to create"
+  type        = list(string)
```
kvenkatachala333: Since ip_names is used in a for_each loop and requires unique values, can you please consider changing the type from list(string) to set(string). This makes the requirement for unique names explicit at the input level.

modules/network/global-static-ip/main.tf
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "google_compute_global_address" "ips" {
+  for_each = toset(var.ip_names)
+  name     = each.value
+  project  = var.project_id
+}
```
kvenkatachala333, line 26: Is it possible to consider adding an optional labels variable (type map(string)) and applying it to the google_compute_global_address resource to allow users to label these resources?

kvenkatachala333: Also, how about adding an optional description field for the google_compute_global_address resource would allow users to provide context for what each IP is being used for?

kvenkatachala333 (approved):

PR #5570, opened 2026-04-29
[Telemetry] Refactor getModules method to use cached standard modules from firestore

kadupoornima:
This pull request refactors the telemetry module retrieval process to rely on **cached data from Firestore for getting pre-defined (standard) modules** rather than performing live network requests to the GitHub API. This change improves the reliability and speed of the telemetry collection process by reducing external dependencies and leveraging pre-cached release metadata.

Changes:

* **Firestore Integration**: Refactored module retrieval to fetch standard modules from Firestore instead of querying the GitHub API directly.
* **Code Cleanup**: Removed obsolete GitHub API client code and associated tests in the config and telemetry packages.
* **Metadata Caching**: Implemented a new mechanism to load release metadata into the Viper registry for improved performance and reliability.

### DIFF
--- a/pkg/telemetry/collector_util.go
+++ b/pkg/telemetry/collector_util.go
@@ -89,12 +89,12 @@ func loadReleaseMetadata() error {
 }
 
 // getStandardModules returns the cached list of official modules from Viper
-func getStandardModules() []string {
+var getStandardModules = func() []string {
 	err := loadReleaseMetadata() // Will move this function call within the root persistent pre-run in future PR to avoid calling multiple times.
-	if err != nil {
-		return []string{}
+	if err == nil && viper.IsSet("standard_modules") {
+		return viper.GetStringSlice("standard_modules")
 	}
-	return viper.GetStringSlice("standard_modules")
+	return []string{}
 }
 
 func getBpModulesList(bp config.Blueprint) []string {
--- a/pkg/telemetry/collector_util.go
+++ b/pkg/telemetry/collector_util.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"cloud.google.com/go/billing/apiv1/billingpb"
+	"github.com/spf13/viper"
 	"github.com/zclconf/go-cty/cty"
 
 	billing "cloud.google.com/go/billing/apiv1"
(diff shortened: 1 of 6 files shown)


pkg/telemetry/collector_util.go
```diff
@@ -52,6 +53,50 @@ func getEventMetadataKVPairs(sourceMetadata map[string]string) []map[string]stri
...
+}
+
+// getStandardModules returns the cached list of official modules from Viper
+var getStandardModules = func() []string {
+	err := loadReleaseMetadata() // Will move this function call within the root persistent pre-run in future PR to avoid calling multiple times.
+	if err == nil && viper.IsSet("standard_modules") {
+		return viper.GetStringSlice("standard_modules")
+	}
+	return []string{}
+}
```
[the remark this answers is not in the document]
kadupoornima: Will move this function call within the root persistent pre-run in future PR to avoid calling multiple times. Also, the loadReleaseMetadata() method already has a timeout implemented.

pkg/telemetry/collector_util.go
```diff
@@ -52,6 +53,50 @@ func getEventMetadataKVPairs(sourceMetadata map[string]string) []map[string]stri
...
+	client, err := config.GetFirestoreClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get firestore client: %v", err)
+	}
+
+	// Try to fetch the cached metadata document for this version
+	doc, err := client.Collection(releasesCollection).Doc(version).Get(ctx)
+	if err != nil {
+		// Document not found or network error.
+		// Here we would trigger the fallback to fetchGitTree() and the GitHub API.
```
SwarnaBharathiMantena, line 81: Will this be implemented in a followup PR?

SwarnaBharathiMantena (approved):
LGTM!

pkg/telemetry/collector_util.go
kadupoornima: Already implemented here - https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/tools/cache_metadata/main.go#L96

PR #5590, opened 2026-05-03
Offload hardware resolution for job submission to hardware.go module

Neelabh94:
This pull request refactors the hardware resolution logic by centralizing it within the pkg/config module. By offloading these responsibilities from the GKE orchestrator and replacing external CLI dependencies with direct API calls (with caching).

This pull request also renames the --accelerator flag to --compute-type and --max-restarts to --restarts. These are just naming changes and the functionality of these flags have not been changed.

### Submission Checklist

### DIFF
--- a/cmd/job/submit.go
+++ b/cmd/job/submit.go
@@ -16,6 +16,7 @@ package job
 
 import (
 	"fmt"
+	"hpc-toolkit/pkg/config"
 	"os"
 	"slices"
 	"time"
@@ -28,18 +29,18 @@ import (
 )
 
 var (
-	imageName       string
-	baseImage       string
-	buildContext    string
-	commandToRun    string
-	acceleratorType string
-	dryRunManifest  string
+	imageName      string
+	baseImage      string
+	buildContext   string
+	commandToRun   string
+	computeType    string
+	dryRunManifest string
 
 	workloadName     string
 	kueueQueueName   string
 	numSlicesOrNodes int
 	vmsPerSlice      int
-	maxRestarts      int
+	restarts         int
 	ttlAfterFinished string
 	gracePeriodStr   string
 
@@ -104,7 +105,7 @@ func init() {
 	SubmitCmd.Flags().StringVarP(&baseImage, "base-image", "B", "", "Name of the base image for Crane to build upon (e.g., python:3.9-slim). Requires --build-context.")
 	SubmitCmd.Flags().StringVarP(&buildContext, "build-context", "b", "", "Path to the build context directory for Crane (e.g., .). Required with --base-image.")
 	SubmitCmd.Flags().StringVarP(&commandToRun, "command", "e", "", "Command to execute in the container (e.g., 'python train.py'). Required.")
-	SubmitCmd.Flags().StringVarP(&acceleratorType, "accelerator", "a", "", "Type of accelerator to request (e.g., 'nvidia-tesla-a100').")
+	SubmitCmd.Flags().StringVar(&computeType, "compute-type", "", "Type of compute to request (e.g., 'n2-standard-32', 'nvidia-l4', 'v6e-8').")
 	SubmitCmd.Flags().StringVarP(&dryRunManifest, "dry-run-out", "o", "", "Path to output the generated Kubernetes manifest instead of applying it.")
 	SubmitCmd.Flags().StringVarP(&platform, "platform", "f", "linux/amd64", "Target platform for the image build (e.g., 'linux/amd64', 'linux/arm64'). Used with --base-image.")
 
@@ -112,7 +113,7 @@ func init() {
 	SubmitCmd.Flags().StringVarP(&kueueQueueName, "queue", "q", "", "Name of the Kueue LocalQueue to submit the workload to. If empty, it will be auto-discovered.")
 	SubmitCmd.Flags().IntVar(&numSlicesOrNodes, "nodes", 1, "Number of JobSet replicas (or Slices for TPUs).")
 	SubmitCmd.Flags().IntVar(&vmsPerSlice, "vms-per-slice", 0, "Number of VMs (pods) per slice. Defaults to auto-calculated value for TPUs.")
-	SubmitCmd.Flags().IntVar(&maxRestarts, "max-restarts", 1, "Maximum number of restarts for the JobSet before failing.")
+	SubmitCmd.Flags().IntVar(&restarts, "restarts", 1, "Maximum number of restarts for the JobSet before failing.")
 	SubmitCmd.Flags().StringVar(&ttlAfterFinished, "gke-ttl-after-finished", "1h", "Time to retain the JobSet after it finishes (e.g. 5m, 1h).")
 	SubmitCmd.Flags().StringVar(&gracePeriodStr, "grace-period", "30s", "Time to wait before forcefully terminating a pod (e.g. 30s, 2m). Gives the workload time to save checkpoints or clean up distributed state during cancellation or preemption events (like Spot VM evictions).")
 
@@ -147,10 +148,13 @@ func init() {
 
 	_ = SubmitCmd.MarkFlagRequired("command")
 	_ = SubmitCmd.MarkFlagRequired("name")
-	_ = SubmitCmd.MarkFlagRequired("accelerator")
+	_ = SubmitCmd.MarkFlagRequired("compute-type")
 }
 
 func runSubmitCmd(cmd *cobra.Command, args []string) error {
+	if err := config.ValidateHardwareRequest(computeType, topology, placementPolicy); err != nil {
+		return err
+	}
 	ttlSeconds, err := parseDurationToSeconds(ttlAfterFinished, "--gke-ttl-after-finished")
 	if err != nil {
 		return err
@@ -181,7 +185,7 @@ func runSubmitCmd(cmd *cobra.Command, args []string) error {
 		BuildContext:                  buildContext,
 		Platform:                      platform,
 		CommandToRun:                  commandToRun,
-		AcceleratorType:               acceleratorType,
+		AcceleratorType:               computeType,
 		DryRunManifest:                dryRunManifest,
 		ProjectID:                     projectID,
 		ClusterName:                   clusterName,
@@ -190,7 +194,7 @@ func runSubmitCmd(cmd *cobra.Command, args []string) error {
 		KueueQueueName:                kueueQueueName,
 		NumSlices:                     numSlicesOrNodes,
 		VmsPerSlice:                   vmsPerSlice,
-		MaxRestarts:                   maxRestarts,
+		MaxRestarts:                   restarts,
 		TtlSecondsAfterFinished:       ttlSeconds,
 		TerminationGracePeriodSeconds: gracePeriodSeconds,
 		PlacementPolicy:               placementPolicy,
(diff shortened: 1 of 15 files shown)


pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -269,13 +212,243 @@ func (g *GKEOrchestrator) resolveMachineName(acceleratorType string) string {
...
+	if err != nil {
+		return err
+	}
+
+	for _, cmt := range clusterMachineTypes {
+		if acceleratorType == cmt {
+			return nil // Valid full machine type found in cluster
+		}
+	}
+	return fmt.Errorf("machine type %q not found in cluster and is not a known shorthand or defined machine type", acceleratorType)
```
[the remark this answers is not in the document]
Neelabh94: We want to keep a hard failure.

cmd/job/submit.go
```diff
@@ -28,18 +29,18 @@ import (
...
+	commandToRun   string
+	computeType    string
+	dryRunManifest string
 
 	workloadName     string
 	kueueQueueName   string
 	numSlicesOrNodes int
 	vmsPerSlice      int
-	maxRestarts      int
+	restarts         int
```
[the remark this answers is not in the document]
Neelabh94: We want to keep it as restarts so that the name is generic across GKE and SLURM

pkg/config/machine_configs.go
```diff
@@ -108,48 +108,77 @@ func parseTPUCount(machineType string) int {
...
+			initErr = fmt.Errorf("failed to initialize compute service: %w", err)
+			return
 		}
-		mt = res
-		machineTypeCache.Store(cacheKey, mt)
+		computeService = s
+	})
+
+	if initErr != nil {
+		return nil, initErr
```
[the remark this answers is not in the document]
Neelabh94: This is out of scope for this PR. Can be picked up in a follow-up PR.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -816,42 +819,18 @@ func (g *GKEOrchestrator) queryMachineType() (string, error) {
...
+			uniqueMap[np.Config.MachineType] = true
 		}
 	}
-
-	return "", fmt.Errorf("could not find a valid topology shape for %d chips with prefix %s", totalChips, prefix)
+	for k := range uniqueMap {
+		unique = append(unique, k)
+	}
+	return unique, nil
 }
```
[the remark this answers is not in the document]
Neelabh94: The function is not redundant with GetMachineType, and it justifies its existence by being used in multiple places. The implementation is also as simple as it can be for extracting unique values in Go.

pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -262,20 +212,221 @@ func (g *GKEOrchestrator) calculateGCPMachineResourceLimits(opts ManifestOptions
...
+		for _, cmt := range clusterMachineTypes {
+			if strings.EqualFold(acceleratorType, cmt) {
+				return acceleratorType, nil
+			}
+		}
+	}
+
+	// 3. Fail fast
+	return "", fmt.Errorf("machine type %q could not be resolved from static maps or cluster state", acceleratorType)
+}
```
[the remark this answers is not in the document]
Neelabh94: The function already uses a map where it can, the other lookup is dynamic based on the cluster's live state.

pkg/config/hardware.go
```diff
@@ -29,6 +30,104 @@ var tpuFamilyDefaults = map[string]int{
...
+	"h100-80gb-2":      "a3-highgpu-2g",
+	"h100-80gb-4":      "a3-highgpu-4g",
+	"h100-80gb-8":      "a3-highgpu-8g",
+	"h100-mega-80gb-8": "a3-megagpu-8g",
+	"h200-141gb-8":     "a3-ultragpu-8g",
+	"b200-8":           "a4-highgpu-8g",
+	"gb200-4":          "a4x-highgpu-4g",
+
+	// TPU mappings
+	"v4-8":    "ct4p-hightpu-4t",
```
SwarnaBharathiMantena, line 71: Is this supposed to be "v4-4"?

pkg/config/hardware.go
```diff
@@ -122,26 +220,22 @@ func calculateAcceleratorNodes(machineType, topology string) (int, error) {
...
+			if val, err := strconv.Atoi(machineType[idx+1 : len(machineType)-1]); err == nil && val > 0 {
+				return val
+			}
 		}
-	}
-
-	// Fallback to known machine family defaults if no explicit "-Nt" suffix
-	if !hasExplicitAccelerators {
+		// Fallback to known machine family defaults
 		for family, defaultAccs := range tpuFamilyDefaults {
```
SwarnaBharathiMantena: Since the orchestrator now has a `machineTypeClient`, this function can be updated to use the `GuestAcceleratorCount` directly from the API object when possible.

pkg/config/hardware.go
SwarnaBharathiMantena: This will also likely break the calculation in `ResolveTopologyForChips` and `CalculateAcceleratorNodes`. These functions expect the numeric suffix of the shorthand to represent total chips.
Also, the unit tests for topology resolution seem to use v4-4 to represent a 4-chip (8-core) setup at pkg/config/hardware_test.go (Lines 229-234).

Should we add instruction somewhere on what the x and y stand for in the notation `vx-y`?

Neelabh94: Good catch, have altered the logic to handle this case specifically.

pkg/config/hardware.go
Neelabh94: Done

SwarnaBharathiMantena (approved):
LGTM

docs/gcluster_job_guide.md
cboneti, line 11: This is only required for developers and not required if the user gets gcluster binary directly.

cboneti, line 15: is there a sensible default?

cboneti, line 13: Why mention a project here? I mean don't you also need a cluster? or will it create one on the fly?

cboneti, line 40: why are we recommending this, vs just obtaining the binary?

cboneti: we should say, for development workflow, build the binary as documented in xyz place.

cboneti, line 59: I doubt the user will know what is Crane. It might be better to mention `gcluster`.

cboneti, line 70: same here, I don't think users should need to know about crane, it is an implementation detail, not the interface.

cboneti, line 74: what should the user do if they already have a cluster?

cboneti, line 101: here, I think the full command reference could be before Troubleshooting. Here, it seems to break the tutorial flow.

cboneti, line 103: this might be the only place where it is ok to mention Crane, and here, I would describe what Crane is. I doubt any user will know otherwise. 
I think Crane is an implementation detail, but since this section is a reference section that goes into detail, it might work to use something like: 
"It can use pre-built images or build images on-the-fly without a local Docker daemon (powered internally by the [Crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/README.md) container utility)."

docs/gcluster_job_guide.md
```diff
@@ -110,18 +110,18 @@ Here are the flags currently supported by `gcluster job submit`:
...
 * `-b, --build-context string`: Path to the build context directory for Crane (e.g., `./job_details`). Required with `--base-image`. Crane will package all files in this directory and append them as a new layer to the base image (it does not require or execute a Dockerfile).
 * `-e, --command string`: Command to execute in the container (e.g., `'python app.py'`). This overrides the `CMD` instruction in your `Dockerfile`. (Required)
-* `-a, --accelerator string`: Type of accelerator to request (e.g., `'nvidia-h100-mega-80gb'` or machine type like `n2-standard-32`). (Required) It also supports shorthand strings for TPUs like `v6e-8` to request total chips; the tool will resolve the machine type and calculate `vms-per-slice` and `topology` automatically.
+* `--compute-type string`: Type of compute to request (e.g., `'n2-standard-32'`, `'nvidia-l4'`, or shorthand strings for TPUs like `v6e-8`). (Required) The tool will resolve the machine type and calculate `vms-per-slice` and `topology` automatically if needed.
```
cboneti, line 113: what about: 
"---compute-type string: The hardware target for the job. Accepts a full GCE machine type (e.g., 'n2-standard-32'), a GKE accelerator type (e.g., 'nvidia-l4'), or a TPU shorthand string representing total chips/cores (e.g., 'v6e-8'). (Required) The tool will automatically resolve the machine type, calculate vms-per-slice, and deduce the correct topology if needed."

Here, topology is only available for tpu, right? so then should we say "the correct TPU topology"?

docs/gcluster_job_guide.md
```diff
@@ -110,18 +110,18 @@ Here are the flags currently supported by `gcluster job submit`:
...
 * `-o, --dry-run-out string`: Path to output the generated Kubernetes manifest instead of applying it directly to the cluster. Useful for inspection.
 * `-c, --cluster string`: Name of the GKE cluster to deploy the job to. Optional if set in configuration.
 * `-l, --location string`: Location (Zone or Region) of the GKE cluster. Optional if set in configuration.
 * `-p, --project string`: Google Cloud Project ID. Optional if set in configuration.
-* `-f, --platform string`: Target platform for the image build (e.g., `linux/amd64`, `linux/arm64`). Used with `--base-image`. (Default: `linux/amd64`)
-* `-w, --name string`: Name of the job (JobSet) to create. This name will be used for Kubernetes resources. (Required)
+* `-f, --platform string`: Target platform for the image build (e.g., `linux/amd64`, `linux/arm64`). Used with --base-image. (Default: `linux/amd64`)
+* `-n, --name string`: Name of the job (JobSet) to create. This name will be used for Kubernetes resources. (Required)
 * `--queue string`: Name of the Kueue LocalQueue to submit the job to. (Default: Auto-discovered from the cluster)
 * `--nodes int`: Number of JobSet replicas (slices). (Default: `1`)
```
cboneti, line 121: This was potentially my mistake in the original design doc. Nodes should really map to total number of VMs. 

I think the above should be called --node-pools or even --num-slices if it cannot be used outside of the TPU world.

docs/gcluster_job_guide.md
```diff
@@ -110,18 +110,18 @@ Here are the flags currently supported by `gcluster job submit`:
...
 * `--vms-per-slice int`: Number of VMs (pods) per slice. (Default: `1`). Can be auto-calculated for TPUs if `--topology` is provided.
```
cboneti, line 122: again, very tpu specific, do you filter that it is only used for TPUs? perhaps we should break down to commands that only work with TPUs, if we cannot generalize them. It would be beneficial to have a table describing what happens in the case of TPU/GPU/CPU if applicable.

cboneti (commented):
partial review, I will continue the review later.

cboneti:
all my comments above are addressed in https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/5618/

PR #5607, opened 2026-05-06
Correctly evaluate Docker credentials prerequisite state

scaliby:
### The Problem
When submitting a job via `gcluster job submit`, the prerequisite check for Docker credentials was getting stuck in a broken state. 
If the user's Docker credentials were not configured, the code would correctly flag the missing prerequisite and populate the `missing` array. It would then update the in-memory state flag (`state.DockerCredsConfigured = true`). However, because the missing array was not empty, the function immediately returned an error and exited before saving the updated state to disk (`~/.gcluster/job_prereq_state.json`). 

  Furthermore, there was no actual logic to dynamically check the user's system to see if the credentials had been configured; it relied entirely on that state file flag. This meant the CLI could never detect if the user had actually run the `gcloud auth configure-docker` commands.

### The Solution
1. Introduced a new function, `isDockerCredsConfigured(region string)`, which actively evaluates the user's system state.
2. The function parses `~/.docker/config.json` and verifies that the `gcloud` credential helper is configured globally via `credsStore`, or specifically for both `"gcr.io"` and `"<region>-docker.pkg.dev"` via `credHelpers`.
3. Updated the `ensurePrerequisites` method to rely on this new dynamic check rather than the flawed state-flag mutation.

### How to Test
1. Remove the `gcloud` credential helpers from your `~/.docker/config.json`.
2. Clear your local state file: `rm ~/.gcluster/job_prereq_state.json`.
3. Run a job submission. Ensure it fails with the correct prompt to run `gcloud auth configure-docker`.
4. Run the suggested `gcloud auth configure-docker ...` commands.
5. Re-run the job submission. The prerequisite check will now pass.

### DIFF
--- a/cmd/job/prereq.go
+++ b/cmd/job/prereq.go
@@ -188,6 +188,37 @@ func checkK8sDependencies(state *PrereqState, missing *[]missingPrereq) {
 	}
 }
 
+// isDockerCredsConfigured checks if Docker is configured to use gcloud credentials for the required registries.
+func isDockerCredsConfigured(region string) bool {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return false
+	}
+	configPath := filepath.Join(homeDir, ".docker", "config.json")
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		return false
+	}
+
+	var config struct {
+		CredHelpers map[string]string `json:"credHelpers"`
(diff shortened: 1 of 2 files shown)


Neelabh94 (approved):
LGTM

kvenkatachala333 (approved):

PR #5628, opened 2026-05-08
Default Kueue config for Pathways

scaliby:
## Objective
Simplify the configuration management for Pathways and remove brittle inline YAML-merging logic.

## Changes
- Removed the complex inline queue-merging logic inside `main.tf`.
- Introduced a dedicated, standalone default template file for Pathways (`kueue/kueue-configuration-pathways.yaml.tftpl`).
- Refactored `main.tf` to conditionally load this template when `enable_pathways_for_tpus` is true and no custom `config_path` is provided.

This cleanup makes the module logic much easier to maintain and establishes a clean pattern for adding default configurations for other advanced features.


SwarnaBharathiMantena (dismissed):
LGTM!

The failing tests are either capacity issues or known issues the team is working on.

SwarnaBharathiMantena (dismissed):

modules/management/kubectl-apply/main.tf
```diff
@@ -15,24 +15,44 @@
...
+
+  kueue_default_config_template = (local.enable_pathways && local.enable_dynamic_slicing) ? "${path.module}/kueue/kueue-configuration-dynamic-slicing-pathways.yaml.tftpl" : (
+    local.enable_pathways ? "${path.module}/kueue/kueue-configuration-pathways.yaml.tftpl" : (
+      local.enable_dynamic_slicing ? "${path.module}/kueue/kueue-configuration-dynamic-slicing.yaml.tftpl" : ""
+    )
+  )
+
+  kueue_config_template_vars = merge(
+    {
+      accelerator_type      = "v4-podslice"
```
Neelabh94: there is no inbuilt way currently to have the accelerator_type flow automatically to the kubectl-apply module without user intervention.

* As the code stands now, any user using a different generation of hardware (like TPU v5p or v7) must explicitly override the value in their blueprint settings to avoid jobs staying pending in Kueue due to mismatched resource flavors

As a resolution, we can:
* Add a variable to modules/management/kubectl-apply/variables.tf, such as tpu_accelerator_type.
* Add a corresponding output to the source module (typically gke-cluster or a node pool module) that exports the actual accelerator type being used.
* Update the local logic in kubectl-apply/main.tf to prefer the value from that new variable over the hardcoded "v4-podslice" default.

modules/management/kubectl-apply/variables.tf
```diff
@@ -121,19 +121,8 @@ variable "apply_manifests" {
 
 variable "kueue" {
   description = "Install and configure [Kueue](https://kueue.sigs.k8s.io/docs/overview/) workload scheduler. A configuration yaml/template file can be provided with config_path to be applied right after kueue installation. If a template file provided, its variables can be set to config_template_vars."
-  type = object({
-    # ATTENTION: If you update the KUEUE's default version below, please also update the corresponding
```
Neelabh94, line 125: this removes strict type checking and might result in user's erroneous inputs going unnoticed.
Is there a specific reason we wish to move to `any` here, instead of expanding the object block?

Neelabh94 (approved):

Neelabh94 (changes requested):
Please address the comments above.

modules/management/kubectl-apply/variables.tf
jamOne-: +1

modules/management/kubectl-apply/kueue/kueue-configuration-dynamic-slicing-pathways.yaml.tftpl
jamOne-, line 1: I think this PR should not add the dynamic-slicing configs as they won't be used.

jamOne- (changes requested):

modules/management/kubectl-apply/main.tf
Neelabh94: If it's not too complicated, we can also look at enhancing this logic to actually convert the machine_type passed by the user to accelerator_type implicitly using the [machine_mappings.json](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/pkg/config/machine_mappings.json) added recently. This would remove the need of a new variable.

modules/management/kubectl-apply/variables.tf
scaliby: I changed this to any based on the automated gemini review, which pointed out that our repository style guide recommends using `type = any` for complex data structures to allow native YAML/HCL configurations. Do you prefer to leave the specific type here?

Neelabh94: Yes, for this use case please let's switch back to the specific object block. 
Apologies for the additional overhead

scaliby: Done. Static schema added.

modules/management/kubectl-apply/main.tf
scaliby: IIUC this will be computed by golang code and we have not decided on how to get this resolved yet. I would skip implementation of that for now and introduce this change in follow-up PR's. @Neelabh94 WDYT?
CC: @SwarnaBharathiMantena

modules/management/kubectl-apply/kueue/kueue-configuration-dynamic-slicing-pathways.yaml.tftpl
scaliby: removed dynamic configs.

modules/management/kubectl-apply/main.tf
jamOne-: I'm fine with skipping for now, but let's don't default to hardcoded `v4-podslice`, but rather add a validation to ensure it's set when we use a default config.

modules/management/kubectl-apply/main.tf
```diff
@@ -19,20 +19,35 @@ locals {
...
+    file(local.kueue_default_config_template)
+  ) : ""
+
+  kueue_user_config_content = var.kueue.config_path != "" && var.kueue.config_path != null ? (
+    endswith(var.kueue.config_path, ".tftpl") || (var.kueue.config_template_vars != null && length(var.kueue.config_template_vars) > 0) ?
+    templatefile(var.kueue.config_path, local.kueue_config_template_vars) :
+    file(var.kueue.config_path)
+  ) : ""
+
+  kueue_config_content = join("\n---\n", compact([local.kueue_default_config_content, local.kueue_user_config_content]))
```
jamOne-: That's not entirely on point, if user provides the `config_path` we should not apply the default config. In this PR the default config should be considered only when pathways and `config_path=null`.

This is important, as this gives a possibility for users to provide their own Kueue setup, without merging anything to it.

jamOne- (changes requested):

modules/management/kubectl-apply/main.tf
scaliby: @jamOne- removed the default and added validation.

modules/management/kubectl-apply/main.tf
scaliby: Done.

modules/management/kubectl-apply/main.tf
```diff
@@ -19,20 +19,34 @@ locals {
...
+      pathways_cpu_quota    = 480
+      pathways_memory_quota = "2000G"
+      tpu_quota             = "999999" # Default high value if not set
+    },
+    var.kueue.config_template_vars != null ? var.kueue.config_template_vars : {}
+  )
+
+  kueue_default_config_content = local.kueue_default_config_template != "" ? (
+    endswith(local.kueue_default_config_template, ".tftpl") ?
+    templatefile(local.kueue_default_config_template, local.kueue_config_template_vars) :
```
jamOne-, line 37: nit: We know that the default config will always be a `templatefile`, if not empty. Please remove the endswith check.

jamOne- (approved):
lgtm

scaliby:
> Please address the comments above.

Done. I also added `accelerator_type` to blueprints as we discussed.

Neelabh94 (approved):
LGTM

SwarnaBharathiMantena (approved):

scaliby:
@SwarnaBharathiMantena can we get this merged?

PR #5653, opened 2026-05-12
Handle empty `machine_type` in TPU reservations

scaliby:
### **Problem**
When using TPU reservations, the `machine_type` (and potentially other properties) in the reservation definitions can sometimes be empty. The module's validation logic previously compared the node pool's VM properties directly against the reservation's properties. If the reservation property was an empty string but the node pool property was not, it incorrectly flagged this as a reservation requirement violation, preventing successful provisioning or causing false validation errors.

### **Solution**
Updated the `specific_reservation_requirement_violations` logic in `modules/compute/gke-node-pool/reservation_definitions.tf` to ignore property mismatch checks when the value coming from the reservation (`local.reservation_vm_properties`) is an empty string. This ensures that empty reservation properties, such as those sometimes encountered with TPU reservations, do not trigger false validation violations.

### DIFF
--- a/modules/compute/gke-node-pool/reservation_definitions.tf
+++ b/modules/compute/gke-node-pool/reservation_definitions.tf
@@ -77,7 +77,10 @@ locals {
   # Compare two maps by counting the keys that mismatch.
   # Know that in map comparison the order of keys does not matter. That is {NVME: x, SCSI: y} and {SCSI: y, NVME: x} are equal
   # As of this writing, there is only one reservation supported by the Node Pool API. So, directly accessing it from the list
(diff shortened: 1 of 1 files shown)


SwarnaBharathiMantena:
The gke-node-pool modules has the [`is_reservation_active` flag](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/8fb2919fde8c7472d5b63b134074b073e5f92c73/modules/compute/gke-node-pool/variables.tf#L176) that a user can set when they are working with reservations that are not active yet.

More info is captured in the [README here](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/main/modules/compute/gke-node-pool/README.md#target-a-future-reservation-or-reservations-that-are-not-fulfilled-yet).

Is the scenario this PR tackles something that can be addressed using the `is_reservation_active` variable?

PR #5654, opened 2026-05-12
Use cluster `id` instead of `self_link` for node pools

scaliby:
## Problem

When referencing a GKE cluster for node pools using a reference type other than "NAME", the module was previously utilizing the `self_link` attribute of the `google_container_cluster` resource. Relying on `self_link` for cluster references can be problematic or generate warnings with newer versions of the Google Cloud Terraform provider, which favor using the resource's `id` for safer and more robust cross-resource dependencies.

## Solution

Updated the `cluster` attribute in both the `system_node_pools` and `cpu_np` node pool resources to use `google_container_cluster.gke_cluster.id` instead of `google_container_cluster.gke_cluster.self_link` when `var.cluster_reference_type` is not set to `"NAME"`. This aligns the module with current Terraform best practices for referencing GCP resources.


modules/scheduler/gke-cluster/variables.tf
```diff
@@ -460,13 +460,13 @@ variable "additional_networks" {
 }
 
 variable "cluster_reference_type" {
-  description = "How the google_container_node_pool.system_node_pools refers to the cluster. Possible values are: {SELF_LINK, NAME}"
-  default     = "SELF_LINK"
+  description = "How the google_container_node_pool.system_node_pools refers to the cluster. Possible values are: {ID, NAME}"
```
SwarnaBharathiMantena, line 463: It is better to let "SELF_LINK" remain as an option to make sure it does not break the existing set up users have and ensure backward compatibility.

We can update the default to "ID".

PR #5658, opened 2026-05-13
[Telemetry] Capture exit code upon fatal command failures

kadupoornima:
This pull request enhances the telemetry system by **ensuring that diagnostic data is captured and sent even when the application encounters a fatal error**. By introducing a hook in the logging infrastructure, the system can now reliably report command failures, providing better visibility into application crashes.

Changes:

* **Telemetry Improvement**: Added a FatalHook mechanism to the logging package to ensure telemetry data is flushed during hard failures.
* **Command Failure Tracking**: Implemented a hook in cmd/root.go that triggers telemetry collection with an exit code of 1 whenever a fatal error occurs.

Tested:

image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -79,9 +79,9 @@ func init() {
 			userConfigExists = true
 
 			// Register the fatal hook to flush telemetry on hard failures
-			logging.FatalHook = func() {
-				if config.IsTelemetryEnabled() {
-					telemetryCollector.Execute(1) // Hard-code the exit code to 1 since logging.Fatal always exits with 1
+			logging.FatalHook = func(exitCode int) {
+				if config.IsTelemetryEnabled() && telemetryCollector != nil {
+					telemetryCollector.Execute(exitCode)
 				}
 			}
 		}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -77,6 +77,13 @@ func init() {
 		if err := config.InitUserConfig(); err == nil {
 			telemetryCollector = telemetry.NewCollector(cmd, args, InstallationMode)
 			userConfigExists = true
+
+			// Register the fatal hook to flush telemetry on hard failures
+			logging.FatalHook = func() {
+				if config.IsTelemetryEnabled() {
+					telemetryCollector.Execute(1) // Hard-code the exit code to 1 since logging.Fatal always exits with 1
+				}
+			}
 		}
 	}
 
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -27,7 +27,7 @@ var (
 	infolog      *log.Logger
 	errorlog     *log.Logger
 	fatallog     *log.Logger
-	FatalHook    func() // FatalHook allows registering a callback to run before the program exits on a fatal error.
+	FatalHook    func(exitCode int) // FatalHook allows registering a callback to run before the program exits on a fatal error.
 	Exit         = os.Exit
 	TsColor      = color.New(color.FgMagenta)
 	WarningColor = color.New(color.FgYellow)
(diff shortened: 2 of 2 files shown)


cmd/root.go
```diff
@@ -77,6 +77,13 @@ func init() {
 		if err := config.InitUserConfig(); err == nil {
 			telemetryCollector = telemetry.NewCollector(cmd, args, InstallationMode)
 			userConfigExists = true
+
+			// Register the fatal hook to flush telemetry on hard failures
+			logging.FatalHook = func() {
+				if config.IsTelemetryEnabled() {
+					telemetryCollector.Execute(1) // Hard-code the exit code to 1 since logging.Fatal always exits with 1
+				}
+			}
```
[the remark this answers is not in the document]
kadupoornima: WAI. We do not want to capture initDependencies failures, we want to capture CLI command failures.

pkg/logging/logging.go
```diff
@@ -66,5 +67,11 @@ func Error(f string, a ...any) {
 func Fatal(f string, a ...any) {
```
SwarnaBharathiMantena, line 67: `defer Exit(1)` can be used at the top of the Fatal function. It ensures that even if the hook panics, the program still terminates with the correct exit code after the panic is handled (or after it finishes unwinding).

```
// pkg/logging/logging.go
func Fatal(f string, a ...any) {
    defer Exit(1)
    msg := fmt.Sprintf(f, a...)
    fatallog.Printf("%s: %s", formatTs(), msg)

    if FatalHook != nil {
        FatalHook()
    }
}
```

cmd/root.go
```diff
@@ -77,6 +77,13 @@ func init() {
 		if err := config.InitUserConfig(); err == nil {
 			telemetryCollector = telemetry.NewCollector(cmd, args, InstallationMode)
 			userConfigExists = true
+
+			// Register the fatal hook to flush telemetry on hard failures
+			logging.FatalHook = func() {
+				if config.IsTelemetryEnabled() {
+					telemetryCollector.Execute(1) // Hard-code the exit code to 1 since logging.Fatal always exits with 1
```
SwarnaBharathiMantena: To make the system more robust and flexible, this can be changed to accept an integer: `FatalHook func(int)`. This allows the logging package to pass the actual exit code to the hook, ensuring telemetry reflects the correct termination status even if `logging.Fatal` is updated to support custom exit codes in the future.

The exit code is typically `2` for panics in Go.

cmd/root.go
```diff
@@ -77,6 +77,13 @@ func init() {
 		if err := config.InitUserConfig(); err == nil {
 			telemetryCollector = telemetry.NewCollector(cmd, args, InstallationMode)
 			userConfigExists = true
+
+			// Register the fatal hook to flush telemetry on hard failures
+			logging.FatalHook = func() {
+				if config.IsTelemetryEnabled() {
```
SwarnaBharathiMantena: The closure assigned to logging.FatalHook uses the `telemetryCollector` variable. Even though it is initialized just before the hook is registered, it is safer to perform a nil check within the hook itself. This prevents a panic if the telemetry initialization failed or if a fatal error occurs in a path where the collector was never created.
```
// cmd/root.go
logging.FatalHook = func() {
    if config.IsTelemetryEnabled() && telemetryCollector != nil {
        telemetryCollector.Execute(1)
    }
}
```

pkg/logging/logging.go
kadupoornima: Done.

cmd/root.go
kadupoornima: Done.

cmd/root.go
kadupoornima: Thanks, done.

SwarnaBharathiMantena (approved):

PR #5664, opened 2026-05-14
feat(job submission): Dynamic topology routing for gke jobs

Neelabh94:
Implements Dynamic Topology Routing in gcluster job submit to allow workloads to fall back to larger idle TPU/GPU pools when their preferred pool is full. This maximizes cluster utilization and reduces job starvation by providing more flexible scheduling constraints.

### Key Changes
 * Support for Pipe Separator in Constraints: Users can now specify multiple allowed values for any node constraint separated by a pipe (e.g., --node-constraint "cloud.google.com/gke-tpu-topology=2x2|4x8").
 * Dynamic NodeAffinity Generation: When multiple values are detected for a constraint, gcluster now generates a Kubernetes nodeAffinity block with the In operator instead of a strict nodeSelector.
 * Smart Merging for Topology: If both a baseline --topology and fallback topologies via --node-constraint are specified, they are automatically merged into a single nodeAffinity list to avoid scheduling conflicts.
 * Documentation: Added examples and updated flag references in docs/gcluster_job_guide.md.

### Tests: 
* Added unit tests in scheduling_test.go covering the new parsing and merging logic. All tests passed.

### Verification Results
Verified via gcluster job submit --dry-run-out that the generated manifest correctly emits nodeAffinity with In operator for fallback values and keeps nodeSelector clean of strict topology requirements when fallbacks are present.

### DIFF
--- a/pkg/orchestrator/gke/scheduling.go
+++ b/pkg/orchestrator/gke/scheduling.go
@@ -16,6 +16,8 @@ package gke
 
 import (
 	"hpc-toolkit/pkg/config"
+	"slices"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -25,6 +27,7 @@ type SchedulingOptions struct {
 	Topology           string
 	Scheduler          string
 	NodeAffinityLabels map[string]string
+	IsDynamicSlicing   bool
 }
 
 func GetNodeSelector(opts SchedulingOptions) map[string]string {
@@ -35,6 +38,14 @@ func GetNodeSelector(opts SchedulingOptions) map[string]string {
 	}
 
 	for k, v := range opts.NodeAffinityLabels {
+		// Skip if it has a pipe (will go to affinity)
+		if strings.Contains(v, "|") {
+			continue
+		}
+		// Skip if it's topology and we have a baseline topology to merge with
+		if k == tpuTopologyLabel && opts.Topology != "" {
+			continue
+		}
 		nodeSelector[k] = v
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/orchestrator/gke/scheduling.go
```diff
@@ -62,6 +72,38 @@ func GetAffinity(opts SchedulingOptions) *corev1.Affinity {
...
+
+		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions = append(
+			affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions,
+			corev1.NodeSelectorRequirement{
+				Key:      k,
+				Operator: corev1.NodeSelectorOpIn,
+				Values:   values,
+			},
+		)
+	}
```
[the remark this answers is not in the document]
Neelabh94: It is correct partially. Have used a pointer and added whitespace trimming logic. However we want to keep nodeconstraint a generic flag that can still be used to pass regular key-value pairs without the pipe separator.

pkg/orchestrator/gke/scheduling.go
```diff
@@ -45,23 +56,66 @@ func GetNodeSelector(opts SchedulingOptions) map[string]string {
...
+			values = append(values, opts.Topology)
+		}
+
+		for _, val := range strings.Split(v, "|") {
+			if trimmed := strings.TrimSpace(val); trimmed != "" {
+				if !slices.Contains(values, trimmed) {
+					values = append(values, trimmed)
+				}
+			}
+		}
```
kvenkatachala333: There is a section of fail-fast approaches in the design doc shared, we should validate,

1. each trimmed value matches a known topology pattern (e.g., using regex) if the key is `cloud.google.com/gke-tpu-topology`

2. If the user provides `2x2|` (a trailing pipe), split yields an empty string which is silently ignored here via `trimmed != ""`. We should fail fast if any split element is empty or invalid

can you please cover these here?

Neelabh94: Thanks for pointing this out. Done.

kvenkatachala333 (approved):

shubpal07 (approved):
LGTM. 


A quick note for our follow-up FR on TAS:

Currently, `IsDynamicSlicing` is resolved inside the gcluster manager strictly by querying GKE's API for the `PROVISION_ONLY` placement policy (autoscaling).

- The Limitation: Pre-provisioned static GCE reservations  may NOT carry the `PROVISION_ONLY` policy. If a user tries to logically partition/sub-slice a static reservation pool, `IsDynamicSlicing` evaluates to false, and GCluster will hardcode the strict topology selector back into nodeSelector, triggering a scheduling deadlock.

- Proposed Solution for Follow-up: In the next follow-up, we can refactor this to use a Topological Intent Check 

1. Instead of relying strictly on GKE's `PROVISION_ONLY` placement policy API check, GCluster's compiler should automatically evaluate the user's topological intent. The Intent Check: If the requested workload topology (e.g., 2x4 requested via --topology) is a proper subset / smaller than the GKE cluster's physically mapped node pool topology (e.g. 4x4), `gcluster` should automatically evaluate IsDynamicSlicing: true.

PR #5671, opened 2026-05-15
Remove the disk size attributes from deployment files

kadupoornima:
This pull request focuses on **cleaning up deployment configurations by removing redundant disk size attributes**. The changes streamline the variable sets used in various GKE and Slurm deployment examples, ensuring a cleaner interface for users deploying these infrastructure templates.

Changes:

* **Cleanup of disk size attributes**: Removed explicit disk size configuration variables from multiple deployment and example files to simplify configuration management.
* **Refactoring variable placement**: Reorganized variable definitions in several YAML files to improve readability and consistency across different deployment templates.


vikramvs-gg:
`gke-tpu-v6e` also has these disk variables in its deployment file and defaults defined in the blueprint file, we can remove from there as well.

vikramvs-gg (approved):

PR #5672, opened 2026-05-15
Remove hpc-slurm-static blueprint

kadupoornima:
This example demonstrates how to create a partition with static compute nodes. However, currently most machine learning examples use static partitions (with reservations), therefore this blueprint is redundant and can be removed.


agrawalkhushi18 (approved):
LGTM!

PR #5676, opened 2026-05-16
[Telemetry] Remove Billing Account Id as a custom metric (it is still collected) 

kadupoornima:
This pull request refactors the telemetry collection logic to remove the billing account ID from the general metadata map while ensuring it is still correctly included in the Concord event construction. This change streamlines the metadata collection process by fetching the billing account ID directly when needed, rather than maintaining it as a persistent custom metric. Note that this data is not lost; it is still being collected and securely sent to Clearcut as a top-level field within the Concord event payload 

Changes:

* **Metric Cleanup**: Removed `BILLING_ACCOUNT_ID` from the general telemetry metadata collection.
* **Event Construction Update**: Updated BuildConcordEvent to fetch the billing account ID directly via getBillingAccountId instead of relying on the cached metadata map.
* **Test Adjustments**: Updated unit tests to reflect the removal of the billing account ID from the expected telemetry metadata.


SwarnaBharathiMantena (approved):
LGTM

PR #5703, opened 2026-05-25
[Telemetry] Fix flaky is_googler metric and improve it by local caching

kadupoornima:
This pull request **optimizes the `is_googler` metric collection by implementing a local caching mechanism and replacing external CLI dependencies with direct file-based configuration parsing**. These changes reduce latency and improve the stability of internal user detection by removing reliance on the `gcloud` binary.

### Background
Currently, the `is_googler` telemetry metric intermittently returns `false` for internal users. The random behavior of `is_googler` returning `false` is caused by a tight execution timeout on a subprocess. When no Application Default Credentials (ADC) are found, the system falls back to spawning a `gcloud` subprocess with a strict 2-second timeout. Because `gcloud` is a Python-based CLI, its startup and execution time can fluctuate depending on the local machine's CPU load, disk I/O, or concurrent processes. When this timeout is hit, the CLI fails open and silently defaults to `false`.

### Changes Introduced
This PR implements a two-pronged solution to make the metric perfectly reliable and highly performant:

1. **Direct Configuration File Parsing (`pkg/telemetry/collector_util.go`)**
   * Removed the flaky `exec.CommandContext(ctx, "gcloud", ...)` subprocess call.
   * Implemented `checkGcloudConfigForInternalUser()`, which directly reads the `gcloud` CLI's local INI configuration files.
   * The logic dynamically resolves the correct configuration path by respecting the `CLOUDSDK_CONFIG` environment variable, falling back to `%APPDATA%\gcloud` on Windows or `~/.config/gcloud` on Linux/macOS.
   * It reads `active_config` and parses `configurations/config_<active>` to extract the `account` email from the `[core]` section.

2. **Persistent Caching (`pkg/config/user_config.go`)**
   * Added an `IsGoogler *bool` field to the `UserConfig` struct. Using a pointer ensures backward compatibility with existing configuration files.
   * Created getter/setter helper functions to read and write this status to the local `telemetry_config.json` cache file.
   * Updated `getIsGoogler()` to check this cache first before attempting any credentials evaluation.

3. **Comprehensive Unit Testing**
   * Replaced the obsolete subprocess-based tests with a new suite in `collector_test.go` (`TestEvaluateIsGoogler`, `TestGetIsGoogler_Cache`, `TestCheckGcloudConfigForInternalUser_*`).
   * Added cache-specific unit tests in `user_config_test.go` to verify backward compatibility with missing fields and proper disk persistence.

### Advantages and ROI
* **Zero Flakiness:** Directly reading the `~/.config/gcloud` text files is nearly instantaneous and completely eliminates the subprocess timeout failures affecting users.
* **High Performance:** Because we persist this value into `telemetry_config.json`, the CLI only has to perform the filesystem read and parsing operation once. Subsequent executions pull the boolean directly from memory, speeding up CLI initialization.
* **Failsafe Privacy:** The fallback mechanism still gracefully returns `false` if files are missing or unreadable, ensuring we never accidentally classify an external user as a Googler.

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"sync"
 )
 
 const configFileName = "telemetry_config.json"
@@ -33,22 +34,31 @@ type UserConfig struct {
 }
 
 // globalUserConfig is the package-level variable holding the state during execution
-var globalUserConfig UserConfig
+var (
+	globalUserConfig UserConfig
+	mu               sync.RWMutex
+)
 
 // InitUserConfig initializes the user's config, prioritizing a local JSON file over defaults.
 func InitUserConfig() error {
 	// Set the defaults
+	mu.Lock()
 	globalUserConfig = UserConfig{
 		UserID:           generateUniqueID(),
 		TelemetryEnabled: true, // Default telemetry state
 	}
+	mu.Unlock()
 
 	configFile := filepath.Join(getLocalDirPath(false), configFileName)
 
 	// Try to read from the local config file
 	if data, err := os.ReadFile(configFile); err == nil {
 		// If the file exists and is valid, overwrite the defaults
-		if err := json.Unmarshal(data, &globalUserConfig); err == nil {
+		mu.Lock()
+		err := json.Unmarshal(data, &globalUserConfig)
+		mu.Unlock()
+
+		if err == nil {
 			return nil
 		}
 	}
@@ -79,12 +89,18 @@ func SetTelemetry(telemetry bool) error {
 
 // GetIsGoogler returns the cached IsGoogler value if it exists, otherwise nil. This refers to whether the user is internal to Google or not.
 func GetIsGoogler() *bool {
+	mu.RLock()
+	defer mu.RUnlock()
+
 	return globalUserConfig.IsGoogler
 }
 
 // SetIsGoogler sets the IsGoogler status and persists it to disk.
 func SetIsGoogler(isGoogler bool) error {
+	mu.Lock()
 	globalUserConfig.IsGoogler = &isGoogler
+	mu.Unlock()
+
 	err := SaveToFile()
 	if err != nil {
 		return fmt.Errorf("failed to save state to file: %v", err)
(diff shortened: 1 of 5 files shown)


pkg/telemetry/collector_util.go
```diff
@@ -254,6 +255,81 @@ var fetchProjectName = func(ctx context.Context, projectID string) (string, erro
...
+		}
+
+		if inCoreSection && strings.HasPrefix(line, "account") {
+			parts := strings.SplitN(line, "=", 2)
+			if len(parts) == 2 {
+				email := strings.TrimSpace(parts[1])
+				return isInternalEmail(email)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/config/user_config.go
SwarnaBharathiMantena: Addition of a sync.RWMutex to the UserConfig struct or using an atomic value to protect access to globalUserConfig can help. The globalUserConfig variable is accessed and modified via GetIsGoogler and SetIsGoogler. If multiple goroutines invoke telemetry collection (e.g., during concurrent operations), there is a risk of a data race because globalUserConfig is a package-level variable without synchronization.

pkg/telemetry/collector_util.go
```diff
@@ -254,6 +255,97 @@ var fetchProjectName = func(ctx context.Context, projectID string) (string, erro
...
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+
+		// Skip empty lines and comments
+		if line == "" || strings.HasPrefix(line, ";") || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		if strings.HasPrefix(line, "[") && strings.HasSuffix(line, "]") {
+			inCoreSection = strings.EqualFold(line, "[core]")
```
SwarnaBharathiMantena, line 341: Please ensure it handles cases where there might be spaces after the section name (e.g., [core]) or inline comments (e.g., account = user@google.com # my email).

pkg/telemetry/collector.go
```diff
@@ -323,30 +320,18 @@ func getBillingAccountId(bp config.Blueprint) string {
...
-		if isInternalEmail(email) {
-			return true
-		}
-	}
-	return false
+	// Evaluate if the user is internal
+	isInternal := evaluateIsGoogler()
+
+	// Cache the evaluated result in the local config for future commands
+	_ = config.SetIsGoogler(isInternal)
```
SwarnaBharathiMantena, line 332: Consider logging the error if SetIsGoogler fails so that users or developers can diagnose why the cache isn't persisting. While failing to save the cache isn't a fatal error, it might lead to repeated evaluations if the file system is read-only or full.

kadupoornima: This makes sense but we are keeping all Telemetry behind the scenes from the user. We do not want to trouble them with warnings which are not relevant for them.

pkg/config/user_config.go
kadupoornima: Done.

pkg/telemetry/collector_util.go
kadupoornima: Done, thanks for catching this!

SwarnaBharathiMantena (approved):
LGTM

PR #5718, opened 2026-05-27
Add GCluster warnings to inform users if their GKE version has vulnerabilities

kadupoornima:
This pull request **introduces a proactive security mechanism for the HPC Toolkit by implementing GKE version vulnerability checks**. The changes enable the toolkit to identify and warn users about potential security risks in their infrastructure configuration prior to deployment. The implementation includes a robust version resolution system that interacts with the GKE API and a centralized security advisory database, ensuring that users are alerted to both patched and pending vulnerabilities.

### Highlights

* **GKE Vulnerability Scanning**: Implemented a new validation layer that checks GKE cluster versions against a database of known security vulnerabilities before deployment.
* **Security Advisories Database**: Added a JSON-based security advisory store, embedded directly into the binary using //go:embed, to track CVEs, their status, and required minimum versions.
* **CLI Integration**: Integrated the vulnerability check into the 'create' and 'deploy' commands, including a new --skip-gke-security-check flag to bypass the validation when necessary.
* **GKE Version Resolution**: Added logic to resolve GKE versions from blueprints, including support for version prefixes by querying the GKE API.


pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,128 @@
...
+	Link            string            `json:"link"`
+}
+
+type VulnerabilityDB struct {
+	Advisories []Advisory `json:"advisories"`
+}
+
+// PerformGkeVersionSecurityChecks
+func PerformGkeVulnerabilitiesCheck(cmd *cobra.Command, args []string) {
+	skipSecurity, _ := cmd.Flags().GetBool("skip-gke-security-checks")
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/config/config.go
```diff
@@ -1237,3 +1238,101 @@ func worker(version string, jobs <-chan string, results chan<- string, wg *sync.
...
+					// Return the prefix itself so the vulnerability check can flag it as EOL or vulnerable.
+					versions = append(versions, versionPrefix)
+				}
+			}
+
+		}
+	})
+
+	return versions, nil
+}
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,128 @@
...
+	"hpc-toolkit/pkg/telemetry"
+
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
+)
+
+const SecurityAdvisoriesFilePath = "pkg/validators/security-advisories.json"
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("no Security Advisories file found: %v", err)
+	}
+
+	if err := json.Unmarshal(securityAdvisories, &db); err != nil {
+		return nil, fmt.Errorf("failed to parse Security advisories: %v", err)
+	}
+
+	return &db, nil
+}
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,123 @@
...
+	Advisories []Advisory `json:"advisories"`
+}
+
+// PerformGkeVersionSecurityChecks
+func PerformGkeVulnerabilitiesCheck(cmd *cobra.Command, args []string) {
+	skipSecurity, _ := cmd.Flags().GetBool("skip-gke-security-check")
+
+	// 3. evaluate security vulnerabilities BEFORE Terraform starts
+	if !skipSecurity {
+		blueprint := telemetry.GetBlueprint(cmd, args)
```
vikramvs-gg: We are importing a telemetry package: `pkg/telemetry` to use telemetry.GetBlueprint(). Can it be extracted out in some common helper utility ? This is because in case, we scrub sensitive data for metrics collection in future in telemetry, it will silently break the validator's logic. Or, can we check if create.go and deploy.go already have the blueprint loaded and `*config.Blueprint` struct directly to `PerformGkeVulnerabilitiesCheck` ?

pkg/config/config.go
```diff
@@ -1237,3 +1239,120 @@ func worker(version string, jobs <-chan string, results chan<- string, wg *sync.
...
+		// 1. Check for min_master_version safely
+		if version := getEvaluatedString("min_master_version", m, bp); version != "" {
+			versions = append(versions, version)
+			return // Proceed to the next module
+		}
+
+		// 2. Check for version_prefix safely
+		if versionPrefix := getEvaluatedString("version_prefix", m, bp); versionPrefix != "" {
+			if releaseChannel := getEvaluatedString("release_channel", m, bp); releaseChannel != "" && releaseChannel != "UNSPECIFIED" {
+				latestVersion, _ := fetchGKEVersionFunc(projectID, region, versionPrefix, releaseChannel)
```
vikramvs-gg: We might need to capture and log the error from GCP instead of using `_` as they would be silently ignored in current scenario. Even though there is fallback if there is error in fetching, but let us also log the error for better debugging.

pkg/validators/gke_vulnerabilities_check.go
kadupoornima: We cannot extract it to the common config package, since it causes a cyclic dependency loop. Scrubbing of data won't be an issue as we will not be modifying any data initially, so all data is collected as is.

pkg/config/config.go
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
arpit974: This introduces an anti pattern. pkg/validators is a core logic package and should not depend on pkg/telemetry, which is a consumer of data. Additionally, calling telemetry.GetBlueprint here causes a redundant and expensive re-parsing of the blueprint YAML.

We should follow the pattern established in previous validation efforts (like the Quota Validator) where the parsed Blueprint is passed as an argument."

pkg/config/config.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
 	"golang.org/x/exp/maps"
+	"google.golang.org/api/container/v1"
```
arpit974: suggestion:
Adding the GKE API client (google.golang.org/api/container/v1) as a dependency makes this package unnecessarily heavy. isn't this logic better suited for the validators or provider packages.

pkg/config/config.go
```diff
@@ -1237,3 +1239,126 @@ func worker(version string, jobs <-chan string, results chan<- string, wg *sync.
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get server config: %w", err)
+	}
+
+	var latestVersion = ""
+	for _, channel := range serverConfig.Channels {
+		if channel.Channel == releaseChannel {
+			// Find the latest valid version in the channel that matches the prefix. This is the version with which the cluster would get created.
+			for _, version := range channel.ValidVersions {
+				if strings.HasPrefix(version, prefix) && version > latestVersion {
```
arpit974: The string comparison version > latestVersion will fail for multi-digit patch versions (e.g., it will incorrectly evaluate "1.29.2" as greater than "1.29.10"). 

you may use semver.Compare for accurate version comparison.

pkg/validators/gke_vulnerabilities_check.go
kadupoornima: Done.

pkg/config/config.go
kadupoornima: Done.

pkg/config/config.go
kadupoornima: Done.

pkg/config/config.go
```diff
@@ -1237,3 +1239,126 @@ func worker(version string, jobs <-chan string, results chan<- string, wg *sync.
...
+					// Return the prefix itself so the vulnerability check can flag it as EOL or vulnerable.
+					versions = append(versions, versionPrefix)
+				}
+			}
+		}
+	})
+	if err == "" {
+		return versions, nil
+	}
+	return versions, fmt.Errorf("%s", err)
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,123 @@
...
+
+		minorVersion := semver.MajorMinor(gkeVersion)
+
+		for _, adv := range db.Advisories {
+			switch adv.Status {
+			case "PENDING":
+				warnings = append(warnings, fmt.Sprintf(
+					"SECURITY WARNING: Your deployment might be vulnerable to %s (%s). "+
+						"Patches are currently PENDING in upstream GKE. See: %s",
+					adv.CVE, adv.Name, adv.Link))
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,123 @@
...
+								"Please upgrade your blueprint to at least %s. See: %s",
+							gkeVersion, adv.CVE, adv.Name, patchedVersion, adv.Link))
+					}
+				}
+			}
+		}
+	}
+
+	return warnings
+}
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/config/config.go
```diff
@@ -1237,3 +1239,126 @@ func worker(version string, jobs <-chan string, results chan<- string, wg *sync.
...
+		if channel.Channel == releaseChannel {
+			// Find the latest valid version in the channel that matches the prefix. This is the version with which the cluster would get created.
+			for _, version := range channel.ValidVersions {
+				if strings.HasPrefix(version, prefix) && version > latestVersion {
+					latestVersion = version
+				}
+			}
+		}
+	}
+	return latestVersion, nil
```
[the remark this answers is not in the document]
kadupoornima: Done.

pkg/config/config.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
 	"golang.org/x/exp/maps"
+	"google.golang.org/api/container/v1"
+	"google.golang.org/api/option"
```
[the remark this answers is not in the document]
kadupoornima: NA.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return &db, nil
+}
+
+// evaluate checks the blueprint's GKE configuration against known advisories.
+func evaluate(db *VulnerabilityDB, gkeVersions []string) []string {
+	var warnings []string
+	for _, gkeVersion := range gkeVersions {
+
+		// Normalize version string for semver evaluation
```
arpit974: You can replace this manual prefix check with the normalizeVersion helper function defined on line 195 to ensure consistent version formatting across the package.

arpit974 (dismissed):
added a small comment, PTAL

pkg/validators/gke_vulnerabilities_check.go
kadupoornima: Done.

pkg/validators/gke_vulnerabilities_check.go
```diff
@@ -0,0 +1,241 @@
...
+			switch adv.Status {
+			case "PENDING":
+				warnings = append(warnings, fmt.Sprintf(
+					"SECURITY WARNING: Your GKE version %s might be vulnerable to %s (%s). "+
+						"Patches are currently PENDING in upstream GKE. See: %s",
+					gkeVersion, adv.CVE, adv.Name, adv.Link))
+			case "PATCHED":
+				if patchedVersion, exists := adv.PatchedVersions[minorVersion]; exists {
+					patchedVersion = normalizeVersion(patchedVersion)
+					if semver.Compare(gkeVersion, patchedVersion) < 0 {
```
shubpal07: Nice work on this check! However, strict `SemVer` (golang.org/x/mod/semver) treats GKE's `-gke.X` suffix as a `pre-release` tag, which has lower precedence than a normal release.

This means `v1.35.3` (normal version) is evaluated as greater than `v1.35.3-gke.1943000` (pre-release). If a user specifies a raw version prefix like `min_master_version: "1.35.3"`, the validator will normalize it to v1.35.3 and falsely pass the safety check against `v1.35.3-gke.1943000`, leaving them vulnerable.

Suggestion: We can implement a custom GKE version comparison helper that treats the -gke suffix as a post-release build number, defaulting a missing suffix to build `0`.

cmd/create.go
```diff
@@ -81,6 +84,8 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 
 func doCreate(cmd *cobra.Command, path string) string {
 	bp, ctx := expandOrDie(cmd, path)
+	skipSecurity, _ := cmd.Flags().GetBool("skip-gke-security-check")
+	validators.PerformGkeVulnerabilitiesCheck(skipSecurity, &bp)
```
shubpal07: gcluster create is designed to be a local, offline template generation step. Adding a dynamic API call here means that in air-gapped environments or for users without active local GCP credentials, this command will block for the full duration of the timeout before failing back.

Since deploy already requires active GCP auth and network access, should we restrict this dynamic API resolution to deploy only? Alternatively, if we keep it in create, we can refine the warning log to make it clear that it's a silent offline fallback to avoid confusing users:
```go
logging.Info("Offline Mode: Skipping GKE vulnerability check (could not reach GCP Container API: %v)", err)
```

pkg/validators/security-advisories.json
```diff
@@ -0,0 +1,37 @@
...
+        "cve": "CVE-2026-46300",
+        "name": "Fragnesia",
+        "affected_images": ["UBUNTU_CONTAINERD"],
+        "status": "PENDING",
+        "patched_versions": {},
+        "link": "https://cloud.google.com/kubernetes-engine/security-bulletins#gcp-2026-033"
+      }
+    ]
+  }
+
```
shubpal07, line 37: This is a great addition to the toolkit's security posture!. The local JSON database (security-advisories.json) is a great MVP to get the validation logic and unit tests merged.

However, looking ahead at *operational scalability* and *maintenance*, a static embedded JSON file has a few long-term drawbacks:

* **Release Coupling**: Security updates will be tied to our release cycles. If a new high-severity GKE CVE is disclosed, we cannot warn users until we merge a CL and users download a new compiled binary.
* **Maintenance Toil**: Manually tracking GKE bulletins and updating this JSON will become an operational burden for the developer team.

_**Proposed Future Iteration**_: We can consider moving to a Hybrid model in a future iteration:

1. Host this `security-advisories.json` in a public Google-owned GCS bucket, updated automatically by a scheduled scraper.

2. Have gcluster attempt to fetch this live JSON during deployment with a tight timeout (e.g., 800ms).
If the user is offline or in an air-gapped environment, the CLI gracefully falls back to the embedded //go:embed database we are introducing in this PR.

shubpal07 (changes requested):
I have left a few forward-looking design thoughts for the next iteration which are completely non-blocking.

However, we do have one critical bug in the version comparison logic (detailed in my inline comment on `gke_vulnerabilities_check.go`) where strict SemVer treats GKE patch suffixes as pre-releases, causing a silent security bypass for raw version strings.

pkg/validators/security-advisories.json
kadupoornima: Thanks for this suggestion! I had considered this but prioritized a simpler and faster implementation for now. Created a ticket for this proposal.

pkg/validators/gke_vulnerabilities_check.go
kadupoornima: Done, thanks for finding this!

cmd/create.go
kadupoornima: Done.

kadupoornima:
> I have left a few forward-looking design thoughts for the next iteration which are completely non-blocking.
> 
> However, we do have one critical bug in the version comparison logic (detailed in my inline comment on `gke_vulnerabilities_check.go`) where strict SemVer treats GKE patch suffixes as pre-releases, causing a silent security bypass for raw version strings.

Thanks for your review comments! I have resolved them, PTAL.

shubpal07 (approved):
Thanks for addressing all the comments @kadupoornima 
LGTM

PR #5749, opened 2026-06-08
Implement GKE Node Auto-Provisioning (NAP) and unified compute consumption options in gcluster job submit

SwarnaBharathiMantena:
### Overview
This PR implements native support for GKE Node Auto-Provisioning (NAP) and introduces unified compute consumption models (spot, on-demand, and reservation) to the gcluster job submit workflow.

Previously, gcluster job submit assumed static, pre-existing node pools and had limited options for targeting provisioning models (e.g. DWS Flex, which is incompatible with GKE NAP). This change allows workloads to dynamically scale up hardware capacity matching GKE Cluster Autoscaler limits while supporting Spot and GCE Reservation consumption models with strict pre-flight validation.

### Key Changes
1. **CLI & Unified Flags**
Introduced --gke-nap-provisioning flag (allowed values: on-demand, spot, reservation).
Introduced --gke-nap-reservation companion flag to target specific GCE reservations when using the reservation provisioning model.
Added strict CLI-level flag compatibility checks and validations (e.g., rejecting reservation names when spot is selected).

2. **GKE Orchestration & Dynamic Kueue Scheduling**
Modified GKE orchestrator to auto-discover GKE autoscaling limits and map hardware families to dynamic Kueue ResourceFlavors and ClusterQueues, enabling Kueue queue routing even when matching physical node pools do not yet exist.
Updated renderClusterQueue to implement a two-pass rendering structure. This guarantees that all flavors under a resource group cover the exact same resource list (injecting nominalQuota: 0 for missing resources to prevent Kueue schema validation failures).

3. **Pre-flight Hybrid & Static Validations**
Implemented validateConsumptionForStaticCluster and isNAPEnabledForMachineType to support hybrid clusters:
If the requested hardware type is covered by GKE NAP autoprovisioning limits, validation is bypassed to allow the GKE autoscaler to dynamically allocate the nodes.
If GKE Node Auto-Provisioning (NAP) flags (`--consumption-model` or `--reservation-name`) are requested, the system validates that the target machine type is covered by the GKE NAP resource limits configured on the cluster. If the machine type is not covered by the autoprovisioning limits, the submission is rejected early with a clear error message. If NAP flags are not provided, or NAP is disabled, the validation is skipped to allow jobs to schedule directly onto matching static node pools.

4. **Manifest Templating**
Added dynamic generation of GKE Spot and Reservation tolerations, matching GKE's default taint configurations.
Configured node selectors in workload templates to map to GKE provisioned node labels.

### Refactoring
**Separation of Concerns**: Encapsulated all GKE NAP-specific logic, limit key mapping, GKE-specific tolerations generators, and static capacity validators into a dedicated file: `pkg/orchestrator/gke/nap.go`, greatly simplifying future debugging of GKE autoprovisioning constraints.

### Documentation
Updated `docs/gcluster_job_guide.md` to document the new flags and provide real-world Spot and Reservation job submission examples on GKE NAP clusters.

### Verification & Testing
1. **Unit Tests**
All GKE package tests compile and pass successfully (PASS in 0.59s):
* TestValidateConsumptionForStaticCluster: Asserts pre-flight validations against static pool constraints, NAP limits, and non-NAP clusters.
* TestSubmitCmd_...: Asserts CLI flag parser constraints.
* TestRenderClusterQueue_NAP: Asserts that Kueue nominal quotas are successfully populated from autoscaling limits.

2. **Manual Verification**
Created a real-world testing directory under containing a testing blueprint and a step-by-step verification guide.


pkg/orchestrator/gke/infra_manager.go
```diff
@@ -313,17 +313,105 @@ func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) er
...
+				}
+				if resName == "nvidia.com/gpu" && !strings.Contains(strings.ToLower(fname), "nvidia") {
+					return 0
+				}
+				if resName == "memory" {
+					return fmt.Sprintf("%dGi", limit/1024) // limit is in MB in GKE describe
+				}
+				return limit
+			}
+		}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed. Refactored getNominalQuota to strip the flavor- prefix to query specific hardware limit keys (with ct and tpu family checking) and added fallback checks to generic keys if specific limits are unconfigured.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,196 @@
...
+	cap, err := g.FetchMachineCapabilities(resolvedType, zone)
+	if err == nil && len(cap.Accelerators) > 0 {
+		key := getGPULimitKey(resolvedType, cap.Accelerators[0].Type)
+		if key != "nvidia.com/gpu" {
+			return g.napLimits[key] > 0
+		}
+		return g.napLimits["nvidia.com/gpu"] > 0
+	}
+
+	return g.napLimits["cpu"] > 0
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed. Updated isNAPEnabledForMachineType to fail-fast and return false immediately on any capability fetch error, preventing invalid configurations from falling back to CPU checks.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,178 @@
...
+					Key:      t.Key,
+					Value:    t.Value,
+					Effect:   corev1.TaintEffect(t.Effect),
+					Operator: corev1.TolerationOpEqual,
+				})
+			}
+		}
+	}
+	return tolerations
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Updated resolveReservationTolerations to always append the standard cloud.google.com/reservation-name toleration first so GKE NAP scheduling works even if the node pool does not exist yet.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (g *GKEOrchestrator) resolveTolerations(acceleratorType string, consumptionModel string, reservationName string) (string, error) {
+	tolerations := GetTolerations(acceleratorType)
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Changed resolveTolerations to copy the slice returned by GetTolerations before appending new tolerations to avoid mutating any shared underlying array.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -399,6 +399,20 @@ func (g *GKEOrchestrator) populateClusterMetadata(job *orchestrator.JobDefinitio
...
+	for _, rl := range clusterDesc.Autoscaling.ResourceLimits {
+		resName := rl.ResourceType
+		g.napLimits[resName] = rl.Maximum
+
+		// Map generic keys for backwards compatibility and generic resource lookups
+		if resName == "gpu" || strings.HasPrefix(resName, "nvidia") {
+			g.napLimits["nvidia.com/gpu"] = rl.Maximum
+		} else if strings.HasPrefix(resName, "tpu") {
+			g.napLimits["google.com/tpu"] = rl.Maximum
+		}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Updated generic key mapping to only update the generic limit if the new value is greater than the existing value, preventing overwrites.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,191 @@
...
+	// Always add the standard GKE reservation toleration to support NAP where the node pool may not exist yet
+	tolerations = append(tolerations, corev1.Toleration{
+		Key:      "cloud.google.com/reservation-name",
+		Operator: corev1.TolerationOpEqual,
+		Value:    extractShortReservationName(reservationName),
+		Effect:   corev1.TaintEffectNoSchedule,
+	})
+
+	for _, np := range g.clusterDesc.NodePools {
+		if strings.EqualFold(np.Config.MachineType, machineType) && np.Config.Labels["cloud.google.com/reservation-name"] == reservationName {
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Comparing short reservation names instead of raw names to handle both short reservation names and full resource URIs.

pkg/orchestrator/gke/nap_test.go
```diff
@@ -0,0 +1,228 @@
...
+// limitations under the License.
+
+package gke
+
+import (
+	"slices"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+)
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Imported the strings package in nap_test.go to support substring assertions.

pkg/orchestrator/gke/nap_test.go
```diff
@@ -0,0 +1,228 @@
...
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			for _, sub := range tt.wantContains {
+				if !slices.ContainsFunc(tt.wantContains, func(s string) bool {
+					return true // simple check for demonstration/assertions
+				}) {
+					t.Errorf("expected output to contain %q, got %q", sub, got)
+				}
+			}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Replaced slices.ContainsFunc with strings.Contains to correctly assert that the generated YAML contains the expected substrings.

pkg/orchestrator/gke/nap_test.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		t.Fatalf("unexpected error on second call: %v", err)
+	}
+
+	// The second result should ONLY have TPU toleration, NOT spot
+	if slices.ContainsFunc([]string{got2}, func(s string) bool {
+		return slices.Contains([]string{got2}, "spot")
+	}) {
+		t.Errorf("second call unexpectedly contains 'spot'. got1: %q, got2: %q", got1, got2)
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Replaced slices.Contains with strings.Contains to correctly assert that the generated YAML does not contain "spot".

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,196 @@
...
+
+	var configuredLimits []string
+	for k, v := range g.napLimits {
+		if v > 0 {
+			configuredLimits = append(configuredLimits, k)
+		}
+	}
+	sort.Strings(configuredLimits)
+	return fmt.Errorf("workload submission rejected. Compute type %q is not configured within your cluster's Node Auto-Provisioning (NAP) limits. Configured limits on cluster: %s", job.ComputeType, strings.Join(configuredLimits, ", "))
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Updated validateConsumptionForStaticCluster to fall back to validating the requested consumption model against matching static node pools when the machine type is not covered by GKE NAP limits, aligning it with the PR description.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -399,6 +399,24 @@ func (g *GKEOrchestrator) populateClusterMetadata(job *orchestrator.JobDefinitio
...
+		if resName == "gpu" || strings.HasPrefix(resName, "nvidia") {
+			if rl.Maximum > g.napLimits["nvidia.com/gpu"] {
+				g.napLimits["nvidia.com/gpu"] = rl.Maximum
+			}
+		} else if strings.HasPrefix(resName, "tpu") {
+			if rl.Maximum > g.napLimits["google.com/tpu"] {
+				g.napLimits["google.com/tpu"] = rl.Maximum
+			}
+		}
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Separated raw limit population and generic GPU/TPU key mapping into two distinct passes to prevent loop-order dependency where specific maximum limits could be overwritten.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return false
+	}
+	if len(cap.Accelerators) > 0 {
+		key := getGPULimitKey(resolvedType, cap.Accelerators[0].Type)
+		if key != "nvidia.com/gpu" {
+			return g.napLimits[key] > 0
+		}
+		return g.napLimits["nvidia.com/gpu"] > 0
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Modified isNAPEnabledForMachineType to check both the specific limit key and the generic GPU/TPU limit fallback keys to prevent false negatives.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -399,6 +399,25 @@ func (g *GKEOrchestrator) populateClusterMetadata(job *orchestrator.JobDefinitio
...
+		if resName == "gpu" || strings.HasPrefix(resName, "nvidia") {
+			if maxVal > g.napLimits["nvidia.com/gpu"] {
+				g.napLimits["nvidia.com/gpu"] = maxVal
+			}
+		} else if strings.HasPrefix(resName, "tpu") {
+			if maxVal > g.napLimits["google.com/tpu"] {
+				g.napLimits["google.com/tpu"] = maxVal
+			}
+		}
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Refactored the mapping loop to iterate over the static resourceLimits slice instead of the map itself to prevent map mutation iteration issues in Go.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,234 @@
...
+		return "nvidia-b200"
+	}
+	if strings.Contains(m, "a4x-highgpu") {
+		return "nvidia-gb200"
+	}
+	if strings.Contains(m, "a2-high") || strings.Contains(m, "a2-mega") || strings.Contains(m, "a2-ultra") || strings.Contains(accelLabel, "a100") {
+		return "nvidia-tesla-a100"
+	}
+	return "nvidia.com/gpu"
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed: Converted accelLabel to lowercase at the start of getGPULimitKey, changed the signature to return (string, error), and implemented a fail-fast validation error for unrecognized accelerator labels.

pkg/orchestrator/gke/infra_manager.go
```diff
@@ -312,18 +312,143 @@ func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) er
...
+	mainCoveredResourcesMap, pathwaysCoveredResourcesMap := g.calculateCoveredResources()
 
+	// Pass 2: Build flavor lists with matched resources
+	var mainFlavors []map[string]interface{}
 	var pathwaysFlavors []map[string]interface{}
-	pathwaysCoveredResourcesMap := make(map[string]bool)
 
-	for name, fc := range g.capacity.Flavors {
-		resources, isPathways := g.buildFlavorResources(name, fc, mainCoveredResourcesMap, pathwaysCoveredResourcesMap)
+	for fname, fc := range g.capacity.Flavors {
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Fixed. Extracted and sorted capacity flavor names alphabetically before iterating in renderClusterQueue to guarantee deterministic output.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,253 @@
...
+					Key:      t.Key,
+					Value:    t.Value,
+					Effect:   corev1.TaintEffect(t.Effect),
+					Operator: corev1.TolerationOpEqual,
+				})
+			}
+		}
+	}
+	return tolerations
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Added a seenTaints tracking map to `resolveReservationTolerations` to prevent duplicate toleration injections if multiple static node pools match the reservation.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,253 @@
...
+func getSubmissionRejectError(job *orchestrator.JobDefinition) error {
+	switch job.GKENAPProvisioning {
+	case "spot":
+		return fmt.Errorf("workload submission rejected. You requested the '--gke-nap-provisioning=spot' option for compute type %q, but the cluster's static node pools for this hardware are configured exclusively as Standard/On-Demand.\nRemediation: Please re-submit your job without the '--gke-nap-provisioning=spot' setting, or enable Node Auto-Provisioning (NAP) limits for this hardware on your cluster to allow dynamic scale-up of Spot resources", job.ComputeType)
+	case "reservation":
+		return fmt.Errorf("workload submission rejected. You requested the '--gke-nap-provisioning=reservation' with reservation name %q for compute type %q, but no static node pools matching this hardware are configured to consume this reservation", job.GKENAPReservation, job.ComputeType)
+	default:
+		return fmt.Errorf("workload submission rejected. No active node pools found matching your consumption model constraints")
+	}
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Added an explicit "on-demand" case to `getSubmissionRejectError` with a clear remediation message if Standard/On-Demand is requested but only Spot is available.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -421,6 +424,28 @@ func (g *GKEOrchestrator) populateClusterMetadata(job *orchestrator.JobDefinitio
...
+		if resName == "gpu" || strings.HasPrefix(resName, "nvidia") {
+			if maxVal > limits["nvidia.com/gpu"] {
+				limits["nvidia.com/gpu"] = maxVal
+			}
+		} else if strings.HasPrefix(resName, "tpu") {
+			if maxVal > limits["google.com/tpu"] {
+				limits["google.com/tpu"] = maxVal
+			}
+		}
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Switched from strings.HasPrefix to strings.Contains for resource name classification in `gke_job_orchestrator.go` and `calculateClusterCapacity` to robustly match GKE resource formats.

cmd/job/submit.go
```diff
@@ -91,6 +94,21 @@ and JobSet/Kueue specific configurations like workload name, queue, nodes, and r
 			return err
 		}
 
+		gkeNapProvisioning = strings.ToLower(gkeNapProvisioning)
```
Neelabh94: To keep the interface clean and extensible, let's extract this into a helper function and call only the function here.

docs/gcluster_job_guide.md
```diff
@@ -298,6 +298,33 @@ Use `--queue` to submit the job to a specific Kueue LocalQueue.
 
 (Note: You would need to ensure a Kueue `LocalQueue` named `my-local-queue` is configured on your cluster.)
 
+### Example 7: Targeting Provisioning Models (Spot/Reservation) on GKE NAP Clusters
+
+> [!NOTE]
+> ### GKE Node Auto-Provisioning (NAP) Only
+> The `--gke-nap-provisioning` and `--gke-nap-reservation` flags are supported **only** on GKE clusters with Node Auto-Provisioning (NAP) enabled. They cannot be used on static GKE clusters where NAP is disabled, and doing so will result in an immediate pre-flight validation error.
```
Neelabh94, line 304: We need to further expand the documentation by adding NAP specific details under GKE advanced feature section in the doc. This section already contains info about Pathways and DynamicSlicing support.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			return false, err
+		}
+		return g.napLimits[key] > 0 || g.napLimits["nvidia.com/gpu"] > 0, nil
+	}
+
+	return g.napLimits["cpu"] > 0, nil
+}
+
+func getTPULimitKey(machineType string) string {
```
Neelabh94: These mappings are already available in pkg/config/machine_mappings.json, can we re-use them?

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,260 @@
...
+	if strings.Contains(m, "v5p") || strings.Contains(m, "ct5p") {
+		return "tpu-v5p-slice"
+	}
+	if strings.Contains(m, "v4") || strings.Contains(m, "ct4p") {
+		return "tpu-v4-podslice"
+	}
+	return "google.com/tpu"
+}
+
+func getGPULimitKey(machineType string, accelLabel string) (string, error) {
```
Neelabh94: same as above, let's try to reuse (and if required expand) the mappings in pkg/config/machine_mappings.json

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,260 @@
...
+	if strings.Contains(m, "a4-highgpu") {
+		return "nvidia-b200", nil
+	}
+	if strings.Contains(m, "a4x-highgpu") {
+		return "nvidia-gb200", nil
+	}
+	if strings.Contains(m, "a2-high") || strings.Contains(m, "a2-mega") || strings.Contains(m, "a2-ultra") || strings.Contains(accel, "a100") {
+		return "nvidia-tesla-a100", nil
+	}
+	if accel != "" {
```
Neelabh94: Why this check only for GPUs? Also, if the name is resolved successfully do we still need this check here?

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return err
+	}
+	if isNAP {
+		return nil // Validated: GKE NAP can dynamically autoprovision this machine type
+	}
+
+	// Fallback: Check if there is a matching static node pool that satisfies the requested consumption model.
+	matchedNodePoolFound := false
+	for _, np := range g.clusterDesc.NodePools {
```
Neelabh94: I am little uncomfortable with this. IIUC, the user can pass gke-nap related flags targeting a machine type that does not have NAP enabled, but in case we find a matching static node-pool, we will silently fallback to that and schedule the workload instead of failing and letting the customer know that the requested NAP machine is not NAP enabled?

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,260 @@
...
+		return "", fmt.Errorf("failed to marshal tolerations: %w", err)
+	}
+	return g.indentYaml(string(b), 16), nil
+}
+
+func extractShortReservationName(resName string) string {
+	if strings.Contains(resName, "/") {
+		parts := strings.Split(resName, "/")
+		return parts[len(parts)-1]
+	}
```
Neelabh94, line 116: Do we tolerate a use rpassing sub-blocks as part of the reservation name?

pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -126,9 +124,13 @@ func (g *GKEOrchestrator) verifyDynamicSlicingActive(opts ManifestOptions) (bool
 	}
 
 	// Check discovered node pools for dynamic-slicing
-	requestedMachineName, err := g.resolveMachineName(opts.ComputeType)
-	if err != nil {
-		return false, err
+	requestedMachineName := opts.MachineType
+	if requestedMachineName == "" {
+		var err error
```
Neelabh94, line 129: I don't think this is needed here. ComputeType is a required variable and the cli fails at the top itself in case it is not supplied

pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -257,8 +259,11 @@ func (g *GKEOrchestrator) calculateGCPMachineResourceLimits(opts ManifestOptions
 	machineName := opts.MachineType
 
 	count, err := g.FetchMachineCapacity(machineName, opts.ClusterLocation)
-	if err != nil {
-		return "", "", "", "", fmt.Errorf("failed to resolve machine type %s: %w", machineName, err)
+	if err != nil || count == 0 {
```
Neelabh94: I doubt this change is needed, FetchMachineCapacity is called with the resolved machine name  already. An error should only indicate that capacity fetching failed.

Neelabh94 (commented):
Left a few comments, PTAL

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: Right, this is not as intended. Fixed it.
The hybrid scenario I considered here is that even if there is static nodepool, the user might be targetting NAP for specific reason. We should not be automatically routing it as they is probably against what they are trying to do. If we fail with a proper message, it is easy to either edit the job submit command to use static nodepool, or update the cluster to enable NAP for that specific machine type.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: TPUs don't use GKE accelerator labels, but I will simplify this function by removing redundant checks if capacity is fully resolved.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: GKE Node Auto-Provisioning (NAP) only supports reservation-level scheduling on pods via the `cloud.google.com/reservation-name` label/taint. GKE NAP does not support scheduling onto a specific reservation block or sub-block via pod template selectors.

I added this info to the `docs/gcluster_job_guide.md` file.

pkg/orchestrator/gke/resource_resolver.go
SwarnaBharathiMantena: Although Cobra makes it required in the CLI, keeping this check ensures defensive safety for programmatic orchestrator calls.

cmd/job/submit.go
SwarnaBharathiMantena: I have extracted this validation logic to the validateGKENAPFlags() helper function in submit.go.

docs/gcluster_job_guide.md
SwarnaBharathiMantena: I have expanded the Advanced GKE Infrastructure Features section (Section 8.3) in docs/gcluster_job_guide.md to document GKE NAP.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: I have updated nap.go to resolve limit keys dynamically using the centralized mappings in machine_mappings.json via config.GetTPULimitKey().

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: I have removed the static node pool fallback; the command now rejects the submission immediately if NAP limits do not cover the requested compute type.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: Removed the local helper functions in nap.go and offloaded all GPU key lookup and validation logic to config.GetGPULimitKey.

pkg/config/hardware.go
```diff
@@ -506,3 +522,60 @@ func CheckTopologyContainment(requested, container string, accelType string) (bo
...
 	return true, nil
 }
+
+// GetTPULimitKey resolves a machine type to its corresponding TPU limit key.
+func GetTPULimitKey(machineType string) string {
+	m := strings.ToLower(machineType)
+
+	var keys []string
+	for k := range parsedMappings.MachineFamilyToLabelMap {
+		if strings.Contains(k, "ct") || strings.Contains(k, "v") {
```
Neelabh94: Is it possible we use the `isTPU` function here ?

pkg/config/hardware.go
```diff
@@ -506,3 +522,60 @@ func CheckTopologyContainment(requested, container string, accelType string) (bo
...
+}
+
+// GetGPULimitKey resolves a machine type and accelerator label to its corresponding GPU limit key.
+func GetGPULimitKey(machineType string, accelLabel string) (string, error) {
+	m := strings.ToLower(machineType)
+	accel := strings.ToLower(accelLabel)
+
+	var keys []string
+	for k := range parsedMappings.MachineFamilyToLabelMap {
+		if !strings.Contains(k, "ct") && !strings.Contains(k, "v") {
```
Neelabh94: What if the machine type is a cpu machine?, we can re-use some existing info about the machine type.

Neelabh94 (commented):
Thanks for resolving the earlier comments.
Some minor nits this time, otherwise LGTM.
Will trigger /gemini review as well to ensure full coverage.

Thanks again for the effort

pkg/config/hardware.go
SwarnaBharathiMantena: Used IsTPU(k) to dynamically identify and group TPU machine families inside init().

pkg/config/hardware.go
```diff
@@ -506,3 +522,60 @@ func CheckTopologyContainment(requested, container string, accelType string) (bo
...
+	if strings.Contains(accel, "l4") {
+		return "nvidia-l4", nil
+	}
+	if strings.Contains(accel, "a100") {
+		return "nvidia-tesla-a100", nil
+	}
+
+	if accel != "" {
+		return "", fmt.Errorf("unknown accelerator label: %q", accelLabel)
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Expanded the fallbacks in GetGPULimitKey to include modern GPU architectures: h100 ("nvidia-h100-80gb"), h200 ("nvidia-h200-141gb"), and b200 ("nvidia-b200").

pkg/config/hardware.go
SwarnaBharathiMantena: Added validation in GetGPULimitKey to reject TPU types and known Google Compute Engine CPU families (e.g., n1-, n2-, e2-, etc.) and return an error.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -498,6 +523,38 @@ func (g *GKEOrchestrator) calculateClusterCapacity(clusterDesc gkeCluster, locat
...
+				if _, ok := flavors[flavorName]; !ok {
+					flavors[flavorName] = FlavorCapacity{
+						NodeLabels: map[string]string{
+							"cloud.google.com/gke-accelerator": resName,
+						},
+					}
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Updated calculateClusterCapacity to set NodeLabels to nil in Kueue's ResourceFlavors when processing generic TPU/GPU limits (google.com/tpu, nvidia.com/gpu).

pkg/config/hardware.go
```diff
@@ -506,3 +522,60 @@ func CheckTopologyContainment(requested, container string, accelType string) (bo
...
+
+	var keys []string
+	for k := range parsedMappings.MachineFamilyToLabelMap {
+		if strings.Contains(k, "ct") || strings.Contains(k, "v") {
+			keys = append(keys, k)
+		}
+	}
+	sort.Slice(keys, func(i, j int) bool {
+		return len(keys[i]) > len(keys[j])
+	})
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Declared package-level variables sortedTPUMachineFamilies and sortedGPUMachineFamilies, populating and sorting them only once inside the package init() function.

pkg/orchestrator/gke/infra_manager.go
```diff
@@ -312,18 +312,150 @@ func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) er
...
+			}
+		}
+	}
+	return mainMap, pathwaysMap
+}
+
+func isTPUFlavor(fname string) bool {
+	lower := strings.ToLower(fname)
+	return strings.Contains(lower, "tpu") || strings.Contains(lower, "ct")
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: The logic is as expected. No changes required.

pkg/config/hardware.go
```diff
@@ -506,3 +540,64 @@ func CheckTopologyContainment(requested, container string, accelType string) (bo
...
+
+// GetGPULimitKey resolves a machine type and accelerator label to its corresponding GPU limit key.
+func GetGPULimitKey(machineType string, accelLabel string) (string, error) {
+	m := strings.ToLower(machineType)
+	accel := strings.ToLower(accelLabel)
+
+	// Reject TPU machine types statically
+	if IsTPU(machineType) {
+		return "", fmt.Errorf("machine type %q is a TPU machine type, not a GPU machine type", machineType)
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: `machineType` is case sensitive. Working as expected.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,190 @@
...
+// It handles standard URIs, simple names, and paths containing reservationBlocks/reservationSubBlocks.
+// E.g.,
+// - "my-res" -> "my-res"
+// - "projects/my-project/reservations/my-res" -> "my-res"
+// - "projects/my-project/reservations/my-res/reservationBlocks/block-1/reservationSubBlocks/subblock-2" -> "my-res"
+// - "my-res/reservationBlocks/block-1/reservationSubBlocks/subblock-2" -> "my-res"
+func extractShortReservationName(resName string) string {
+	if !strings.Contains(resName, "/") {
+		return resName
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: We will align with "GCP reservation names are strictly lowercase" and not support uppercase reservation names.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,190 @@
...
+		"cloud.google.com/reservation-name": true,
+	}
+
+	shortResName := extractShortReservationName(reservationName)
+	for _, np := range g.clusterDesc.NodePools {
+		lblVal := np.Config.Labels["cloud.google.com/reservation-name"]
+		if lblVal == "" {
+			continue
+		}
+		if strings.EqualFold(np.Config.MachineType, machineType) && extractShortReservationName(lblVal) == shortResName {
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Done

pkg/orchestrator/gke/infra_manager.go
```diff
@@ -312,18 +312,150 @@ func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) er
...
+	if resName == "google.com/tpu" && !isTPUFlavor(fname) {
+		return 0, true
+	}
+	if resName == "nvidia.com/gpu" && !isGPUFlavor(fname) {
+		return 0, true
+	}
+	if resName == "memory" {
+		return fmt.Sprintf("%dGi", limit/1024), true // limit is in MB in GKE describe
+	}
+	return limit, true
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Addressed

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// NAP flags were requested. Validate strictly against GKE NAP limits.
+	isNAP, err := g.isNAPEnabledForMachineType(job.MachineType, job.ClusterLocation)
+	if err != nil {
+		return err
+	}
+	if !isNAP {
+		return g.getConfiguredLimitsError(job.ComputeType)
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: The code and test is as expected. The PR description needs an update.

.pre-commit-config.yaml
```diff
@@ -143,7 +143,7 @@ repos:
 #   hooks:
 #   - id: go-critic
 - repo: https://github.com/golangci/golangci-lint
-  rev: v1.63.4
+  rev: v2.4.0
```
SwarnaBharathiMantena, line 146: I am unable to fix the precommit failures without this fix. Is this acceptable update? @Neelabh94

Neelabh94: Yes, this is fine. This is supported for Go 1.24.6

Can you instead raise a separate PR for this? I think this is failing for everyone!

SwarnaBharathiMantena: This is added as part of #5803

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return nil
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: The code is working as expected. I updated the PR description.

pkg/orchestrator/gke/infra_manager.go
```diff
@@ -312,18 +312,144 @@ func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) er
...
+	// Only apply TPU/GPU limit to matching flavors
+	if resName == "google.com/tpu" && !isTPUFlavor(fname) {
+		return 0, true
+	}
+	if resName == "nvidia.com/gpu" && !isGPUFlavor(fname) {
+		return 0, true
+	}
+	if resName == "memory" {
+		return fmt.Sprintf("%dG", limit), true // limit is in GB in GKE describe
+	}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: We verified GKE's Node Auto-Provisioning (NAP) API behavior on a live GKE cluster:

When setting --max-memory 100 (which is 100 GB in gcloud), GKE's describe API returns "maximum": "100" under resourceLimits for memory.
This indicates GKE's autoscaling API provides the limits in Gigabytes (GB) directly, not Megabytes (MB).
Dividing the API limit by 1024 would result in 0Gi for a standard 100 GB limit, which would set the ClusterQueue nominal memory capacity to zero and block all workloads from scheduling.

We have reverted the format to %dGi (without division) to maintain consistency with the develop branch's existing capacity units while accurately representing GKE's NAP limit.

cmd/job/submit.go
```diff
@@ -352,10 +363,28 @@ func validateBuildContext() error {
...
 	}
 	return nil
 }
+
+func validateGKENAPFlags() error {
+	gkeNapProvisioning = strings.ToLower(gkeNapProvisioning)
+	if gkeNapProvisioning != "" {
+		validModels := []string{"on-demand", "spot", "reservation"}
+		if !slices.Contains(validModels, gkeNapProvisioning) {
+			return fmt.Errorf("invalid --gke-nap-provisioning %q. Allowed values: %s", gkeNapProvisioning, strings.Join(validModels, ", "))
```
Neelabh94: nit: can be rephrased for clarity
"invalid value %q for --gke-nap-provisioning. Allowed ..."

cmd/job/submit.go
```diff
@@ -352,10 +363,28 @@ func validateBuildContext() error {
...
+		}
+
+		if gkeNapProvisioning == "reservation" && gkeNapReservation == "" {
+			return fmt.Errorf("--gke-nap-reservation is required when --gke-nap-provisioning=reservation")
+		}
+	}
+	if gkeNapProvisioning != "reservation" && gkeNapReservation != "" {
+		return fmt.Errorf("--gke-nap-reservation is only valid when --gke-nap-provisioning=reservation")
```
Neelabh94: nit: can be rephrased for clarity
--gke-nap-reservation should only be provided/ passed  when

docs/gcluster_job_guide.md
```diff
@@ -298,6 +298,33 @@ Use `--queue` to submit the job to a specific Kueue LocalQueue.
 
 (Note: You would need to ensure a Kueue `LocalQueue` named `my-local-queue` is configured on your cluster.)
 
+### Example 7: Targeting Provisioning Models (Spot/Reservation) on GKE NAP Clusters
+
+> [!NOTE]
+> ### GKE Node Auto-Provisioning (NAP) Only
```
Neelabh94: nit: This heading does not convey much, can we rephrase or remove this?

pkg/config/hardware.go
```diff
@@ -538,3 +570,78 @@ func Is3DTorusTPU(machineType string) bool {
 	}
 	return false
 }
+
+// GetTPULimitKey resolves a machine type to its corresponding TPU limit key.
+func GetTPULimitKey(machineType string) string {
```
Neelabh94: This is probably not needed, we have a similar function `GenerateGKENodeSelectorLabel`  in `pkg/orchestrator/gke/gke_job_orchestrator.go`. 
Please see if that can be used instead of defining this new function.

pkg/config/hardware.go
```diff
@@ -538,3 +570,78 @@ func Is3DTorusTPU(machineType string) bool {
...
+	for _, k := range sortedTPUMachineFamilies {
+		if strings.Contains(m, k) {
+			return parsedMappings.MachineFamilyToLabelMap[k]
+		}
+	}
+	return "google.com/tpu"
+}
+
+// GetGPULimitKey resolves a machine type and accelerator label to its corresponding GPU limit key.
+func GetGPULimitKey(machineType string, accelLabel string) (string, error) {
```
Neelabh94: same as above

pkg/config/hardware.go
```diff
@@ -538,3 +570,78 @@ func Is3DTorusTPU(machineType string) bool {
...
+	}
+
+	if accel != "" {
+		return "", fmt.Errorf("unknown accelerator label: %q", accelLabel)
+	}
+
+	return "nvidia.com/gpu", nil
+}
+
+func resolveGPULabelFallback(accel string) (string, bool) {
```
Neelabh94: if we decide to reuse (and possible expand) `GenerateGKENodeSelectorLabel`, we won't need this map.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -422,6 +425,28 @@ func (g *GKEOrchestrator) populateClusterMetadata(job *orchestrator.JobDefinitio
 	return nil
 }
 
+func parseNAPLimits(autoscaling gkeClusterAutoscaling) map[string]int64 {
```
Neelabh94: Do you think it might be better if we move these NAP specific functions in nap.go?

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -509,6 +538,51 @@ func (g *GKEOrchestrator) calculateClusterCapacity(clusterDesc gkeCluster, locat
 	return capacity, nodePoolSAs, nil
 }
 
+func (g *GKEOrchestrator) populateNAPFlavors(flavors map[string]FlavorCapacity) error {
```
Neelabh94: same as above, move to nap.go?

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		return g.napLimits[key] > 0 || g.napLimits["nvidia.com/gpu"] > 0, nil
+	}
+
+	return g.napLimits["cpu"] > 0, nil
+}
+
+func (g *GKEOrchestrator) checkNAPFlagsSupported(hasNAPFlags bool, job *orchestrator.JobDefinition) error {
+	if !g.napEnabled && hasNAPFlags {
+		return fmt.Errorf("GKE NAP provisioning options (--gke-nap-provisioning=%q, --gke-nap-reservation=%q) are only supported on GKE clusters with Node Auto-Provisioning (NAP) enabled. The current cluster does not have NAP enabled.\nRemediation: Enable Node Auto-Provisioning on your cluster to use these options, or submit your job without them", job.GKENAPProvisioning, job.GKENAPReservation)
```
Neelabh94: we don't actually use the "=" sign while passing these flags, consider removing them from here too

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,204 @@
...
+		if v > 0 {
+			configuredLimits = append(configuredLimits, k)
+		}
+	}
+	sort.Strings(configuredLimits)
+	return fmt.Errorf("workload submission rejected. Compute type %q is not configured within your cluster's Node Auto-Provisioning (NAP) limits. Configured limits on cluster: %s", computeType, strings.Join(configuredLimits, ", "))
+}
+
+func (g *GKEOrchestrator) validateConsumptionForStaticCluster(job *orchestrator.JobDefinition) error {
+	hasNAPFlags := job.GKENAPProvisioning != "" || job.GKENAPReservation != ""
```
Neelabh94: We might be okay with just checking the job.GKENAPProvisioning != "", since job.GKENAPReservation cannot be empty only if job.GKENAPProvisioning == reservation.

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+	if !isNAP {
+		return g.getConfiguredLimitsError(job.ComputeType)
+	}
+
+	return nil
+}
+
+func (g *GKEOrchestrator) resolveReservationTolerations(machineType, reservationName string) []corev1.Toleration {
```
Neelabh94: this function might be better suited for manifest_generator.go, please consider moving it there

pkg/orchestrator/gke/nap.go
```diff
@@ -0,0 +1,204 @@
...
+					Effect:   effect,
+					Operator: corev1.TolerationOpEqual,
+				})
+			}
+		}
+	}
+	return tolerations
+}
+
+func (g *GKEOrchestrator) resolveTolerations(acceleratorType string, consumptionModel string, reservationName string, indent int) (string, error) {
```
Neelabh94: saem as above, please consider moving it to manifest_generator.go

pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -417,20 +419,29 @@ func (g *GKEOrchestrator) resolveHardwareRequirements(job *orchestrator.JobDefin
 		return JobProfile{}, false, false, err
 	}
 	job.MachineType = machineName
-
 	if config.IsTPU(machineName) {
 		isDynamicSlicing, isStaticSlicing, err = g.resolveTPURequirements(job)
 		if err != nil {
 			return JobProfile{}, false, false, err
 		}
+		if g.napEnabled {
```
Neelabh94: might want to rebase this and see how we want to handle static sub-slicing

Neelabh94 (changes requested):

pkg/config/hardware.go
SwarnaBharathiMantena: Updated

pkg/config/hardware.go
SwarnaBharathiMantena: Updated

pkg/config/hardware.go
SwarnaBharathiMantena: Updated

pkg/orchestrator/gke/gke_job_orchestrator.go
SwarnaBharathiMantena: Agree, I will separate this out into nap.go

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: Agree

pkg/orchestrator/gke/resource_resolver.go
SwarnaBharathiMantena: You mean static subslicing in combination with NAP?

Neelabh94 (approved):
LGTM! Please make sure you test this once before submitting.

pkg/orchestrator/gke/resource_resolver.go
SwarnaBharathiMantena: This will be addressed in a followup PR.

PR #5750, opened 2026-06-08
Loosen Terraform version constraints in modules and add version warning

parulbajaj01:
This PR loosens the strict Terraform version constraint (= 1.12.2) across all modules in the repository to required_version = ">= 1.12.2". It also updates the cluster-toolkit's dependency resolver to perform pre-flight version checks, warning users when they use non-standard versions rather than blocking execution.

### Submission Checklist

### DIFF
--- a/pkg/dependencies/resolver.go
+++ b/pkg/dependencies/resolver.go
@@ -18,6 +18,7 @@ package dependencies
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
@@ -55,7 +56,7 @@ func PatchPath() error {
 	}
 
 	currentPath := os.Getenv("PATH")
-	newPath := currentPath + string(os.PathListSeparator) + tfCacheDir + string(os.PathListSeparator) + packerCacheDir
+	newPath := tfCacheDir + string(os.PathListSeparator) + packerCacheDir + string(os.PathListSeparator) + currentPath
 	os.Setenv("PATH", newPath)
 
 	return nil
@@ -74,8 +75,30 @@ func EnsureDependencies(decision DownloadDecision) error {
 }
 
 func ensureBinary(binaryName, version string, decision DownloadDecision) error {
-	if _, err := exec.LookPath(binaryName); err == nil {
-		return nil
+	path, err := exec.LookPath(binaryName)
+	if err == nil {
+		if binaryName == "terraform" {
+			installedVersion, err := getInstalledTfVersion(path)
+			if err != nil {
+				fmt.Printf("Warning: Could not determine installed terraform version: %v. Proceeding to download recommended version %s.\n", err, version)
+			} else {
+				cmp, err := compareVersions(installedVersion, version)
+				if err != nil {
+					fmt.Printf("Warning: Could not parse installed terraform version %q: %v. Proceeding to download recommended version %s.\n", installedVersion, err, version)
+				} else if cmp == 0 {
+					return nil // exact match
+				} else if cmp > 0 {
+					// installed version is newer
+					fmt.Printf("WARNING: Terraform version %s is currently installed. We recommend using version %s for compatibility with all features.\n", installedVersion, version)
+					return nil // proceed with newer version
+				} else {
+					// installed version is older (incompatible)
+					fmt.Printf("Installed terraform version %s is older than required version %s.\n", installedVersion, version)
+				}
+			}
+		} else {
+			return nil // for packer, just check existence for now
+		}
 	}
 
 	if err := confirmDownload(binaryName, version, decision); err != nil {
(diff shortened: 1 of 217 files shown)


pkg/dependencies/resolver.go
```diff
@@ -18,6 +18,7 @@ package dependencies
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
```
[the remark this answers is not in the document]
parulbajaj01: It's already present on line 26

pkg/dependencies/resolver.go
```diff
@@ -90,6 +113,69 @@ func ensureBinary(binaryName, version string, decision DownloadDecision) error {
...
+//	-1 if v1 < v2
+//	 0 if v1 == v2
+//	 1 if v1 > v2
+func compareVersions(v1, v2 string) (int, error) {
+	v1 = strings.TrimPrefix(v1, "v")
+	v2 = strings.TrimPrefix(v2, "v")
+
+	// Strip pre-release info
+	v1 = strings.Split(v1, "-")[0]
+	v2 = strings.Split(v2, "-")[0]
```
shubpal07: By stripping everything after the hyphen, this implementation treats pre-releases as exactly equal to their stable counterparts. Per [SemVer](https://github.com/Masterminds/semver) 2.0.0, a pre-release version has lower precedence than the stable release (e.g., `1.12.2−beta<1.12.2`). 

While treating them as equal might be an _"safe"_ choice for warnings, but we can consider using a robust library like `github.com/Masterminds/semver` to handle these standard precedence rules automatically.

pkg/dependencies/resolver.go
```diff
@@ -90,6 +113,69 @@ func ensureBinary(binaryName, version string, decision DownloadDecision) error {
...
+
+	var major1, minor1, patch1 int
+	var major2, minor2, patch2 int
+
+	_, err := fmt.Sscanf(v1, "%d.%d.%d", &major1, &minor1, &patch1)
```
shubpal07: NIT:

The use of `Sscanf` with a fixed `%d.%d.%d` pattern is quite rigid. If Terraform reports a version with only two segments (e.g., '1.13'), this will fail and trigger an unnecessary download fallback. A more resilient approach would be to use a standard SemVer library or split the string by . and parse the resulting segments. This ensures compatibility with abbreviated version strings which may sometimes occur in documentation and tools.

pkg/dependencies/resolver.go
```diff
@@ -74,8 +75,30 @@ func EnsureDependencies(decision DownloadDecision) error {
 }
 
 func ensureBinary(binaryName, version string, decision DownloadDecision) error {
-	if _, err := exec.LookPath(binaryName); err == nil {
-		return nil
+	path, err := exec.LookPath(binaryName)
+	if err == nil {
+		if binaryName == "terraform" {
```
shubpal07: The logic here is starting to get quite deep with nested if-else blocks.  We may consider flattening this function using early returns for the non-Terraform or error cases.

Additionally, using a switch block for the cmp results would make the three possible outcomes (equal, newer, or older) much more readable than the current if/else if chain

pkg/dependencies/resolver.go
```diff
@@ -90,6 +113,69 @@ func ensureBinary(binaryName, version string, decision DownloadDecision) error {
...
+		return 1, nil
+	}
+	if patch1 != patch2 {
+		if patch1 < patch2 {
+			return -1, nil
+		}
+		return 1, nil
+	}
+	return 0, nil
+}
```
shubpal07: While this logic could be flattened to reduce nesting (e.g., using direct major1 < major2 checks), a most robust may be to replace this entire block with the Compare method from the `github.com/Masterminds/semver/v3` library https://github.com/Masterminds/semver#basic-comparisons

Manual comparison is often fragile. For instance, the current Sscanf pattern will return an error for abbreviated versions like "1.13". Adopting the library handles these edge cases and pre-release precedence natively, significantly reducing boilerplate while increasing reliability.

```go
// Using the semver library reduces this entire section to:
return v1Parsed.Compare(v2Parsed), nil
```

shubpal07 (commented):
Thanks @parulbajaj01  for putting this PR!

Left some non-blocking technical suggestions below to simplify the control flow and make the version parsing more resilient to varied versioning formats. 

Please feel free to defer some suggestions in the next iterations.

pkg/dependencies/resolver.go
parulbajaj01: thanks for pointing that out. Done!

pkg/dependencies/resolver.go
parulbajaj01: done

shubpal07 (approved):
Thanks for the fast revision @parulbajaj01 !!
LGTM

PR #5776, opened 2026-06-12
Enhance Slurm-GCP resilience to capacity constraints and quota errors

vikramvs-gg:
## Description
This PR makes the Slurm-GCP provisioner resilient to GCP capacity constraints (stockouts) and quota limitations. It prevents wasted budget from partially provisioned jobs and handles failures gracefully by pushing jobs back into the queue without wedging the partition.
### Key Changes
1. **Atomic Provisioning (`minCount` omission):** 
   - `resume.py` now omits `minCount` during the `bulkInsert` call for job-triggered requests. This enforces an "All or Nothing" provisioning behavior from GCP for job requests, entirely preventing the "partially provisioned" nodes issue that wastes budget while jobs hang.
2. **Transparent Error Handling (`error_handler.py`):** 
   - Introduced a standalone error handler script to accurately parse GCP API error codes (e.g., `ZONE_RESOURCE_POOL_EXHAUSTED`, `QUOTA_EXCEEDED`). 
   - Slurm jobs are now updated with the *exact* raw GCP error message via `AdminComment` for immediate user visibility.
3. **Graceful Job Requeuing & Backoff:** 
   - Instead of marking nodes `DOWN` forever on a capacity error, `resume.py` executes a "Double Update" (`state=down` followed by `state=power_down`) to instantly clear the `POWERING_UP` state, bypassing Slurm's rigid `ResumeTimeout`.
   - The job is safely requeued to `PENDING`, and its `StartTime` is pushed back with a randomized jitter (3-5 minutes) to prevent aggressive API hammering during a prolonged stockout.
 
4. **Array Jobs / Heterogenous Jobs:**
   - If the Job Array/HetJob hits a stockout/quota exceeded error, the ArrayJobID or HetJobID is taken as a seed for the random jitter. This ensures all associated tasks share the exact same jitter and are not scattered across different scheduled time slots.
5. **Batching:**
   - Instead of running job updates individually, which can cause the controller to hang if there are thousands of pending jobs, the updates are processed in batches.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,6 +489,52 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+            for _ in range(10):
+                try:
+                    res = subprocess.run([lookup().scontrol, "show", "job", str(job.job_id)], capture_output=True, text=True)
+                    if "JobState=PENDING" in res.stdout:
+                        break
+                except Exception:
+                    pass
+                time.sleep(1)
+            
+            run(f"{lookup().scontrol} update JobId={job.job_id} StartTime={backoff_time}", check=False)
```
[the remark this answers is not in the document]
vikramvs-gg: Incorporated

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -17,10 +17,15 @@
 
 from typing import List, Optional, Dict, Any
 import argparse
+import random
+import time
+import subprocess
+import datetime
 from datetime import timedelta
```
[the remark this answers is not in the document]
vikramvs-gg: Done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +487,62 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+    admin_reason_quoted = shlex.quote(admin_comment)
+    fallback_reason_quoted = shlex.quote(reason)
+
+    nodelist = util.to_hostlist(nodes)
+    if action == error_handler.Action.REQUEUE:
+        log.error(f"Resetting nodes {nodelist} to power_down. Reason: {reason}")
+        # 1. Force DOWN to instantly strip the POWERING_UP (#) flag and bypass ResumeTimeout
+        run(f"{lookup().scontrol} update nodename={nodelist} state=down reason='Force clear booting state'", check=False)
+        # 2. Return to power_down so Slurm transitions it safely to idle~
+        run(f"{lookup().scontrol} update nodename={nodelist} state=power_down", check=False)
```
abbas1902, line 517: An alternative path is running power_down_job so job allocations are immediately revoked and the nodes are returned to ~idle as soon as possible.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +487,62 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+                        pending_jobs.add(job.job_id)
+                except Exception:
+                    pass
+            if len(pending_jobs) == len(jobs_to_requeue):
+                break
+            time.sleep(1)
+
+        for job in jobs_to_requeue:
+            # To prevent the requeued job from immediately retrying and hammering
+            # the GCP API during a stockout, we calculate a randomized backoff (e.g. 3-5 mins) and
```
abbas1902: Does this means that the order of jobs running might be shuffled around between initial attempt and the requeue since Slurm will strictly adhere to the new startTime?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
```diff
@@ -0,0 +1,38 @@
...
+
+def classify_gcp_error(error_code: str, error_message: str) -> tuple[Action, str]:
+    """
+    Classifies a GCP API error into an actionable Slurm response.
+    Returns: (Action Enum, Normalized Reason String)
+    """
+    msg_lower = error_message.lower()
+    
+    # 1. Quota Errors -> Transient (active usage limits), require REQUEUE
+    if "quotaexceeded" in error_code.lower() or "quotaexceeded" in msg_lower or ("quota" in msg_lower and "exceeded" in msg_lower) or error_code == "QUOTA_EXCEEDED":
```
abbas1902: nit: Do both `error_code.lower() == "quota_exceeded"` and `quotaexceeded` occur (first and last conditionals?)

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
```diff
@@ -0,0 +1,38 @@
...
+    REQUEUE = "REQUEUE"
+    IGNORE = "IGNORE"
+
+def classify_gcp_error(error_code: str, error_message: str) -> tuple[Action, str]:
+    """
+    Classifies a GCP API error into an actionable Slurm response.
+    Returns: (Action Enum, Normalized Reason String)
+    """
+    msg_lower = error_message.lower()
+
```
abbas1902, line 28: nit: Could we do the same for error_code?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +487,62 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+    for job in jobs:
+        if not (set(job.nodes_alloc) & nodes_set):
+            continue
+            
+        run(f"{lookup().scontrol} update jobid={job.job_id} admincomment={admin_reason_quoted}", check=False)
+
+        if action == error_handler.Action.REQUEUE:
+            jobs_to_requeue.append(job)
+
+    if jobs_to_requeue:
```
abbas1902, line 532: Could we use `scontrol requeue <job_id>` to short circuit this 10 second wait?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -353,7 +367,9 @@ def resume_nodes(nodes: List[str], resume_data: Optional[ResumeData]):
         failed_reqs = [str(e) for e in failed.items()]
         log.error("bulkInsert API failures: {}".format("; ".join(failed_reqs)))
         for ident, exc in failed.items():
-            down_nodes_notify_jobs(grouped_nodes[ident].nodes, f"GCP Error: {exc._get_reason()}", resume_data) # type: ignore
+            reason = exc._get_reason() if hasattr(exc, "_get_reason") else str(exc) # type: ignore
```
abbas1902, line 370: nit: We might be able to extract `error_details` from exc so we don't have to pass in `reason` twice.

AdarshK15: Done. Added Error detail.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: Using `scontrol requeue` creates a race condition with Slurm's internal node-failure handler, which is already automatically requeuing the job because we marked the nodes `DOWN`.
The 10 sec polling loop is needed to let Slurm finish its internal teardown and state cleanup first, ensuring the job is fully `PENDING` before we inject the `StartTime` backoff.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: Yes, some job shuffling is expected, but it's an intentional trade-off to prevent a thundering herd effect. The randomized jitter ensures that if multiple jobs fail simultaneously, they don't wake up at the exact same moment and overwhelm the GCP API.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: While `power_down_force` does revoke the allocation quickly for active nodes, applying it during the POWERING_UP (#) phase doesn't always cleanly strip the booting flag or guarantee a clean failure signal to Slurm's scheduler. The two-step down -> power_down approach is used here to force a clear "failed boot" signal that triggers an immediate native requeue while keeping the node healthy and available for the next job.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
AdarshK15: Done.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
AdarshK15: Yes, both occur depending on when it fails. Immediate API rejections return `quotaExceeded`, while delayed background failures return `QUOTA_EXCEEDED`. 
Anyway I have refactored this line to check for "quota" and "exceeded" separately, which catches all variations.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: I don't really understand why the randomness is necessary, we could do a consistent backoff if we want jobs spaced out and to not overwhelm a given API.

Jobs can often times be queued with the intention that they are dependent on the previous job's success. Yes while a user could combine these into a single job script, I think it probably teeters into unexpected behavior if jobs are run out of order of submission unless the reason can be justified.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: Yes it is predicated on the earlier alternative being implemented [here](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/5776#discussion_r3456401974).  My logic is this gives you a greater amount of control on what job gets requeued when, but if you'd to proceed with the initial double state transition than my comment is obsolete.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: > applying it during the POWERING_UP (#) phase doesn't always cleanly strip the booting flag

Hmm sounds like potentially a bug? I'm not aware of such caveat but if you observe this please file a SchedMD ticket as this sounds like potentially unexpected behavior.

In any event this is merely a potential way of getting to the desired state quicker (no concerns with this present approach), so feel free to ignore this comment.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: If we use a consistent backoff and 50 jobs fail at the same time, they will all wake up at the exact same second later on. That just creates another massive spike of API requests that will probably hit the quota limit again. The randomness is just there to prevent them from syncing up.

As for job ordering, Slurm doesn't actually guarantee submission order anyway since the backfill scheduler runs jobs out of order when nodes free up. If the jobs actually depend on each other, the user has to use dependency flag. If they do that, the jitter doesn't matter because that keeps the next job blocked until the first one finishes.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: Makes sense! Since we're keeping the double state transition to wipe those # flags, we will leave this polling loop as is to avoid race condition.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
abbas1902: Was thinking more so linear rather than consistent. It is true Slurm doesn't make many guarantees over submission order, however I am concerned about arbitrary queue fragmentation.

Same idea with priority inversion, low priority jobs running before high priority jobs because of an unfavorable number pull.

I understand these are trade-offs you may be willing to make, but if I could suggest a few items:

1. Some sort of guide/notice/documentation update about this behavior somewhere, it appears to not be opt in so it would be good if we had somewhere we could point to regarding why this behavior may occur. We can do some sort of follow up for this part.

2. Could we narrow the 3-5 minute window? Maybe closer to 1 minute or less?

3. Can we potentially preserve user intention around priority weighting? As you mention backfill will attempt to run jobs whenever it can. Maybe we can extract the priority information about the job and use that as part of the backoff calculation?

4. What about heterogenous jobs/job arrays. I think this may cause some issues for such cases. For array jobs, this result in more ResumeProgram runs than initially expected right (due to the job scattering)? 

It's good that this is directly connected to cases where we can match the error to quota exhaustion , I think if we can tighten the extremities a bit the drawbacks can be minimized.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +498,62 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+            if len(pending_jobs) == len(jobs_to_requeue):
+                break
+            time.sleep(1)
+
+        for job in jobs_to_requeue:
+            # To prevent the requeued job from immediately retrying and hammering
+            # the GCP API during a stockout, we calculate a randomized backoff (e.g. 3-5 mins) and
+            # explicitly delay the job's next evaluation via the StartTime parameter.
+            delay_seconds = random.randint(180, 300)
+            backoff_time = (datetime.now() + timedelta(seconds=delay_seconds)).strftime("%Y-%m-%dT%H:%M:%S")
```
arpit974: Per the design doc, `sbatch` jobs should requeue with backoff, but interactive `srun` jobs should **fail fast** to the user terminal instead of hanging on a delayed `StartTime`. 

Can we check `batch_flag == 0` and skip the `StartTime` backoff for interactive jobs?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +498,62 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+            continue
+            
+        run(f"{lookup().scontrol} update jobid={job.job_id} admincomment={admin_reason_quoted}", check=False)
+
+        if action == error_handler.Action.REQUEUE:
+            jobs_to_requeue.append(job)
+
+    if jobs_to_requeue:
+        # Wait up to 10 seconds for all jobs to natively requeue to PENDING
+        pending_jobs = set()
```
arpit974, line 539: Nit: `scontrol show job` is called individually per job inside the polling loop, which can spawn dozens of subprocesses per second and spike CPU.

Can we batch this into a single query per loop iteration?
```python
res = run(f"{lookup().scontrol} show job {','.join(remaining_ids)}", check=False)

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py [hunk 11]
```diff
@@ -0,0 +1,40 @@
...
+def classify_gcp_error(error_code: str, error_message: str) -> tuple[Action, str]:
+    """
+    Classifies a GCP API error into an actionable Slurm response.
+    Returns: (Action Enum, Normalized Reason String)
+    """
+    msg_lower = error_message.lower()
+    code_lower = error_code.lower()
+    
+    # 1. Quota Errors -> Transient (active usage limits), require REQUEUE
+    if (("quota" in code_lower and "exceeded" in code_lower) or
```
arpit974, line 30: GCP API throttling and quota rejections often return status codes like `RESOURCE_EXHAUSTED` or `RATE_LIMIT_EXCEEDED` without including the word `"quota"` in the message string. 

Could we update `classify_gcp_error` to check `code_lower` for `resource_exhausted` and `rate_limit_exceeded` as well, so these errors don't fall through to `Action.IGNORE`?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
arpit974: +1 on point 2nd suggestions. Narrowing the backoff window to 60–120s (1–2 minutes) feels like a great middle ground here.

It keeps enough jitter to protect the GCP Compute API from a thundering herd of retries, while addressing the queue delay and priority inversion concerns raised above.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
AdarshK15: Added resource_exhausted to the list.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: Done, added batch processing for querying job statuses.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: When an srun job fails and error handler requests to requeue, Slurm instantly kills it instead of putting it back in PENDING. As we are only checking jobs in PENDING state, interactive jobs skip the backoff and fail fast.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: 1. Added a note on this behaviour in `docs/slurm-troubleshooting.md`.
2. Yes, narrowed it down to 1-2 minutes.
3. If we take priority into calculating jitter, and if there are multiple huge high priority tasks, they will all receive a shorter jitter and may cause starvation for other jobs, and may even cause a thundering herd problem. The scope is to just prevent this by adding random jitter, As we have reduced the delay window, it will not keep jobs waiting in queue for too long.
4. Yes it would have caused multiple runs, I have updated the logic to fetch the array/pack id and use it as a seed value for generating random jitter, so that all jobs in an array have same jitter and also added batching to reduce number of queries/updates.

abbas1902 (dismissed):
LGTM

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -470,8 +498,94 @@ def down_nodes_notify_jobs(nodes: List[str], reason: str, resume_data: Optional[
...
+                        raise RuntimeError(f"scontrol show job returned empty output. (code={getattr(res, 'returncode', 'None')}, stderr={getattr(res, 'stderr', 'None')})")
+                        
+                    data = json.loads(res.stdout)
+                    for job_dict in data.get("jobs", []):
+                        state = job_dict.get("job_state", "")
+                        # Unlike sbatch, interactive srun jobs fail immediately and won't be in PENDING state
+                        if "PENDING" in state:
+                            job_id = str(job_dict.get("job_id"))
+                            pending_jobs.add(job_id)
+                            parent_job_ids[job_id] = str(job_dict.get("array_job_id") or job_dict.get("pack_job_id") or job_id)
```
arpit974: In some Slurm JSON schemas, unset `array_job_id` / `pack_job_id` fields return string `"0"` or `"0_0"` instead of integer `0` or `None`. 
Since string `"0"` is truthy in Python, `parent_job_ids[job_id]` would evaluate to `"0"` for all regular non-array jobs. If all regular jobs get seed `"0"`, they will generate the exact same random delay, destroying the jitter.
Can we explicitly check against `"0"` to be safe?
```python
arr_id = job_dict.get("array_job_id")
pack_id = job_dict.get("pack_job_id")
parent_id = arr_id if (arr_id and str(arr_id) != "0") else (pack_id if (pack_id and str(pack_id) != "0") else job_id)
parent_job_ids[job_id] = str(parent_id)

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py [hunk 13]
```diff
@@ -0,0 +1,40 @@
+# Copyright 2026 "Google LLC"
```
arpit974, line 1:
```markdown
The error classification and batching improvements look great! 
Before merging, could we add a lightweight unit test file (e.g. `test_error_handler.py`) asserting that error codes like `quotaExceeded`, `ZONE_RESOURCE_POOL_EXHAUSTED`, `RESOURCE_EXHAUSTED`, and `RATE_LIMIT_EXCEEDED` map cleanly to `Action.REQUEUE`? 
This will prevent regressions to error classification in future releases.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
AdarshK15: Done, Added check to ignore "0" and "0_0" for array ids and pack ids.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/error_handler.py
AdarshK15: Done, Added unit test file to verify the classification.

AdarshK15:
### Testing & Validation

Validated end-to-end on live GCP clusters across different job types (Exclusive, Non-Exclusive, Job Arrays, Heterogeneous, and Interactive `srun`/`salloc`) and different Availability scenarios (Full, Partial, Zero/Stockout):

#### Scenarios
| Scenario |  System Behavior & Outcome |
| :--- | :--- |
| **System Jobs** | **Greedy fulfillment:** Slurm background jobs use `minCount=1` so it can provision as many nodes as are available. |
| **User Jobs** | **Atomic protection:** By default, all user jobs are submitted as exclusive/Atomic jobs (no `minCount`). |
| **Full Availability / Reservation** | As Slurm can assign the jobs to all nodes requested, there is no GCP error thrown. No backoff is applied. In the case of Job Arrays/Heterogeneous jobs where we can't assign all jobs at a time, the jobs wait in queue for other jobs to complete.|
| **Errors:** `VM_MIN_COUNT_NOT_REACHED`, `QUOTA_EXCEEDED`| **Atomic protection:** `error_handler.py` requeues the job. `resume.py` applies AdminComment and a random `StartTime` delay to the job to prevent it from being requeued again immediately. A **double-update** safely forces nodes from `booting` to `idle` state. |
| **Array Jobs & Heterogeneous Jobs**| If the Job Array/HetJob hits a stockout/quota exceeded error, the ArrayID or HetID are taken as a seed for the random jitter so all the jobs have the same jitter and are not scattered across different time slots.|
| **Interactive Shells:** `srun`/`salloc` | **Released Instantly:** Slurm natively refuses to requeue interactive jobs upon failure. Directly forced to `FAILED`, skipping `error_handler.py`. |
| **Non-Capacity Errors** | **Fails Fast:** For GCP errors like `INVALID_ARGUMENT` and `Invalid topology network` (HttpError 400), `error_handler.py` returns `Action.IGNORE`. No jitter/requeue applied to prevent masking fatal bugs. |
| **Jobs taking more than 10s to requeue** | If there are too many jobs and they take more than 10 seconds to move to a pending state, jitter and the AdminComment are not added as `resume.py` only waits for 10 seconds. However, the job is requeued and nodes are back to idle. Eventually, as the backlog reduces, the jitter and AdminComment are added. This is a tradeoff to prevent hanging the controller. |
| **GCP Allocation Timeout** | In some cases where the nodes are available and GCP accepts the `bulkInsert` request, but due to spot preemption or node placement it is unable to provision nodes, GCP takes more than 15 mins to timeout. As the default Slurm `ResumeTimeout` is 900s, Slurm requeues the job but doesn't call `error_handler.py` as it did not receive any error from GCP at that point.|

AdarshK15:
Failing Checks:
1. The following tests are failing due to capacity:
  - `PR-test-gke-tpu-v5e`
  - `PR-test-ml-a3-highgpu-slurm`
  - `PR-test-ml-a3-ultragpu-custom-2404-blueprint-test`
  - `PR-test-ml-a3-ultragpu-custom-blueprint-test`
  - `PR-test-ml-a3-ultragpu-onspot-slurm`
  - `PR-test-ml-a4x-highgpu-custom-blueprint-test`
  - `PR-test-ml-a4x-highgpu-slurm`
  - `PR-test-slurm-gcp-v6-tpu`
2. Test `PR-test-spack-gromacs` is failing due to timeout during image building.
3. Test `PR-test-ml-a3-highgpu-onspot-slurm` is failing due to DCGM installation during image building.

All the above mentioned test failures are known issues and are not caused by the code changes in this PR.

arpit974 (approved):
LGTM

PR #5790, opened 2026-06-16
Slurm ha controller (adoptiong MIGs for controller)

arpit974:
1. Control Plane Architecture (MIG Integration)
Zonal & Regional Schedulers: Introduced support for both zonal and regional HA deployments controlled via the controller_ha_type variable. Schedulers are distributed across zones to protect against single-zone outages.
Stateful IP Preservation: Integrated google_compute_per_instance_config (zonal) and google_compute_region_per_instance_config (regional). On VM recreation or auto-healing events, GCE automatically preserves the exact static IP bindings (controller-0 and controller-1) and persistent stateful templates.
Init Target Size: Configured target_size = 0 during MIG resource definition. This ensures instances are exclusively provisioned using the explicit per-instance config rules, preventing the generation of random-suffix helper VMs.

2. Networking & Host Resolution
FQDN Support: Refactored boot-time hostname checks in setup.py to isolate the short hostname before resolving node roles, ensuring compatibility with corporate DNS suffixes and FQDN networks.
Self-Cleaning /etc/hosts: Added intelligent regex cleanup logic to standard setup scripts. Reboots and VM recreations dynamically purge stale local hostname entries and update DNS entries cleanly without duplicate headers.
GCE Metadata Fallback: Added a local configuration checking safety fallback to setup.py. If the GCE Metadata Server is slow or rate-limiting on startup, the system extracts roles based on local configs and hostname endings, completely preventing split-brain database states.

3. Safety & Configuration Constraints
Upgrade Safety (MIG Update Policy):
Zonal MIG: Set max_unavailable_fixed = 1 for zero-downtime rolling upgrades.
Regional MIG: Configured max_unavailable_fixed = 2 to satisfy the GCP regional MIG API validation check requiring max_unavailable to be at least equal to the number of zones when using RECREATE.
Database & IP Validations: Enforced compile-time validation checks requiring var.cloudsql != null (external DB connection parameters) and at least 2 static IPs when enable_backup_controller is true.
Default Port Configuration: Assured slurm_control_host_port is always populated (defaulting to "6818"), eliminating potential null-parsing issues across client nodes.

### Submission Checklist

### DIFF
--- a/community/examples/hpc-slurm-ha.yaml
+++ b/community/examples/hpc-slurm-ha.yaml
@@ -161,6 +161,7 @@ deployment_groups:
     use: [debug_partition]
     settings:
       enable_backup_controller: true
+      controller_ha_type: regional
       slurm_cluster_name: hpcslurmha
       backup_zone: us-central1-f
       # Disable PD state disk, rely on /var/spool/slurm filestore share
--- a/community/examples/hpc-slurm-ha.yaml
+++ b/community/examples/hpc-slurm-ha.yaml
@@ -21,6 +21,7 @@ vars:
   deployment_name: hpcslurmha
   region: us-central1
   zone: us-central1-a
+  enable_slurm_auth: true
 
 terraform_backend_defaults:
   type: gcs
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
@@ -88,6 +88,7 @@ To enable High Availability:
    Filestore, NFS) that is mounted on both controllers at `/var/spool/slurm`.
 4. Configure the Managed Instance Group distribution type using `controller_ha_type`
    (set to either `"zonal"` or `"regional"`, defaults to `"zonal"`).
+5. Provide at least two static IP addresses in `static_ips` (one for the primary controller and one for the backup controller).
 
 > **CRITICAL**: The `StateSaveLocation` MUST be a shared network filesystem
 > (like Filestore or NFS) that supports simultaneous read/write access from both
@@ -107,6 +108,9 @@ To enable High Availability:
     enable_backup_controller: true
     controller_ha_type: regional
     controller_state_disk: null
+    static_ips:
+    - 10.80.0.10
+    - 10.80.0.11
     network_storage:
     - server_ip: $(controller_fs.network_storage.server_ip)
       remote_mount: $(controller_fs.network_storage.remote_mount)
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
@@ -89,6 +89,7 @@ To enable High Availability:
 4. Configure the Managed Instance Group distribution type using `controller_ha_type`
    (set to either `"zonal"` or `"regional"`, defaults to `"zonal"`).
 5. Provide at least two static IP addresses in `static_ips` (one for the primary controller and one for the backup controller).
+6. Configure an external database (`cloudsql`) to manage Slurm accounting and state across the controllers.
 
 > **CRITICAL**: The `StateSaveLocation` MUST be a shared network filesystem
 > (like Filestore or NFS) that supports simultaneous read/write access from both
@@ -111,6 +112,11 @@ To enable High Availability:
     static_ips:
     - 10.80.0.10
     - 10.80.0.11
+    cloudsql:
+      server_ip: $(slurm_db.cloudsql.server_ip)
+      user: $(slurm_db.cloudsql.user)
+      password: $(slurm_db.cloudsql.password)
+      db_name: $(slurm_db.cloudsql.db_name)
     network_storage:
     - server_ip: $(controller_fs.network_storage.server_ip)
       remote_mount: $(controller_fs.network_storage.remote_mount)
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
@@ -61,7 +61,10 @@ The following are examples of updates that can be made to a running cluster:
 
 The `schedmd-slurm-gcp-v6-controller` module supports High Availability (HA) for
 the Slurm controller. This feature deploys a secondary "Backup" controller that
-allows the cluster to continue operating if the primary controller fails.
+allows the cluster to continue operating if the primary controller fails. The module
+manages these controllers using a Google Compute Managed Instance Group (MIG).
+By default, the MIG is **zonal** but can be configured as **regional** to increase
+fault tolerance across multiple zones.
 
 ### Architecture
 
@@ -83,6 +86,8 @@ To enable High Availability:
    disk.
 3. Ensure `StateSaveLocation` points to a shared network mount (e.g.,
    Filestore, NFS) that is mounted on both controllers at `/var/spool/slurm`.
+4. Configure the Managed Instance Group distribution type using `controller_ha_type`
+   (set to either `"zonal"` or `"regional"`, defaults to `"zonal"`).
 
 > **CRITICAL**: The `StateSaveLocation` MUST be a shared network filesystem
 > (like Filestore or NFS) that supports simultaneous read/write access from both
@@ -100,7 +105,7 @@ To enable High Availability:
   - homefs
   settings:
     enable_backup_controller: true
-    backup_machine_type: c2-standard-8
+    controller_ha_type: regional
     controller_state_disk: null
     network_storage:
     - server_ip: $(controller_fs.network_storage.server_ip)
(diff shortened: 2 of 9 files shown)


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,112 @@ resource "google_compute_instance_from_template" "controller" {
...
-  subnetwork = var.subnetwork_self_link
+# REGIONAL INSTANCE GROUP MANAGER (deployed if ha_type is regional)
+resource "google_compute_region_instance_group_manager" "controller_regional_mig" {
+  count                     = (var.enable_backup_controller && var.controller_ha_type == "regional") ? 1 : 0
+  name                      = "${local.slurm_cluster_name}-controller-mig"
+  base_instance_name        = "${local.slurm_cluster_name}-controller"
+  region                    = var.region
+  project                   = local.controller_project_id
+  distribution_policy_zones = [var.zone, local.backup_zone]
+  target_size               = 0
```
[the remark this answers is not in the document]
arpit974: not needed, we have already tested this...
once the controllers are created this automatically takes into account

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,112 @@ resource "google_compute_instance_from_template" "controller" {
...
 
-  source_image_family  = local.source_image_family
-  source_image_project = local.source_image_project_normalized
-  source_image         = local.source_image
+  lifecycle {
+    ignore_changes = [
+      target_size,
+    ]
+  }
+}
```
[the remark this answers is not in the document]
arpit974: this is a known pattern because of a gap in MIGs:
Identified a resource duplication issue: the deployment spins up 4 controller VMs in the Managed Instance Group (MIG) instead of the expected 2 (Primary + Backup). This doubles the compute cost of the scheduler layer.
Technical Root Cause The issue is caused by a declarative ordering limitation in the Terraform Google Cloud Provider when configuring Stateful MIGs:
Initial Group Creation: Terraform creates the MIG with a target size of 2. GCE immediately boots 2 instances with random suffixes (e.g., controller-t7v7, controller-8g3j) to satisfy the target size.
Stateful Configuration: Terraform then applies the stateful per-instance configurations to bind the pre-allocated static IPs to specific names (controller-0 and controller-1).
Group Expansion: GCE treats these configurations as new instances and boots controller-0 and controller-1, bringing the total running count to 4. GCE does not automatically terminate the original 2 random-suffix instances.

for this setting target size to 0, then have the stateful configuration, and it will change the target size to 2. Also need to make terraform not look at target size so it doesn't destroy the vms everytime we reapply.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,112 @@ resource "google_compute_instance_from_template" "controller" {
...
+    max_unavailable_fixed        = 2
+  }
 
-  tags = concat([local.slurm_cluster_name], var.tags)
+  lifecycle {
+    ignore_changes = [
+      target_size,
+    ]
+  }
 }
```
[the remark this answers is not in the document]
arpit974: same as above

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,116 @@ resource "google_compute_instance_from_template" "controller" {
...
+  version {
+    instance_template = module.slurm_controller_template.self_link
+  }
+  update_policy {
+    type                         = "PROACTIVE"
+    minimal_action               = "REPLACE"
+    replacement_method           = "RECREATE"
+    instance_redistribution_type = "NONE"
+    max_surge_fixed              = 0
+    max_unavailable_fixed        = 2
```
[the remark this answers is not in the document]
arpit974: For regional MIGs using the RECREATE replacement method (which we must use to preserve stateful IPs and name suffixes like -0 and -1), max_unavailable cannot be less than the number of zones in the group.
If we set max_unavailable_fixed = 1 for a 2-zone regional MIG, Terraform apply will fail with a GCP API error: "max_unavailable cannot be less than the number of zones (2)".

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,116 @@ resource "google_compute_instance_from_template" "controller" {
...
-  preemptible         = var.preemptible
-  service_account     = local.service_account
+# ZONAL INSTANCE GROUP MANAGER (deployed if ha_type is zonal)
+resource "google_compute_instance_group_manager" "controller_zonal_mig" {
+  count              = (var.enable_backup_controller && var.controller_ha_type == "zonal") ? 1 : 0
+  name               = "${local.slurm_cluster_name}-controller-mig"
+  base_instance_name = "${local.slurm_cluster_name}-controller"
+  zone               = var.zone
+  project            = local.controller_project_id
+  target_size        = 0
```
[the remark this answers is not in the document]
arpit974: target_size 0 is intentional,
already explained in comment above.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf
```diff
@@ -226,83 +218,116 @@ resource "google_compute_instance_from_template" "controller" {
...
-  subnetwork = var.subnetwork_self_link
+# REGIONAL INSTANCE GROUP MANAGER (deployed if ha_type is regional)
+resource "google_compute_region_instance_group_manager" "controller_regional_mig" {
+  count                     = (var.enable_backup_controller && var.controller_ha_type == "regional") ? 1 : 0
+  name                      = "${local.slurm_cluster_name}-controller-mig"
+  base_instance_name        = "${local.slurm_cluster_name}-controller"
+  region                    = var.region
+  project                   = local.controller_project_id
+  distribution_policy_zones = [var.zone, local.backup_zone]
+  target_size               = 0
```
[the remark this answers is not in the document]
arpit974: target_size 0 is intentional,
already explained in comment above.

community/examples/hpc-slurm-ha.yaml
LAVEEN, line 1: Please add a test for this blueprint which covers your changes as well so that if blueprint breaks then can be fixed immediately.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -492,6 +493,19 @@ def setup_controller(is_primary: bool):
     if not lkp.cfg.cloudsql_secret:
         configure_mysql(lkp)
 
+    if not is_primary:
+        db_sentinel = Path(slurmdirs.state / "db_initialized")
+        timeout = 300
```
LAVEEN, line 498: isn't the retries too high?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
```diff
@@ -100,8 +107,16 @@ To enable High Availability:
   - homefs
   settings:
     enable_backup_controller: true
-    backup_machine_type: c2-standard-8
+    controller_ha_type: regional
     controller_state_disk: null
+    static_ips:
+    - 10.80.0.10
```
LAVEEN, line 113: wont this ip conflict if i deploy two times the same blueprint with different deployment name?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
arpit974: The 5-minute timeout is a safe buffer. Cloud SQL instance setup, slurmdbd startup, and initial DB schema creation on the primary controller can take up to 2-3 minutes in GCE. If the backup controller times out too early and starts slurmctld before registration completes, the slurmctld daemon will fail to start.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -706,16 +778,35 @@ def main():
...
-        pass
+    if role == "controller":
+        try:
+            ha_role = util.instance_metadata("attributes/slurm_ha_role", silent=True)
+            if ha_role == "backup":
+                role = "backup_controller"
+            elif ha_role == "dynamic":
+                hostname = socket.gethostname()
+                short_hostname = hostname.split(".")[0]
+                if short_hostname.endswith("-0"):
```
AdarshK15, line 795: Here we are checking hostname doesn't end with `"-0"` to set `role = "backup_controller"`.
but if primary controller goes down, backup will promote itself. When the failed controller is recreated it will still have `-0` and it may have conflicting role because of this check.
should we check this using the shared state directory - `/var/spool/slurm/ha_state`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/README.md
arpit974: By default, each deployment creates its own isolated VPC network and subnet based on the deployment_name. Since these are internal IPs, they are completely isolated and will not conflict across different deployments.

If a user is deploying multiple clusters into the same shared/pre-existing VPC subnet, they will need to override the static_ips variable in the blueprint settings of their second deployment to assign a different set of IPs.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
arpit974: In Slurm's native HA design, the roles of the primary and backup controllers are statically defined in slurm.conf (by hostname). When a recreated primary (controller-0) comes back online, its slurmctld daemon starts up, contacts the active backup (controller-1), and the backup automatically yields active control back to the primary.

Therefore, controller-0 must always be configured as the primary controller during setup so that it can trigger this native failback. Dynamically swapping roles at the setup script level would conflict with the static host declarations in slurm.conf.

community/examples/hpc-slurm-ha.yaml
arpit974: since it is in community folder,
Will take it in separate PR.

LAVEEN (approved):
LGTM , thanks for addressing comments.

PR #5818, opened 2026-06-19
fix(job): Make priorityClass installation conditional

Neelabh94:
This PR modifies the installation logic to ensure PriorityClasses are only checked and installed when explicitly requested by the user in their job configuration.


agrawalkhushi18 (approved):
LGTM.

PR #5820, opened 2026-06-20
fix(job): restrict TPU topology node constraints & deprecated rbac-proxy images

Neelabh94:
### 1. TPU Topology Validation & Restrictions

* **GKE Warden Compliance**: The toolkit now strictly blocks the use of multiple values (via the pipe | operator) in `--node-constraints` for the `cloud.google.com/gke-tpu-topology` label. This prevents GKE clusters from rejecting pod templates that target multiple physical topologies.

* **Removal of "Smart Merging"**: Automatic merging of the `--topology` flag value into the node affinity constraints has been removed to ensure scheduling logic is explicit and predictable.

* **Case Insensitivity**: All TPU topology values are now canonicalized to lowercase at the point of submission and within scheduling logic to ensure case-insensitive matching with node labels.

### 2. Dynamic Image Replacement

* **GCR Deprecation Fix**: Implemented a dynamic helper function that replaces deprecated gcr.io/kubebuilder/kube-rbac-proxy images with the supported quay.io/brancz/kube-rbac-proxy versions in JobSet manifests.

### 3. Documentation Updates

* **Scheduling Guidance**: Updated docs/gcluster_job_guide.md with new examples for TPU sub-slicing and fallback scheduling.

* **Best Practices**: Added instructions on leveraging GKE Kueue’s tryNextFlavor mechanism for dynamic resource allocation when topology constraints are omitted.

### DIFF
--- a/cmd/job/submit.go
+++ b/cmd/job/submit.go
@@ -195,6 +195,8 @@ func runSubmitCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("--num-nodes cannot be used with TPU jobs (it is calculated automatically from topology)")
 	}
 
+	jobTopology := strings.ToLower(strings.TrimSpace(topology))
+
 	jobDef := orchestrator.JobDefinition{
 		ImageName:                     imageName,
 		BaseImage:                     baseImage,
@@ -219,7 +221,7 @@ func runSubmitCmd(cmd *cobra.Command, args []string) error {
 		RestartOnExitCodes:            restartOnExitCodes,
 		ImagePullSecrets:              imagePullSecrets,
 		ServiceAccountName:            serviceAccountName,
-		Topology:                      topology,
+		Topology:                      jobTopology,
 		GKEScheduler:                  gkeScheduler,
 		AwaitJobCompletion:            awaitJobCompletion,
 		UseParallelContainers:         !gkeDisableParallelContainers,
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -1744,10 +1744,12 @@ func (g *GKEOrchestrator) addTopologyLabel(nodeSelector map[string]string, sched
 
 func (g *GKEOrchestrator) buildNodeSelector(schedOpts SchedulingOptions, job orchestrator.JobDefinition, isCPUMachine bool) (string, error) {
 	nodeSelector := make(map[string]string)
-	if existing := GetNodeSelector(schedOpts); existing != nil {
-		for k, v := range existing {
-			nodeSelector[k] = v
-		}
+	existing, err := getNodeSelector(schedOpts)
+	if err != nil {
+		return "", err
+	}
+	for k, v := range existing {
+		nodeSelector[k] = v
 	}
 
 	// Inject unified consumption options
--- a/pkg/orchestrator/gke/scheduling.go
+++ b/pkg/orchestrator/gke/scheduling.go
@@ -33,7 +33,7 @@ type SchedulingOptions struct {
 	IsStaticSlicing    bool
 }
 
-func GetNodeSelector(opts SchedulingOptions) map[string]string {
+func getNodeSelector(opts SchedulingOptions) (map[string]string, error) {
 	nodeSelector := make(map[string]string)
 
 	if opts.PlacementPolicy != "" {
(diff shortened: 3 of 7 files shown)


cmd/job/submit.go
```diff
@@ -195,6 +195,8 @@ func runSubmitCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("--num-nodes cannot be used with TPU jobs (it is calculated automatically from topology)")
 	}
 
+	topology = strings.ToLower(strings.TrimSpace(topology))
```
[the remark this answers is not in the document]
Neelabh94: done

pkg/orchestrator/gke/scheduling.go
```diff
@@ -117,28 +125,29 @@ func GetAffinity(opts SchedulingOptions) (*corev1.Affinity, error) {
...
+		if k == tpuTopologyLabel {
+			trimmed = strings.ToLower(trimmed)
+		}
+		if k == tpuTopologyLabel && !config.TopologyRegex.MatchString(trimmed) {
+			return nil, fmt.Errorf("invalid topology format %q for key %s", trimmed, k)
+		}
+		if !slices.Contains(values, trimmed) {
+			values = append(values, trimmed)
 		}
 	}
```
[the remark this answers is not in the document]
Neelabh94: done

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -1744,10 +1744,12 @@ func (g *GKEOrchestrator) addTopologyLabel(nodeSelector map[string]string, sched
...
-	if existing := GetNodeSelector(schedOpts); existing != nil {
-		for k, v := range existing {
-			nodeSelector[k] = v
-		}
+	existing, err := getNodeSelector(schedOpts)
+	if err != nil {
+		return "", err
+	}
+	for k, v := range existing {
+		nodeSelector[k] = v
```
shubpal07, line 1752: **Suggestion (Future Work)**: If a user requests a sub-sliced topology (logical < physical) during a --dry-run-out or on a cluster where Kueue is installed but the Topology CRDs are missing (causing hasKueueTopologies() to return false), the job will schedule but JAX will crash at runtime due to lack of TAS partitioning (Test Case 2). In the next iteration, we should add a client-side warning here to catch these specific fallback paths.

```
Warning: You are attempting to run a sub-sliced TPU job (%s logical on %s physical) without Kueue TAS. Without Kueue's mesh partitioning, your JAX/libtpu workload will crash at runtime with connection timeouts. Please ensure Kueue is installed and active on your queue.
```

pkg/orchestrator/gke/scheduling.go
```diff
@@ -117,28 +121,23 @@ func GetAffinity(opts SchedulingOptions) (*corev1.Affinity, error) {
...
+func parseAffinityValues(k string, v string) ([]string, error) {
+	if v == "" {
 		return nil, nil
 	}
 
-	var values []string
-	if isTopologyMerge {
-		values = append(values, topology)
+	if k == tpuTopologyLabel && strings.Contains(v, "|") {
+		return nil, fmt.Errorf("multiple topologies in node constraints are not supported for key %s: got %q. GKE Warden validation rejects templates targeting multiple topologies", k, v)
```
shubpal07, line 130: nit: Since `parseAffinityValues` is a private helper that is only invoked by GetAffinity when `hasPipe` is already guaranteed to be true (see [L97-100](https://github.com/Neelabh94/cluster-toolkit/blob/6eb68e45203f4965b46842a11903d9d25f276742/pkg/orchestrator/gke/scheduling.go#L97-L103)), the `strings.Contains(v, "|")` check here is redundant.
We can simplify this to:
```go
if k == tpuTopologyLabel {
    return nil, fmt.Errorf("multiple topologies in node constraints are not supported for key %s: got %q. GKE Warden validation rejects templates targeting multiple topologies", k, v)
}
```

pkg/orchestrator/gke/scheduling_test.go
```diff
@@ -22,16 +22,91 @@ import (
...
-	selector := GetNodeSelector(opts)
-	if selector["cloud.google.com/gke-placement-group"] != "compact-placement" {
-		t.Errorf("Expected placement policy label, got %v", selector["cloud.google.com/gke-placement-group"])
+	tests := []struct {
+		name      string
+		opts      SchedulingOptions
+		wantKey   string
+		wantValue string
+		wantErr   bool
+	}{
```
shubpal07, line 31: nit:
To make the test assertions more idiomatic and readable, we could structure the test cases to verify the entire expected map output rather than using `wantKey` / `wantValue` combinations with implicit "_empty string means exclusion_" rules.

For example, we could define the expected output as a map:

```go
tests := []struct {
    name         string
    opts         SchedulingOptions
    wantSelector map[string]string
    wantErr      bool
}{
```
This makes it clear from the test cases table what the exact expected nodeSelector map should look like.

pkg/orchestrator/gke/scheduling_test.go
```diff
@@ -239,38 +305,6 @@ func TestGetAffinity_ConstraintsAndMerging(t *testing.T) {
...
-			name: "invalid topology format",
-			opts: SchedulingOptions{
-				NodeAffinityLabels: map[string]string{
-					"cloud.google.com/gke-tpu-topology": "invalid",
-				},
-			},
-			wantErr: true,
-		},
 		{
 			name: "empty element in list",
```
shubpal07, line 309: **Future work:**
To prevent future regressions, it would be great to add a dry-run manifest generation test here that explicitly asserts the final pod template has only a single topology value in `nodeSelector` and none in `nodeAffinity` when a sub-sliced job is compiled.

shubpal07 (approved):
LGTM with some minor non-blocking suggestions / future work.
Thanks for the PR @Neelabh94

Neelabh94:
> LGTM with some minor non-blocking suggestions / future work. Thanks for the PR @Neelabh94

Thanks for the suggestions Shubham, will take these up in a quick follow up PR.

PR #5821, opened 2026-06-21
Fixing A3H Issue

LAVEEN:
This PR resolves an issue where A3 High compute nodes in the Slurm cluster consistently entered a `DRAIN` state immediately after job execution, which prevented multi-node workloads and NCCL tests from running. 

The root cause was an API version mismatch between the NVIDIA Driver 580 (CUDA 13.0) and the legacy Datacenter GPU Manager (DCGM) v3.3.9 package. The unversioned `datacenter-gpu-manager` dependency resolved to an older version that is incompatible with CUDA 13, causing Slurm's mandatory post-job health check (`gpu-test` epilog) to fail.

This PR implements the permanent fix by explicitly pairing the Driver 580 upgrade with the CUDA 13-compatible DCGM v4 package (`datacenter-gpu-manager-4-cuda13`). 

Additionally, it updates the daily cloud build tests to run with 2 nodes instead of 4

### DIFF
--- a/examples/machine-learning/a3-highgpu-8g/a3high-slurm-blueprint.yaml
+++ b/examples/machine-learning/a3-highgpu-8g/a3high-slurm-blueprint.yaml
@@ -191,10 +191,10 @@ deployment_groups:
               package_filename: /tmp/{{ package_url | basename }}
               nvidia_packages:
               - cuda-toolkit-12-8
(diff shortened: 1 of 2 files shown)


LAVEEN:
95r5ka9UMTk2a4a

AdarshK15 (approved):

PR #5825, opened 2026-06-22
Fix(validators): Support shared reservations in blueprint validation

SwarnaBharathiMantena:
Resolve owner project for shared reservations defined in modules.

When validating if a reservation exists, the validator previously only looked at the global 'project_id'. If a shared reservation was defined inside a module's 'reservation_affinity' setting with a specific 'project' field, the validator would fail to find it because it searched in the deployment project.

This CL adds a helper 'findReservationOwnerProject' that scans the blueprint modules for 'reservation_affinity' settings matching the reservation name. If a specific owner project is defined, it uses that project for the validation.

### DIFF
--- a/pkg/validators/cloud.go
+++ b/pkg/validators/cloud.go
@@ -408,42 +408,90 @@ func TestReservationExists(ctx context.Context, reservationProjectID string, zon
 	return fmt.Errorf("reservation %q was not found in any zone of project %q", reservationName, reservationProjectID)
 }
 
+// findReservationOwnerProject scans the blueprint's modules to see if a specific
+// reservation is configured as a shared reservation (meaning it has an explicit
+// 'project' field defined in its 'reservation_affinity' settings).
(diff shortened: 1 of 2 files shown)


pkg/validators/cloud.go
```diff
@@ -403,6 +408,44 @@ func TestReservationExists(ctx context.Context, reservationProjectID string, zon
 	return fmt.Errorf("reservation %q was not found in any zone of project %q", reservationName, reservationProjectID)
 }
 
+func findReservationOwnerProject(bp config.Blueprint, reservationName string) string {
```
saara-tyagi27, line 415: Could we add a few inline comments to explain the workflow? It'll serve as a great reference point for future.

SwarnaBharathiMantena: Sure, added comments to the go file.

saara-tyagi27 (approved):
LGTM

PR #5830, opened 2026-06-23
failure-triage-agent: Integrate into daily integration test ansible pipeline

hritik0101:
**Overview**
This PR integrates the new Failure Triage Agent into our daily integration test pipelines. When a test fails, this PR ensures that before the cluster infrastructure is torn down, a secure webhook is fired to the agent to immediately begin log analysis, root-cause determination and potential next steps.

This significantly reduces the manual effort required to debug daily test failures by providing inline forensic summaries directly in the Cloud Build logs.

**Key Changes**

1. Build Context Injection
Modified the Cloud Build test configurations (e.g., ansible-vm.yaml, gke-*.yaml, etc.) to securely pass the $BUILD_ID down to the Ansible playbooks via the full_build_id extra variable. This allows the triage agent to map its analysis back to the exact failing Cloud Build run.

2. Playbook Orchestration (rescue_gcluster_failure.yml)
Hooked the triage agent trigger into the always block of the base-integration-test.yml by calling it inside rescue_gcluster_failure.yml. This ensures the agent is only invoked upon test failure, and guarantees it runs before the infrastructure is destroyed so logs and states remain intact for the agent.

3. Trigger Logic (trigger_failure_triage_agent.yml) Created a robust new Ansible task dedicated to invoking the agent:
_Feature Flag/Kill Switch:_ Validates the presence of gs://g-ift-agent-bucket/ENABLE_TRIAGE_AGENT. If missing, the triage agent is skipped. This allows us to disable the agent globally without requiring a code merge.
_Secure Authentication:_ Impersonates the triage-invoker service account to generate a secure OIDC token, ensuring the Cloud Run endpoint remains fully authenticated and protected.
_Inline Reporting (Polling):_ Instead of a fire-and-forget approach, the task implements an asynchronous wait (polling state.json every 60s for up to ~30 mins). Once the analysis completes, it parses the JSON and prints the LLM-generated executive_summary directly to the Cloud Build stdout. Developers can see the root cause without leaving the Google Cloud console.

**Testing/Validation**
Verified that the trigger task accurately captures the full_build_id and checks the GCS feature flag.
Verified that the identity token is correctly generated using the impersonated service account.
Verified the playbook correctly hits the Cloud Run endpoint via an authenticated POST request.
Verified the polling loop correctly blocks until status: completed and prints the summary report inline.

**Notes for Reviewers**
Timeouts: Because the playbook now waits for the triage agent to finish its analysis, the total duration of failing builds will increase by up to 15 minutes.

Tested the following daily test trigger with these changes:
1. [DAILY-test-ml-a3-ultragpu-onspot-slurm](https://pantheon.corp.google.com/cloud-build/builds;region=global/3e11b4b1-55dc-48a5-b475-62f71f41f679?e=-13802955&mods=monitoring_api_prod&project=hpc-toolkit-dev)
2. [DAILY-test-chrome-remote-desktop](https://pantheon.corp.google.com/cloud-build/builds;region=global/72ef8a21-1496-4130-95f9-5eaa30127ba3?e=-13802955&mods=monitoring_api_prod&project=hpc-toolkit-dev)
3. [DAILY-test-ml-h4d-onspot-slurm](https://pantheon.corp.google.com/cloud-build/builds;region=global/14305e58-e7d0-41b4-a438-0d36e073b271?e=-13802955&mods=monitoring_api_prod&project=hpc-toolkit-dev)
4. [DAILY-test-gke-tpu-7x](https://pantheon.corp.google.com/cloud-build/builds;region=global/dc912a4b-a3e0-460a-b495-073978362351?e=-13802955&mods=monitoring_api_prod&project=hpc-toolkit-dev)
5. [DAILY-test-htcondor](https://pantheon.corp.google.com/cloud-build/builds;region=global/93a7b986-ed88-4b93-b8b3-591088b71f74?e=-13802955&mods=monitoring_api_prod&project=hpc-toolkit-dev)

Next Steps: The application code for the Failure Triage Agent itself (the Cloud Run service) is being reviewed and merged into a separate Gerrit PR.

### Submission Checklist


tools/cloud-build/daily-tests/ansible_playbooks/multigroup-integration-test.yml
```diff
@@ -34,6 +34,9 @@
       environment:
         TF_IN_AUTOMATION: "TRUE"
     always:
+    - name: Include Failure Triage Agent trigger tasks
+      ansible.builtin.include_tasks: tasks/trigger_failure_triage_agent.yml
```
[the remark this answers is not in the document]
hritik0101: By Desgin

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      HTTP_STATUS=$(curl -sS -o /dev/null -w "%{http_code}" -X POST "{{ triage_cloud_run_url }}/trigger" \
+        -H "Authorization: Bearer $TOKEN" \
+        -H "Content-Type: application/json" \
+        -d "{\"build_id\": \"$TRIAGE_BUILD_ID\", \"project_id\": \"{{ triage_project_id }}\"}")
+
+      if [ "$HTTP_STATUS" != "200" ] && [ "$HTTP_STATUS" != "202" ]; then
+        echo "Failed to trigger agent. HTTP Status: $HTTP_STATUS" >&2
+        exit 1
+      fi
```
kvenkatachala333: One minor nit: While `$TRIAGE_BUILD_ID` is correctly passed via the `environment` parameter into the shell tasks to prevent shell interpolation issues, other configuration variables like `{{ triage_gcs_bucket }}`, `{{ triage_cloud_run_url }}`, `{{ triage_project_id }}`, and `{{ triage_invoker_sa }}` are still interpolated directly into the shell scripts via Jinja templates.

For best practises, consider passing all of these variables via the `environment` parameter as well and referencing them natively as bash variables (e.g., `$TRIAGE_GCS_BUCKET`, `$TRIAGE_PROJECT_ID`), wdyt?

LAVEEN:
https://storage.cloud.google.com/g-ift-agent-bucket/20ad5942-2241-4397-9995-e07eb919b8e0/report.tx
im not able to view this file.

hritik0101:
> https://storage.cloud.google.com/g-ift-agent-bucket/20ad5942-2241-4397-9995-e07eb919b8e0/report.tx im not able to view this file.

https://storage.cloud.google.com/g-ift-agent-bucket/20ad5942-2241-4397-9995-e07eb919b8e0/report.txt
this the whole link
you missed the "txt"

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: Yes, I will update this.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+    ignore_errors: true
+
+  - name: Print Triage Report
+    delegate_to: localhost
+    ansible.builtin.debug:
+      msg: |
+        {% if agent_state.failed or (agent_state.stdout | default('{}', true) | from_json).status | default('') != 'completed' %}
+        Failure Triage Agent testing did not complete in time or failed internally.
+        {% else %}
+        TRIAGE EXECUTIVE SUMMARY:
```
AdarshK15: Nit: Triage Agent Summary?

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+      exit 0
+    fi
+
+    if ! gcloud storage buckets describe "gs://$TRIAGE_GCS_BUCKET" >/dev/null 2>&1; then
+      echo "SKIPPED: Triage Agent bucket '$TRIAGE_GCS_BUCKET' does not exist." >&2
+      exit 0
+    fi
+
+    CONFIG_CONTENT=$(gcloud storage cat "gs://$TRIAGE_GCS_BUCKET/config_triage_agent.env" 2>/dev/null || echo "enable_agent=false")
+    if ! echo "$CONFIG_CONTENT" | grep -qi '^[[:space:]]*enable_agent[[:space:]]*=[[:space:]]*true'; then
```
AdarshK15, line 50: Nit: Instead of grep, could we use .properties file, (or json you can add comment as description) and use python to parse it to get value for this field. wdyt?

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+      echo "SKIPPED: Failure Triage Agent is currently disabled in config_triage_agent.env." >&2
+      exit 0
+    fi
+
+    echo "PROCEED: Agent is enabled and build ID is present."
+  register: triage_init
+
+- name: Execute Triage Agent Pipeline
+  when: "'PROCEED' in triage_init.stdout"
```
AdarshK15, line 60: Nit: As we are linking both ansible tasks based on word 'PROCEED' in echo output, can we use `ansible.builtin.set_fact` instead?

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+    environment:
+      TRIAGE_BUILD_ID: "{{ triage_build_id }}"
+      TRIAGE_GCS_BUCKET: "{{ triage_gcs_bucket }}"
+    ansible.builtin.shell: |
+      # Wait for Cloud Run to start and for the initial state file to be copied
+      sleep 60
+      for i in {1..30}; do
+        STATE_JSON=$(gcloud storage cat "gs://$TRIAGE_GCS_BUCKET/$TRIAGE_BUILD_ID/state.json" 2>/dev/null || echo '{}')
+        STATUS=$(echo "$STATE_JSON" | python3 -c "import sys, json; print(json.load(sys.stdin).get('status', ''))" 2>/dev/null)
+        if [ "$STATUS" = "completed" ] || [ "$STATUS" = "failed" ]; then
```
AdarshK15, line 121: if state file was not created due to any issue even after 60 sec initial buffer time, then it is a failure case because agent also can't update it if not present.
In this case we should fail fast, currently it will wait for 30 mins before getting timed out.
Add a check if `"$STATUS" = ""` or  `-z "$STATE_JSON"` then directly exit 1 or maybe exit after 5 polls.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+    changed_when: false
+    args:
+      executable: /bin/bash
+    environment:
+      TRIAGE_BUILD_ID: "{{ triage_build_id }}"
+      TRIAGE_GCS_BUCKET: "{{ triage_gcs_bucket }}"
+    ansible.builtin.shell: |
+      # Wait for Cloud Run to start and for the initial state file to be copied
+      sleep 60
+      for i in {1..30}; do
```
AdarshK15, line 118: Nit: do we need 30 mins timeout, can we reduce it to 15 mins.
may be keep the loop count same but reduce sleep to 30?

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+      exit 1
+    register: agent_state
+    ignore_errors: true
+
+  - name: Print Triage Report
+    delegate_to: localhost
+    ansible.builtin.debug:
+      msg: |
+        {% if agent_state.failed or (agent_state.stdout | default('{}', true) | from_json).status | default('') != 'completed' %}
+        Failure Triage Agent testing did not complete in time or failed internally.
```
AdarshK15, line 137: should we add a link to state file here?

LAVEEN (dismissed):
LGTM

tools/cloud-build/daily-tests/builds/gke-a2-highgpu-kueue-onspot.yaml
```diff
@@ -36,6 +36,7 @@ steps:
   name: us-central1-docker.pkg.dev/$PROJECT_ID/hpc-toolkit-repo/test-runner
   entrypoint: /bin/bash
   env:
+  - "BUILD_ID=$BUILD_ID"
```
AdarshK15: check line 48.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: No need.Line 123-126 does exactly that.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: I'd prefer to stick with .env and grep. .env supports native # comments, which we need for a clear 'DO NOT DELETE' warning (JSON doesn't). Also, since this is just a single flag, grep keeps the Bash block lightweight without needing to invoke Python.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: Good point! I considered using set_fact, but because our checks are encapsulated inside a single Bash block, Ansible doesn't have a built-in way to know the outcome other than reading the stdout text. To use set_fact cleanly, I'd have to break that Bash script apart into multiple separate Ansible tasks(log cluttering). Since this approach keeps the logic contained in one small block, I went with the string match.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: okay I will make that change.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
hritik0101: Yes that would be better I will make a check so that it can fail fast in case the state file is not present.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,126 @@
...
+        {% else %}
+        TRIAGE EXECUTIVE SUMMARY:
+        {{ (agent_state.stdout | default('{}', true) | from_json).executive_summary | default('No summary available.') | wordwrap(100) }}
+
+        Full diagnostic report available at:
+        https://storage.cloud.google.com/{{ triage_gcs_bucket }}/{{ triage_build_id }}/report.txt
+        {% endif %}
+
+        For detailed intermediate state information, please review the state file:
```
AdarshK15: Nit: diagnostics state or triage state? 
just state file can be ambiguous, wdyt?

tools/cloud-build/daily-tests/builds/gke-a2-highgpu-kueue-onspot.yaml
hritik0101: Yeah will undo similar changes.

AdarshK15 (dismissed):
LGTM, please run few daily tests and PR test with this commit to test.

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,133 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: Set Triage Agent Configuration
+  ansible.builtin.set_fact:
+    triage_gcs_bucket: "{{ triage_gcs_bucket_override | default('g-ift-agent-bucket') }}"
+    triage_project_number: "{{ triage_project_number_override | default('508417052821') }}"
+    triage_invoker_sa: "{{ triage_invoker_sa_override | default('triage-invoker@hpc-toolkit-dev.iam.gserviceaccount.com') }}"
+    triage_cloud_run_url: "{{ triage_cloud_run_url_override | default('https://failure-triage-agent-508417052821.us-central1.run.app') }}"
```
kvenkatachala333: Since cluster-toolkit is an open-source, it is not recommended to hardcode internal GCP project numbers (508417052821), buckets, and service accounts as fallback defaults. 

Please remove these internal defaults (defaulting them to '' instead) and pass them exclusively via extra-vars in the Cloud Build CI configuration. You can add a quick check in the prerequisite step to skip the agent if these variables are missing

tools/cloud-build/daily-tests/ansible_playbooks/tasks/trigger_failure_triage_agent.yml
```diff
@@ -0,0 +1,133 @@
...
+      executable: /bin/bash
+    environment:
+      TRIAGE_BUILD_ID: "{{ triage_build_id }}"
+      TRIAGE_GCS_BUCKET: "{{ triage_gcs_bucket }}"
+    ansible.builtin.shell: |
+      # Wait for Cloud Run to start and for the initial state file to be copied
+      sleep 60
+
+      # Fail fast if state.json is not present after initial buffer
+      if ! gcloud storage ls "gs://$TRIAGE_GCS_BUCKET/$TRIAGE_BUILD_ID/state.json" >/dev/null 2>&1; then
```
kvenkatachala333, line 113: this fixed sleep 60 is still an unconditional 1-minute block for every test failure. If the triage agent starts quickly (e.g., in 2-3 seconds), we are needlessly waiting a full minute.

Instead of a fixed sleep, consider wrapping the check in a short retry loop. For example:

```
      # Wait for Cloud Run to start and for the initial state file to be copied
      for i in {1..12}; do
        if gcloud storage ls "gs://$TRIAGE_GCS_BUCKET/$TRIAGE_BUILD_ID/state.json" >/dev/null 2>&1; then
          break
        fi
        sleep 5
      done
      
      if ! gcloud storage ls "gs://$TRIAGE_GCS_BUCKET/$TRIAGE_BUILD_ID/state.json" >/dev/null 2>&1; then
        echo "Agent failed to start: state.json was not created within 60 seconds." >&2
        exit 1
      fi
```

This ensures we fail-fast if it takes longer than 60s, but we proceed immediately if it takes only 5s.

hritik0101: okay will change that.

tools/cloud-build/daily-tests/ansible_playbooks/multigroup-integration-test.yml
LAVEEN, line 38: If the triage agent trigger tasks fail for any reason, they will return a non-zero exit code. Since tasks inside an 'always:' block are executed sequentially, any fatal failure in the trigger steps will cause Ansible to immediately halt the playbook execution. This will result in the 'Destroy deployment' task being completely skipped, leading to leftover/leaked cloud resources. I think you need to  add 'ignore_errors: true' to the triage agent's trigger tasks to guarantee that diagnostic failures never block the critical destroy phase.

Please test this case and make sure that destroy is always working even if failures propagate  via agent else it would leave leftover on the project.

hritik0101: okay  yeah got it,I will add 'ignore_errors: true' .
thanks

LAVEEN (approved):
LGTM

AdarshK15 (approved):

kvenkatachala333 (approved):

PR #5841, opened 2026-06-24
Fix(kueue): Unify Kueue resource groups for Pathways workloads

SwarnaBharathiMantena:
Unifies the Kueue ClusterQueue resource groups when Pathways is active. This prevents Kueue from merging incompatible node selectors (TPU nodes vs CPU nodes) for TPU worker pods, which request both TPU and CPU/Memory resources.

Specifically:
- Updated Terraform templates to use a single unified resource group covering ["google.com/tpu", "cpu", "memory"] and aligned both flavors.
- Added default high CPU/Memory quotas for the TPU flavor in main.tf.
- Updated the Go orchestrator (infra_manager.go) to programmatically detect Pathways and unify the resource groups at runtime.
- Updated GKE orchestrator tests to assert the unified queue behavior.

Resolves scheduling conflicts for Pathways workloads on GKE.

### DIFF
--- a/modules/management/kubectl-apply/main.tf
+++ b/modules/management/kubectl-apply/main.tf
@@ -31,9 +31,11 @@ locals {
 
   kueue_config_template_vars = merge(
     {
-      pathways_cpu_quota    = 480
-      pathways_memory_quota = "2000G"
-      tpu_quota             = "999999" # Default high value if not set
+      pathways_cpu_quota      = 480
+      pathways_memory_quota   = "2000G"
+      tpu_flavor_cpu_quota    = "999999" # High default to avoid limiting TPU pods by CPU
+      tpu_flavor_memory_quota = "999999T"
+      tpu_quota               = "999999" # Default high value if not set
     },
     var.kueue.config_template_vars != null ? var.kueue.config_template_vars : {}
   )
--- a/pkg/orchestrator/gke/infra_manager_test.go
+++ b/pkg/orchestrator/gke/infra_manager_test.go
@@ -68,10 +68,12 @@ func TestRenderClusterQueue_Pathways(t *testing.T) {
 				"flavor-tpu": { // TPU flavor (name contains 'tpu')
 					CPUs:     100,
 					MemoryGi: 400,
+					TPUs:     8, // Physical TPU capacity
 				},
-				"pathways-flavor": { // Pathways flavor
+				"pathways-flavor": { // Pathways flavor (CPU-only)
 					CPUs:     480,
 					MemoryGi: 2000,
+					TPUs:     0, // No TPUs
 				},
 			},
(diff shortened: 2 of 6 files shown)


pkg/orchestrator/gke/infra_manager_test.go
```diff
@@ -92,10 +92,18 @@ func TestRenderClusterQueue_Pathways(t *testing.T) {
 		t.Errorf("expected nominalQuota: 480 for CPU, got %s", output)
 	}
 
-	// Verify TWO resource groups
+	// Verify SINGLE resource group (unified when Pathways is active)
```
shubpal07, line 114: **Suggestion (Non-blocking)**: It would be great to expand `TestRenderClusterQueue_Pathways` or add a hybrid unit test that actually contains both a TPU-enabled flavor (e.g. flavor-default with TPUs > 0) and pathways-flavor (CPU-only).

The current test only defines CPU/Memory capacities for both, so it doesn't explicitly verify that `pathways-flavor` gets successfully injected with `google.com/tpu: 0` in the unified resource group. Adding that coverage would make the Go orchestrator tests 100% complete for the actual Pathways TPU worker use case.

modules/management/kubectl-apply/main.tf
```diff
@@ -31,9 +31,11 @@ locals {
...
   kueue_config_template_vars = merge(
     {
-      pathways_cpu_quota    = 480
-      pathways_memory_quota = "2000G"
-      tpu_quota             = "999999" # Default high value if not set
+      pathways_cpu_quota      = 480
+      pathways_memory_quota   = "2000G"
+      tpu_flavor_cpu_quota    = "999999" # High default to avoid limiting TPU pods by CPU
+      tpu_flavor_memory_quota = "999999T"
+      tpu_quota               = "999999" # Default high value if not set
```
shubpal07, line 38: (Non-blocking): The default quotas (`tpu_flavor_cpu_quota` = `"999999"` and `tpu_flavor_memory_quota` = `"999999T"`) are great for ensuring things work without artificial scheduling limits.

Let's make sure we update the module's `README.md` (or the input variables documentation in a follow-up) to explicitly mention that users who want to enforce strict capacity sharing on the TPU pool can override these variables via `config_template_vars`.

shubpal07 (dismissed):
LGTM to merge, with some minor, non-blocking suggestions for unit test coverage and some documentation update(s).
Thanks for working on this critical fix!

pkg/orchestrator/gke/infra_manager_test.go
SwarnaBharathiMantena: Great point! I will add this test and update the PR.

shubpal07 (approved):

PR #5846, opened 2026-06-24
cleanup(gke): remove redundant switch in GenerateGKENodeSelectorLabel

kadupoornima:
Simplify `GenerateGKENodeSelectorLabel` by Removing Redundant Switch.

#### The Problem
The `GenerateGKENodeSelectorLabel` function maps accelerator types to GKE node selector labels. It contains a hardcoded fallback switch statement for "direct values":
```go
// Fallback for direct values
switch resolvedLower {
case "nvidia-tesla-a100", "tpu-v4-podslice", "tpu-v6e-slice", "tpu-v5p-slice", "tpu-v5-lite-podslice":
    return acceleratorType
}
```
This switch statement is **completely redundant**. If any of these values (or their prefixes) are passed to the function, they do not match any keys in the `machineFamilyToLabelMap`. As a result, the function naturally falls through all checks and returns `acceleratorType` unmodified at the end of the function anyway. 

Maintaining this hardcoded list of direct values in a switch statement in addition to the map increases maintenance overhead and leads to duplication.

#### The Solution
Remove the redundant `switch` block entirely. Let the function rely on its natural fall-through behavior, which already returns the input string if no map keys match.

#### The Impact
*   **Code Simplicity**: Reduces the size and complexity of the function, making it easier to read and maintain.
*   **Maintainability**: Eliminates the risk of developers forgetting to keep the switch statement in sync with the map when adding new accelerators.

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -1141,7 +1141,6 @@ func (g *GKEOrchestrator) generateAndApplyManifest(opts ManifestOptions, profile
 	return g.ApplyManifest(gkeManifestContent, outputManifestPath, opts.WorkloadName)
 }
 
-// TODO Use a map
 var machineFamilyToLabelMap = map[string]string{
 	"g2-standard":   "nvidia-l4",
 	"a3-highgpu":    "nvidia-h100-80gb",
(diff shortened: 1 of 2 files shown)


pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: nit: Can you please remove this stray TODO as well. 
LGTM otherwise

kadupoornima: Done.

Neelabh94 (approved):

PR #5885, opened 2026-07-01
Fix(gcluster): enforce strict GKE NAP accelerator validation

SwarnaBharathiMantena:
### Problem
GKE Node Auto-Provisioning (NAP) validation in gcluster was too permissive. If a cluster was configured with specific accelerator limits (e.g., H100 only), gcluster would still allow submissions for unconfigured accelerator types (e.g., L4).

This happened because the parser propagated specific limits to the generic `nvidia.com/gpu` and `google.com/tpu` keys in memory. The validator saw these generic limits were greater than zero and incorrectly approved the workload. These jobs were accepted locally but hung in a Pending state indefinitely on GKE because the autoscaler had no authority to provision the unconfigured hardware.

### Solution
Refined the validation logic in `isNAPEnabledForMachineType` (nap.go):

**Strict Matching**: Added helper functions (isSpecificGPUKey and isSpecificTPUKey) to detect if the cluster has opted into specific accelerator limits.
Selective Fallback: If specific limits are configured (e.g., H100), the validator now bypasses the generic nvidia.com/gpu / google.com/tpu fallback for unconfigured types (e.g., L4) and strictly rejects the submission.
**Preserved Propagation**: We preserved the generic limit propagation in parseNAPLimits to ensure that Kueue resource activation and cluster capacity calculations continue to function correctly.

### Verification Results
**Unit Tests**: Ran the GKE orchestrator test suite (go test -v ./pkg/orchestrator/gke/...). All tests passed, including TestPopulateClusterMetadata_NAPLimitsLoopOrder (which verifies generic limit propagation).
**Live Manual Tests**: Verified on the deployed GKE NAP cluster :
* Rejection Test: Requesting an unconfigured L4 GPU (l4-1) was successfully blocked:
text
```
Error: workload submission rejected. Compute type "l4-1" is not configured within your cluster's Node Auto-Provisioning (NAP) limits. Configured limits on cluster: cpu, google.com/tpu, memory, nvidia-h100-mega-80gb, nvidia.com/gpu, tpu-v6e-slice
```
* Acceptance Test: Requesting a configured TPU (v6e-4) was successfully accepted and generated the manifest.

### DIFF
--- a/pkg/orchestrator/gke/nap.go
+++ b/pkg/orchestrator/gke/nap.go
@@ -30,22 +30,7 @@ func (g *GKEOrchestrator) isNAPEnabledForMachineType(machineType, zone string) (
 	resolvedType := config.ResolveMachineType(machineType)
 
 	if config.IsTPU(resolvedType) {
-		key := strings.ToLower(g.GenerateGKENodeSelectorLabel(resolvedType))
-		if limit, exists := g.napLimits[key]; exists {
-			return limit > 0, nil
-		}
-		// Fallback to generic TPU limit ONLY if no specific TPU limits are configured.
-		hasSpecificTPULimits := false
-		for k := range g.napLimits {
-			if isSpecificTPUKey(k) {
-				hasSpecificTPULimits = true
-				break
-			}
-		}
-		if hasSpecificTPULimits {
-			return false, nil
-		}
-		return g.napLimits["google.com/tpu"] > 0, nil
+		return g.validateTPUNAPLimit(resolvedType)
 	}
 
 	cap, err := g.FetchMachineCapabilities(resolvedType, zone)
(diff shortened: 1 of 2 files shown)


pkg/orchestrator/gke/nap.go
```diff
@@ -30,28 +30,64 @@ func (g *GKEOrchestrator) isNAPEnabledForMachineType(machineType, zone string) (
...
+func (g *GKEOrchestrator) validateTPUNAPLimit(resolvedType string) (bool, error) {
+	key := strings.ToLower(g.GenerateGKENodeSelectorLabel(resolvedType))
+	if limit, exists := g.napLimits[key]; exists {
+		return limit > 0, nil
+	}
+	// Fallback to generic TPU limit ONLY if no specific TPU limits are configured.
+	hasSpecificTPULimits := false
+	for k := range g.napLimits {
+		if isSpecificTPUKey(k) {
+			hasSpecificTPULimits = true
```
Neelabh94: we can "return false, nil " directly here, and then won't need the code at line 53, 56 to 62

pkg/orchestrator/gke/nap.go
```diff
@@ -30,28 +30,64 @@ func (g *GKEOrchestrator) isNAPEnabledForMachineType(machineType, zone string) (
...
+	if !isKnownGKEAccelerator(key) {
+		return false, fmt.Errorf("unknown accelerator label: %q", cap.Accelerators[0].Type)
+	}
+	if limit, exists := g.napLimits[key]; exists {
+		return limit > 0, nil
+	}
+	// Fallback to generic GPU limit ONLY if no specific GPU limits are configured.
+	hasSpecificGPULimits := false
+	for k := range g.napLimits {
+		if isSpecificGPUKey(k) {
```
Neelabh94, line 74: same as above, we can do a return here directly

Neelabh94 (changes requested):
small comments around code improvement. PTAL

pkg/orchestrator/gke/nap.go
```diff
@@ -126,6 +154,14 @@ func extractShortReservationName(resName string) string {
 	return parts[len(parts)-1]
 }
 
+func isSpecificGPUKey(key string) bool {
+	return strings.HasPrefix(key, "nvidia-") && key != "nvidia.com/gpu"
+}
+
+func isSpecificTPUKey(key string) bool {
+	return (strings.HasPrefix(key, "tpu-") || strings.HasPrefix(key, "ct")) && key != "google.com/tpu"
+}
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Simplified both functions.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: Thanks! Refactored the code.

pkg/orchestrator/gke/nap.go
SwarnaBharathiMantena: Thanks! Refactored the code.

agrawalkhushi18 (approved):

Neelabh94 (approved):
LGTM

PR #5897, opened 2026-07-02
fix(destroy): recover when GKE is unreachable 

Neelabh94:
This PR helps improve the destroy mechanism by removing kubernetes/helm/kubectl resources from state if the GKE control plane becomes inaccessible during the destroy phase.

Since these are logical entities that live solely inside GKE's control plane database (etcd). Since the GKE cluster itself is being deleted, all of these objects are deleted with it.

### Submission Checklist


pkg/shell/terraform.go
```diff
@@ -502,3 +518,52 @@ func TfVersion() (string, error) {
...
+
+	resources := getResourcesRecursively(state.Values.RootModule)
+	for _, res := range resources {
+		if strings.HasPrefix(res.Type, "kubernetes_") ||
+			strings.HasPrefix(res.Type, "helm_") ||
+			strings.HasPrefix(res.Type, "kubectl_") {
+			logging.Info("GKE cluster unreachable: removing resource %s from state", res.Address)
+			if err := tf.StateRm(ctx, res.Address); err != nil {
+				logging.Error("failed to remove resource %s from state: %v", res.Address, err)
+			}
```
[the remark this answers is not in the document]
Neelabh94: This is an acceptable trade-off to ensure that the clean-up is as thorough as possible before failing.

pkg/shell/terraform.go
```diff
@@ -330,6 +331,21 @@ func applyOrDestroy(tf *tfexec.Terraform, b ApplyBehavior, of OutputFormat, dest
...
+			if rmErr := RemoveKubernetesResourcesFromState(tf); rmErr != nil {
+				return false, fmt.Errorf("failed to clean unreachable Kubernetes resources from state: %w (original error: %v)", rmErr, err)
+			}
+			return planModule(tf, planPath, destroy)
+		}
+		return false, err
+	}
+	return wantsChange, nil
+}
+
```
shubpal07: Modifying state (`RemoveKubernetesResourcesFromState`) during the plan phase can permanently alters the state file before user confirmation. If a user runs `gcluster destroy` interactively, encounters a transient network hiccup, and subsequently declines/aborts the destroy plan at the prompt, their state will be left out-of-sync with the healthy GKE cluster.

We should only run this cleanup automatically in non-interactive mode (when `--auto-approve` / AutomaticApply is enabled). For interactive runs, prompt the user first (e.g., using PromptYesNo) before running the state cleanup

pkg/shell/terraform.go
```diff
@@ -502,3 +518,52 @@ func TfVersion() (string, error) {
...
 	return version.TerraformVersion, nil
 }
+
+// IsKubernetesUnreachableError returns true if the error indicates GKE is unreachable
+func IsKubernetesUnreachableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "dial tcp [::1]:80") ||
```
shubpal07: **Fragile loopback check:** Hardcoding [::1]:80 can miss connection failures if the local machine resolves to 127.0.0.1 or uses a different default fallback port (like 8080).

Can we think to generalize to match loopback prefixes dynamically?

pkg/shell/terraform.go
```diff
@@ -502,3 +518,52 @@ func TfVersion() (string, error) {
...
+		strings.Contains(msg, "Kubernetes cluster unreachable") ||
+		strings.Contains(msg, "failed to create kubernetes rest client") ||
+		strings.Contains(msg, "no configuration has been provided, try setting KUBERNETES_MASTER")
```
shubpal07: Using case-sensitive checks on err.Error() can fail on minor message differences across provider versions (e.g. "Failed to create Kubernetes REST client" vs "failed to create kubernetes rest client").

We may convert the error string to lowercase once (`strings.ToLower(err.Error())`) before running the matching checks.

shubpal07 (changes requested):

pkg/shell/terraform.go
Neelabh94: Good catch, made it more robust by wrapping this inside the --robust flag and asking for an explicit user permission before cleaning up the state.

pkg/shell/terraform.go
Neelabh94: Done

pkg/shell/terraform.go
Neelabh94: Good point, done!

pkg/shell/terraform.go
```diff
@@ -502,3 +503,55 @@ func TfVersion() (string, error) {
...
+	// Convert to lowercase to ensure case-insensitive matching
+	msg := strings.ToLower(err.Error())
+	return dialLoopbackRegex.MatchString(msg) ||
+		strings.Contains(msg, "kubernetes cluster unreachable") ||
+		strings.Contains(msg, "failed to create kubernetes rest client") ||
+		strings.Contains(msg, "no configuration has been provided, try setting kubernetes_master")
+}
+
+// RemoveKubernetesResourcesFromState removes all kubernetes/helm/kubectl provider resources from the state
+func RemoveKubernetesResourcesFromState(tf *tfexec.Terraform) error {
```
parulbajaj01: Could we add unit tests for`RemoveKubernetesResourcesFromState` in pkg/shell/terraform_test.go (verifying resource filtering with mock/stubbed state data)?

cmd/destroy.go
parulbajaj01, line 226: Could we add unit test for this as well in cmd/destroy_test.go (testing behavior with --robust enabled/disabled and auto-approve vs interactive prompt)?

cmd/destroy.go
```diff
@@ -200,9 +200,40 @@ func destroyTerraformGroup(groupDir string) error {
...
+				} else {
+					return err
+				}
+
+				if forceDestroyApproved {
+					logging.Warn("GKE cluster is unreachable during destroy. Attempting to remove Kubernetes-related resources from state and retrying.")
+					if rmErr := shell.RemoveKubernetesResourcesFromState(tf); rmErr != nil {
+						return fmt.Errorf("failed to clean unreachable Kubernetes resources from state: %w (original error: %v)", rmErr, err)
+					}
+					return shell.Destroy(tf, shell.AutomaticApply, shell.TextOutput)
```
parulbajaj01: If the user explicitly declines ([y/n] prompt -> n) state modification during interactive destroy, destroyTerraformGroup returns an error, which triggers destroyRunner's --robust retry loop (up to 3 attempts). This causes gcluster to re-prompt the user up to 3 times for the same decision.

When the user enters n, we should prevent further retries for this run or abort the retry loop for that group.

pkg/shell/terraform.go
```diff
@@ -502,3 +503,55 @@ func TfVersion() (string, error) {
...
+
+	resources := getResourcesRecursively(state.Values.RootModule)
+	for _, res := range resources {
+		if strings.HasPrefix(res.Type, "kubernetes_") ||
+			strings.HasPrefix(res.Type, "helm_") ||
+			strings.HasPrefix(res.Type, "kubectl_") {
+			logging.Info("GKE cluster unreachable: removing resource %s from state", res.Address)
+			if err := tf.StateRm(ctx, res.Address); err != nil {
+				logging.Error("failed to remove resource %s from state: %v", res.Address, err)
+			}
```
parulbajaj01, line 552: To prevent silent failure can we aggregate any StateRm errors (e.g. using errors.Join or an error slice) and return a combined error at the end if one or more resources failed to remove?

pkg/shell/terraform.go
```diff
@@ -502,3 +503,55 @@ func TfVersion() (string, error) {
 
 	return version.TerraformVersion, nil
 }
+
+var dialLoopbackRegex = regexp.MustCompile(`dial tcp (?:127\.0\.0\.1|\[::1\]|localhost):[0-9]+`)
```
parulbajaj01: It currently only matches loopback addresses (127.0.0.1, [::1], localhost). 
Should we also add checks for i/o timeout, connection refused, and context deadline exceeded when talking to the Kubernetes API?

parulbajaj01 (dismissed):

pkg/shell/terraform.go
Neelabh94: I think a best-effort approach is safer here. Returning an error and aborting the cleanup on a single StateRm failure (which could be a non-critical or transient error) would bypass the subsequent retry of terraform destroy entirely. Instead, logging the errors while proceeding allows us to maximize the chance of a successful teardown of the remaining infrastructure.

pkg/shell/terraform.go
Neelabh94: Adding generic network checks like i/o timeout or connection refused is too broad and risky. Since IsKubernetesUnreachableError triggers a destructive force-cleanup (removing GKE resources from the state file permanently), false positives caused by transient API timeouts or rate limits on other providers could cause accidental state corruption. We should stick to specific GKE/Kubernetes-related error signatures to ensure safety.

cmd/destroy.go
Neelabh94: Great catch. Corrected this

cmd/destroy.go
Neelabh94: Added

pkg/shell/terraform.go
Neelabh94: Done

pkg/shell/terraform.go
```diff
@@ -502,3 +503,61 @@ func TfVersion() (string, error) {
...
+			strings.HasPrefix(res.Type, "helm_") ||
+			strings.HasPrefix(res.Type, "kubectl_") {
+			logging.Info("GKE cluster unreachable: removing resource %s from state", res.Address)
+			if err := tf.StateRm(ctx, res.Address); err != nil {
+				logging.Error("failed to remove resource %s from state: %v", res.Address, err)
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Neelabh94: N/A, replied above

pkg/shell/terraform_test.go
```diff
@@ -50,3 +54,100 @@ func (s *MySuite) TestFindTerraform(c *C) {
...
 }
+
+func (s *MySuite) TestIsKubernetesUnreachableError(c *C) {
+	c.Assert(IsKubernetesUnreachableError(nil), Equals, false)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("some other error")), Equals, false)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("dial tcp [::1]:80: connect: connection refused")), Equals, true)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("Error: Kubernetes cluster unreachable: invalid configuration")), Equals, true)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("failed to create kubernetes rest client for read")), Equals, true)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("no configuration has been provided, try setting KUBERNETES_MASTER environment variable")), Equals, true)
+}
```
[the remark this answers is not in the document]
Neelabh94: Done

pkg/shell/terraform_test.go
```diff
@@ -50,3 +54,104 @@ func (s *MySuite) TestFindTerraform(c *C) {
 	os.Setenv("PATH", pathEnv)
 	c.Assert(err, NotNil)
 }
+
+func (s *MySuite) TestIsKubernetesUnreachableError(c *C) {
+	c.Assert(IsKubernetesUnreachableError(nil), Equals, false)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("some other error")), Equals, false)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("kubernetes_namespace.ns: dial tcp 10.128.0.2:443: connect: connection refused")), Equals, true)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("helm_release.release: dial tcp [::1]:80: connect: connection refused")), Equals, true)
+	c.Assert(IsKubernetesUnreachableError(fmt.Errorf("kubernetes_namespace.ns: dial tcp 35.184.1.2:443: i/o timeout")), Equals, true)
```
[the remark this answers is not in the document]
Neelabh94: done

pkg/shell/terraform.go
```diff
@@ -502,3 +503,66 @@ func TfVersion() (string, error) {
...
```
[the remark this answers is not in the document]
Neelabh94: I think a best-effort approach is safer here. Returning an error and aborting the cleanup on a single StateRm failure (which could be a non-critical or transient error) would bypass the subsequent retry of terraform destroy entirely. Instead, logging the errors while proceeding allows us to maximize the chance of a successful teardown of the remaining infrastructure.

pkg/shell/terraform.go
shubpal07: Thanks @Neelabh94

shubpal07 (approved):
LGTM! Thanks for addressing the feedback @Neelabh94 , particularly the safe interactive prompt logic and the comprehensive test coverage.

PR #5908, opened 2026-07-06
Fix(gke-cluster): use system_node_pool_machine_type for initial default node pool

SwarnaBharathiMantena:
## Description
When creating a GKE cluster with `remove_default_node_pool = true`, Terraform initializes a temporary default node pool that is immediately deleted after the cluster bootstrapping completes. Previously, this temporary node pool hardcoded `machine_type = "e2-medium"` unless confidential nodes were enabled.

In specialized accelerator-only or TPU-focused zones, general-purpose E2 machine shapes do not exist. Consequently, cluster bootstrapping failed with `GCE_NOT_FOUND` even when the user explicitly configured a valid `system_node_pool_machine_type`.

This PR replaces the hardcoded `"e2-medium"` fallback with `var.system_node_pool_machine_type` in the default node pool's `node_config`. 

## Why this is safe
Because `remove_default_node_pool = true` and `initial_node_count = 1` are set, this node pool exists for only ~60 seconds during cluster creation before being deleted by Terraform. Using `var.system_node_pool_machine_type` introduces no ongoing cost or quota overhead while ensuring cluster bootstrapping respects user-configured zone capabilities.

## Checklist
- [x] Verified Terraform syntax and formatting via `terraform fmt -check`
- [x] Verified that cluster creation uses the configured system machine type for initial bootstrapping


shubpal07 (dismissed):

modules/scheduler/gke-cluster/main.tf
```diff
@@ -359,7 +359,7 @@ resource "google_container_cluster" "gke_cluster" {
   }
 
   node_config {
-    machine_type = var.enable_confidential_nodes ? var.system_node_pool_machine_type : "e2-medium"
+    machine_type = var.system_node_pool_machine_type
```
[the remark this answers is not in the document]
SwarnaBharathiMantena: Done

agrawalkhushi18 (dismissed):
LGTM

modules/scheduler/gke-cluster/main.tf
```diff
@@ -406,8 +406,8 @@ resource "google_container_cluster" "gke_cluster" {
       error_message = "FQDN Network Policy requires GKE Dataplane V2 to be enabled."
     }
     precondition {
-      condition     = !var.enable_confidential_nodes || !var.system_node_pool_enabled || can(regex("^(n2d-|c2d-|c3d?-|t2d-|g4-)", var.system_node_pool_machine_type))
-      error_message = "The system_node_pool_machine_type must be a confidential-compatible machine type (e.g., n2d, c2d, c3d, c3, t2d, g4) when enable_confidential_nodes is true and system_node_pool_enabled is true."
+      condition     = !var.enable_confidential_nodes || can(regex("^(n2d-|c2d-|c3d?-|t2d-|g4-)", var.system_node_pool_machine_type))
```
Neelabh94: From a UX perspective, since we already provide a default for non-confidential mode, should we also provide a default for confidential mode too?

shubpal07 (dismissed):

modules/scheduler/gke-cluster/main.tf
SwarnaBharathiMantena: Absolutely! Thanks for suggesting.

SwarnaBharathiMantena:
image
The failing tests are not related to the PR.

agrawalkhushi18 (approved):

PR #5917, opened 2026-07-07
integrate cluster health check into toolkit.

leiyiz:
passive cluster healthcheck daemon


community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,623 @@
...
+			// First check, don't fail immediately
+			allZeroDiff = false
+		}
+		lastRxPackets[match] = val
+	}
+	if allZeroDiff {
+		return SeverityWarning, fmt.Errorf("rx_packets difference is 0 on all GPU RDMA interfaces")
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
leiyiz: There are always background `rx_packets` even on idle nodes.

this tool is meant for A3+ and the assumption is that the node has RDMA.

community/cluster-health-check/deployment/dcgm-healthcheck.yaml
```diff
@@ -0,0 +1,119 @@
...
+        image: <YOUR_REGISTRY>/<YOUR_REPO>/cluster-health-check:latest
+        imagePullPolicy: Always
+        args: ["-check-interval=15s"]
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: LD_LIBRARY_PATH
+          value: /usr/local/nvidia/lib64
```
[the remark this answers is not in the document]
leiyiz: Will remove `LD_LIBRARY_PATH` from Dockerfile. it's left over from previous iterations

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,623 @@
...
+		parts := strings.Split(line, ",")
+		if len(parts) >= 2 {
+			width, _ := strconv.Atoi(strings.TrimSpace(parts[1]))
+			if width > 0 && width < 16 {
+				return SeverityWarning, fmt.Errorf("PCIe link width degraded on GPU %d: %d", i, width)
+			}
+		}
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
leiyiz: I think all A3+ machine types have x16 PCIe

community/cluster-health-check/README.md
```diff
@@ -0,0 +1,49 @@
+# DCGM Health Check
```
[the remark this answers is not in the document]
leiyiz: the `modules/README.md` doesn't contain reference to `cos-nvidia-bug-report`

is this check still relevant

community/cluster-health-check/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+RUN CGO_ENABLED=0 \
+    GOARCH=${TARGETARCH} \
+    GOOS=${TARGETOS} \
+    GOTOOLCHAIN=local \
+    go build -mod=vendor -o dcgm-healthcheck ./cmd/dcgm-healthcheck
+
+# -----------------------------------------------------------------------------
+# Base Debian Libraries
+# -----------------------------------------------------------------------------
+FROM gke.gcr.io/debian-base:bookworm-v1.0.2-gke.0@sha256:0c094a459519c81e5e6dcd969c38b865fee1f470196134a4a044fffcabd70fc0 AS base
```
bytetwin: Why is the image pinned to specific sha/commit versions instead of semver version ?

community/cluster-health-check/deployment/dcgm-healthcheck.yaml
```diff
@@ -0,0 +1,132 @@
...
+        app: dcgm-healthcheck
+    spec:
+      serviceAccountName: dcgm-healthcheck
+      hostNetwork: true
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: cloud.google.com/gke-accelerator
```
bytetwin, line 66: Will this label be present on tpu nodes too ? If yes we should restrict to gpu

community/cluster-health-check/deployment/daemonset.yaml
```diff
@@ -0,0 +1,132 @@
...
+                operator: Exists
+      initContainers:
+      - name: nvml-wait
+        image: gke.gcr.io/gke-distroless/bash:gke_distroless_20240307.00_p0
+        command:
+        - bash
+        - -c
+        - |
+          echo "Waiting for NVML drivers..."
+          while true; do
```
bytetwin: This will enter into an infinite loop if the drivers havent loaded

community/cluster-health-check/deployment/dcgm-healthcheck.yaml
```diff
@@ -0,0 +1,132 @@
+# Copyright 2026 Google LLC
```
bytetwin, line 1: This file installs Role bindings along with daemonset. Should this be renamed something other than daemonset eg) passive-healthcheck.yaml ?

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,738 @@
...
+	log.Println("Starting nv-hostengine...")
+	hostEngine := exec.CommandContext(ctx, "nv-hostengine", "-n")
+	if err := hostEngine.Start(); err != nil {
+		log.Fatalf("Failed to start nv-hostengine: %v", err)
+	}
+
+	go func() {
+		err := hostEngine.Wait()
+		if ctx.Err() != nil {
+			log.Printf("nv-hostengine exited during shutdown: %v", err)
```
bytetwin: shutdown ?

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,738 @@
...
+		log.Printf("Error patching node metadata %s: %v", nodeName, err)
+	} else {
+		log.Printf("Successfully updated node %s with status %s", nodeName, finalStatus)
+	}
+}
+
+var xidRegex = regexp.MustCompile(`(?i)(?:xid|sxid)(?:\s*\(.*?\))?:\s*(\d+)`)
+
+func checkKernelLogsForXidSxid(ctx context.Context, fatalXids map[int]struct{}) (string, error) {
+	out, err := exec.CommandContext(ctx, "dmesg").CombinedOutput()
```
bytetwin: dmesg reads from the kernel buffer which is circular buffer. 

Running plain dmesg might result in previous XID errors being reported repeatedly until they are cleared.

Option would be to use dmesg -c that reads and immediately clears the buffers. The flipside is that other tools that may depend on dmesg would lose those messages. Find if there is a better way to do this or parse dmesg output's timestamp and read the ones since last check

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,738 @@
...
+		// {"NIC heartbeat", checkNICHeartbeat},
+		{"XID/SXID monitoring", func(c context.Context) (string, error) { return checkKernelLogsForXidSxid(c, fatalXids) }},
+		{"ECC errors", checkECCErrors},
+		{"PCIe link health", checkPCIe},
+		{"InfiniBand links", checkIB},
+		{"GPU temperature", checkTemperature},
+		// {"HCA Firmware", checkHCAFW},
+		// {"GPU Power draw", checkPower},
+	}
+	for _, chk := range passiveChecks {
```
bytetwin, line 218: Could these checks be run independently using go routines/wait groups instead of being sequential ?

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,738 @@
...
+func checkIB(ctx context.Context) (string, error) {
+	out, err := exec.CommandContext(ctx, "ibstat").CombinedOutput()
+	if err != nil {
+		return "", nil
+	}
+	if strings.Contains(string(out), "Down") {
+		return SeverityWarning, fmt.Errorf("ibstat reported Down")
+	}
+
+	outDiag, errDiag := exec.CommandContext(ctx, "ibdiagnet").Output()
```
bytetwin: ibdiagnet queries the entire fabric for the health and running this across all the nodes will stress the fabric.

Instead we should check the local node on the health using `ibstat` which checks the state of the links plugged into the node and the nic status

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -0,0 +1,738 @@
...
+	}
+	return fatalXids
+}
+
+func runHealthCheck(mainCtx context.Context, clientset *kubernetes.Clientset, nodeLister corev1listers.NodeLister, nodeName string) {
+	ctx, cancel := context.WithTimeout(mainCtx, 5*time.Minute)
+	defer cancel()
+
+	var errorMessages []string
+	highestSeverity := SeverityWarning
```
bytetwin, line 200: Setting this to warning will result in info also escalated to warning. While the health checks do not throw info, this bug will go unnoticed if we modify the health checks in the future.

Lets set it to some baseline

bytetwin (changes requested):

community/cluster-health-check/deployment/dcgm-healthcheck.yaml
leiyiz: TPU nodes have `cloud.google.com/gke-tpu-accelerator`

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: This is intentional - to start with we want to preserve every XID error because we don't want XID errors to be cleared until the recovery action is done.

To begin with we will read the full ring buffer and once the node is rebooted or recreated to recover the fault, the buffer should be empty again.

In the future we will limit read range.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: warning is the baseline. the design is that if nothing goes wrong no `Node Condition` message will be generated and patched to the node object.

Adding some comments to make that clear.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: sg, removing ibdiagnet.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: I chose to sequentially run the tests to have a deterministic sequence of error messages if the same set of tests fail.

community/cluster-health-check/deployment/daemonset.yaml
leiyiz: thanks for the catch, adding a time limit and exit (1)

community/cluster-health-check/Dockerfile
leiyiz: this is leftover from internal build pipeline. changing to semver `v1.0.2-gke.0`

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: added comments to explain this.

basically when `hostEngine.Wait()` returns, we want to know if the overall container is shutting down.

community/cluster-health-check/deployment/dcgm-healthcheck.yaml
leiyiz: done.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
bytetwin: The problem is when the node isnt rebooted but the gpu is healthy, then this method will still read the previous XID error and mark it as failure.

If you dont want to clear, probably you should check timestamp based

community/cluster-health-check/Dockerfile
bytetwin: Same as above - pinned to sha

bytetwin (changes requested):

community/cluster-health-check/Dockerfile
leiyiz: done.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: I added a `--since` filter with timestamp of the container startup to align with how `dcgmi` handles xid - it doesn't clear out the xid errors by default

mufaqam-gcl (dismissed):

community/cluster-health-check/README.md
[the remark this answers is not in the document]
leiyiz: closing.

simrankaurb (dismissed):

bytetwin (dismissed):

community/cluster-health-check/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+![Platform: GKE](https://img.shields.io/badge/Platform-GKE-green.svg)
+
+This tool runs as a Kubernetes DaemonSet to perform passive health checks on NVIDIA GPUs within a GKE cluster using DCGM (Data Center GPU Manager) and NVML.
+
+It continuously monitors GPU nodes for hardware errors, XID errors, and InfiniBand/network issues. When an issue is detected, the agent patches the Kubernetes Node object to report the failure, applying the appropriate severity label (`Warning`, `Failure` or `Fatal`).
+
+--------------------------------------------------------------------------------
+
+## ✨ Key Features
```
Jiaqicao257, line 11: Can we list the exact tests / logs that are included?

leiyiz: will address in follow up pr

mufaqam-gcl (dismissed):

simrankaurb (dismissed):

Neelabh94 (approved):

mufaqam-gcl (approved):

PR #5923, opened 2026-07-09
fix(job): add shared-tmp volumes and log exit codes in GKE Pathways jobs

Neelabh94:
### Summary
This PR adds default `shared-tmp` `hostPath` volume configuration and workload container exit code logging to Pathways job templates.

### Changes
- **Templates (`pathways_jobset.tmpl`)**:
  - Added the `shared-tmp` `hostPath` volume definition pointing to `/tmp` with `type: DirectoryOrCreate` under the `volumes` section of both the `pathways-head` and `worker` replicated jobs.
  - Mounted the `shared-tmp` volume at `/tmp` inside the `workload-container` container (on the head pod) and the `pathways-worker` container (on the worker pods).
  - Appended exit code logging output (`echo "Exit code: $EXIT_CODE"`) right after the workload run ends in the main container command.
  
- **Tests (`gke_job_orchestrator_test.go`)**:
  - Updated `TestGeneratePathwaysManifest` assertions to verify the presence of the `shared-tmp` volume/mount definitions and exit code print in GKE manifest generation.

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator_test.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator_test.go
@@ -491,6 +491,10 @@ func TestGeneratePathwaysManifest(t *testing.T) {
 		`cpu: "2"`,
 		`memory: "8Gi"`,
 		"kill -SIGTERM $PID",
+		"echo \"Exit code: $EXIT_CODE\"",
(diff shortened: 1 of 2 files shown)


agrawalkhushi18 (approved):
LGTM

PR #5925, opened 2026-07-09
feat(job): add job inspect command for easy diagnostics

Neelabh94:
This pull request introduces a diagnostic tool for GKE clusters, allowing users to quickly capture cluster state, node health, and workload configurations. The new command, gcluster job inspect, generates a timestamped log file and provides optional console output, significantly simplifying the troubleshooting process for scheduling delays or resource exhaustion issues.

### Changes

* **New 'inspect' command:** Added a new gcluster job inspect command to perform comprehensive diagnostic sweeps of GKE clusters and workloads.
* **Diagnostic logging:** Implemented automated log generation that captures cluster infrastructure, node status, Kueue/JobSet configurations, and relevant console links.
* **Orchestrator interface update:** Updated the JobOrchestrator interface to include InspectCluster and added necessary InspectOptions.
* **Documentation:** Updated the job guide to include instructions and examples for using the new inspection tool.


pkg/orchestrator/gke/inspector_test.go
```diff
@@ -0,0 +1,167 @@
...
+		"kubectl get workloads -A":                                        {{ExitCode: 0, Stdout: "workloads-list"}},
+		"kubectl describe jobsets":                                        {{ExitCode: 0, Stdout: "jobset-config"}},
+		"kubectl describe workloads":                                      {{ExitCode: 0, Stdout: "workload-config"}},
+	}
+}
+
+func TestInspectCluster_Success(t *testing.T) {
+	clusterName := "test-cluster"
+	location := "us-central1-a"
+	project := "test-project"
```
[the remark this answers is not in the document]
Neelabh94: done

pkg/orchestrator/gke/inspector.go
```diff
@@ -0,0 +1,299 @@
...
+	if opts.WorkloadName != "" {
+		workloadLink := struct {
+			desc string
+			url  string
+		}{
+			desc: fmt.Sprintf("Cloud Console for workload %s", opts.WorkloadName),
+			url:  fmt.Sprintf("https://console.cloud.google.com/kubernetes/service/%s/%s/%s/%s/details?project=%s", opts.ClusterLocation, opts.ClusterName, workloadNamespace, opts.WorkloadName, opts.ProjectID),
+		}
+		links = append(links, workloadLink)
+	}
```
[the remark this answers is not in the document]
Neelabh94: done

docs/gcluster_job_guide.md
```diff
@@ -253,6 +253,35 @@ Verify that the Kubernetes JobSet ran successfully on your GKE cluster.
...
+    ```
+
+    To view logs output in the console in addition to saving them to the diagnostic file, add the `--show` (or `-s`) flag:
+
+    ```bash
+    ./gcluster job inspect --name my-python-app-job --show
+    ```
+
+    The tool will create a timestamped log file `gcluster-inspect-<cluster>-<timestamp>.log` in your current working directory containing:
+
```
mufaqam-gcl, line 276: Shall we include a sample screenshot of the log?

pkg/orchestrator/gke/inspector.go
```diff
@@ -0,0 +1,445 @@
...
+	writer.runAndLog("Local Setup: Active gcloud configuration", "gcloud", "config", "list")
+
+	// --- 2. GKE Infrastructure ---
+	writer.runAndLog("GKE: Cluster Details", "gcloud", "container", "clusters", "describe", opts.ClusterName, "--location", opts.ClusterLocation, "--project", opts.ProjectID, "--format=yaml")
+	writer.runAndLog("GKE: Node Pool Details", "gcloud", "container", "node-pools", "list", "--cluster", opts.ClusterName, "--location", opts.ClusterLocation, "--project", opts.ProjectID)
+
+	// ConfigMaps (graceful handle if not present)
+	metadataCM := fmt.Sprintf("%s-metadata", opts.ClusterName)
+	resourcesCM := fmt.Sprintf("%s-resources", opts.ClusterName)
+	writer.runAndLog("GKE: Cluster Metadata ConfigMap Details", "kubectl", "get", "configmap", metadataCM, "-n", "default", "-o", "yaml")
```
SwarnaBharathiMantena: The namespace might not be `default` all the time. A user might be creating these resources in a different namespace.
One way to tackle this is providing an option to pass a specific namespace with the inspector command.

pkg/orchestrator/gke/inspector.go
```diff
@@ -0,0 +1,445 @@
...
+	writer.runAndLog("GKE: Cluster Resources ConfigMap Details", "kubectl", "get", "configmap", resourcesCM, "-n", "default", "-o", "yaml")
```
SwarnaBharathiMantena: Same as the above suggestion.

pkg/orchestrator/gke/inspector.go
```diff
@@ -0,0 +1,445 @@
...
+
+type inspectWriter struct {
+	writer   io.Writer
+	show     bool
+	executor Executor
+}
+
+func (w *inspectWriter) runAndLog(description string, command string, args ...string) {
+	outputStr := fmt.Sprintf("Description: %s\nCommand: %s %v\n", description, command, args)
+	if w.show {
```
SwarnaBharathiMantena: Instead of checking `show` status on every print block, we can construct an `io.MultiWriter` once inside `InspectCluster` if `opts.Show` is active.

This lets us refactor `runAndLog` to write to `w.writer` unconditionally, removing several if branches and simplifying the method:

pkg/orchestrator/gke/inspector.go
Neelabh94: I agree, I am however a little hesitant to add a GKE specific flag --namespace here. I believe like already done in other commands, we can overcome this by fetching the current namespace of the user and then fetching from there? 

I believe in enterprise settings, user's access might be restricted to specific namespaces they are working in and hence all their resources should be in the same namespace. 
Do you think this can work?

SwarnaBharathiMantena: Fetching the active namespace from kubeconfig context as a default is a great idea for UX, and I think we should still support an explicit `--namespace / -n` flag as an override.

A user's local kubeconfig context might still be set to `default` while they are inspecting a job or ConfigMap created in another namespace or when an admin inspects another team's namespace. As namespace is a core Kubernetes concept (not GKE-specific), so having `--namespace` with a fallback to the active context is an advantage from the UX perspective and gives the flexibility.

Neelabh94: Discussed offline, this can be taken up as a follow-up PR among other enhancements.

SwarnaBharathiMantena (approved):
LGTM!

PR #5934, opened 2026-07-10
feat(slurm): support expedited requeue and start-only health checks for Slurm 25.11

AdarshK15:
## Summary
This PR introduces support for the following Slurm 25.11 features: 
* **Expedited Requeue**: If a job is running and its node crashes or gets preempted, this feature automatically moves the failed job to the front of the queue to run again as soon as possible, instead of sending it to the back of the line. Users can trigger this by submitting jobs with `sbatch --requeue=expedite`.

* **Start-Only Health Checks**: If a cluster has a custom health-check script, Slurm will run it continuously in the background on a timer. This feature tells Slurm to only run the script exactly once when a node first boots up, and then stop. This prevents the script from constantly interrupting the compute nodes while they are busy running jobs.

### Key Changes
* **Terraform Variables:** Added `enable_expedited_requeue` and `enable_health_check_start_only` as boolean settings in the controller module to be set in blueprint.
* **Config Generation:** Updated the Config generator (conf_v2511.py) to automatically inject Slurm settings (SlurmctldParameters and HealthCheckNodeState=START_ONLY) based on blueprint settings.
* **Unit Tests:** Added parameterized tests in tests/test_conf.py to ensure these flags only ever get applied to Slurm 25.11 clusters.

### Testing & Cluster Validation
Tested manually with a live cluster with Slurm 25.11.
1. Configuration Check
```bash
[adarsh2511-controller ~]$ scontrol show config | grep HealthCheckNodeState
HealthCheckNodeState    = START_ONLY
[adarsh2511-controller ~]$ scontrol show config | grep SlurmctldParameters
SlurmctldParameters     = cloud_dns,enable_configless,idle_on_node_suspend,enable_expedited_requeue
```
2. **Functional Test: `enable_expedited_requeue`**
Submitted a job with `--requeue=expedite`  and after provisioning, deleted the node manually. the job was requeued with highest priority and was assigned to a new node.

    1. Submitted sbatch -p dyn --requeue=expedite --wrap="sleep 600" (Job 3), which allocated onto node-1.
    2. Forcefully deleted the compute node (adarsh2511-dynamicnodeset-1) via the GCP UI to simulate a fatal hardware failure.
    3. **Observation:** slurmctld marked node down*, triggered the failover logic, and instantly auto-requeued the job in CF state on a new node (node-3) gracefully instead of failing outright.
    ```bash
    [adarsh2511-controller ~]$ sinfo
    PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
    dyn          up   infinite      1 alloc# adarsh2511-dynamicnodeset-3
    dyn          up   infinite      1  down* adarsh2511-dynamicnodeset-1
    dyn          up   infinite      1  alloc adarsh2511-dynamicnodeset-2
    [adarsh2511-controller ~]$ squeue
    JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
        3       dyn     wrap adarsh   CF       0:10      1 adarsh2511-dynamicnodeset-3
        1       dyn     wrap adarsh   R        6:54      1 adarsh2511-dynamicnodeset-2
    ```
3. **Functional Test: `enable_health_check_start_only`**
Verified that setting `HealthCheckNodeState=START_ONLY` functionally restricts health-check operations strictly to initial node-boot occurrences, successfully overriding internal loop dependencies.

    1. Added a dummy health script (`HealthCheckProgram=$HOME/dummy_health.sh`) to `slurm.conf` and set it to run every 30 seconds (`HealthCheckInterval=30`).
    2. **Observation:** For nodes that were already online, the log stayed completely empty, proving it completely ignores the 30-second timer and won't interrupt active jobs. To test a fresh boot, I submitted a job to wake up a suspended (`idle~`) dynamic node. The log captured exactly one timestamp right as the node powered on, confirming the check only runs during startup.
    ```bash
    [adarsh2511-controller ~]$ sbatch -p dyn --wrap="sleep 60"
    Submitted batch job 7
    [adarsh2511-controller ~]$ sinfo
    PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
    dyn          up   infinite      1 alloc# adarsh2511-dynamicnodeset-2
    dyn          up   infinite      1  idle~ adarsh2511-dynamicnodeset-3
    # ... waited until alloc# completed ...
    [adarsh2511-controller ~]$ cat /home/health_log.txt
    Healthcheck ran on adarsh2511-dynamicnodeset-2 at Fri Jul 10 04:09:57 UTC 2026
    # ... checked file contents after 5 mins, no other entries added as no new nodes were created ...
    ```
    
#### Backward Compatibility
Verified backwards compatibility by setting these flags in the blueprint for a Slurm 25.05 cluster. The version-checking logic for config generation ignored the unsupported settings and safely left the legacy defaults in place (`HealthCheckNodeState=ANY`).

## Usage
1. **Expedited Requeue:** When set to true in your blueprint's controller block, any job submitted with the `sbatch --requeue=expedite` flag will instantly be requeued with maximum priority if the compute node crashes or gets preempted.

```yaml
- id: slurm_controller
  settings:
    enable_expedited_requeue: true
```
2. **Health Check Start Only:** When set to true, any custom-defined HealthCheckProgram script will be strictly restricted to running exactly once during a node's boot-up sequence. It fully bypasses HealthCheckInterval, protecting running jobs from background script interference.

```yaml
- id: slurm_controller
  settings:
    enable_health_check_start_only: true
```


LAVEEN:
Hi @AdarshK15 , what happens in the below cases ? are those tested? If yes , can you share the findings?
1. If we send a shutdown signal forcefully to remove a node on which job was running (instead of deleting from UI) ? 
   Does it wait till completing the job or reschedule to new node?
2. What happens in case of dynamic nodeset?

AdarshK15:
Yes, the two cases mentioned are tested.
1. If we send a shutdown signal to a node running a job, Slurm will cancel the job instead of requeuing it. Slurm treats this as an intentional action because the user explicitly shut the node down. Expedited requeue only triggers on hard, unexpected node failures like spot preemptions or node crashes.
2. The same behaviour applies to static nodes and dynamic nodes, i have tested on both node types.

LAVEEN (approved):
LGTM

AdarshK15:
Test failures are due to insufficient capacity and known issues(vpc peering), these failures are not caused by code changes in this PR.

PR #5940, opened 2026-07-13
fix(healthcheck): correct logic bug in temperature check

Neelabh94:
This pull request addresses a logic bug in the GPU health check mechanism. 
The previous implementation incorrectly interpreted the temperature limit as a countdown value, leading to inaccurate reporting. The changes ensure that the system correctly compares the current temperature against the defined threshold and provides a more accurate warning message when the limit is exceeded.

### DIFF
--- a/community/cluster-health-check/cmd/dcgm-healthcheck/main.go
+++ b/community/cluster-health-check/cmd/dcgm-healthcheck/main.go
@@ -684,16 +684,19 @@ func checkTemperature(ctx context.Context) (string, error) {
 			tempStr := strings.TrimSpace(parts[0])
 			limitStr := strings.TrimSpace(parts[1])
 
-			countdown, err := strconv.Atoi(limitStr)
-			if err == nil {
+			temp, errTemp := strconv.Atoi(tempStr)
+			limit, errLimit := strconv.Atoi(limitStr)
+			if errTemp == nil && errLimit == nil {
+				countdown := limit - temp
 				if countdown <= 0 {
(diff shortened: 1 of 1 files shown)


community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -684,16 +684,19 @@ func checkTemperature(ctx context.Context) (string, error) {
 			tempStr := strings.TrimSpace(parts[0])
 			limitStr := strings.TrimSpace(parts[1])
 
-			countdown, err := strconv.Atoi(limitStr)
-			if err == nil {
+			temp, errTemp := strconv.Atoi(tempStr)
+			limit, errLimit := strconv.Atoi(limitStr)
+			if errTemp == nil && errLimit == nil {
+				countdown := limit - temp
```
bytetwin, line 690: nit: countdown is a timer variable and not suitable for temperature intent

bytetwin (approved):

leiyiz:
`temperature.gpu,temperature.gpu.tlimit` is a countdown to the temperature when the GPU starts to throttle:

[ref 1](https://docs.nvidia.com/deploy/nvidia-smi/index.html#:~:text=T.Limit%3A%20The%20T.Limit%20sensor%20measures%20the%20current%20margin%20in%20degree%20Celsius%20to%20the%20maximum%20operating%20temperature)
[discussion](https://forums.developer.nvidia.com/t/nvidia-smi-gpu-t-limit-gpu-shutdown-t-limit-temp/292006)

Neelabh94:
As highlighted in NVIDIA discussions, the `temperature.gpu.tlimit` is indeed a countdown to the temperature where GPUs will throttle, hence this change is not needed.

PR #5942, opened 2026-07-13
add nvlink checks to cluster health check daemon

leiyiz:
This PR introduces NVLink health monitoring specifically targeted at A4x and GB2xx node families (GB200/GB300).

**Key Changes:**
* Adds `isA4x` helper to detect nodes based on `gke-accelerator` (`gb200`, `gb300`) or `machine-family` (`a4x`) labels.
* Implements `checkNVLinkSmi` which monitors both NVLink bandwidth and error rates.
  * **Bandwidth Checks**: Parses output from `nvidia-smi nvlink -s` to ensure per-link bandwidth meets the expected `A4xExpectedNvlinkSpeed`.
  * **Error Rate Checks**: Parses output from `nvidia-smi nvlink -e` to flag hardware errors. It safely ignores benign `FEC Errors` while alerting on elevated `Effective BER` (>1e-9), `Symbol BER` (>1e-25), and other non-packet/byte error counters.

### DIFF
--- a/community/cluster-health-check/README.md
+++ b/community/cluster-health-check/README.md
@@ -116,7 +116,7 @@ bash build-and-push-cluster-health-check.sh \
     -v "0.0.3"
 ```
 
-## Active Tests and Logs
+## Tests and Logs
 
 The healthcheck daemon actively monitors the following items by running specific commands on the node:
 
@@ -126,4 +126,4 @@ The healthcheck daemon actively monitors the following items by running specific
 - **PCIe link health**: Monitors PCIe link width consistency via `nvidia-smi --query-gpu=pcie.link.width.current,pcie.link.width.max --format=csv,noheader,nounits`.
 - **InfiniBand links**: Verifies InfiniBand port state and link layer status via `ibstat`.
 - **GPU temperature**: Monitors GPU temperatures against predefined thermal limits via `nvidia-smi --query-gpu=temperature.gpu,temperature.gpu.tlimit --format=csv,noheader,nounits`.
-- **NVLink health**: Checks for fatal NVLink errors and alignment issues via `nvidia-smi nvlink -s` and `nvidia-smi nvlink -e` (A4x nodes only).
+- **NVLink health**: Checks for fatal NVLink errors and alignment issues via `nvidia-smi nvlink -s` and `nvidia-smi nvlink -e` (A4x family nodes only).
--- a/community/cluster-health-check/README.md
+++ b/community/cluster-health-check/README.md
@@ -115,3 +115,15 @@ bash build-and-push-cluster-health-check.sh \
     -l "us-east1" \
     -v "0.0.3"
 ```
+
+## Active Tests and Logs
+
+The healthcheck daemon actively monitors the following items by running specific commands on the node:
+
+- **dcgmi health**: Detects hardware issues via DCGM using `dcgmi health -c -j`.
+- **XID/SXID monitoring**: Scans kernel logs for NVIDIA XID and NVSwitch SXID errors via `dmesg --since <daemon-start-time>`.
+- **ECC errors**: Checks for uncorrectable ECC memory errors via `nvidia-smi --query-gpu=ecc.errors.corrected.volatile.total,ecc.errors.uncorrected.volatile.total --format=csv,noheader,nounits`.
+- **PCIe link health**: Monitors PCIe link width consistency via `nvidia-smi --query-gpu=pcie.link.width.current,pcie.link.width.max --format=csv,noheader,nounits`.
+- **InfiniBand links**: Verifies InfiniBand port state and link layer status via `ibstat`.
+- **GPU temperature**: Monitors GPU temperatures against predefined thermal limits via `nvidia-smi --query-gpu=temperature.gpu,temperature.gpu.tlimit --format=csv,noheader,nounits`.
+- **NVLink health**: Checks for fatal NVLink errors and alignment issues via `nvidia-smi nvlink -s` and `nvidia-smi nvlink -e` (A4x nodes only).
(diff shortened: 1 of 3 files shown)


leiyiz:
Thanks for the review! I have replaced the `strings.SplitSeq` calls with standard `strings.Split` to ensure compatibility with older Go toolchains. I have pushed a new commit with these fixes.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -741,3 +750,101 @@ func clearNodeHealth(ctx context.Context, clientset *kubernetes.Clientset, nodeN
...
+	if err != nil {
+		return SeverityWarning, fmt.Errorf("nvidia-smi nvlink -s failed: %v", err)
+	}
+
+	for _, line := range strings.Split(string(outS), "\n") {
+		matches := nvlinkStatusRegex.FindStringSubmatch(line)
+		if len(matches) > 1 {
+			if speed, err := strconv.ParseFloat(matches[1], 64); err == nil {
+				if speed < A4xExpectedNvlinkSpeed {
+					return SeverityWarning, fmt.Errorf("unexpected nvlink bandwidth: got %v GB/s, expected %v GB/s", speed, A4xExpectedNvlinkSpeed)
```
Jiaqicao257: Will it be helpful to log the impacted GPU / Link as well?

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -57,6 +57,8 @@ const (
 	SeverityFailure = "failure"
 	SeverityFatal   = "fatal"
 	SeverityInfo    = "info"
+
+	A4xExpectedNvlinkSpeed = 53.125
```
Jiaqicao257: Should we relax this a little? Does the same value apply for A4X and A4X MAX?

community/cluster-health-check/README.md
```diff
@@ -115,3 +115,15 @@ bash build-and-push-cluster-health-check.sh \
     -l "us-east1" \
     -v "0.0.3"
 ```
+
+## Active Tests and Logs
```
Jiaqicao257: should we call this passive tests instead?

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: 53.125 GB/s from my searches are the expected nvlink speed for A4x and A4x Max. in reality it should always be this number.

if we want to be 100% sure we can relax to 50GB/s to align with marketing claims of Nvidia. wdyt

community/cluster-health-check/README.md
leiyiz: Done.

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: Link number probably isn't helpful because we return at the first time we see unexpected bandwidth. perhaps all the links are unhealthy for this GPU.

included GPU in the error.

Jiaqicao257 (approved):

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -741,3 +752,124 @@ func clearNodeHealth(ctx context.Context, clientset *kubernetes.Clientset, nodeN
...
+}
+
+func checkNVLinkSmi(ctx context.Context) (string, error) {
+	if sev, err := checkNVLinkBandwidth(ctx); err != nil {
+		return sev, err
+	}
+	return checkNVLinkErrors(ctx)
+}
+
+var nvlinkStatusRegex = regexp.MustCompile(`^\s*Link\s+\d+:\s+([0-9.]+)\s+GB/s`)
```
parulbajaj01, line 773: should we update this regex to capture the link identifier (e.g., Link 0) and also match inactive link states:
`regexp.MustCompile(`^\s*(Link\s+\d+):\s+(Inactive|Disconnected|Down)`)`

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
```diff
@@ -741,3 +752,124 @@ func clearNodeHealth(ctx context.Context, clientset *kubernetes.Clientset, nodeN
...
+func isA4x(node *corev1.Node) bool {
+	if node == nil || node.Labels == nil {
+		return false
+	}
+	accel := strings.ToLower(node.Labels["cloud.google.com/gke-accelerator"])
+	machine := strings.ToLower(node.Labels["cloud.google.com/machine-family"])
+	return strings.Contains(accel, "gb300") || strings.Contains(accel, "gb200") || strings.Contains(machine, "a4x")
+}
+
+func checkNVLinkSmi(ctx context.Context) (string, error) {
```
parulbajaj01: should we avoid early returning if checkNVLinkBandwidth returns an error. If a node suffers from both degraded bandwidth and error counter spikes (CRC/BER errors), we should aggregate issues from both checks into a single report rather than masking checkNVLinkErrors.

parulbajaj01 (changes requested):

community/cluster-health-check/cmd/dcgm-healthcheck/main.go
leiyiz: it already is tolerant of trailing text and extra whitespace after  GB/s .

Because `nvlinkStatusRegex`  uses the anchor  `^`  but omits  end of line anchor `$` , the engine will match from the start of the line up to the string  `GB/s` , and it will ignore whatever comes after that.

leiyiz: I added a test case for this.

parulbajaj01 (dismissed):
LGTM

saara-tyagi27 (dismissed):

mufaqam-gcl (approved):

Neelabh94 (approved):

PR #5948, opened 2026-07-13
feat(job): Enable MTC for Pathways Workloads

Neelabh94:
This pull request introduces support for Multi-Tier Checkpointing (MTC) in Pathways workloads within the HPC Toolkit. The changes include new CLI flags for enabling MTC, updates to the GKE orchestrator to handle MTC-specific volume mounts and sidecar configurations, and refinements to the JobSet templates to improve reliability and integration with GKE infrastructure.


pkg/orchestrator/gke/templates/pathways_jobset.tmpl
```diff
@@ -258,6 +258,35 @@ spec:
             priorityClassName: {{.PriorityClassName}}
 {{- end }}
             terminationGracePeriodSeconds: {{.TerminationGracePeriodSeconds}}
+            {{- if .Pathways.ColocatedPythonSidecarImage}}
+            initContainers:
+            - name: colocated-python-sidecar
+              image: {{.Pathways.ColocatedPythonSidecarImage}}
+              imagePullPolicy: Always
+              restartPolicy: Always
```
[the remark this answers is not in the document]
Neelabh94: The current label is helpful for quick iterative development since it prevents the risk of cached images getting out of sync with the actual new image.

pkg/orchestrator/gke/templates/pathways_jobset.tmpl
```diff
@@ -258,9 +263,39 @@ spec:
...
+              ports:
+              - containerPort: 50051
+                protocol: TCP
+              resources: {}
+              env:
+              - name: TCMALLOC_RELEASE_RATE
+                value: "10"
+              - name: GRPC_SERVER_ADDRESS
+                value: '0.0.0.0:50051'
+              - name: CLOUD_PATHWAYS_SIDECAR_SHM_DIRECTORY
```
agrawalkhushi18, line 281: The environment variable CLOUD_PATHWAYS_SIDECAR_SHM_DIRECTORY is injected unconditionally into the colocated-python-sidecar container. Since the JAX worker only gets the corresponding flag when MTCEnabled is true, do you think we should probably make this env var conditional in the sidecar template as well to keep it clean for non-MTC sidecar use cases?

```
- name: GRPC_SERVER_ADDRESS
                value: '0.0.0.0:50051'
              {{- if .Pathways.MTCEnabled}}
              - name: CLOUD_PATHWAYS_SIDECAR_SHM_DIRECTORY
                value: /tmp/sidecar
              {{- end}}
              - name: PYTHONUNBUFFERED
```

Neelabh94: If you look at the pathways_utils repo, It is being added unconditionally there too, meaning that it is a standard requirement not restricted by MTC:
https://github.com/search?q=repo%3AAI-Hypercomputer%2Fpathways-utils%20CLOUD_PATHWAYS_SIDECAR_SHM_DIRECTORY&type=code

agrawalkhushi18: Got it. Thanks for clarifying that!

agrawalkhushi18 (approved):
LGTM

PR #5963, opened 2026-07-16
fix(job): improve slicing topology handling

Neelabh94:
This PR fixes slicing topology handling and caches detection.

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -1111,6 +1111,14 @@ func (g *GKEOrchestrator) validateRequestedTopology(requested string, topologies
 			return fmt.Errorf("failed to check topology containment: %w", err)
 		}
 		if fit {
+			if !g.hasSlicingTopologies() {
+				var valid []string
+				for t := range topologies {
+					valid = append(valid, t)
+				}
+				slices.Sort(valid)
+				return fmt.Errorf("requested topology %s fits inside discovered limits but no slicing labels found. It must match discovered limits exactly: %v", requested, valid)
+			}
 			logging.Info("Validated provided Topology: %s", requested)
 			return nil
 		}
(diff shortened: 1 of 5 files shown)


pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -1068,13 +1068,25 @@ func (g *GKEOrchestrator) validateRequestedTopology(requested string, topologies
...
 		if !contained {
-			var valid []string
-			for t := range topologies {
-				valid = append(valid, t)
-			}
 			return fmt.Errorf("requested topology %s is not valid for cluster. It must match or fit inside discovered limits: %v", requested, valid)
 		}
+
+		if !g.hasSlicingTopologies() {
+			if g.napEnabled {
```
shubpal07: Redundant `g.napEnabled` Check. `ValidateRequestedTopology` exits early at the top of the function if `g.napEnabled` is true.

Thus, this inner if g.napEnabled block is unreachable and redundant. We should clean it up and only return the error when !g.hasSlicingTopologies().

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -1068,13 +1068,25 @@ func (g *GKEOrchestrator) validateRequestedTopology(requested string, topologies
 				break
 			}
 		}
+
+		var valid []string
+		for t := range topologies {
+			valid = append(valid, t)
+		}
+
+		slices.Sort(valid)
```
shubpal07: perf nit:
We can only sort the list when a validation actually fails and we are about to return an error. This way we can keep the happy path fast by wrapping this logic in a helper function and calling sorting operation only inside the error blocks.
```go
if !fits {
    return fmt.Errorf("... limits: %v", g.getSortedTopologies(topologies))
}
```

shubpal07 (dismissed):
LGTM

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: Good point, the top level check was recently added via a different PR. Will rebase my changes on to upstream/develop and then remove the redundancy.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: the max topology count in general would be in single digits. wouldn't be a noticeable performance improvement.

shubpal07 (approved):

PR #5965, opened 2026-07-16
fix: GKE Pathways Kueue auto-configuration and simplify job orchestrator logic

agrawalkhushi18:
This PR resolves issues in the GKE job orchestrator CLI (gcluster) that led to misconfigured Kueue ResourceFlavors and ClusterQueues during job submission, causing workloads to remain stuck in Pending state with NoFit errors. It also refactors the code to satisfy linter limits and adds comprehensive unit tests.

## Problem
1.  **Kueue Template Discrepancies :** 
    *   The CPU head node pool was missing the `cloud.google.com/gke-nodepool` label, causing head workloads to fail scheduling.
    *   The TPU worker flavor name was hardcoded to `tpu-flavor` instead of matching `flavor-${accelerator_type}` expected by GCluster.
    *   TPU topology labels were not dynamically rendered in the static Kueue configuration templates, leading to scheduling mismatches.
2.  **GPU Topology Bug:** The Go orchestrator applied `cloud.google.com/gke-tpu-topology` to any pool with a placement policy and a topology specified, which incorrectly tagged GPU pools.

### 1. Kueue Configuration Template Updates
*   Updated `kueue-configuration-pathways.yaml.tftpl` and `kueue-configuration-dynamic-slicing-pathways.yaml.tftpl` to:
    *   Rename the TPU flavor to `flavor-${accelerator_type}` in order bring parity with compiler.
    *   Dynamically render `cloud.google.com/gke-tpu-topology` if `tpu_topology` is provided.
    *   Add `cloud.google.com/gke-nodepool: cpu-np` label to the CPU ResourceFlavor (`pathways-flavor`).
*   Updated `modules/management/kubectl-apply/main.tf` to define default `tpu_topology = ""` and wrap Kueue template variables with safety checks.
*   Updated the main blueprint `examples/gke-tpu-v5e/gke-tpu-v5e.yaml` to pass the `tpu_topology` variable and also removed custom kueue config path because it should default to use `kueue-configuration-pathways.yaml.tftpl` .

### 2. GCluster Go Orchestrator Refactoring
*   **Correct Head Node Pool Mapping:** Deferred the capacity calculation until after the head node pool is resolved, ensuring the CPU pool is correctly identified and mapped to the `pathways-flavor`.
*   **Automatic Node Pool Labeling:** Added logic to automatically attach the `cloud.google.com/gke-nodepool: <node-pool-name>` label to all non-accelerated node pools (like `cpu-np`), enabling correct scheduling of head/coordinator pods.
*   **TPU Topology Guard:** Restricted the `cloud.google.com/gke-tpu-topology` label assignment to node pools that actually host TPUs (`tpus > 0`), ensuring GPU pools with placement policies do not receive TPU-specific labels.

## Testing
*   **Unit Tests:** Added new unit tests `TestProcessNodePoolCapacity_FlavorsAndLabels` in `gke_job_orchestrator_test.go` and verified all GKE package tests pass (`go test -v ./pkg/orchestrator/gke/...`).

### Submission Checklist

### DIFF
--- a/examples/gke-tpu-v5e/gke-tpu-v5e.yaml
+++ b/examples/gke-tpu-v5e/gke-tpu-v5e.yaml
@@ -56,9 +56,6 @@ vars:
   # Enable Pathways for TPUs (provisions CPU node pool and configures Kueue quotas)
   enable_pathways_for_tpus: true
 
-  # Kueue configuration
-  kueue_configuration_path: $(ghpc_stage("./kueue-configuration.yaml.tftpl"))
-
   # The namespace to be configured for running workloads
   user_namespace: default
 
@@ -167,10 +164,10 @@ deployment_groups:
     settings:
       kueue:
         install: true
-        config_path: $(vars.kueue_configuration_path)
         config_template_vars:
           tpu_quota: $(vars.num_slices * gke-tpu-v5e-pool.node_count_static * gke-tpu-v5e-pool.tpu_chips_per_node)
           accelerator_type: $(vars.accelerator_type)
+          tpu_topology: $(vars.tpu_topology)
           namespace: $(vars.user_namespace)
       jobset:
         install: true
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -512,9 +512,6 @@ func (g *GKEOrchestrator) autoDetectCPUNodePool() string {
 }
 
 func (g *GKEOrchestrator) isSystemPool(np gkeJobNodePool) bool {
-	if np.Name == "system" {
-		return true
-	}
 	for _, taint := range np.Config.Taints {
 		if taint.Key == "components.gke.io/gke-managed-components" && taint.Value == "true" {
 			return true
@@ -658,8 +655,19 @@ func (g *GKEOrchestrator) resolveAccelerators(np gkeJobNodePool, cap MachineType
 		if err != nil {
 			return 0, 0, "flavor-default", nil, fmt.Errorf("in node pool %s: %w", np.Name, err)
 		}
+		for _, acc := range np.Config.Accelerators {
+			if g.acceleratorToMachineType == nil {
+				g.acceleratorToMachineType = make(map[string]string)
+			}
+			g.acceleratorToMachineType[strings.ToLower(acc.AcceleratorType)] = np.Config.MachineType
+		}
 	} else if len(cap.Accelerators) > 0 {
-		gpus, tpus, flavor, nodeLabels = g.processCapabilitiesAccelerators(np, cap, nodeCount)
+		gpus, tpus, flavor, nodeLabels = g.processCapabilitiesAccelerators(cap, nodeCount)
+		accType := cap.Accelerators[0].Type
+		if g.acceleratorToMachineType == nil {
+			g.acceleratorToMachineType = make(map[string]string)
+		}
+		g.acceleratorToMachineType[strings.ToLower(accType)] = np.Config.MachineType
 	} else {
 		return 0, 0, "flavor-default", make(map[string]string), nil
 	}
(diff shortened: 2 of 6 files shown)


pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -625,33 +631,70 @@ func (g *GKEOrchestrator) processNodePoolCapacity(np gkeJobNodePool, location st
...
-		}
-		g.acceleratorToMachineType[strings.ToLower(accType)] = np.Config.MachineType
+	if len(cap.Accelerators) > 0 {
+		gpus, tpus, flavor, nodeLabels = g.processCapabilitiesAccelerators(np, cap, nodeCount)
+		return gpus, tpus, flavor, nodeLabels, nil
 	}
 
-	return cpus, memMb, gpus, tpus, flavor, nodeLabels, sa, nil
+	return 0, 0, "flavor-default", nodeLabels, nil
+}
```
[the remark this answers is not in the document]
agrawalkhushi18: Addressed 1. and 2.

GKE API & GKE Terraform Provider: Both use tpu_topology (or tpuTopology in JSON) inside the placement policy block, even when GKE temporarily reuses this field to pass GPU topologies in some contexts. There is no accelerator_topology field in GKE's NodePool resource. 
GCE Resource Policy: GCE has a workload_policy which does use accelerator_topology, but this is a different API and resource.
Therefore, we should not rename it to AcceleratorTopology. Keeping it as TpuTopology is correct because it maintains exact alignment with the GKE API response we are unmarshaling and the Terraform GKE provider configuration.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -512,6 +512,9 @@ func (g *GKEOrchestrator) autoDetectCPUNodePool() string {
 }
 
 func (g *GKEOrchestrator) isSystemPool(np gkeJobNodePool) bool {
+	if np.Name == "system" {
```
Neelabh94: This might be a little too strict. If a user names their nodepool "system", this will cause an error. Relying on the standard taints and labels might be better here.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -625,33 +631,66 @@ func (g *GKEOrchestrator) processNodePoolCapacity(np gkeJobNodePool, location st
...
+		tpus = count * nodeCount
+		nodeLabels["cloud.google.com/gke-tpu-accelerator"] = accType
+	} else {
+		gpus = count * nodeCount
+		nodeLabels["cloud.google.com/gke-accelerator"] = accType
+	}
+	if g.acceleratorToMachineType == nil {
+		g.acceleratorToMachineType = make(map[string]string)
+	}
+	g.acceleratorToMachineType[strings.ToLower(accType)] = np.Config.MachineType
```
Neelabh94: nit: The whole block R689-R692 should be moved out of `processCapabilitiesAccelerators` and instead added to the `resolveAccelerators` (parent) function, since this does not align with the core functionality of this method.

examples/gke-tpu-v5e/gke-tpu-v5e.yaml
```diff
@@ -56,9 +56,6 @@ vars:
   # Enable Pathways for TPUs (provisions CPU node pool and configures Kueue quotas)
   enable_pathways_for_tpus: true
 
-  # Kueue configuration
```
Neelabh94, line 59: can we make these changes for other TPU blueprints as well (v6e and 7x) ? It is okay to do them as a separate PR too!

Neelabh94 (commented):
Nit comments. otherwise LGTM

examples/gke-tpu-v5e/gke-tpu-v5e.yaml
agrawalkhushi18: Yea, I will test the config once for v6e and 7x and raise a separate PR for that.

pkg/orchestrator/gke/gke_job_orchestrator.go
agrawalkhushi18: Removed the hardcoded name. Thanks!

pkg/orchestrator/gke/gke_job_orchestrator.go
agrawalkhushi18: Yes, I have moved out of processCapabilitiesAccelerators and instead added to the resolveAccelerators.

Neelabh94 (approved):
LGTM

PR #5968, opened 2026-07-17
feat(config): support ai_assisted telemetry marker in blueprint config

Grimreaper00025:
Add support for the `ai_assisted` boolean field at the root level of the blueprint YAML configuration.

This field serves as a telemetry/provenance marker to indicate that a blueprint was generated or modified by an AI agent. Internal AI agents will be programmed to automatically add this field, setting the value to `true`. 

Changes:

- Add `AiAssisted` field to the Blueprint struct in `pkg/config/config.go` with `yaml:"ai_assisted,omitempty"` tag.
- Add `AiAssisted` field to the rootPath struct in `pkg/config/path.go` with `path:"ai_assisted"` tag to support path-based configuration resolution.
- Add unit tests in `pkg/config/config_test.go` (TestParseBlueprint_AiAssisted) to verify parsing of the `ai_assisted` field (covering true, false, and missing/default cases).
- Update unit tests in `pkg/config/path_test.go` (TestPath) to verify path resolution for the `ai_assisted` field.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -295,6 +295,7 @@ func (m Module) InfoOrDie() modulereader.ModuleInfo {
 type Blueprint struct {
 	BlueprintName            string      `yaml:"blueprint_name"`
 	GhpcVersion              string      `yaml:"ghpc_version,omitempty"`
+	AiAssisted               bool        `yaml:"ai_assisted,omitempty"`
 	Validators               []Validator `yaml:"validators,omitempty"`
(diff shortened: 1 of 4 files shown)


kadupoornima (commented):
Pls check that it doesn't throw an error if the field is not set. By default, the value should be considered "false".

shubpal07 (approved):

kadupoornima (approved):

PR #5997, opened 2026-07-22
feat(job): Enable MTC at job submission stage

Neelabh94:
This PR introduces support for Multi-Tier Checkpointing (MTC) at the job submission level and makes MTC flags globally available for all job types, rather than being restricted to Pathways workloads.

### Key Changes
* Global MTC Flags: Promoted --gke-mtc-enabled and --gle-mtc-ramdisk-dir to global job submission flags in submit.go, making MTC available for standard workloads.
* Cluster Addon Validation: Added a pre-submission check in gke_job_orchestrator.go that queries the target GKE cluster to ensure the HighScaleCheckpointing add-on is enabled. Job submission will safely abort with a helpful error message if the user requests MTC on an unsupported cluster.
* Dynamic Volume Injection: Centralized MTC CSI volume and mount configuration into storage.go. This removes the need for hardcoded MTC volume blocks in the raw Go templates.
* Template Cleanup: Removed redundant MTC csi volume and mount definitions from `pathways_jobset.tmpl` to prevent duplicate volume mounting.

### Testing
* Passed all unit tests.
* Successfully verified MTC job submission (`gcluster job submit`) and manifest generation on a live GKE cluster with the `HighScaleCheckpointing` addon enabled.


pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -369,8 +390,8 @@ func (g *GKEOrchestrator) GeneratePathwaysManifest(job orchestrator.JobDefinitio
 		// WorkerImage defaults to ServerImage if not explicitly set
 		job.Pathways.WorkerImage = job.Pathways.ServerImage
 	}
-	if job.Pathways.MTCEnabled && job.Pathways.RamdiskDirectory == "" {
-		job.Pathways.RamdiskDirectory = "/tmp/mtc_checkpoints"
+	if job.GKEMTCEnabled && job.GKEMTCRamdiskDirectory == "" {
+		job.GKEMTCRamdiskDirectory = "/tmp/mtc_checkpoints"
```
SwarnaBharathiMantena: This default of `"/tmp/mtc_checkpoints"` is being set in the `pkg/orchestrator/gke/storage.go` file as well. In order to prevent future inconsistency, can we assign a default at a single location?

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -133,6 +135,25 @@ func (g *GKEOrchestrator) SubmitJob(job orchestrator.JobDefinition) error {
 	return nil
 }
 
+func (g *GKEOrchestrator) checkMTCAddonEnabled(projectID, location, clusterName string) error {
+	ctx := context.Background()
+	svc, err := container.NewService(ctx)
```
SwarnaBharathiMantena: This container call requires an additional permissions `container.clusters.get`. Can we avoid requiring Google Cloud IAM permissions (container.clusters.get) by validating the presence of Multi-Tier Checkpointing (MTC) entirely through the Kubernetes API?

We can check for the Kubernetes resources that the MTC addon installs, rather than querying the Google Cloud API for the cluster's configuration.

My intension behind this is to avoid requiring addition permissions. In case there is already a requirement for the `container.clusters.get` permission elsewhere, please ignore this.

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -133,6 +135,25 @@ func (g *GKEOrchestrator) SubmitJob(job orchestrator.JobDefinition) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create cluster manager client: %v", err)
+	}
+
+	name := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", projectID, location, clusterName)
+	resp, err := svc.Projects.Locations.Clusters.Get(name).Context(ctx).Do()
+	if err != nil {
+		return fmt.Errorf("failed to get cluster details for MTC validation: %v", err)
+	}
+
```
SwarnaBharathiMantena: One more potential validation to add: GKE MTC requires nodes with Local SSDs. If a user submits an MTC job on a standard machine type without SSDs, the pods will likely fail to schedule. Should we add a quick check against the --compute-type flag to warn the user early before we generate and apply the manifest?

I think this check can be here and also at the cluster creation checks to let a user know of supported machines.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: Thanks for pointing this out, have refactored the code since this is actually redundant.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: Removed this entirely, since we are fetching all cluster details at the initialize() step. Have removed this function entirely.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: How would you like to handle the local SSD check to validate MTC compatibility?"

We have a few options:

1. **Hardcode machine types:** Write a function that checks for strings like a3- or ct5p-hightpu which we know come with local SSDs by default.
2. **Just trust GKE/Kubernetes:** Skip the validation at the gcluster level entirely. If MTC is requested on a non-SSD machine, Kubernetes simply won't be able to schedule the pod or the MTC CSI driver will fail to mount.
3. **Parse more metadata:** Map EphemeralStorageLocalSsdConfig.LocalSsdCount in types.go as well and check if LocalSsdCount > 0 (though this might still fail for some accelerator types with built-in NVMe that don't technically report a LocalSsdCount).

I prefer option 2 since for most cases checking mtc enabled should be good enough validation. Wdyt?

SwarnaBharathiMantena: Sure, we can go ahead with 2 at this time. If we observe issues in future, we can reconsider.

SwarnaBharathiMantena (approved):
LGTM

PR #6005, opened 2026-07-23
feat(job): Enhance mount flag to pass additional mount options for gcsfuse mounts

Neelabh94:
## Description
This PR adds support for passing custom mount options when submitting a job, bringing `gcluster` into closer parity with `xpk` behavior.

**Key Changes:**
- **Added `options` key to the `mount` flag:** Users can now pass custom storage options (e.g., `logging:severity:info,enable-atomic-rename-object:true`) to `gcluster job submit`.
- **Added Validation:** Included a run validation step that ensures the `options` key is currently only utilized for GCS fuse volumes (i.e. it enforces that all supplied `--mount` values start with `gs://` when options are present).
- **Updated Documentation:** Added the new flag and usage examples to the `docs/gcluster_job_guide.md`.


cmd/job/submit.go
```diff
@@ -378,6 +386,20 @@ func validateGKENAPFlags() error {
 	return nil
 }
 
+func validateMountOptions() error {
+	if mountOptions != "" {
+		if len(volumeStr) == 0 {
+			return fmt.Errorf("--mount-options requires at least one GCS mount")
+		}
+		for _, m := range volumeStr {
+			if !strings.HasPrefix(m, "gs://") {
```
agrawalkhushi18: If a user runs a job that mounts a GCS bucket alongside a local disk or NFS drive, the command will be completely rejected. In storage.go, the backend handles mixed workloads perfectly by explicitly checking if `volType == "gcsfuse"`.  Should we update the validation to just confirm that at least one GCS bucket is mounted when --mount-options is used, rather than an aggressive front-end validation?

example snippet:
```
func validateMountOptions() error {
	if mountOptions != "" {
		hasGCSMount := false
		for _, m := range volumeStr {
			if strings.HasPrefix(m, "gs://") {
				hasGCSMount = true
				break
			}
		}
		if !hasGCSMount {
			return fmt.Errorf("--mount-options was provided, but no GCS fuse volumes (gs://...) were mounted")
		}
	}
	return nil
}
```

Neelabh94: Nice catch, fixed. PTAL

agrawalkhushi18 (dismissed):
LGTM

agrawalkhushi18 (approved):
LGTM

PR #6009, opened 2026-07-24
fix(slurm-controller): add HA single-active guard and defer scontrol …

andybubu:
### Summary
Fixes controller race conditions in Slurm High Availability (HA) setups and eliminates `slurm_load_ctl_conf error: Socket timed out` during reconfiguration. 

This PR introduces an **HA Single-Active Guard** in `slurmsync.py`, defers `scontrol reconfigure` calls to a single pass, and polls daemon readiness (`scontrol ping`) with exponential backoff.

---

#### 1. Dual-Active Execution in Slurm HA Setups
In dual-controller setups (`slurmctld-0` Primary and `slurmctld-1` Backup):
- `slurmsync.py` ran concurrently on **both** controllers every minute.
- Both controllers issued duplicate GCP Compute API calls (`instances.start()`, `delete_instances()`, `repair.call_rr_api()`), leading to API rate-limit errors (`429 Quota Exceeded`), race conditions in VM lifecycle management, and corrupted repair tracking.

#### 2. Double Restarts & Socket Timeouts
- When both cloud configuration (`cloud.conf`) and placement topology (`topology.conf`) changed in a single sync pass, `slurmctld` was restarted twice sequentially, tearing down mid-registration compute nodes.
- Executing `scontrol reconfigure` immediately after `systemctl restart` caused `Socket timed out on send/recv operation` on large clusters while `slurmctld` initialized state.

---

### Architectural Changes Introduced

1. **Active Controller Detection (`util.is_active_controller`)**:
   - Non-HA clusters: always returns `True` for the single controller.
   - HA clusters: queries `scontrol ping` to determine if the local instance is Primary Active (`-0`) or Backup Takeover (`-1`). Returns `False` on Standby.
2. **Single-Active Guard (`slurmsync.py`)**:
   - Both controllers synchronize configuration files (`cloud.conf` / `topology.conf`).
   - Only the **Active controller** executes GCP instance management, node repairs, placement group sync, and `scontrol reconfigure`. Standby controllers log standby status and exit cleanly.
3. **Single Deferred Reconfigure & Polled Readiness**:
   - Coalesces config and topology changes into a single `util.scontrol_reconfigure()` call at the end of `slurmsync.py`.
   - `wait_slurmctld_up()` polls `scontrol ping` with exponential backoff before executing `scontrol reconfigure`.

---

### Modified & Added Files

- **`community/.../scripts/util.py`**: Added `is_active_controller()` and `wait_slurmctld_up()`; updated `scontrol_reconfigure()`.
- **`community/.../scripts/slurmsync.py`**: Added `is_active_controller` guard inside `main()`; deferred `scontrol_reconfigure()` execution.
- **`community/.../scripts/setup.py`**: Added `check_slurmdbd_ready` for slurmdbd readiness check.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -2381,12 +2381,74 @@ def update_config(cfg: NSDict) -> None:
...
+            or "primary controller(up)" in line
+            or ("slurmctld(primary) at" in line and "is up" in line)
+            for line in output.splitlines()
+        )
+        backup_up = any(
+            "backup controller: up" in line
+            or "backup controller(up)" in line
+            or ("slurmctld(backup) at" in line and "is up" in line)
+            for line in output.splitlines()
+        )
```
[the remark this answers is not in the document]
andybubu: This not true standard out of SLURM scontrol ping uses: 
```
Slurmctld at <hostname> is UP
Slurmctld(primary) at <hostname> is UP
Slurmctld(backup) at <hostname> is UP
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -2381,12 +2381,81 @@ def update_config(cfg: NSDict) -> None:
     global _lkp
     _lkp = Lookup(cfg)
 
+def _is_target_controller_up(output: str, target_role: str) -> bool:
```
arpit974, line 2419: In single-controller (non-HA) deployments, `scontrol ping` outputs:

`Slurmctld at <hostname> is UP`

Notice there is no (primary) tag when a backup controller is not configured).

Because target_role is set to "primary", target_role in line_lower evaluates to False on non-HA clusters. This causes wait_slurmctld_up() to loop for 60 seconds and raise 
`TimeoutError: slurmctld (primary) is not fully up after 60 seconds`
on all single-controller clusters during reconfigure.

suggestion:
Pass is_ha = bool(backup_name) into _is_target_controller_up() so that on single-controller clusters, any is UP line is recognized as ready:

```
def _is_target_controller_up(output: str, target_role: str, is_ha: bool = True) -> bool:
    for line in output.splitlines():
        line_lower = line.lower()
        is_up = ("is up" in line_lower or ": up" in line_lower or "(up)" in line_lower) and "down" not in line_lower
        if not is_ha and is_up:
            return True
        if target_role in line_lower and is_up:
            return True
    return False

```

Could we also add unit tests in test_util.py covering:

_is_target_controller_up (single controller format, HA primary, HA backup).
is_active_controller (Primary active, Backup takeover, Backup standby, Non-HA).
check_slurmdbd_ready.

andybubu: @arpit974  `scontrol ping` format is consistent either  in single or HA controllers setup, see image attached.

 Hence the  `_is_target_controller_up()`  and `is_active_controller()` guard and defer controls works as expected.

See attached test verification carried to show `scontrol reconfigure` works as expected in single controller deployments as well.

scontrol ping on single controller setup:
image

scontrol reconfigure issued successfully by slurmsync after scheduler parameters: default_queue_depth was updated from 100 to 150
image

arpit974: Thanks for providing the test logs and screenshots! That confirms that `scontrol ping` consistently reports `(primary)` even on single-controller setups with `SlurmctldHost` configured. 
The fix looks good and addresses the HA race conditions and reconfigure timeouts cleanly. LGTM!

arpit974:
Tests caught a regression, please check.

arpit974:
I investigated the `slurm-gke` daily test failure (`srun: error: Unable to allocate resources: Requested topology configuration is not available`), and identified the root cause in `slurmsync.py`.

### Cause of the failure
1. At the beginning of `main()`, `lkp = lookup()` is captured with the initial config.
2. When new partition/nodeset configs are synced (like `gkenodeset`), `reconfigure_slurm()` calls `util.update_config(cfg_new)`, which updates the global lookup in `util.py`.
3. However, `main()`'s local variable `lkp` is not refreshed.
4. Because `reconfigure_slurm()` now defers the `slurmctld` restart:
   - `update_topology(lkp)` is called *before* the restart, using the **stale `lkp`**.
   - `_generate_topology(lkp)` generates a topology missing `gkenodeset` and overwrites `cloud_topology.yaml` on disk.
   - Then, the single deferred reconfigure restarts `slurmctld`, which loads `slurm.conf` (has GKE nodes) + `cloud_topology.yaml` (missing GKE nodes).
   - `slurmctld` logs:
     `warning: switches lack access to 63996 nodes: slurm929-gkenodeset-[0-1]`
     `warning: TOPOLOGY: no switch can reach all nodes through its descendants.`
   - Any multi-node job on the GKE partition subsequently fails with `Requested topology configuration is not available`.

### Suggested Fix
Refresh `lkp = lookup()` when `config_changed` is true so `update_topology(lkp)` and all downstream functions use the updated lookup:

```python
    config_changed = False
    if lookup().is_controller or lookup().instance_role_safe == "compute" or lookup().is_login_node:
        try:
            config_changed = bool(reconfigure_slurm())
            if config_changed:
                lkp = lookup()
        except Exception:
            log.exception("failed to reconfigure Slurm")

**Comment by sriram-quad** [comment]
Please NOTE : We recently upgraded our repository's Go version to 1.26 (https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/6173). We strongly recommend rebasing your open branches onto the latest develop branch to avoid or resolve any PR test failures.

Thank you!

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py (hunk 2 above)
Comment (arpit974) [inline community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py:2419]:
```markdown
Could we also add unit test assertions in `community/.../scripts/tests/test_util.py` covering the new helper functions:
- `_is_target_controller_up()`: single controller ping format (`Slurmctld (primary)...`), HA primary, HA backup.
- `is_active_controller()`: Primary active, Backup takeover, Backup standby, Non-HA.
- `check_slurmdbd_ready()`: fast success on ready, retry/timeout on failure.

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py [hunk 3]
```diff
@@ -2381,12 +2381,81 @@ def update_config(cfg: NSDict) -> None:
...
+
+
+def wait_slurmctld_up(lkp: Lookup, timeout: float = 60) -> None:
+    """Wait for local slurmctld daemon to respond to scontrol ping and report UP status.
+    This ensures scontrol reconfigure does not fail due to slurmctld not being ready
+    after a service restart.
+    """
+    log.info("Waiting for slurmctld to be fully up...")
+    hostname = socket.gethostname().split(".")[0]
+    backup_name = lkp.cfg.get("slurm_backup_controller_name")
```
arpit974, line 2437: In is_active_controller(), there is an early return (if not backup_name: return True), but here in wait_slurmctld_up(), is_backup evaluates hostname.endswith("-1") even in non-HA setups. If a non-HA single-controller VM hostname happens to end with -1 (e.g. cluster-1 or slurm-1), is_backup evaluates to True, causing wait_slurmctld_up() to poll for "backup" on a single controller and fail with a 60-second TimeoutError.

Suggested fix: Align with is_active_controller by checking bool(backup_name):

```python
backup_name = lkp.cfg.get("slurm_backup_controller_name")
is_backup = bool(backup_name) and (hostname == backup_name or hostname.endswith("-1"))
target_role = "backup" if is_backup else "primary"

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py [hunk 4]
```diff
@@ -662,8 +674,18 @@ def main():
     lkp = lookup()
     if util.should_mount_slurm_bucket() and not lkp.is_controller:
         return
+
+    if lkp.is_controller and not util.is_active_controller(lkp):
```
arpit974: Currently, the standby guard is placed before reconfigure_slurm(). When cluster configuration is updated in GCS:

The active primary controller updates its local /etc/slurm/ files and issues scontrol reconfigure.
In Slurm HA, scontrol reconfigure signals the backup slurmctld daemon to reload its configuration from disk.
However, because the standby node exited before reconfigure_slurm(), its local /etc/slurm/ files on disk remain stale.
Since reconfigure_slurm() now only regenerates configuration files on disk (the service restart and scontrol reconfigure have been deferred to the end of main()), moving the standby guard immediately after reconfigure_slurm() ensures the standby controller's /etc/slurm/ files are always up-to-date with GCS without performing any active GCP instance management or reconfigures:

```python
    config_changed = False
    try:
        config_changed = reconfigure_slurm()
        if config_changed:
            lkp = lookup()
    except Exception:
        log.exception("failed to reconfigure slurm")
    if lkp.is_controller and not util.is_active_controller(lkp):
        log.info("Local controller is in STANDBY mode. Skipping slurmsync pass.")
        return

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py [hunk 5]
```diff
@@ -2381,12 +2381,81 @@ def update_config(cfg: NSDict) -> None:
     global _lkp
     _lkp = Lookup(cfg)
 
+def _is_target_controller_up(output: str, target_role: str) -> bool:
+    """Check if a specific controller role ('primary' or 'backup') is UP in scontrol ping output."""
+    for line in output.splitlines():
+        line_lower = line.lower()
+        if target_role in line_lower:
```
arpit974: `if target_role in line_lower:` is an unanchored substring search across the entire line. If a cluster name or controller hostname contains `"primary"` or `"backup"` (e.g., `primary-cluster-controller-1` or `cluster-primary-0`), the role will match the hostname on every line:
For example, on `primary-cluster`, the backup ping line:
`Slurmctld(backup) at primary-cluster-controller-1 is UP`
contains `"primary"` (from the hostname), `"is up"`, and no `"down"`. `_is_target_controller_up(output, "primary")` therefore evaluates to `True` even if the primary controller is actually DOWN. This would prevent the backup controller from taking over during a real failover.
We should match the role specifically inside the delimiter `({target_role}` or before `controller:`:
```suggestion
        if f"({target_role}" in line_lower or f"{target_role} controller" in line_lower:

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py [hunk 6]
```diff
@@ -459,6 +459,63 @@ def self_report_controller_address(lkp: util.Lookup) -> None:
...
+            log.info("Timed out waiting for sackd to start; waiting for controller slurmctld to become ready...")
+        except Exception as e:
+            log.debug(f"sackd check error: {e}")
+
+        log.info(f"Waiting for sackd to be ready... (try {try_cnt}/{maxtries}, retrying in {int(wait)}s)")
+        time.sleep(wait)
+        wait = min(wait * 1.5, interval)
+
+    log.error(f"sackd is still not ready after {maxtries} tries. Expecting issues!")
+    run("systemctl status sackd", timeout=30)
```
arpit974: `util.run()` runs with `check=True` by default. When `sackd` fails to come up after `maxtries`, `systemctl status sackd` exits with return code 3 (inactive/failed). 
Because `check=True`, `run()` raises `subprocess.CalledProcessError`, crashing `setup_login()` completely and regressing login node provisioning. Diagnostic status checks should pass `check=False`:
```suggestion
    run("systemctl status sackd", timeout=30, check=False)

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py [hunk 7]
```diff
@@ -2381,12 +2381,81 @@ def update_config(cfg: NSDict) -> None:
...
+    return False
+
+
+def wait_slurmctld_up(lkp: Lookup, timeout: float = 60) -> None:
+    """Wait for local slurmctld daemon to respond to scontrol ping and report UP status.
+    This ensures scontrol reconfigure does not fail due to slurmctld not being ready
+    after a service restart.
+    """
+    log.info("Waiting for slurmctld to be fully up...")
+    hostname = socket.gethostname().split(".")[0]
```
arpit974, line 2436: `hostname = socket.gethostname().split(".")[0]` strips the domain, but `backup_name = lkp.cfg.get("slurm_backup_controller_name")` is not stripped. 
If `slurm_backup_controller_name` is passed as an FQDN (e.g. `cluster-controller-1.c.project.internal`) or custom hostname not ending in `-1`, `hostname == backup_name` evaluates to `False`, and the backup controller will misidentify itself as primary (triggering dual-active execution).
Both sides should normalize with `.split(".")[0]`:
```suggestion
    backup_short = backup_name.split(".")[0] if backup_name else ""
    is_backup_instance = (backup_short and hostname == backup_short) or hostname.endswith("-1")

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py [hunk 8]
```diff
@@ -2381,12 +2381,81 @@ def update_config(cfg: NSDict) -> None:
...
+    backup_name = lkp.cfg.get("slurm_backup_controller_name")
+    is_backup = bool(backup_name) and (hostname == backup_name or hostname.endswith("-1"))
+    target_role = "backup" if is_backup else "primary"
+
+    for wait in backoff_delay(0.5, timeout=timeout):
+        try:
+            res = run(f"{lkp.scontrol} ping", check=False, timeout=5)
+            if res.returncode == 0:
```
arpit974: `if res.returncode == 0:`.
In HA failover scenarios where the primary controller is dead or unreachable, `scontrol ping` can return a non-zero exit code while still reporting the local backup controller as `UP`. Because of `if res.returncode == 0:`, `wait_slurmctld_up()` ignores the `UP` status and loops until it hits a 60-second `TimeoutError`.
Notice that in `is_active_controller()`, `output = (res.stdout + res.stderr).lower()` is parsed directly without checking `returncode == 0`. We should align `wait_slurmctld_up` to parse output directly:
```suggestion
        try:
            res = run(f"{lkp.scontrol} ping", check=False, timeout=5)
            output = (res.stdout + res.stderr).lower()
            if _is_target_controller_up(output, target_role):
                log.info(f"slurmctld ({target_role}) is fully up.")
                return
        except Exception as e:

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py [hunk 9]
```diff
@@ -768,3 +769,105 @@ def test_get_reservation_details_error(mocker):
         )
     
     lkp._get_reservation.assert_called_once_with("my-project", "us-central1-a", "my-reservation")
+
```
arpit974, line 772: Could we add test assertions covering:
1. Role matching when the hostname contains the role name (e.g., verifying `_is_target_controller_up("Slurmctld(backup) at primary-cluster-1 is UP", "primary")` is `False`).
2. Unit tests for `wait_slurmctld_up()` asserting that it retries on exit code 1 and raises `TimeoutError` on timeout:
```python
def test_is_target_controller_up_hostname_containing_role():
    line = "Slurmctld(backup) at primary-cluster-controller-1 is UP"
    assert util._is_target_controller_up(line, "primary") is False
    assert util._is_target_controller_up(line, "backup") is True
def test_wait_slurmctld_up_timeout(mocker):
    mocker.patch("socket.gethostname", return_value="slurmctld-0")
    mocker.patch("util.run", return_value=Mock(returncode=1, stdout="", stderr="DOWN"))
    mocker.patch("util.sleep")
    lkp = util.Lookup(TstCfg())
    with pytest.raises(TimeoutError, match=r"slurmctld \(primary\) is not fully up"):
        util.wait_slurmctld_up(lkp, timeout=2)

**Review by arpit974 — DISMISSED** [review]

**Review by arpit974 — DISMISSED** [review]

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py [hunk 10]
```diff
@@ -662,10 +674,19 @@ def main():
...
+    config_changed = False
     try:
-        reconfigure_slurm()
+        config_changed = reconfigure_slurm()
+        if config_changed:
+            lkp = lookup()
     except Exception:
         log.exception("failed to reconfigure slurm")
+    if lkp.is_controller and not util.is_active_controller(lkp):
+        log.info("Local controller is in STANDBY mode. Skipping slurmsync pass.")
```
Neelabh94, line 688: If the standby controller returns here before `update_topology()`, its local `cloud_topology.yaml` is not updated when partition or nodeset definitions change. When the active controller subsequently issues `scontrol reconfigure`, Slurm signals the backup daemon to reload its configuration from disk, causing it to load a stale topology file and log missing switch/node warnings.

Since `update_topology()` now only writes the local file and does not restart daemons or call cloud APIs, moving `update_topology(lkp)` above this standby check will ensure both controllers keep their topology files synchronized on disk:

```python
    topology_changed = False
    topology_summary = None
    if lkp.is_controller:
        try:
            topology_changed, topology_summary = update_topology(lkp)
        except Exception:
            log.exception("failed to update topology")

    if lkp.is_controller and not util.is_active_controller(lkp):
        log.info("Local controller is in STANDBY mode. Skipping slurmsync pass.")
        if topology_changed and topology_summary is not None:
            topology_summary.dump(lkp)
        return
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -451,32 +451,40 @@ def sync_instances():
...
-        log.info("Restarting slurmctld to make changes take effect.")
+        # Defer slurmctld restart and scontrol reconfigure:
+        # Running them now would cause slurmctld to temporarily load new node defs with old topology.
+        # This causes slurmctld to hang/timeout when update_topology tries to reconfigure it subsequently.
+        log.info("Config files regenerated. Deferring scontrol reconfigure until after topology update.")
         try:
-            # TODO: consider removing "restart" since "reconfigure" should restart slurmctld as well
-            run("sudo systemctl restart slurmctld.service", check=False)
-            util.scontrol_reconfigure(lookup())
+            util.run(f"wall '{update_msg}'", timeout=30)
```
Neelabh94, line 483: Because service restart and `scontrol reconfigure` are now deferred to `main()`, broadcasting the wall message here notifies users before `slurmctld` actually reloads the configuration.

We can move the broadcast to `main()` right after `scontrol_reconfigure()` succeeds, and use `check=False` instead of the `try/except` block:
```python
util.run(f"wall '{update_msg}'", timeout=30, check=False)
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -462,6 +462,63 @@ def self_report_controller_address(lkp: util.Lookup) -> None:
...
+
+    log.error(f"Slurmdbd is still not ready after {maxtries} tries. Expecting issues!")
+    raise RuntimeError("Slurmdbd is not ready")
+
+def check_sackd_ready(maxtries: int = 40, interval: float = 15.0) -> None:
+    """Poll sackd until it successfully connects to slurmctld and becomes active."""
+    run("systemctl enable sackd", timeout=30)
+    wait = 10.0
+    for try_cnt in range(1, maxtries + 1):
+        try:
```
Neelabh94, line 490: For consistency with the rest of `setup.py` (such as line 519), we can use `run("systemctl restart sackd", timeout=60, check=False)` instead of calling `subprocess.run` directly.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
```diff
@@ -839,3 +840,145 @@ def test_compute_service_custom_universe_domain(mocker):
...
+        "util.run",
+        return_value=Mock(
+            returncode=0,
+            stdout="Slurmctld(primary) at primary-cluster-controller-0 is DOWN\nSlurmctld(backup) at primary-cluster-controller-1 is UP",
+            stderr="",
+        ),
+    )
+    assert util.is_active_controller(lkp_primary_cluster) is True
+
+    # Backup standby: Backup node (-1), primary UP
```
Neelabh94, line 941: In `test_is_active_controller`, the backup standby test case relies on the `socket.gethostname` patch from the preceding test case. Re-mocking or parameterizing `socket.gethostname` for each scenario ensures each test case is isolated.

Could we also add a unit test for `check_sackd_ready()`?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
Neelabh94: @andybubu I believe we should add topology_Summary.dump(lkp) here as well(as suggested in the previous comment)

IIUC `topology_summary.dump(lkp)` writes `/etc/slurm/cloud_topology.summary.json` on disk. Because the standby controller returns early at line 712, `topology_summary.dump(lkp)` is never executed on the standby controller. 

In the current implementation, on the very next 1-minute cron pass on the standby controller: 
`update_topology(lkp)` loads `prev_summary = TopologySummary.load(lkp)`. Because `cloud_topology.summary.json` was never dumped, prev_summary is stale! Therefore, `requires_reconfigure(prev_summary)` returns `True` every single minute on the standby controller, needlessly regenerating the topology data on disk every 60 seconds indefinitely.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -462,6 +462,51 @@ def self_report_controller_address(lkp: util.Lookup) -> None:
     with blob.open('w') as f:
         f.write(yaml.dump(data))
 
+def check_slurmdbd_ready(maxtries: int = 30, interval: float = 15.0) -> None:
```
Neelabh94: nit: since interval here is used as the maximum backoff cap, we can maybe rename it to max_interval for clarity. Renaming `interval` to `max_interval` would clarify that it represents the ceiling for the exponential backoff rather than a fixed sleep duration.

Neelabh94 (changes requested):
Thanks for addressing the comments @andybubu . I have added two follow-up suggestions, PTAL.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -2488,12 +2488,83 @@ def update_config(cfg: NSDict) -> None:
     global _lkp
     _lkp = Lookup(cfg)
 
+def _is_target_controller_up(output: str, target_role: str) -> bool:
+    """Check if a specific controller role ('primary' or 'backup') is UP in scontrol ping output."""
+    role = target_role.lower()
+    for line in output.splitlines():
+        line_lower = line.lower()
+        if f"({role}" in line_lower or f"{role} controller" in line_lower:
+            if ("is up" in line_lower or ": up" in line_lower or "(up)" in line_lower) and "down" not in line_lower:
```
Neelabh94, line 2497: `"down" not in line_lower` scans the full `scontrol ping` line, including the hostname (`Slurmctld(primary) at <hostname> is UP`). If a cluster or controller hostname contains `"down"` (e.g., `downscale-controller-0`), `_is_target_controller_up` returns `False`, causing both controllers to enter `STANDBY` mode and `wait_slurmctld_up()` to time out.

Check for explicit down status tokens instead:
```python
        if f"({role}" in line_lower or f"{role} controller" in line_lower:
            is_up = any(s in line_lower for s in ("is up", ": up", "(up)"))
            is_down = any(s in line_lower for s in ("is down", ": down", "(down)"))
            if is_up and not is_down:
                return True
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -2488,12 +2488,83 @@ def update_config(cfg: NSDict) -> None:
...
+
+        primary_up = _is_target_controller_up(output, "primary")
+        backup_up = _is_target_controller_up(output, "backup")
+
+        if not is_backup_instance:
+            return primary_up
+        else:
+            return (not primary_up) and backup_up
+    except Exception as e:
+        log.warning(f"Failed to query scontrol ping for HA active check: {e}")
```
Neelabh94, line 2564: When the primary controller GCE VM is stopped or preempted (TCP SYN blackhole rather than `Connection refused`), `scontrol ping` blocks on `SlurmctldHost[0]` up to `MessageTimeout` (`60s` in `slurm.conf.tpl`). With `timeout=5`, `run()` raises `TimeoutExpired` and this `except` block returns `not is_backup_instance` (`False` on the backup controller), leaving the backup controller stuck in `STANDBY` mode when the primary VM dies.

Consider falling back to a quick socket check on the backup controller:
```python
    except Exception as e:
        log.warning(f"Failed to query scontrol ping for HA active check: {e}")
        if is_backup_instance:
            port = int(lkp.control_host_port or 6817)
            primary_up = _is_host_port_open(lkp.control_host, port, timeout=2.0)
            backup_up = _is_host_port_open(hostname, port, timeout=2.0)
            return (not primary_up) and backup_up
        return True
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -710,10 +737,23 @@ def main():
...
+        # Placed here so it runs even if topology generation failed above.
+        if config_changed or topology_changed:
+            reasons = []
+            if config_changed:
+                reasons.append("config updated")
+            if topology_changed:
+                reasons.append("topology changed")
+            try:
+                log.info(f"Reconfiguring Slurm ({', '.join(reasons)}).")
+                util.scontrol_reconfigure(lkp)
```
Neelabh94, line 750: `reconfigure_slurm()` calls `util.fetch_config()`, which immediately writes `/slurm/scripts/.config.hash` to disk. If `util.scontrol_reconfigure(lkp)` fails or times out here, the next `slurmsync` pass sees `.config.hash` already matches GCS (`config_changed = False`), so `scontrol reconfigure` is never retried (unlike `topology_changed`, which defers `topology_summary.dump(lkp)` until reconfigure succeeds).

Consider tracking pending reconfigures with a marker file (e.g. `dirs.scripts / ".reconfigure_pending"`) created when `config_changed` is `True` and unlinked only after `util.scontrol_reconfigure(lkp)` succeeds.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -462,6 +462,51 @@ def self_report_controller_address(lkp: util.Lookup) -> None:
...
+            log.debug(f"slurmdbd check error: {e}")
+
+        log.info(f"Waiting for slurmdbd to be ready... (try {try_cnt}/{maxtries}, retrying in {int(wait)}s)")
+        time.sleep(wait)
+        wait = min(wait * 1.5, max_interval)
+
+    log.error(f"Slurmdbd is still not ready after {maxtries} tries. Expecting issues!")
+    raise RuntimeError("Slurmdbd is not ready")
+
+def check_sackd_ready(maxtries: int = 40, interval: float = 15.0) -> None:
```
Neelabh94, line 485: Nit: Rename `interval: float = 15.0` to `max_interval: float = 15.0` to match `check_slurmdbd_ready()` above, and consider guarding `time.sleep(wait)` with `if try_cnt < maxtries:` in both functions to avoid sleeping up to 240s after the final retry fails.

Neelabh94 (approved):
LGTM

arpit974 (approved):

PR #6010, opened 2026-07-27
[Telemetry] Add machine_category metric for hardware classification (CPU/GPU/TPU)

kadupoornima:
### Overview
This PR introduces a new telemetry metric `machine_category` to categorize instantiated compute machines dynamically by identifying their underlying hardware components (e.g., CPU, GPU, TPU). The metric outputs the detected categories similarly to existing telemetry structures maintaining consistency with execution flows. For example, the output traces configurations like `a3-highgpu-8g:GPU,tpu7x-standard-4t:TPU,c2-standard-4:CPU`.
Here is a section you can append directly to your PR description summarizing the categorization logic clearly for reviewers:

### Categorization Logic
The telemetry extraction (`detectMachineCategory`) identifies the hardware bucket of an instance via a strict, prioritized sequence to handle ambiguous instance shaping (such as `-standard-` suffixes shared across CPU and GPU families):

1. **TPU Inference:** Evaluates exactly against the toolkit's native `IsTPU()` identifier, correctly capturing `ct*`, `tpu`, and `v*` pod slice shorthands directly.
2. **GPU Lookup:** Resolves natively against `config.GetMachineMappings().MachineFamilyToLabelMap`. Ambiguous GPU nodes like `g2-standard` or `rtx-6000` (via shorthand substitution) natively map to `"nvidia"` payload labels and are flagged correctly as `GPU`. 
3. **GPU String Keyword:** Acts as a safety net explicitly catching standard types directly featuring `"gpu"` implicitly (like `a2-highgpu-1g`).
4. **CPU Lookup:** With accelerators ruled out, we evaluate against all known standard CPU families using `config.GetMachineMappings().CpuMachineFamilies` (e.g. `n1-`, `c2-`).
5. **CPU Fallback:** Finally, it checks against uniform GCP structural classifiers (`-standard-`, `-highmem-`, `-highcpu-`, `custom-`). This operates purely as a fallback to guarantee that completely undocumented new CPU families securely classify as `CPU` rather than `Other` until mappings update.

### Changes

* **New Telemetry Metric**: Introduced the `MACHINE_CATEGORY` metric to dynamically categorize compute machines as CPU, GPU, or TPU within telemetry data.
* **Classification Logic**: Implemented a prioritized detection mechanism in collector_util.go that leverages existing toolkit mappings and naming conventions to identify hardware types.
* **Robust Testing**: Added comprehensive unit tests in collector_test.go to validate classification accuracy across various machine types, including shorthand mappings and edge cases.


pkg/telemetry/collector_util.go
```diff
@@ -1075,3 +1075,52 @@ func getModuleDynamicNodeCounts(m config.Module, bp config.Blueprint, targetKeys
...
+	}
+
+	return "Other"
+}
+
+// isCPUFallback checks if the machine type contains any of the standard CPU identifiers.
+func isCPUFallback(mType string) bool {
+	cpuKeywords := []string{
+		"-standard-", "-highmem-", "-highcpu-",
+		"-megamem-", "-ultramem-", "custom-",
```
SwarnaBharathiMantena: I checked that there are a few more strings we can add in this list: "-micro", "-small", "-medium", "-metal". These correspond to some of the existing GCE supported CPUs.

pkg/telemetry/collector.go
```diff
@@ -231,6 +233,21 @@ func getMachineType(bp config.Blueprint) string {
 	return strings.Join(machineTypes, ",")
 }
 
+// getMachineCategory maps each machine type to its inferred hardware category (CPU/GPU/TPU/Other)
+func getMachineCategory(bp config.Blueprint) string {
```
SwarnaBharathiMantena, line 235: Is there a way we can reuse the `getMachineType` function here? The current `getMachineCategory` function will probably get simpler if we use `getMachineType`.

pkg/telemetry/collector_test.go
```diff
@@ -3010,3 +3013,163 @@ func TestGetIsAIAssisted(t *testing.T) {
...
+									"disk_type": cty.StringVal("pd-standard"),
+								}),
+							},
+						},
+					},
+				},
+			},
+			want: "",
+		},
+		{
```
SwarnaBharathiMantena, line 3161: Can we add a test to verify the `isCPUFallback`?

pkg/telemetry/collector_util.go
kadupoornima: Done.

pkg/telemetry/collector_test.go
kadupoornima: Done.

pkg/telemetry/collector.go
kadupoornima: Done, thanks.

SwarnaBharathiMantena (approved):
LGTM

PR #6031, opened 2026-07-29
feat: automatic GKE namespace detection for job commands

Neelabh94:
This pull request refactors the GKE job orchestrator to scope Kubernetes operations (such as listing jobsets, resolving Kueue queues, and retrieving logs) to the current active namespace instead of querying across all namespaces or hardcoding the "default" namespace. 
* Removes the GetJobNamespace method and updates the `KubeClient` interface and its implementations accordingly. 
* Iintroduces logic in `configureKubectl` to preserve and restore the active namespace context across `gcloud credential `refreshes, which otherwise reset the context to "default".

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -338,12 +338,18 @@ func (g *GKEOrchestrator) printConsoleLinks(job orchestrator.JobDefinition) {
 	if job.IsPathwaysJob {
 		jobName = job.WorkloadName + "-pathways-head-0"
 	}
-	gkeLink := fmt.Sprintf("https://console.cloud.google.com/kubernetes/job/%s/%s/default/%s/details?project=%s",
-		job.ClusterLocation, job.ClusterName, jobName, job.ProjectID)
+
+	ns, err := g.getCurrentNamespace()
+	if err != nil {
+		ns = "default"
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -145,9 +145,14 @@ func (g *GKEOrchestrator) ListJobs(opts orchestrator.ListOptions) ([]orchestrato
 		return nil, err
 	}
 
-	list, err := g.kubeClient.ListJobSets("gcluster.google.com/workload")
+	ns, err := g.getCurrentNamespace()
 	if err != nil {
-		return nil, fmt.Errorf("failed to list jobsets across all namespaces: %w", err)
+		return nil, fmt.Errorf("failed to get current namespace: %w", err)
```
vikramvs-gg, line 150: While fetching namespace using `getCurrentNamespace()`, there are some cases where namespace falls back to "default" (ex: in `configureKubectl` method) while in some cases we are throwing hard error (like in this case). Should we check if this behaviour can be standardized or if there is functional logic for different behaviour, then lets add some descriptive comment for the same.

Neelabh94: Updated, also created PR #6049 that adds a new namespace flag that further simplifies the logic. PTAL at both PRs together.

PR #6052, opened 2026-07-31
Proactive check and creation of missing GCS backend buckets

PayalJakhar:
## Description

This PR improves the user experience by preventing `terraform init` crashes caused by missing remote state buckets. 
Instead of waiting for Terraform to fail during deployment, `gcluster create` now proactively checks if the specified GCS backend bucket exists. If it is missing, it interactively prompts the user to create it on the spot.

### Key Changes

*   **Pre-flight Check:** Uses the GCS Go SDK to verify if the bucket specified in `terraform_backend_defaults` actually exists.
*   **Interactive Prompt:** If missing, prompts the user `(y/N)` to automatically create the bucket before writing the `.tf` files.
*   **Auto-Provisioning:** Uses the blueprint's `project_id` to instantly create the bucket if the user accepts.
*   **Typo Protection:** If the user declines, the command aborts so flawed `.tf` files aren't generated.

### Testing Performed
*   **Prompt Declined:** Verified that selecting `N` safely aborts the command, protecting the user from typos and preventing invalid `.tf` file generation.
*   **Prompt Accepted:** Verified that selecting `Y` successfully provisions the bucket into the exact `project_id` specified in the blueprint.
*   **Existing Bucket (Authorized):** Verified that if the bucket already exists and the user has IAM permissions, the prompt is seamlessly bypassed and generation succeeds.
*   **Existing Bucket (Unauthorized/403):** Verified the permission edge case. If the globally unique bucket name is already taken by another project and the user lacks access, the program safely catches the `403 Permission Denied` error and aborts immediately.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -174,7 +174,9 @@ func doCreate(cmd *cobra.Command, path string) string {
 	deplDir := filepath.Join(createFlags.outputDir, bp.DeploymentName())
 	logging.Info("Creating deployment folder %q ...", deplDir)
 	checkErr(checkOverwriteAllowed(deplDir, bp, createFlags.overwriteDeployment, createFlags.forceOverwrite), ctx)
-	checkErr(promptAndCreateGcsBuckets(bp), ctx)
+	if os.Getenv("GHPC_SKIP_BUCKET_CREATION") != "true" {
+		checkErr(promptAndCreateGcsBuckets(bp), ctx)
+	}
 	checkErr(modulewriter.WriteDeployment(bp, deplDir), ctx)
 	return deplDir
 }
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -171,10 +171,10 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 
 func doCreate(cmd *cobra.Command, path string) string {
 	bp, ctx := expandOrDie(cmd, path)
-	checkErr(promptAndCreateGcsBuckets(bp), ctx)
 	deplDir := filepath.Join(createFlags.outputDir, bp.DeploymentName())
 	logging.Info("Creating deployment folder %q ...", deplDir)
 	checkErr(checkOverwriteAllowed(deplDir, bp, createFlags.overwriteDeployment, createFlags.forceOverwrite), ctx)
+	checkErr(promptAndCreateGcsBuckets(bp), ctx)
 	checkErr(modulewriter.WriteDeployment(bp, deplDir), ctx)
 	return deplDir
 }
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -73,7 +73,7 @@ var (
 )
 
 func verifyGcsBuckets(ctx context.Context, bp config.Blueprint) error {
-	buckets, err := GetUniqueGcsBuckets(bp)
+	buckets, err := modulewriter.GetUniqueGcsBuckets(bp)
 	if err != nil {
 		return err
 	}
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -89,7 +89,7 @@ func verifyGcsBuckets(ctx context.Context, bp config.Blueprint) error {
 		if client == nil {
 			client, err = storage.NewClient(ctx)
 			if err != nil {
-				logging.Warn("Failed to initialize GCS client to verify bucket '%s': %v. Ensure you have the right permissions and network access.", bucketName, err)
+				logging.Warn("Failed to initialize GCS client to verify bucket '%s': %v. Ensure you have the necessary permissions and network access.", bucketName, err)
 				continue
 			}
 		}
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -89,9 +89,7 @@ func promptAndCreateGcsBuckets(bp config.Blueprint) error {
 		bucketName := g.TerraformBackend.Configuration.Get("bucket").AsString()
 
 		if client == nil {
-			ctxInit, cancelInit := context.WithTimeout(ctx, 30*time.Second)
-			client, err = storage.NewClient(ctxInit)
-			cancelInit()
+			client, err = storage.NewClient(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to initialize GCS client: %w", err)
 			}
@@ -134,11 +132,11 @@ func createGcsBucket(ctx context.Context, client *storage.Client, bucketName str
 		err = bucketHandle.Create(ctxCreate, projectID, bucketAttrs)
 		cancelCreate()
 
-		logging.Info("Note: The GCS bucket %s will not be deleted by gcluster destroy command and needs manual deletion.", bucketName)
-
 		if err != nil {
 			return err
 		}
+
+		logging.Info("Note: The GCS bucket %s will not be deleted by gcluster destroy command and needs manual deletion.", bucketName)
 	} else if err != nil {
 		return fmt.Errorf("failed to verify GCS bucket %q: %w", bucketName, err)
 	}
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -20,6 +20,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"hpc-toolkit/pkg/config"
 	"hpc-toolkit/pkg/logging"
@@ -31,6 +32,7 @@ import (
 	"strings"
 	"time"
 
+	"cloud.google.com/go/storage"
 	"github.com/spf13/cobra"
 )
 
@@ -70,6 +72,46 @@ var (
 	})
 )
 
+func promptAndCreateGcsBuckets(bp config.Blueprint) error {
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = client.Close()
+	}()
+
+	for _, g := range bp.Groups {
+		if g.TerraformBackend.Type != "gcs" || !g.TerraformBackend.Configuration.Has("bucket") {
+			continue
+		}
+		bucketName := g.TerraformBackend.Configuration.Get("bucket").AsString()
+
+		bucketHandle := client.Bucket(bucketName)
+		if _, err := bucketHandle.Attrs(ctx); errors.Is(err, storage.ErrBucketNotExist) {
+			fmt.Printf("Bucket '%s' missing. Create it? (y/N): ", bucketName)
+			var response string
+			_, _ = fmt.Scanln(&response)
+
+			if strings.ToLower(response) != "y" && strings.ToLower(response) != "yes" {
+				return fmt.Errorf("user aborted")
+			}
+
+			if !bp.Vars.Has("project_id") {
+				return fmt.Errorf("cannot create bucket: project_id is missing from blueprint vars")
+			}
+			projectID := bp.Vars.Get("project_id").AsString()
+			if err := bucketHandle.Create(ctx, projectID, nil); err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func runCreateCmd(cmd *cobra.Command, args []string) {
 	deplDir := doCreate(cmd, args[0])
 	logging.Info("To deploy your infrastructure please run:")
@@ -81,6 +123,7 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 
 func doCreate(cmd *cobra.Command, path string) string {
 	bp, ctx := expandOrDie(cmd, path)
+	checkErr(promptAndCreateGcsBuckets(bp), ctx)
 	deplDir := filepath.Join(createFlags.outputDir, bp.DeploymentName())
 	logging.Info("Creating deployment folder %q ...", deplDir)
 	checkErr(checkOverwriteAllowed(deplDir, bp, createFlags.overwriteDeployment, createFlags.forceOverwrite), ctx)
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -34,6 +34,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/spf13/cobra"
+	"github.com/zclconf/go-cty/cty"
 )
 
 func addCreateFlags(c *cobra.Command) *cobra.Command {
@@ -74,7 +75,6 @@ var (
 
 func promptAndCreateGcsBuckets(bp config.Blueprint) error {
 	var client *storage.Client
-	var err error
 	ctx := context.Background()
 	defer func() {
 		if client != nil {
@@ -82,11 +82,28 @@ func promptAndCreateGcsBuckets(bp config.Blueprint) error {
 		}
 	}()
 
+	seenBuckets := make(map[string]bool)
+
 	for _, g := range bp.Groups {
 		if g.TerraformBackend.Type != "gcs" || !g.TerraformBackend.Configuration.Has("bucket") {
 			continue
 		}
-		bucketName := g.TerraformBackend.Configuration.Get("bucket").AsString()
+		evaluatedConfig, err := bp.EvalDict(g.TerraformBackend.Configuration)
+		if err != nil {
+			return fmt.Errorf("failed to evaluate terraform backend configuration: %w", err)
+		}
+		bucketVal := evaluatedConfig.Get("bucket")
+		if bucketVal.IsNull() || bucketVal.Type() != cty.String {
+			return fmt.Errorf("GCS backend bucket name cannot be empty")
+		}
+		bucketName := bucketVal.AsString()
+		if bucketName == "" {
+			return fmt.Errorf("GCS backend bucket name cannot be empty")
+		}
+		if seenBuckets[bucketName] {
+			continue
+		}
+		seenBuckets[bucketName] = true
 
 		if client == nil {
 			client, err = storage.NewClient(ctx)
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -62,14 +62,14 @@ var (
 		forceOverwrite      bool
 	}{}
 
-	createCmd = addAutoApproveFlag(addCreateFlags(&cobra.Command{
+	createCmd = addCreateFlags(&cobra.Command{
 		Use:               "create <BLUEPRINT_FILE>",
 		Short:             "Create a new deployment.",
 		Long:              "Create a new deployment based on a provided blueprint.",
 		Run:               runCreateCmd,
 		Args:              cobra.MatchAll(cobra.ExactArgs(1), checkExists),
 		ValidArgsFunction: filterYaml,
-	}))
+	})
 )
 
 func verifyGcsBuckets(ctx context.Context, bp config.Blueprint) error {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -102,7 +102,7 @@ func verifyGcsBuckets(ctx context.Context, bp config.Blueprint) error {
 		if errors.Is(err, storage.ErrBucketNotExist) {
 			logging.Warn("GCS backend bucket '%s' does not exist. It will be created during the deploy phase.", bucketName)
 		} else if err != nil {
-			logging.Warn("Failed to connect to GCS to verify bucket '%s': %v. Ensure you have the right permissions and network access.", bucketName, err)
+			logging.Warn("Failed to connect to GCS to verify bucket '%s': %v. Ensure you have the necessary permissions and network access.", bucketName, err)
 		}
 	}
 	return nil
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -89,7 +89,9 @@ func promptAndCreateGcsBuckets(bp config.Blueprint) error {
 		bucketName := g.TerraformBackend.Configuration.Get("bucket").AsString()
 
 		if client == nil {
-			client, err = storage.NewClient(ctx)
+			ctxInit, cancelInit := context.WithTimeout(ctx, 30*time.Second)
+			client, err = storage.NewClient(ctxInit)
+			cancelInit()
 			if err != nil {
 				return fmt.Errorf("failed to initialize GCS client: %w", err)
 			}
@@ -132,13 +134,14 @@ func createGcsBucket(ctx context.Context, client *storage.Client, bucketName str
 		err = bucketHandle.Create(ctxCreate, projectID, bucketAttrs)
 		cancelCreate()
 
+		logging.Info("Note: The GCS bucket %s will not be deleted by gcluster destroy command and needs manual deletion.", bucketName)
+
 		if err != nil {
 			return err
 		}
 	} else if err != nil {
 		return fmt.Errorf("failed to verify GCS bucket %q: %w", bucketName, err)
 	}
-	logging.Info("Note: The GCS bucket %s will not be deleted by gcluster destroy command and needs manual deletion.", bucketName)
 	return nil
 }
 
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -62,20 +62,25 @@ var (
 		forceOverwrite      bool
 	}{}
 
-	createCmd = addCreateFlags(&cobra.Command{
+	createCmd = addAutoApproveFlag(addCreateFlags(&cobra.Command{
 		Use:               "create <BLUEPRINT_FILE>",
 		Short:             "Create a new deployment.",
 		Long:              "Create a new deployment based on a provided blueprint.",
 		Run:               runCreateCmd,
 		Args:              cobra.MatchAll(cobra.ExactArgs(1), checkExists),
 		ValidArgsFunction: filterYaml,
-	})
+	}))
 )
 
 func promptAndCreateGcsBuckets(bp config.Blueprint) error {
 	var client *storage.Client
 	var err error
 	ctx := context.Background()
+	defer func() {
+		if client != nil {
+			_ = client.Close()
+		}
+	}()
 
 	for _, g := range bp.Groups {
 		if g.TerraformBackend.Type != "gcs" || !g.TerraformBackend.Configuration.Has("bucket") {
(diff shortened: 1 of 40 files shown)


cmd/create.go
```diff
@@ -70,6 +72,46 @@ var (
...
+			projectID := bp.Vars.Get("project_id").AsString()
+			if err := bucketHandle.Create(ctx, projectID, nil); err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
PayalJakhar: Updated to defer client initialization and use config.GetKeyFromBlueprint

cmd/create.go
```diff
@@ -70,6 +72,51 @@ var (
...
+			continue
+		}
+		bucketName := g.TerraformBackend.Configuration.Get("bucket").AsString()
+
+		if client == nil {
+			client, err = storage.NewClient(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to initialize GCS client: %w", err)
+			}
+			defer func() {
```
AdarshK15: can we move this defer function outside the loop? if needed we can have `client != nil` condition inside it.

cmd/create.go
```diff
@@ -81,6 +128,7 @@ func runCreateCmd(cmd *cobra.Command, args []string) {
 
 func doCreate(cmd *cobra.Command, path string) string {
 	bp, ctx := expandOrDie(cmd, path)
+	checkErr(promptAndCreateGcsBuckets(bp), ctx)
```
AdarshK15: As bucket creation takes place before terraform apply phase, it will not be deleted with ./gcluster destroy command. 
This behaviour should be explicitly mentioned in deployment logs, readme files(of all blueprints that uses gcs buckets), and instructions.txt similar to Packer Images.

cmd/create.go
```diff
@@ -70,6 +72,51 @@ var (
...
+				return fmt.Errorf("failed to initialize GCS client: %w", err)
+			}
+			defer func() {
+				_ = client.Close()
+			}()
+		}
+
+		bucketHandle := client.Bucket(bucketName)
+		if _, err := bucketHandle.Attrs(ctx); errors.Is(err, storage.ErrBucketNotExist) {
+			fmt.Printf("Bucket '%s' missing. Create it? (y/N): ", bucketName)
```
AdarshK15: If  `--auto-approve` is set, it should automatically select `y` instead of waiting for user input. Also in headless mode, it will be stuck here expecting user input.

can we check feasibility of using `confirmAction` and `flagAutoApprove` here similar to 
https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/ceb976a251777ed40208a53b9fc1f48287163982/cmd/destroy.go#L363-L369

AdarshK15: Nit: Mention that this bucket will not be deleted by gcluster destroy command and needs manual deletion.

cmd/create.go
```diff
@@ -70,6 +72,51 @@ var (
...
+
+			if strings.ToLower(response) != "y" && strings.ToLower(response) != "yes" {
+				return fmt.Errorf("user aborted")
+			}
+
+			projectID := config.GetKeyFromBlueprint("project_id", bp)
+			if projectID == "" {
+				return fmt.Errorf("cannot create bucket: project_id is missing or invalid in blueprint vars")
+			}
+			if err := bucketHandle.Create(ctx, projectID, nil); err != nil {
```
AdarshK15: Nit: using `nil` here creates the bucket in default GCP locations, Can we extract `region` from blueprint and use it here?

AdarshK15: Should we add a timeout for this call? instead of using global ctx with timeout (which might add timeout for user input) we should use a new context using `context.WithTimeout` separately for this call.
This also applies to `bucketHandle.Attrs(ctx)`.

cmd/create.go
PayalJakhar: Done, added the flagAutoApprove check

cmd/create.go
PayalJakhar: Done, added logging, updated instructions.txt to mirror the Packer setup, and added a warning to all relevant READMEs.

cmd/create.go
PayalJakhar: added a 30-second context timeout for both network calls.

AdarshK15 (dismissed):

rahimkhan19:
@PayalJakhar Some tests are failing. 
Could you please check to ensure these failures weren't introduced by this change?

cmd/create.go
```diff
@@ -60,16 +63,103 @@ var (
...
+		}
+	}()
+
+	seenBuckets := make(map[string]bool)
+
+	for _, g := range bp.Groups {
+		if g.TerraformBackend.Type != "gcs" || !g.TerraformBackend.Configuration.Has("bucket") {
+			continue
+		}
+		evaluatedConfig, err := bp.EvalDict(g.TerraformBackend.Configuration)
```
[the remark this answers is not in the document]
PayalJakhar: Not needed. The blueprint is already expanded by config.ExpandConfig before reaching this point, meaning expandBackend() has already populated the group's TerraformBackend with the defaults.

pkg/modulewriter/modulewriter.go
```diff
@@ -454,9 +456,22 @@ func writeDestroyInstructions(w io.Writer, bp config.Blueprint, deploymentDir st
 		if grp.Kind() == config.PackerKind {
 			packerManifests = append(packerManifests, filepath.Join(grpPath, string(grp.Modules[0].ID), "packer-manifest.json"))
 		}
+		if grp.TerraformBackend.Type == "gcs" && grp.TerraformBackend.Configuration.Has("bucket") {
+			evaluatedConfig, err := bp.EvalDict(grp.TerraformBackend.Configuration)
+			if err == nil {
```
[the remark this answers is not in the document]
PayalJakhar: Not needed. The blueprint is fully expanded before reaching the modulewriter, meaning TerraformBackendDefaults are already automatically copied into grp.TerraformBackend

cmd/create.go
```diff
@@ -60,16 +63,103 @@ var (
...
 		Run:               runCreateCmd,
 		Args:              cobra.MatchAll(cobra.ExactArgs(1), checkExists),
 		ValidArgsFunction: filterYaml,
-	})
+	}))
 )
 
+func promptAndCreateGcsBuckets(bp config.Blueprint) error {
+	var client *storage.Client
+	ctx := context.Background()
```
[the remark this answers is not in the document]
PayalJakhar: Since createGcsBucket internally enforces a strict 30-second context timeout, plumbing the root command context down isn't strictly necessary for this PR. Keeping as-is.

cmd/create.go
```diff
@@ -84,6 +174,9 @@ func doCreate(cmd *cobra.Command, path string) string {
 	deplDir := filepath.Join(createFlags.outputDir, bp.DeploymentName())
 	logging.Info("Creating deployment folder %q ...", deplDir)
 	checkErr(checkOverwriteAllowed(deplDir, bp, createFlags.overwriteDeployment, createFlags.forceOverwrite), ctx)
+	if os.Getenv("GHPC_SKIP_BUCKET_CREATION") != "true" {
+		checkErr(promptAndCreateGcsBuckets(bp), ctx)
```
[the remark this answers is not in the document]
PayalJakhar: Keeping context.Background() internally for now, as the operation is bounded by an internal 30-second timeout anyway.

cmd/create_test.go
```diff
@@ -188,3 +189,39 @@ func (s *MySuite) TestIsOverwriteAllowed_Present(c *C) {
...
+					Type: "gcs",
+					Configuration: config.NewDict(map[string]cty.Value{
+						"bucket": cty.NullVal(cty.String),
+					}),
+				},
+			},
+		},
+	}
+
+	err := promptAndCreateGcsBuckets(bp)
```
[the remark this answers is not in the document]
PayalJakhar: Skipping this since the function signature isn't being changed.

cmd/create_test.go
```diff
@@ -188,3 +189,39 @@ func (s *MySuite) TestIsOverwriteAllowed_Present(c *C) {
...
+	if err == nil {
+		t.Errorf("expected error when bucket is null, got nil")
+	} else if err.Error() != "GCS backend bucket name cannot be empty" {
+		t.Errorf("expected error 'GCS backend bucket name cannot be empty', got: %v", err)
+	}
+
+	bp.Groups[0].TerraformBackend.Configuration = config.NewDict(map[string]cty.Value{
+		"bucket": cty.StringVal(""),
+	})
+	err = promptAndCreateGcsBuckets(bp)
```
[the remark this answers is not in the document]
PayalJakhar: Skipping this since the function signature isn't being changed.

pkg/modulewriter/modulewriter_test.go
```diff
@@ -667,3 +668,48 @@ func (s *zeroSuite) TestStageFile(c *C) {
...
+		Groups: []config.Group{
+			{
+				Name: "group1",
+				TerraformBackend: config.TerraformBackend{
+					Type: "gcs",
+					Configuration: config.NewDict(map[string]cty.Value{
+						"bucket": cty.StringVal("my-bucket-name"),
+					}),
+				},
+			},
```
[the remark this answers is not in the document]
PayalJakhar: Since the fallback logic happens automatically during blueprint expansion and is already covered there, we don't need redundant test coverage for it in the modulewriter.

cmd/create.go
```diff
@@ -60,16 +63,103 @@ var (
...
+			bucketAttrs = &storage.BucketAttrs{Location: region}
+		}
+
+		ctxCreate, cancelCreate := context.WithTimeout(ctx, 30*time.Second)
+		err = bucketHandle.Create(ctxCreate, projectID, bucketAttrs)
+		cancelCreate()
+
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
PayalJakhar: The GCP storage client already returns a highly descriptive error on failure, so additional error wrapping isn't strictly needed here.

AdarshK15:
Please test and ensure different combinations of the following scenarios are working as expected:
1. gcluster create with/without auto-approve -> gcluster deploy with/without auto-approve
2. gcluster deploy without create with/without auto-approve
3. bucket existing/not existing
4. multiple groups using same/different bucket

AdarshK15 (dismissed):
LGTM, please ensure the test scenarios mentioned are working as expected.

tools/validate_configs/golden_copies/validate.sh
```diff
@@ -51,6 +52,7 @@ run_test() {
 		cd "${tmpdir}"
 	fi
 	${GHPC_PATH} create -l ERROR \
+		--auto-approve \
```
Neelabh94: I don't this this is needed, we don't and shouldn't need to pass --auto-approve for create.

cmd/create.go
```diff
@@ -60,16 +62,52 @@ var (
 		forceOverwrite      bool
 	}{}
 
-	createCmd = addCreateFlags(&cobra.Command{
+	createCmd = addAutoApproveFlag(addCreateFlags(&cobra.Command{
```
Neelabh94: I believe this change is not needed. Can you please confirm once.

cmd/deploy.go
```diff
@@ -61,22 +66,28 @@ func runDeployCmd(cmd *cobra.Command, args []string) {
 		deplRoot = args[0]
 		// check that no "create" flags were specified
 		cmd.Flags().VisitAll(func(f *pflag.Flag) {
-			if f.Changed && createCmd.LocalFlags().Lookup(f.Name) != nil {
+			_, isShared := f.Annotations["shared"]
```
Neelabh94: If we remove the --auto-approve flag change from create.go, this change can be reverted too

cmd/deploy.go
```diff
@@ -61,22 +66,28 @@ func runDeployCmd(cmd *cobra.Command, args []string) {
...
 			}
 		})
 	}
 	skipSecurity, _ := cmd.Flags().GetBool("skip-gke-security-check")
-	doDeploy(deplRoot, skipSecurity)
+	doDeploy(cmd, deplRoot, skipSecurity)
 }
 
-func doDeploy(deplRoot string, skipSecurity bool) {
+func doDeploy(cmd *cobra.Command, deplRoot string, skipSecurity bool) {
```
Neelabh94, line 78: I think instead of passing the whole cmd object, we can just pass the context here, since that is the only thing being used? wdyt?

cmd/deploy.go
```diff
@@ -61,22 +66,28 @@ func runDeployCmd(cmd *cobra.Command, args []string) {
...
-			if f.Changed && createCmd.LocalFlags().Lookup(f.Name) != nil {
+			_, isShared := f.Annotations["shared"]
+			if f.Changed && createCmd.LocalFlags().Lookup(f.Name) != nil && !isShared {
 				checkErr(fmt.Errorf("cannot specify flag %q with DEPLOYMENT_DIRECTORY provided", f.Name), nil)
 			}
 		})
 	}
 	skipSecurity, _ := cmd.Flags().GetBool("skip-gke-security-check")
-	doDeploy(deplRoot, skipSecurity)
+	doDeploy(cmd, deplRoot, skipSecurity)
```
Neelabh94, line 75: if we decide to change the function signature below, we need to only pass the context here, instead of the whole cmd object.

cmd/deploy.go
```diff
@@ -163,3 +174,64 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
 	}
 	return shell.ExportOutputs(tf, artifactsDir, applyBehavior, outputFormat)
 }
+
+func createGcsBucketsIfMissing(ctx context.Context, bp config.Blueprint) error {
```
Neelabh94, line 192: This is very similar to the function in create.go with the major difference around error handling (warning v/s hard failures). See if we can create a standard helper function in utils.go and use that at both places (unless that leads to unreadable code, in which case, please leave it as is).

cmd/deploy.go
```diff
@@ -163,3 +174,64 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
...
+				prompt := fmt.Sprintf("The required Terraform state bucket '%s' is missing. We can create it now, but for data safety, gcluster destroy will not delete it later—you must delete it manually when done. Create it automatically? [y/N]: ", bucketName)
+				if !confirmActionFunc(prompt) {
+					return fmt.Errorf("user aborted")
+				}
+			} else {
+				logging.Info("The required Terraform state bucket '%s' is missing. Auto-approving creation. Note: for data safety, gcluster destroy will not delete it later—you must delete it manually when done.", bucketName)
+			}
+
+			projectID := config.GetKeyFromBlueprint("project_id", bp)
+			if projectID == "" {
```
Neelabh94, line 229: Will this validation not occur at an earlier stage itself? If not, then I think we should throw a similar error at line 221 below as well if region is missing

cmd/deploy.go
```diff
@@ -163,3 +174,64 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
...
+		bucketHandle := client.Bucket(bucketName)
+		ctxTimeout, cancel := context.WithTimeout(ctx, 30*time.Second)
+		_, err = bucketHandle.Attrs(ctxTimeout)
+		cancel()
+
+		if errors.Is(err, storage.ErrBucketNotExist) {
+			if !flagAutoApprove {
+				prompt := fmt.Sprintf("The required Terraform state bucket '%s' is missing. We can create it now, but for data safety, gcluster destroy will not delete it later—you must delete it manually when done. Create it automatically? [y/N]: ", bucketName)
+				if !confirmActionFunc(prompt) {
+					return fmt.Errorf("user aborted")
```
Neelabh94, line 222: Let's make it slightly more informative. We can may be ask the user to create it manually via gcloud command and then update the blueprint once done.

cmd/deploy.go
```diff
@@ -163,3 +174,64 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
...
+			}
+
+			ctxCreate, cancelCreate := context.WithTimeout(ctx, 30*time.Second)
+			err = bucketHandle.Create(ctxCreate, projectID, bucketAttrs)
+			cancelCreate()
+
+			if err != nil {
+				return err
+			}
+		} else if err != nil {
```
Neelabh94, line 246: I am slightly torn here, if we are unable to verify the bucket for any reason, should we block the user or log a warning and let it fail at terraform if it is indeed a problem?

This might help in cases where the IAM permissions are extremely tailored. For example:
The `bucketHandle.Attrs(ctx)` call explicitly requires the `storage.buckets.get` IAM permission. However, to manage state files, Terraform only strictly requires object-level permissions (like `roles/storage.objectAdmin`).

cmd/utils.go
```diff
@@ -52,6 +53,7 @@ func getApplyBehavior() shell.ApplyBehavior {
 
 func addAutoApproveFlag(c *cobra.Command) *cobra.Command {
 	c.Flags().BoolVar(&flagAutoApprove, "auto-approve", false, "Automatically approve proposed changes")
+	_ = c.Flags().SetAnnotation("auto-approve", "shared", []string{"true"})
```
Neelabh94: If we remove --auto-approve from create.go (as suggested above), this change might not be needed.

community/examples/README.md
```diff
@@ -5,3 +5,5 @@ These can be used by the Toolkit to provision your infrastructure.
 
 They are documented in the [the core examples folder](../../examples/README.md)
 along with [community support guidelines](../../examples/README.md#blueprint-descriptions).
+
+**Note:** GCS buckets created for Terraform state are not deleted by the `./gcluster destroy` command and must be deleted manually.
```
Neelabh94: In many of the readmes this instruction/ Note looks completely out of place. Please consider adding it to selected READMEs which actually talk about/ give example of the ./gcluster destroy command, instead of adding it in all READMEs at the end of the file.

tools/validate_configs/golden_copies/expectations/igc_tf/instructions.txt
```diff
@@ -31,3 +31,10 @@ Infrastructure should be destroyed in reverse order of creation:
 
 terraform -chdir=golden_copy_deployment/one destroy
 terraform -chdir=golden_copy_deployment/zero destroy
+
+Please browse to the Cloud Console to remove GCS buckets used for Terraform state.
+The following buckets were used and will not be automatically deleted by gcluster destroy:
+
+us-east4-c
```
Neelabh94, line 37: Why us-east4-c ?

Neelabh94 (changes requested):
Some nit comments about code improvements. PTAL

cmd/deploy.go
PayalJakhar: I tried combining them, but the vastly different error handling and user prompts made a shared helper too messy, so I left them separate to keep the code readable as you suggested.

cmd/deploy.go
PayalJakhar: Both values are strictly validated earlier, so they should be safe here. I kept the project_id check just as a final safeguard because the Google Cloud API strictly requires it to make a bucket. However, region is entirely optional for GCS buckets (it just defaults to a multi region if omitted), so we definitely shouldn't throw a hard failure if the region is missing.

tools/validate_configs/golden_copies/expectations/igc_tf/instructions.txt
PayalJakhar: dummy test file (igc_tf.yaml) sets the bucket name using the `$(vars.zone)` variable instead of a real bucket name. Since `$(vars.zone)` evaluates to us-east4-c in this test, my code just picks that up and prints it as the bucket name

examples/gke-a3-megagpu/README.md
```diff
@@ -171,3 +171,5 @@ Refer to [Deploy and run NCCL test with Topology Aware Scheduling (TAS)](https:/
 ### Additional Consumption Options
 The Cluster Toolkit supports alternative consumption options such as Spot VMs or Dynamic Workload Scheduler (DWS) Flex-start.
 Refer to step 5 of [Create a cluster using Cluster Toolkit](https://docs.cloud.google.com/ai-hypercomputer/docs/create/gke-ai-hypercompute#use-cluster-toolkit) for general instructions on other consumption options. Similar configuration settings can be used for GKE-A3M cluster as well.
+
+**Note:** GCS buckets created for Terraform state are not deleted by the `./gcluster destroy` command and must be deleted manually.
```
Neelabh94: Please move it to line 164 right under the destroy command

examples/gke-tpu-7x/README.md
```diff
@@ -461,3 +461,5 @@ The blueprint includes a sample job (`shared-fs-job`) that demonstrates how two
     ```
 
 The logs will display content from `shared_output.txt`, showing timestamps and hostnames from both pods, confirming that the filesystem is truly shared.
+
```
Neelabh94: Please move it before the Appendix Section (make it a part of the clean-up section)

examples/gke-tpu-v6e/README.md
```diff
@@ -387,3 +387,5 @@ The blueprint includes a sample job (`shared-fs-job`) that demonstrates how two
     ```
 
 The logs will display content from `shared_output.txt`, showing timestamps and hostnames from both pods, confirming that the filesystem is truly shared.
+
```
Neelabh94: Same here as tpu-7x readme

examples/machine-learning/a3-highgpu-8g/README.md
```diff
@@ -291,3 +291,5 @@ sbatch run-nccl-tests.sh
 [tkdeps]: https://cloud.google.com/cluster-toolkit/docs/setup/install-dependencies
 [tkinstall]: https://github.com/GoogleCloudPlatform/cluster-toolkit/#quickstart
 [deprecated-badge]: https://img.shields.io/badge/-deprecated-%23fea2a2?style=plastic
+
```
Neelabh94: Even though it's a little scope creep, but can you please add this section here:
### Destroy the VMs

```bash
#!/bin/bash
./gcluster destroy <DEPLOYMENT_FOLDER> --auto-approve
```

examples/machine-learning/a4x-highgpu-4g/README.md
```diff
@@ -138,3 +138,5 @@ to manage data.
    storage.
 1. `/gcs-model-serving` is an optimized mount for serving models, which
    downloads model weights in parallel to local ssd.
+
```
Neelabh94: Move the note closer to the destroy command section

pkg/modulewriter/modulewriter.go
```diff
@@ -454,9 +456,22 @@ func writeDestroyInstructions(w io.Writer, bp config.Blueprint, deploymentDir st
 		if grp.Kind() == config.PackerKind {
 			packerManifests = append(packerManifests, filepath.Join(grpPath, string(grp.Modules[0].ID), "packer-manifest.json"))
 		}
+		if grp.TerraformBackend.Type == "gcs" && grp.TerraformBackend.Configuration.Has("bucket") {
```
Neelabh94: Can we see if reusing the function defined in `cmd/utils.go` is possible.

Neelabh94 (changes requested):

examples/science/af3-slurm/README.md
```diff
@@ -542,3 +542,5 @@ reverse order. Leave out the `--auto-approve` if you want control over each depl
 > associated costs. Also, the buckets you may have created via the cloud console or CLI will
 > not be destroyed by the above command (they would be, however, destroyed if you deleted the project).
 > For deleting the buckets consult [Delete buckets](https://cloud.google.com/storage/docs/deleting-buckets).
+
+**Note:** GCS buckets created for Terraform state are not deleted by the `./gcluster destroy` command and must be deleted manually.
```
Neelabh94: If you read through line 542 and 543 above this, you'll see that this is already mentioned 
```text
Also, the buckets you may have created via the cloud console or CLI will
> not be destroyed by the above command (they would be, however, destroyed if you deleted the project)
```
I think we can skip the added note here and keep this readme unchanged

examples/README.md
```diff
@@ -2105,3 +2105,5 @@ This means that you can redeploy a blueprint (`gcluster deploy <blueprint> -w`)
 Slurm-GCP v5 users should read [Slurm-GCP v5 EOL](../docs/slurm-gcp-support.md)
 for information on v5 retirement and feature highlights for v6. Slurm-GCP v6 is
 only supported option within the Toolkit.
+
+**Note:** GCS buckets created for Terraform state are not deleted by the `./gcluster destroy` command and must be deleted manually.
```
Neelabh94: This should be moved to the section where the terraform backend buckets are discussed. I think a good position will be somewhere between line 147 and 157.

pkg/modulewriter/modulewriter.go
```diff
@@ -475,3 +480,64 @@ func WritePackerDestroyInstructions(w io.Writer, manifests []string) {
...
 }
+
+// WriteGcsDestroyInstructions prints our best effort guidance to the user on
+// deleting GCS buckets used for Terraform state.
+func WriteGcsDestroyInstructions(w io.Writer, buckets []string) {
+	if len(buckets) == 0 {
+		return
+	}
+
+	uniqueBuckets := []string{}
```
Neelabh94: This is redundant since The buckets array passed into this function is populated by calling `GetUniqueGcsBuckets(bp)`.
`GetUniqueGcsBuckets` already returns a deduplicated list

pkg/modulewriter/modulewriter.go
```diff
@@ -475,3 +480,64 @@ func WritePackerDestroyInstructions(w io.Writer, manifests []string) {
...
+	seen := map[string]bool{}
+	for _, b := range buckets {
+		if !seen[b] {
+			seen[b] = true
+			uniqueBuckets = append(uniqueBuckets, b)
+		}
+	}
+
+	_, _ = fmt.Fprintln(w)
```
Neelabh94, line 491: Currently, the code produces this output:

```text
Please browse to the Cloud Console to remove GCS buckets used for Terraform state.
The following buckets were used and will not be automatically deleted by gcluster destroy:

my-bucket-name

https://console.cloud.google.com/storage/browser
```

It feels a little awkward because the call-to-action ("Please browse to the Cloud Console...") is separated from the actual link by the list of buckets. Please consider replacing it with

```shell
	_, _ = fmt.Fprintln(w)
	_, _ = fmt.Fprintln(w, "The following buckets were used for Terraform state and will not be automatically deleted by gcluster destroy:")
	_, _ = fmt.Fprintln(w)
	for _, bucket := range uniqueBuckets {
		_, _ = fmt.Fprintln(w, bucket)
	}
	_, _ = fmt.Fprintln(w)
	_, _ = fmt.Fprintln(w, "Please browse to the Cloud Console to remove them:")
	_, _ = fmt.Fprintln(w, "https://console.cloud.google.com/storage/browser")
```

Neelabh94 (changes requested):

cmd/deploy.go
PayalJakhar: cmd itself is passed directly into checkDependencies below, so we still need the full object here.

cmd/deploy.go
PayalJakhar: Since we're keeping the downstream function as-is for now, we still need to pass the full cmd object here

Neelabh94 (dismissed):
LGTM

cmd/create.go
```diff
@@ -70,6 +72,42 @@ var (
...
+
+		bucketHandle := client.Bucket(bucketName)
+		ctxTimeout, cancel := context.WithTimeout(ctx, 30*time.Second)
+		_, err = bucketHandle.Attrs(ctxTimeout)
+		cancel()
+
+		if errors.Is(err, storage.ErrBucketNotExist) {
+			logging.Warn("GCS backend bucket '%s' does not exist. It will be created during the deploy phase.", bucketName)
+		} else if err != nil {
+			logging.Warn("Failed to connect to GCS to verify bucket '%s': %v. Ensure you have the right permissions and network access.", bucketName, err)
```
AdarshK15: Nit: `right permissions` -> `necessary permissions`?

cmd/deploy.go
```diff
@@ -179,3 +188,64 @@ func deployTerraformGroup(groupDir string, artifactsDir string, applyBehavior sh
...
+		}
+
+		bucketHandle := client.Bucket(bucketName)
+		ctxTimeout, cancel := context.WithTimeout(ctx, 30*time.Second)
+		_, err = bucketHandle.Attrs(ctxTimeout)
+		cancel()
+
+		if errors.Is(err, storage.ErrBucketNotExist) {
+			if !flagAutoApprove {
+				prompt := fmt.Sprintf("The required Terraform state bucket '%s' is missing. We can create it now, but for data safety, gcluster destroy will not delete it later—you must delete it manually when done.\nAlternatively, you can create it manually via `gcloud storage buckets create gs://%s` and then re-run this command.\nCreate it automatically? [y/N]: ", bucketName, bucketName)
```
AdarshK15: Nit: 
`gcluster destroy` -> `'gcluster destroy'`
`later—you` -> `later. You`

AdarshK15 (dismissed):
LGTM

AdarshK15 (approved):
LGTM

PR #6056, opened 2026-08-01
feat: optimize dependency management and improve UX

Neelabh94:
This PR optimizes dependency management by making checks context-aware and improves UX with actionable hints for missing tools (Git, Packer, GCloud).

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -56,8 +56,6 @@ func runExportCmd(cmd *cobra.Command, args []string) {
 
 	bp, ctx := artifactBlueprintOrDie(artifactsDir)
 
-	checkDependencies(cmd, "terraform")
-
 	checkErr(shell.ValidateDeploymentDirectory(bp.Groups, deplRoot), ctx)
 
 	group, err := bp.Group(groupName)
@@ -69,6 +67,8 @@ func runExportCmd(cmd *cobra.Command, args []string) {
 	if group.Kind() != config.TerraformKind {
 		checkErr(errors.New("export command is supported for Terraform modules only"), ctx)
 	}
+
+	checkDependencies(cmd, "terraform")
 
 	tf, err := shell.ConfigureTerraform(groupDir)
 	checkErr(err, ctx)
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -56,6 +56,8 @@ func runExportCmd(cmd *cobra.Command, args []string) {
 
 	bp, ctx := artifactBlueprintOrDie(artifactsDir)
 
+	checkDependencies(cmd, "terraform")
+
 	checkErr(shell.ValidateDeploymentDirectory(bp.Groups, deplRoot), ctx)
 
 	group, err := bp.Group(groupName)
--- a/pkg/dependencies/resolver.go
+++ b/pkg/dependencies/resolver.go
@@ -66,14 +66,39 @@ func PatchPath() error {
 	return nil
 }
 
-// EnsureDependencies checks if terraform and packer are accessible in the PATH.
+// HasBinary checks if a binary is available in the PATH.
+func HasBinary(name string) bool {
+	_, err := exec.LookPath(name)
+	return err == nil
+}
+
+// EnsureDependencies checks if requested tools are accessible in the PATH.
 // If not, it handles downloading them according to the decision.
-func EnsureDependencies(decision DownloadDecision) error {
-	if err := ensureBinary("terraform", TerraformVersion, decision); err != nil {
-		return err
+// It returns an error if no tools are specified.
+func EnsureDependencies(decision DownloadDecision, tools ...string) error {
+	if len(tools) == 0 {
+		return fmt.Errorf("no tools specified for dependency check")
 	}
-	if err := ensureBinary("packer", PackerVersion, decision); err != nil {
-		return err
+
+	seen := make(map[string]bool)
+	for _, tool := range tools {
+		if seen[tool] {
+			continue
+		}
+		seen[tool] = true
+
+		var err error
+		switch tool {
+		case "terraform":
+			err = ensureBinary("terraform", TerraformVersion, decision)
+		case "packer":
+			err = ensureBinary("packer", PackerVersion, decision)
+		default:
+			return fmt.Errorf("unknown tool requested: %s", tool)
+		}
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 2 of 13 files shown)


pkg/dependencies/resolver.go
```diff
@@ -64,14 +64,45 @@ func PatchPath() error {
...
+			err = ensureBinary("packer", PackerVersion, decision)
+		default:
+			return fmt.Errorf("unknown tool requested: %s", tool)
+		}
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
```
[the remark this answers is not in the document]
Neelabh94: This is in direct contradiction to a previous review comment which suggested adding this loop to ensure all dependencies are scanned to prevent partial downloads.

Ignoring this since the performance hit is minimal (2-3 items at max) but the fail-fast gain is higher

cmd/export.go
```diff
@@ -56,6 +56,8 @@ func runExportCmd(cmd *cobra.Command, args []string) {
 
 	bp, ctx := artifactBlueprintOrDie(artifactsDir)
 
+	checkDependencies(cmd, "terraform")
```
saara-tyagi27: Can we consider moving `checkDependencies(cmd, "terraform")` down after line 69 (right before `shell.ConfigureTerraform(groupDir)`).

Currently, if a user runs `export-outputs` with an invalid group name or on a Packer group when Terraform is missing, it might prompt/fail on Terraform before reporting that the module kind is unsupported.

pkg/sourcereader/goget.go
```diff
@@ -18,12 +18,23 @@ import (
...
 
+// GitMissingError indicates that git is required but missing from PATH.
+type GitMissingError struct {
+	Source string
+}
+
+func (e GitMissingError) Error() string {
+	return fmt.Sprintf("'git' is required to download remote module %q but not found in PATH", e.Source)
+}
+
```
saara-tyagi27, line 46: Error message suggestion:

```suggestion
func (e GitMissingError) Error() string {
	return fmt.Sprintf("'git' is required to download remote module %q but not found in PATH. Please install git via your OS package manager (e.g., `apt install git` or `brew install git`).", e.Source)
}

File: pkg/sourcereader/goget.go [hunk 4]
```diff
@@ -54,6 +65,29 @@ func getterClient(source string, dst string) getter.Client {
 
 // GetModule copies the git source to a provided destination (the deployment directory)
 func (r GoGetterSourceReader) GetModule(source string, dst string) error {
+	detectors := []getter.Detector{
+		new(getter.GitHubDetector),
```
saara-tyagi27: This `detectors` slice is similar to the one declared in `getterClient`.
Can we consider defining a package-level variable at the top of the file:
```go
var defaultDetectors = []getter.Detector{
	new(getter.GitHubDetector),
	new(getter.GitDetector),
	new(getter.GCSDetector),
}
```

pkg/sourcereader/goget.go
```diff
@@ -54,6 +65,29 @@ func getterClient(source string, dst string) getter.Client {
...
+		return fmt.Errorf("failed to get working directory: %w", err)
+	}
+
+	detected, err := getter.Detect(source, pwd, detectors)
+	isGit := false
+	if err == nil {
+		isGit = strings.HasPrefix(detected, "git::") || strings.HasPrefix(detected, "git://") || strings.HasPrefix(detected, "git+")
+	}
+
+	if isGit {
```
saara-tyagi27:
```suggestion
	if isGit && !dependencies.HasBinary("git") {
		return GitMissingError{Source: source}
	}
```

Can we reuse the `dependencies.HasBinary("git")` helper introduced in this PR (`pkg/dependencies/resolver.go`) instead of calling `exec.LookPath("git")` directly.

cmd/export.go
Neelabh94: Agreed. Done

pkg/sourcereader/goget.go
Neelabh94: Thanks, updated.

pkg/sourcereader/goget.go
Neelabh94: Done.

pkg/sourcereader/goget.go
Neelabh94: Good catch, done.

pkg/sourcereader/goget.go
```diff
@@ -54,17 +69,48 @@ func getterClient(source string, dst string) getter.Client {
...
+
+	if !isSupported {
+		return fmt.Errorf("unsupported module source protocol %q (detected from %q); "+
+			"supported protocols are Git (e.g., git::https://..., git+ssh://...) and GCS (e.g., gcs::...)",
+			u.Scheme, source)
+	}
+
+	if isGit && !dependencies.HasBinary("git") {
+		return GitMissingError{Source: source}
+	}
```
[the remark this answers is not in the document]
Neelabh94: We do **NOT** need the Protocol Splitter.

The current approach of `getter.Detect` followed by `url.Parse` is actually more robust and introduces fewer edge cases than the manual splitting suggestion above, because:

* We avoid the IPv6 literal bug.
* We rely on the standard library for parsing.
* `getter.Detect` already does the job of canonicalizing shorthand URLs into one of the explicit schemes we support.

The reviewer's concern about `gcs::https://...` breaking `url.Parse` was based on a misunderstanding of how Go's URL parser handles custom schemes (it handles them fine).

Sticking with the current `url.Parse` approach. It is safer and more idiomatic.

pkg/sourcereader/goget.go
```diff
@@ -54,17 +69,48 @@ func getterClient(source string, dst string) getter.Client {
...
+
 	tmp, err := os.MkdirTemp("", "get-module-*")
 	defer os.RemoveAll(tmp)
 	if err != nil {
 		return err
```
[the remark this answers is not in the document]
Neelabh94: Done

pkg/sourcereader/goget.go
```diff
@@ -17,13 +17,32 @@ package sourcereader
...
+	"hpc-toolkit/pkg/dependencies"
+
 	"github.com/hashicorp/go-getter"
 )
 
+var defaultDetectors = []getter.Detector{
+	new(getter.GitHubDetector),
+	new(getter.GitDetector),
+	new(getter.GCSDetector),
+}
```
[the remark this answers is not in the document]
Neelabh94: Done

Neelabh94:
@saara-tyagi27 thanks for the in-depth review.
Have resolved all your comments + went through another round of gemini reviews as well.
Please review again when possible.

saara-tyagi27 (approved):

PR #6113, opened 2026-08-10
fix(test): update get_real_rack to extract cluster and rack ID

sriram-quad:
The Fix: Updated the get_real_rack() function in tools/python-integration-tests/slurm_topology.py to extract both the cluster and the rack components from the physicalHost path (e.g., parts[1]/parts[2]). This ensures the test evaluates the physical node topology exactly the same way the Slurm topology generator does.

### DIFF
--- a/tools/python-integration-tests/slurm_topology.py
+++ b/tools/python-integration-tests/slurm_topology.py
@@ -15,6 +15,7 @@
 from deployment import Deployment
 from collections import defaultdict
 import logging
+import time
 import test
 
 logging.basicConfig(level=logging.INFO) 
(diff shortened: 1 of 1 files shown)


tools/python-integration-tests/slurm_topology.py
```diff
@@ -51,7 +51,8 @@ def get_real_rack(self, node: str):
         result = self.run_command(f"gcloud compute instances describe {node} --zone={self.deployment.zone} --project={self.deployment.project_id} --format='value(resourceStatus.physicalHost)'")
         physicalHost = result.stdout
         log.info(f"physicalHost for {node}: {physicalHost.strip()}")
-        return physicalHost.split("/")[1]
+        parts=physicalHost.strip().split("/")
```
rahimkhan19: Please fix these issues:
1. NameError: The variable is defined as parts but referenced as part.
2. TypeError / Formatting issue: The expression {part[1]/part[2]} inside the f-string will attempt to mathematically divide the two strings rather than concatenate them with a slash.
The correct code should be:

```
parts = physicalHost.strip().split("/")
return f"{parts[1]}/{parts[2]}"      
```

tools/python-integration-tests/slurm_topology.py
```diff
@@ -51,11 +51,12 @@ def get_real_rack(self, node: str):
         result = self.run_command(f"gcloud compute instances describe {node} --zone={self.deployment.zone} --project={self.deployment.project_id} --format='value(resourceStatus.physicalHost)'")
         physicalHost = result.stdout
         log.info(f"physicalHost for {node}: {physicalHost.strip()}")
-        return physicalHost.split("/")[1]
+        parts = physicalHost.strip().split("/")
+        return f"{parts[1]}/{parts[2]}"
```
AdarshK15: Nit: Can we add a length check for parts here along with a fallback or a descriptive exception if the condition is not met? 
Even though we expect GCE physicalHost path to always contain these components, it would be better to add this defensively to prevent `IndexErrors` if the format changes or value is empty.
example:
```python
        parts = [p for p in physicalHost.split("/") if p]
        if len(parts) > 1:
            return f"{parts[0]}/{parts[1]}"
        elif len(parts) > 0:
            return parts[0]
        else:
            raise ValueError(f"Unexpected physicalHost format or empty value received for {node}: '{physicalHost}'.")
```

tools/python-integration-tests/slurm_topology.py
sriram-quad: Thanks for catching that! We have fixed both issues by defining the variable as `parts` and updating the f-string to correctly concatenate the components with a slash (`f"{parts[1]}/{parts[2]}"`).

AdarshK15 (dismissed):

tools/python-integration-tests/slurm_topology.py
sriram-quad: Thanks for the suggestion! Added the fallback logic as recommended.

tools/cloud-build/daily-tests/builds/slurm-gcp-v6-topology.yaml
```diff
@@ -22,239 +22,24 @@ tags:
 - slurm6
 
 substitutions:
-  _TEST_PREFIX: "" # Default to no prefix
```
rahimkhan19, line 25: _TEST_PREFIX is required
we should not remove this

sriram-quad: Sure Rahim reverted the changes

aslam-quad:
Hi @rahimkhan19 

1.The topology depth check was flaky because Slurm sometimes reports depth 1 while the topology is still being updated, even though the expected depth is 2. Added a retry mechanism in get_slurm_rack() to check again after 30 seconds, up to 5 attempts, before failing the test.

    def get_slurm_rack(self, node: str):
        max_retries = 5     
        wait_seconds = 30    

        for attempt in range(max_retries):
            stdin, stdout, stderr = self.ssh_client.exec_command(f"scontrol show topology node={node} | tail -1 | cut -d' ' -f1")
            switch_name = stdout.read().decode().strip()
            err = stderr.read().decode()
            
            if err:
                raise Exception(f"Slurm topology error for {node}: {err}") 

            depth = self.get_node_depth(switch_name)
            
            # If the depth is exactly 2, the map is ready! Stop waiting and return it.
            if depth == 2:
                [log.info](http://log.info/)(f"Slurm rack for {node} ready: {switch_name}")
                return switch_name
                
            # If depth is not 2, log a message and wait before trying again.
            [log.info](http://log.info/)(f"Attempt {attempt + 1}: Topology depth for {node} is {depth}. Waiting {wait_seconds}s for Slurm to update...")
            time.sleep(wait_seconds)
            
        # If the loop finishes and it still didn't find a depth of 2, THEN fail the test.
        self.assert_equal(depth, 2, f"{node} has a topology depth of {depth} which does not match expected topology depth of 2.")
        return switch_name
    
    
 2.As per my understanding, the rack information is being taken from the wrong part of the physical host path. Because of this, the nodes may not be grouped correctly, which can cause the Google Cloud rack grouping to not match the Slurm grouping. If we change split("/")[1] to split("/")[2], it should get the correct rack-level information, group the nodes correctly, and make the rack comparison match Slurm.

 ```
 def get_real_rack(self, node: str):
        result = self.run_command(f"gcloud compute instances describe {node} --zone={self.deployment.zone} --project={self.deployment.project_id} --format='value(resourceStatus.physicalHost)'")
        physicalHost = result.stdout
        [log.info](http://log.info/)(f"physicalHost for {node}: {physicalHost.strip()}")
        return physicalHost.split("/")[2]
```

**Comment by rahimkhan19** [comment]
@aslam-quad 
**Regarding Point 1 (Retry mechanism for depth):** 
Great catch! I completely agree and we should implement this retry logic. When nodes first spin up and `physicalHost` isn't fully processed yet, `conf.py` temporarily places them in a fallback path (`Root -> Nodeset`). Slurm compresses this to a switch named `s0_X` (which has 1 underscore, i.e., depth 1). Once the actual physical placement is resolved, it updates to the true `Root -> Cluster -> Rack` path, compressed to `s0_X_Y` (depth 2). Adding the polling mechanism ensures we don't accidentally pass the test while the cluster is still in its transient fallback state.

**Regarding Point 2 (Using `split("/")[2]`):** 
We should stick with the current implementation (`parts[0]/parts[1]`). Inside `conf.py`, the `_make_physical_path` function explicitly uses `parts[:2]` to build the topology tree, meaning Slurm partitions the nodes by **both** Cluster and Rack. 

If we switch to `split("/")[2]`, we discard the Cluster component. In a scenario where multiple clusters happen to share a Rack ID (e.g., `/mac-1/rack-A` and `/mac-2/rack-A`), our test would mistakenly merge them into one giant group, while Slurm would correctly separate them into different leaf switches (`s0_0_0` vs `s0_1_0`). Returning `parts[0]/parts[1]` exactly mirrors Slurm's equivalence classes and prevents this false failure, while also gracefully handling edge cases where the path might be shorter.

cc: @sriram-quad

File: tools/python-integration-tests/slurm_topology.py [hunk 4]
```diff
@@ -48,20 +49,51 @@ def get_node_depth(self, switch_name: str):
...
+            
+            if depth == 2:
+                log.info(f"Slurm rack for {node} ready: {switch_name}")
+                return switch_name
+                
+            log.info(f"Attempt {attempt + 1}: Topology depth for {node} is {depth}. Waiting {wait_seconds}s for Slurm to update...")
+            time.sleep(wait_seconds)
+            
         self.assert_equal(depth, 2, f"{node} has a topology depth of {depth} which does not match expected topology depth of 2.")
         return switch_name
```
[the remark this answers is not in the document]
sriram-quad: Changes added in PR

rahimkhan19 (approved):
LGTM

AdarshK15 (approved):

PR #6115, opened 2026-08-10
Add Service Account ID validation

agrawalkhushi18:
This PR validates the combined length and format of the deployment name and service account name to ensure they comply with Google Cloud's 30-character limit and naming rules. This validation runs during the pre-flight phase, preventing late-stage deployment failures in Terraform.
The validator is safe-guarded against null or unknown values to prevent panics during blueprint expansion, and uses a pre-compiled regular expression for efficiency.

### Submission Checklist

### DIFF
--- a/modules/project/service-account/metadata.yaml
+++ b/modules/project/service-account/metadata.yaml
@@ -34,3 +34,7 @@ ghpc:
       vars: [prefix]
       deprecated: true
     error_message: "Unsupported variable: 'prefix'. Please use 'deployment_name' instead."
+  - validator: service_account_id
+    inputs:
+      deployment_name: deployment_name
+      name: name
--- a/modules/project/service-account/metadata.yaml
+++ b/modules/project/service-account/metadata.yaml
@@ -34,7 +34,10 @@ ghpc:
       vars: [prefix]
       deprecated: true
     error_message: "Unsupported variable: 'prefix'. Please use 'deployment_name' instead."
-  - validator: service_account_id
+  - validator: regex
     inputs:
-      deployment_name: deployment_name
-      name: name
+      vars: [deployment_name, name]
+      concat: true
+      separator: "-"
+      pattern: "^[a-z][-a-z0-9]{4,28}[a-z0-9]$"
+    error_message: "The combined service account ID (deployment_name + name) must be between 6 and 30 characters, start with a lowercase letter, contain only lowercase alphanumeric characters and dashes, and cannot end with a dash."
--- a/pkg/validators/metadata_validators.go
+++ b/pkg/validators/metadata_validators.go
@@ -37,7 +37,7 @@ func resolveSettingToString(
 	values, path, err := getModuleSettingValues(bp, group, modIdx, mod, settingName)
 	if err != nil {
 		if optional {
-			return "", true, path, nil
+			return "", false, path, nil
 		}
 		missingPath := config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Settings.Dot(settingName)
 		return "", false, missingPath, config.BpError{
@@ -47,7 +47,7 @@ func resolveSettingToString(
 	}
 	if len(values) == 0 || values[0].Type() != cty.String {
 		if len(values) == 0 && optional {
-			return "", true, path, nil
+			return "", false, path, nil
 		}
 		return "", false, path, config.BpError{Err: fmt.Errorf("setting %q must be a string", settingName), Path: path}
 	}
@@ -101,6 +101,10 @@ func (r *RegexValidator) validateConcat(
 		if val != "" {
 			parts = append(parts, val)
 		}
+	}
+
+	if len(parts) == 0 {
+		return nil
 	}
 
 	joined := strings.Join(parts, separator)
--- a/pkg/validators/metadata_validators.go
+++ b/pkg/validators/metadata_validators.go
@@ -70,7 +70,7 @@ func (r *RegexValidator) validateConcat(
 	optional bool,
 ) error {
 	varsList, _ := parseStringList(rule.Inputs["vars"])
-	separator, _ := rule.Inputs["separator"].(string)
+	separator, _ := parseString(rule.Inputs["separator"])
 	allowNullList, _ := parseStringList(rule.Inputs["allow_null"])
 	allowNullMap := make(map[string]struct{})
 	for _, v := range allowNullList {
@@ -179,10 +179,11 @@ func (r *RegexValidator) Validate(
 		}
 	}
 
-	optional := true
-	if v, ok := rule.Inputs["optional"]; ok {
-		if b, ok := v.(bool); ok {
-			optional = b
+	optional, err := parseBoolInput(rule.Inputs, "optional", true)
+	if err != nil {
+		return config.BpError{
+			Err:  fmt.Errorf("failed to parse 'optional' input: %w", err),
+			Path: config.Root.Groups.At(bp.GroupIndex(group.Name)).Modules.At(modIdx).Source,
 		}
 	}
 
@@ -607,19 +608,19 @@ func (c *ConditionalRegexValidator) Validate(
 		return nil
 	}
 
-	dependentVal, depPath := getSettingWithFallback(bp, group, modIdx, mod, dependent)
-	if dependentVal == cty.NilVal || !dependentVal.IsKnown() || dependentVal.IsNull() || dependentVal.Type() != cty.String {
+	valStr, known, depPath, err := resolveSettingToString(bp, group, modIdx, mod, dependent, true, false)
+	if err != nil || !known {
 		return nil
 	}
 
-	matched := re.MatchString(dependentVal.AsString())
+	matched := re.MatchString(valStr)
 	if matched != matchExpected {
 		msg := rule.ErrorMessage
 		if msg == "" {
 			if matchExpected {
-				msg = fmt.Sprintf("variable %q value %q must match pattern %q when conditions are met", dependent, dependentVal.AsString(), re.String())
+				msg = fmt.Sprintf("variable %q value %q must match pattern %q when conditions are met", dependent, valStr, re.String())
 			} else {
-				msg = fmt.Sprintf("variable %q value %q must not match pattern %q when conditions are met", dependent, dependentVal.AsString(), re.String())
+				msg = fmt.Sprintf("variable %q value %q must not match pattern %q when conditions are met", dependent, valStr, re.String())
 			}
 		}
 		return config.BpError{Err: fmt.Errorf("%s", msg), Path: depPath}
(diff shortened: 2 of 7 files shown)


pkg/validators/metadata_validators.go
```diff
@@ -590,3 +590,67 @@ func evalTriggers(
 	}
 	return true
 }
+
+// ServiceAccountIDValidator validates if the combined deployment_name and name
+// form a valid GCP Service Account ID.
+type ServiceAccountIDValidator struct{}
```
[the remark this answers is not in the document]
agrawalkhushi18: done

pkg/validators/metadata_validators.go
```diff
@@ -590,3 +590,67 @@ func evalTriggers(
...
+	}
+
+	depNameValues, depNamePath, err := getModuleSettingValues(bp, group, modIdx, mod, depNameSetting)
+	if err != nil {
+		return config.BpError{Err: fmt.Errorf("failed to resolve setting %q: %w", depNameSetting, err), Path: depNamePath}
+	}
+	if len(depNameValues) == 0 || depNameValues[0].Type() != cty.String {
+		return config.BpError{Err: fmt.Errorf("setting %q must be a string", depNameSetting), Path: depNamePath}
+	}
+	depName := depNameValues[0].AsString()
```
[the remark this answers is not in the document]
agrawalkhushi18: done

pkg/validators/metadata_validators.go
```diff
@@ -590,3 +590,67 @@ func evalTriggers(
...
+
+	if len(saID) < 6 || len(saID) > 30 {
+		return config.BpError{
+			Err:  fmt.Errorf("generated service account ID %q is invalid: length must be between 6 and 30 characters, but got %d (deployment_name: %d, name: %d)", saID, len(saID), len(depName), len(name)),
+			Path: namePath,
+		}
+	}
+
+	re := regexp.MustCompile(`^[a-z][-a-z0-9]*[a-z0-9]$`)
+	if !re.MatchString(saID) {
```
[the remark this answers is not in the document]
agrawalkhushi18: done

pkg/validators/metadata_validators.go
```diff
@@ -590,3 +590,67 @@ func evalTriggers(
...
+	depName := depNameValues[0].AsString()
+
+	nameValues, namePath, err := getModuleSettingValues(bp, group, modIdx, mod, nameSetting)
+	if err != nil {
+		return config.BpError{Err: fmt.Errorf("failed to resolve setting %q: %w", nameSetting, err), Path: namePath}
+	}
+	if len(nameValues) == 0 || nameValues[0].Type() != cty.String {
+		return config.BpError{Err: fmt.Errorf("setting %q must be a string", nameSetting), Path: namePath}
+	}
+	name := nameValues[0].AsString()
```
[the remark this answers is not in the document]
agrawalkhushi18: Done

pkg/validators/registry.go
```diff
@@ -16,11 +16,12 @@ package validators
...
-	"conditional":       &ConditionalValidator{},
-	"conditional_regex": &ConditionalRegexValidator{},
+	"regex":              &RegexValidator{},
+	"allowed_enum":       &AllowedEnumValidator{},
+	"range":              &RangeValidator{},
+	"exclusive":          &ExclusiveValidator{},
+	"required":           &RequiredValidator{},
+	"conditional":        &ConditionalValidator{},
+	"conditional_regex":  &ConditionalRegexValidator{},
+	"service_account_id": &ServiceAccountIDValidator{},
```
AdarshK15: Implementation of the new validator is good, but AFAIU, `ServiceAccountIDValidator` is used only for `modules/project/service-account`. 
Instead of creating an entirely new generic validator for single use case, could we enhance existing validators like `regex` to support concatenation of multiple values?
If this is not feasible, adding a pre-conditional check in the module to validate this would also fail fast.

modules/project/service-account/metadata.yaml
```diff
@@ -34,3 +34,7 @@ ghpc:
       vars: [prefix]
       deprecated: true
     error_message: "Unsupported variable: 'prefix'. Please use 'deployment_name' instead."
+  - validator: service_account_id
+    inputs:
+      deployment_name: deployment_name
+      name: name
```
AdarshK15: Since we are defaulting them anyway, do we need to explicitly use these fields here?

AdarshK15: Nit: Should we add a custom error_message here?

pkg/validators/registry.go
agrawalkhushi18: Agreed. Thanks for the suggestion. I will leverage the existing validator 'regex' for adding this validation.

modules/project/service-account/metadata.yaml
agrawalkhushi18: Added an error message with the modified implementation.

pkg/validators/metadata_validators.go
```diff
@@ -25,8 +25,126 @@ import (
...
+	mod config.Module,
+	rule modulereader.ValidationRule,
+	group config.Group,
+	modIdx int,
+	re *regexp.Regexp,
+	patternRaw string,
+	optional bool,
+) error {
+	varsList, _ := parseStringList(rule.Inputs["vars"])
+	separator, _ := rule.Inputs["separator"].(string)
```
AdarshK15: Nit: can we use the `parseString` method here?

pkg/validators/metadata_validators_test.go
```diff
@@ -139,6 +139,327 @@ func TestRegexValidator(t *testing.T) {
...
+		wantErrSub  string
+	}{
+		{
+			name:    "passes_on_valid_id",
+			depName: cty.StringVal("my-dep"),
+			saName:  cty.StringVal("sa-name"),
+		},
+		{
+			name:       "fails_when_too_long",
+			depName:    cty.StringVal("a-very-long-deployment-name-that-exceeds"),
```
AdarshK15, line 275: Minor Nit: This is a good set of unit tests. Can we add one more test with the exact max length?

pkg/validators/metadata_validators.go
```diff
@@ -490,7 +608,7 @@ func (c *ConditionalRegexValidator) Validate(
 	}
 
 	dependentVal, depPath := getSettingWithFallback(bp, group, modIdx, mod, dependent)
-	if dependentVal.IsNull() || dependentVal == cty.NilVal || dependentVal.Type() != cty.String {
+	if dependentVal == cty.NilVal || !dependentVal.IsKnown() || dependentVal.IsNull() || dependentVal.Type() != cty.String {
```
AdarshK15: Instead of these additional checks, can we reuse the new method `resolveSettingToString`.

modules/project/service-account/metadata.yaml
```diff
@@ -34,3 +34,10 @@ ghpc:
       vars: [prefix]
       deprecated: true
     error_message: "Unsupported variable: 'prefix'. Please use 'deployment_name' instead."
+  - validator: regex
+    inputs:
+      vars: [deployment_name, name]
+      concat: true
+      separator: "-"
+      pattern: "^[a-z][-a-z0-9]{4,28}[a-z0-9]$"
```
AdarshK15, line 42: We are introducing excellent new features (concat, separator, allow_null) to the regex metadata validator, could we update the Regex Validator section in `docs/blueprint-validation.md`?

pkg/validators/metadata_validators.go
```diff
@@ -53,28 +171,25 @@ func (r *RegexValidator) Validate(
...
-	err = IterateRuleTargets(bp, mod, rule, group, modIdx, func(t Target) error {
-		return validateValues(t.Values, t.Path)
-	})
-	return err
+	optional := true
+	if v, ok := rule.Inputs["optional"]; ok {
+		if b, ok := v.(bool); ok {
+			optional = b
+		}
+	}
```
AdarshK15, line 188: Can we use `parseBoolInput` instead of this block?
```go
	optional, _ := parseBoolInput(rule.Inputs, "optional", true)
```

Nit: can we also update the same in the `IterateRuleTargets` method?

pkg/validators/metadata_validators.go
```diff
@@ -566,6 +684,20 @@ func getSettingWithFallback(
...
 
+func matchTrigger(actualVal, expectedVal cty.Value) bool {
+	if actualVal == cty.NilVal || actualVal.IsNull() {
+		if !isVarSet([]cty.Value{expectedVal}) {
+			return true
+		}
+		return expectedVal != cty.NilVal && expectedVal.IsKnown() && !expectedVal.IsNull() && expectedVal.Type() == cty.Bool && expectedVal.False()
+	}
+	if !actualVal.IsKnown() || !expectedVal.IsKnown() || expectedVal == cty.NilVal {
+		return false
```
AdarshK15, line 699: Good catch! Checking for `cty.NilVal` and `IsKnown()` will make validators more robust.

pkg/validators/metadata_validators.go
agrawalkhushi18: Fixed.

pkg/validators/metadata_validators.go
agrawalkhushi18: Fixed.

modules/project/service-account/metadata.yaml
agrawalkhushi18: Added, done.

pkg/validators/metadata_validators_test.go
agrawalkhushi18: Added.

pkg/validators/metadata_validators.go
agrawalkhushi18: done.

pkg/validators/metadata_validators.go
```diff
@@ -489,19 +608,19 @@ func (c *ConditionalRegexValidator) Validate(
 		return nil
 	}
 
-	dependentVal, depPath := getSettingWithFallback(bp, group, modIdx, mod, dependent)
-	if dependentVal.IsNull() || dependentVal == cty.NilVal || dependentVal.Type() != cty.String {
+	valStr, known, depPath, err := resolveSettingToString(bp, group, modIdx, mod, dependent, true, false)
```
AdarshK15: Nit: can we rename `valStr` to `dependentVal` for readability?

pkg/validators/metadata_validators.go
```diff
@@ -489,19 +608,19 @@ func (c *ConditionalRegexValidator) Validate(
...
+	if err != nil || !known {
```
AdarshK15: can we change this to throw error?
something like this:
```go
if err != nil {
	return err
}
if !known {
	return nil
}
```

AdarshK15 (approved):
LGTM

PR #6118, opened 2026-08-11
fix(daily-tests): Add --auto-approve to CHS integration tests

kvenkatachala333:
fixing CHS tests using auto approve flag

### DIFF
--- a/tools/cloud-build/daily-tests/builds/gke-a3-megagpu.yaml
+++ b/tools/cloud-build/daily-tests/builds/gke-a3-megagpu.yaml
@@ -70,13 +70,13 @@ steps:
 
     bash tools/add_ttl_label.sh "$${EXAMPLE_BP}"
     ansible-playbook tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml \
-        --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT} full_build_id=$BUILD_ID" \
+        --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT} full_build_id=$BUILD_ID chs_repo=$$CHS_REPO" \
         --extra-vars="@tools/cloud-build/daily-tests/tests/gke-a3-megagpu.yml" \
         --extra-vars="triage_gcs_bucket_override=$$TRIAGE_GCS_BUCKET" \
         --extra-vars="triage_project_number_override=$$TRIAGE_PROJECT_NUMBER" \
         --extra-vars="triage_invoker_sa_override=$$TRIAGE_INVOKER_SA" \
         --extra-vars="triage_cloud_run_url_override=$$TRIAGE_CLOUD_RUN_URL"
-  secretEnv: ['GCLUSTER_GCS_PATH', 'TRIAGE_GCS_BUCKET', 'TRIAGE_PROJECT_NUMBER', 'TRIAGE_INVOKER_SA', 'TRIAGE_CLOUD_RUN_URL']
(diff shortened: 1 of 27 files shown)


tools/cloud-build/daily-tests/tests/gke-a3-megagpu-onspot.yml
```diff
@@ -43,3 +43,9 @@ post_deploy_tests:
 - test-validation/test-gke-job.yml
 - test-validation/test-gke-a3-mega.yml
 - test-validation/test-gke-kueue-config.yml
+
+- test-validation/test-chs.yml
+scheduler: "gke"
+chs_suites:
```
sarthakag: Do we want to run both lite and standard suites? It would increase time for CTK tests by an hour atleast

tools/cloud-build/daily-tests/builds/gke-a3-megagpu.yaml
```diff
@@ -70,13 +70,13 @@ steps:
 
     bash tools/add_ttl_label.sh "$${EXAMPLE_BP}"
     ansible-playbook tools/cloud-build/daily-tests/ansible_playbooks/base-integration-test.yml \
-        --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT} full_build_id=$BUILD_ID" \
+        --user=sa_106486320838376751393 --extra-vars="project=${PROJECT_ID} build=$${BUILD_ID_SHORT} full_build_id=$BUILD_ID chs_repo=\$$CHS_REPO" \
```
sarthakag: \$$CHS_REPO - why do we need the backslash? I don't see that in other test runs

tools/cloud-build/daily-tests/tests/gke-a3-megagpu-onspot.yml
sarthakag: The existing tests did not run these suites.

tools/cloud-build/daily-tests/tests/slurm-v6-tpu.yml
```diff
@@ -32,6 +32,12 @@ login_node: "{{ slurm_cluster_name }}-slurm-login-*"
 controller_node: "{{ slurm_cluster_name }}-controller"
 post_deploy_tests:
 - test-slurm-v6-tpu.yml
+- test-validation/test-chs.yml
```
sarthakag: Why are we adding chs for slurm tpu?

sarthakag (changes requested):

tools/cloud-build/daily-tests/tests/slurm-v6-tpu.yml
kvenkatachala333: removed it

tools/cloud-build/daily-tests/builds/gke-a3-megagpu.yaml
kvenkatachala333: I've stripped the backslash across all the build definitions to align with standard Cloud Build syntax ($$CHS_REPO)

tools/cloud-build/daily-tests/tests/gke-a3-megagpu-onspot.yml
kvenkatachala333: The reason the existing tests ran the standard suite when undefined is because the underlying test-chs.yml playbook sets ['standard'] as the default fallback for the chs_suites variable. 

To maintain strict parity with existing tests and avoid the time penalty of running both suites, I have removed the explicit chs_suites declarations from all the newly added configurations. They will now all rely on the same default behavior and only run standard

sarthakag:
Thanks for the changes, can you also please trigger the modified PR tests?

PR #6127, opened 2026-08-12
 Optional Managed Instance Group (MIG) support for static compute nodes

arpit974:
## Overview

This pull request introduces native, optional support for provisioning **static Slurm compute nodes** using Google Cloud **Managed Instance Groups (MIGs)**.

Previously, static compute nodes were provisioned as standalone Compute Engine VMs using bulk insertion. With this change, users can choose to have static nodes managed by Google Cloud MIGs, unlocking automatic infrastructure self-healing and centralized group management while keeping Slurm job scheduling and MPI communication fully intact.

* **Intelligent `AUTO` Behavior**: By default (`provisioning_engine = "AUTO"`), the toolkit automatically routes DWS Flex-Start workloads to MIGs and standard compute workloads to Bulk Insert with zero user configuration. Existing clusters operate completely unchanged, while users who want automated hardware self-healing for their dedicated static nodesets can simply set `provisioning_engine = "MIG"`.

---

## Why Managed Instance Groups?

1. **Automatic Health Repair & Recovery**: If an underlying physical server experiences a hardware failure or a VM encounters an unrecoverable fault, Google Cloud's auto-healing automatically replaces the instance. Slurm detects this repair in real-time, avoids scheduling jobs on the recovering node, and brings it back to `idle` once healthy.
2. **Deterministic Naming & Job State**: Even inside a managed group, nodes retain their exact, sequential Slurm hostnames (e.g., `cluster-nodeset-0`, `cluster-nodeset-1`), maintaining full compatibility with Slurm configs and MPI communication.
3. **Large Cluster Scalability**: For clusters with large static nodesets exceeding 1,000 instances, the system automatically distributes nodes across multiple indexed groups (e.g., `<nodeset>-mig-0`, `<nodeset>-mig-1`) to remain well within Google Cloud's recommended group limits.
4. **100% Backward Compatible**: The default setting remains `"AUTO"`, meaning existing blueprints, dynamic auto-scaling nodes, and standard bulk-insert workflows continue to function with zero changes.

---

## How It Works

* **The `AUTO` User Scenario**: Users do not need to configure anything by default. `"AUTO"` preserves standard bulk provisioning for general workloads and automatically engages MIGs for DWS Flex-Start. Users explicitly specify `"MIG"` only when they want GCE-managed auto-repair on their static compute nodes.
* **Declarative MIGs with Per-Instance Configs**: MIGs are declared in Terraform with a target size of `0`. When the cluster boots, the Slurm controller scripts call GCE APIs to add instances to the group using deterministic names and configurations.
* **Smart Power & Suspend Management**: When Slurm powers down or suspends static nodes, `suspend.py` removes them from the group. When resumed, `resume.py` recreates them in the group with the current instance template and binds the Slurm Job ID to instance metadata.
* **Auto-Repair Sync**: `slurmsync.py` checks whether GCE is actively repairing a VM. If a node is in a `REPAIRING` state, Slurm marks it as `DOWN` with the reason `"MIG Auto-Healing instance repair in progress"` so no jobs are sent to it until it recovers.
* **FQDN & Hostname Resilience**: Suspension and status synchronization seamlessly support both short hostnames and Fully Qualified Domain Names (FQDNs) used in private DNS or OS Login configurations.

---

## Summary of Changes

* **Terraform Modules**:
  * `schedmd-slurm-gcp-v6-nodeset`: Added `provisioning_engine` input variable with validation. Added safety guardrail ensuring pure static sizing when `MIG` is selected.
  * `schedmd-slurm-gcp-v6-partition`: Passed through `provisioning_engine` configuration to partition definitions.
  * `schedmd-slurm-gcp-v6-controller`: Added declarative regional MIG resources, automatic multi-group indexing, and safety preconditions to prevent unsupported engine mutations on active nodes.
* **Controller Python Scripts**:
  * `resume.py`: Added `resume_mig_nodes()` to provision per-instance configurations into the target MIG.
  * `suspend.py`: Added `suspend_mig_nodes()` with FQDN short-hostname normalization to remove instances via `deleteInstances`.
  * `slurmsync.py`: Added GCE Auto-Repair status detection to mark repairing nodes `DOWN` in Slurm.
  * `util.py`: Added MIG lookup helpers (`is_node_mig`, `node_mig_name`, `get_mig_instances`, and group size expansion math).
* **Automated Unit Tests**:
  * Added unit test coverage in `test_util.py` and `test_resume.py` covering multi-MIG indexing, FQDN node resolution, API schema variations, and auto-repair status reporting.

---

### 1. Automated Test Suites
* All **183 unit tests passed** in Python 3.13 (`pytest-check`).
* Pre-commit checks passed cleanly across the entire codebase (`mypy`, `terraform_fmt`, `terraform_validate`, `codespell`, `duplicate-diff`, `module-label-check`).

### 2. Live GCP Deployment & Workload Testing
Verified on a live multi-partition Slurm cluster deployed on Google Cloud covering:
* **Pure Static On-Demand MIGs**: Executed parallel multi-node MPI jobs (`srun`, `sbatch`) with `ExitCode 0:0`.
* **Pure Static Spot MIGs**: Verified spot VM creation, state persistence, and job execution with `ExitCode 0:0`.
* **Multi-MIG Scaling**: Verified jobs co-scheduling across multiple distinct MIGs.
* **Hybrid Partitions**: Successfully ran concurrent jobs across a mixed partition containing both static MIG nodes and dynamic bulk-insert nodes.
* **Dynamic Scale-to-Zero Partitions**: Verified standard dynamic burst nodes remain 100% operational on `BULK_INSERT`.
* **DWS Flex Partitions**: Verified Flex-Start dynamic provisioning continues to operate normally.

### 3. Fault Tolerance & Disaster Recovery
* **Simulated Hardware Failure**: Deleted an active static node out-of-band directly in Google Cloud. GCE automatically triggered an auto-repair recreation, `slurmsync` kept Slurm informed, the node safely transitioned back to `idle`, and subsequent parallel MPI jobs ran with zero errors.
* **Controller Reboot**: Hard-rebooted the controller VM (`sudo reboot`). Services restarted cleanly, reconnected to all managed instances, and resumed job scheduling without manual intervention.
* **Clean Teardown**: Verified `gcluster destroy` cleans up all instances, MIGs, disks, and templates with 0 orphaned cloud resources.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -92,12 +92,51 @@ module "nodeset_cleanup" {
...
+  project = var.project_id
+  region  = var.region
+}
+
+resource "google_compute_region_instance_group_manager" "nodeset_mig" {
+  for_each           = var.provisioning_engine == "MIG" ? local.nodeset_map : {}
+  name               = "${local.slurm_cluster_name}-${each.value.nodeset_name}-mig"
+  base_instance_name = each.value.nodeset_name
+  region             = coalesce(each.value.region, var.region)
+  target_size        = each.value.node_count_static
```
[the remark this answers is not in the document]
arpit974: done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -92,12 +92,51 @@ module "nodeset_cleanup" {
...
+  }
+
+  distribution_policy_zones        = length(coalesce(each.value.zone_policy_allow, [])) > 0 ? each.value.zone_policy_allow : (var.zone != null ? [var.zone] : null)
+  distribution_policy_target_shape = "ANY_SINGLE_ZONE"
+
+  update_policy {
+    type                         = "OPPORTUNISTIC"
+    minimal_action               = "REPLACE"
+    instance_redistribution_type = "NONE"
+    max_surge_fixed              = length(coalesce(each.value.zone_policy_allow, (var.zone != null ? [var.zone] : data.google_compute_zones.available.names)))
```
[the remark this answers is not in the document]
arpit974: fixed

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -84,13 +84,83 @@ def delete_instances(instances):
...
+                if zone:
+                    zone_name = zone.split("/")[-1]
+                    links.append(f"zones/{zone_name}/instances/{node}")
+                else:
+                    nodeset = lkp.node_nodeset(node)
+                    zone_name = getattr(nodeset, "zone", None)
+                    if zone_name:
+                        links.append(f"zones/{zone_name}/instances/{node}")
+                    else:
+                        links.append(f"zones/us-central1-a/instances/{node}")
```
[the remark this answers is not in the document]
arpit974: addressed

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -343,6 +381,8 @@ def resume_nodes(nodes: List[str], resume_data: Optional[ResumeData]):
             tpu_chunks.append(chunk.nodes)
         elif lkp.is_flex_node(model):
             flex_chunks.append(chunk)
+        elif lkp.is_mig_engine():
+            mig_chunks.append(chunk)
```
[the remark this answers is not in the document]
arpit974: addressed.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -84,13 +84,83 @@ def delete_instances(instances):
...
     lkp = lookup()
     other_nodes, tpu_nodes = util.separate(lkp.node_is_tpu, nodes)
     bulk_nodes, flex_nodes = util.separate(lkp.is_flex_node, other_nodes)
 
     mig_flex.suspend_flex_nodes(flex_nodes, lkp)
-    delete_instances(bulk_nodes)
+    if lkp.is_mig_engine():
+        suspend_mig_nodes(bulk_nodes, lkp)
+    else:
+        delete_instances(bulk_nodes)
```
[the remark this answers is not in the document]
arpit974: addressed.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -84,13 +84,83 @@ def delete_instances(instances):
...
+            project=lkp.project,
+            region=region,
+            instanceGroupManager=mig_name,
+            body={
+                "instances": links,
+                "decreaseTargetSize": True
+            }
+        )
+        try:
+            res = util.execute_with_futures(lambda r: r.execute(), [req])
```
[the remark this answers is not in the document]
arpit974: addressed.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -303,6 +303,44 @@ def chunk_nodes(nodes: List[str]):
...
+
+    # 2. Per-Instance Config (PIC) - Lightweight instance name binding for static Slurm hostnames
+    pic_instances = [{"name": node} for node in nodes]
+    pic_req = lkp.compute.regionInstanceGroupManagers().createInstances(
+        project=lkp.project,
+        region=region,
+        instanceGroupManager=mig_name,
+        body={"instances": pic_instances}
+    )
+    res = execute_with_futures(lambda req: req.execute(), [pic_req])
```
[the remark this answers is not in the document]
arpit974: addressed.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -303,6 +303,44 @@ def chunk_nodes(nodes: List[str]):
...
+    template_link = getattr(nodeset, "instance_template", None)
+    if template_link:
+        try:
+            aic_req = lkp.compute.regionInstanceGroupManagers().setInstanceTemplate(
+                project=lkp.project,
+                region=region,
+                instanceGroupManager=mig_name,
+                body={"instanceTemplate": template_link}
+            )
+            execute_with_futures(lambda req: req.execute(), [aic_req])
```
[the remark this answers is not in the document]
arpit974: done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -92,12 +92,51 @@ module "nodeset_cleanup" {
...
+  name               = "${local.slurm_cluster_name}-${each.value.nodeset_name}-mig"
+  base_instance_name = each.value.nodeset_name
+  region             = coalesce(each.value.region, var.region)
+  target_size        = each.value.node_count_static
+
+  version {
+    instance_template = module.slurm_nodeset_template[each.value.nodeset_name].self_link
+  }
+
+  distribution_policy_zones        = length(coalesce(each.value.zone_policy_allow, [])) > 0 ? each.value.zone_policy_allow : (var.zone != null ? [var.zone] : null)
```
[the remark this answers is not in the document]
arpit974: done

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -683,3 +683,13 @@ variable "machine_configs" {
   type        = any
   default     = {}
 }
+
+variable "provisioning_engine" {
+  description = "Compute node provisioning engine: 'AUTO', 'MIG', or 'BULK_INSERT'."
```
Neelabh94: should we have a validation in metadata.yaml ensuring a fail fast for this?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -303,6 +303,80 @@ def chunk_nodes(nodes: List[str]):
...
+            try:
+                mig = lkp.get_mig(lkp.project, region, mig_name)
+                if mig:
+                    current_template = (
+                        mig.get("instanceTemplate")
+                        or (mig.get("versions", [{}])[0].get("instanceTemplate") if mig.get("versions") else None)
+                    )
+                    if current_template and util.trim_self_link(current_template) != util.trim_self_link(template_link):
+                        log.info(f"Updating MIG {mig_name} template: {current_template} -> {template_link}")
+                        aic_req = lkp.compute.regionInstanceGroupManagers().setInstanceTemplate(
```
Neelabh94: Because the regional MIG is declared in Terraform using a `version { ... }` block, Google Compute Engine treats it as a versioned MIG. Calling `setInstanceTemplate` on a MIG configured with `versions` is rejected by the GCE API with `"The resource cannot have both versions and instanceTemplate"`.

To update the template on a versioned MIG, consider using `patch` instead:
```python
aic_req = lkp.compute.regionInstanceGroupManagers().patch(
    project=lkp.project,
    region=region,
    instanceGroupManager=mig_name,
    body={"versions": [{"instanceTemplate": template_link}]},
)
ensure_execute(aic_req)
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -93,12 +106,83 @@ module "nodeset_cleanup" {
...
+  }
+
+  distribution_policy_zones        = length(try(each.value.zone_policy_allow, [])) > 0 ? each.value.zone_policy_allow : (var.zone != null ? [var.zone] : null)
+  distribution_policy_target_shape = "ANY_SINGLE_ZONE"
+
+  update_policy {
+    type                         = "OPPORTUNISTIC"
+    minimal_action               = "REPLACE"
+    instance_redistribution_type = "NONE"
+    max_surge_fixed              = length(try(each.value.zone_policy_allow, [])) > 0 ? length(each.value.zone_policy_allow) : 1
```
Neelabh94: When `zone_policy_allow` is empty and `var.zone` is null, `distribution_policy_zones` defaults to `null` (spanning all zones in the region). GCE requires that for a regional MIG, `maxSurge` must be either `0` or greater than or equal to the number of zones in the group, so `max_surge_fixed = 1` will fail API validation.

Furthermore, because Slurm compute nodes use fixed deterministic names, instances cannot surge. We recommend setting `max_surge_fixed = 0` and `replacement_method = "RECREATE"`, similar to the controller MIG in `controller.tf`:
```terraform
update_policy {
  type                         = "OPPORTUNISTIC"
  minimal_action               = "REPLACE"
  replacement_method           = "RECREATE"
  instance_redistribution_type = "NONE"
  max_surge_fixed              = 0
  max_unavailable_fixed        = length(try(each.value.zone_policy_allow, [])) > 0 ? length(each.value.zone_policy_allow) : 1
}
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -303,6 +303,80 @@ def chunk_nodes(nodes: List[str]):
...
+                    "metadata": {
+                        "slurm_job_id": str(excl_job_id)
+                    }
+                }
+            pic_instances.append(inst)
+
+        for chunk in chunked(pic_instances, n=ZONAL_MIG_SIZE_LIMIT):
+            chunk_nodes = [item["name"] for item in chunk]
+            try:
+                pic_req = lkp.compute.regionInstanceGroupManagers().createInstances(
```
Neelabh94, line 445: If any node in `mig_nodes` already exists in the MIG (e.g. after a controller restart or an API retry), `createInstances` rejects the entire chunk with `ALREADY_EXISTS`. Because the exception is caught at the chunk level, `down_nodes_notify_jobs` marks all nodes in that chunk `DOWN` in Slurm.

Similar to the defensive approach in `suspend_mig_nodes`, consider querying `lkp.get_mig_instances()` first and filtering out instances that are already members of the MIG before calling `createInstances`.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -683,3 +683,13 @@ variable "machine_configs" {
   type        = any
   default     = {}
 }
+
+variable "provisioning_engine" {
```
Neelabh94, line 687: Should we also add an `allowed_enum` validator for `provisioning_engine` in `community/modules/compute/schedmd-slurm-gcp-v6-nodeset/metadata.yaml`?
```yaml
- validator: allowed_enum
  inputs:
    vars: [provisioning_engine]
    allowed:
    - AUTO
    - MIG
    - BULK_INSERT
    allow_null: false
  error_message: "Allowed values for provisioning_engine are 'AUTO', 'MIG', or 'BULK_INSERT'."
```
This ensures typos fail fast during `gcluster validate` before Terraform runs.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1776,6 +1776,42 @@ def node_is_tpu(self, node_name=None):
     def nodeset_is_tpu(self, nodeset_name=None) -> bool:
         return self.cfg.nodeset_tpu.get(nodeset_name) is not None
 
+    def is_mig_engine(self) -> bool:
```
Neelabh94, line 1814: `is_mig_engine()` checks `getattr(self.cfg, "provisioning_engine", "BULK_INSERT")`, but `provisioning_engine` is configured per nodeset in Terraform rather than as a top-level attribute on `self.cfg`. In production, `self.cfg.provisioning_engine` is not set and will always return `False`.

Suggest relying directly on `nodeset.mig_name is not None` or `nodeset.provisioning_engine == "MIG"` inside `is_nodeset_mig()` and removing the global fallback.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -343,13 +417,24 @@ def resume_nodes(nodes: List[str], resume_data: Optional[ResumeData]):
...
     for chunk in flex_chunks:
-        mig_flex.resume_flex_chunk(chunk.nodes, chunk.excl_job_id, lkp, chunk.placement_group)
+        try:
+            mig_flex.resume_flex_chunk(chunk.nodes, chunk.excl_job_id, lkp, chunk.placement_group)
+        except Exception:
+            log.exception(f"failed to resume flex chunk {chunk.nodes}")
+
+    for chunk in mig_chunks:
+        try:
+            resume_mig_nodes(chunk.nodes, chunk.excl_job_id, lkp, resume_data)
```
Neelabh94, line 531: For MIG nodes, `chunk.placement_group` is not passed to `resume_mig_nodes()`. As a result, if a user specifies `enable_placement = true` on a MIG nodeset, compact placement is silently ignored at runtime.

Could we either add a precondition in `outputs.tf` disallowing `enable_placement = true` when `provisioning_engine = "MIG"`, or attach a static `resource_policy` to `module.slurm_nodeset_template` in `partition.tf` when placement is enabled for MIG nodesets?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1976,18 +2012,48 @@ def _get_reservation(self, project: str, zone: str, name: str) -> Any:
...
+            )
+            res = ensure_execute(req)
+            all_instances.extend(res.get("managedInstances", []) if isinstance(res, dict) else [])
+            page_token = res.get("nextPageToken") if isinstance(res, dict) else None
+            if not page_token:
+                break
+        return {"managedInstances": all_instances}
+
+    @lru_cache()
+    def get_mig_repairing_instances(self, project: str, region: str, self_link: str) -> Set[str]:
```
Neelabh94, line 2078: During instance creation or state transitions, GCP may return `"instance": null` in `managedInstances`:
1. In `util.py:2043`, `m_inst.get("instance", "")` will return `None` if the key exists with a `null` value, causing `None.split("/")` to raise an `AttributeError`. Consider using `(m_inst.get("instance") or "").split("/")[-1]`.
2. Similarly in `suspend.py:121`, `if "instance" in item:` evaluates to `True` even when the value is `None`. Suggest changing to `if item.get("instance"):`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/slurmsync.py
```diff
@@ -292,7 +292,19 @@ def get_node_action(nodename: str) -> NodeAction:
...
+    short_nodename = nodename.split(".")[0]
+    inst = lkp.instance(short_nodename)
+
+    # For MIG compute nodes, detect active GCE Auto-Healing repairs
+    if lkp.is_node_mig(nodename):
+        mig_name = lkp.node_mig_name(nodename)
+        region = lkp.node_region(nodename)
+        try:
+            if short_nodename in lkp.get_mig_repairing_instances(lkp.project, region, mig_name):
+                return NodeActionDown(reason="MIG Auto-Healing instance repair in progress")
```
Neelabh94: While a node is in `REPAIRING` state, `get_node_action()` will return `NodeActionDown` on every 60-second cycle of `slurmsync.py`, repeatedly executing `scontrol update nodename=... state=down`.

Consistent with the convention used elsewhere in this file (e.g. lines 226, 241, 321, 337), consider guarding this with `if state is not None and state.base != "DOWN":` so that `NodeActionDown` is only emitted when transitioning state.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -303,6 +303,80 @@ def chunk_nodes(nodes: List[str]):
...
+        for chunk in chunked(pic_instances, n=ZONAL_MIG_SIZE_LIMIT):
+            chunk_nodes = [item["name"] for item in chunk]
+            try:
+                pic_req = lkp.compute.regionInstanceGroupManagers().createInstances(
+                    project=lkp.project,
+                    region=region,
+                    instanceGroupManager=mig_name,
+                    body={"instances": chunk}
+                )
+                res = ensure_execute(pic_req)
```
Neelabh94, line 451: `createInstances` returns an asynchronous GCP Operation (`res`). If the operation fails asynchronously (e.g., due to quota exhaustion or stockout in the zone), `resume.py` currently exits successfully, causing Slurm to wait for the full `ResumeTimeout` (typically 5–10 minutes) before marking nodes not responding.

Consider calling `util.wait_for_operation(res)` so that asynchronous creation errors can be caught and handled immediately.

Neelabh94 (commented):
Thanks for putting together this PR to support Managed Instance Groups for static compute nodes! This is a great addition for enabling hardware auto-repair and self-healing for long-running Slurm workloads.

Overall the architecture is well thought out. I've left a few inline comments on areas around GCE API compatibility, validation, idempotency, and edge-case handling.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/README.md
```diff
@@ -242,6 +242,7 @@ modules. For support with the underlying modules, see the instructions in the
 | <a name="input_placement_max_distance"></a> [placement\_max\_distance](#input\_placement\_max\_distance) | Maximum distance between nodes in the placement group. Requires enable\_placement to be true. Values must be supported by the chosen machine type. | `number` | `null` | no |
 | <a name="input_preemptible"></a> [preemptible](#input\_preemptible) | Should use preemptibles to burst. | `bool` | `false` | no |
 | <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Project ID to create resources in. | `string` | n/a | yes |
+| <a name="input_provisioning_engine"></a> [provisioning\_engine](#input\_provisioning\_engine) | Compute node provisioning engine: 'AUTO', 'MIG', or 'BULK\_INSERT'. Note: When using 'MIG', 'node\_count\_dynamic\_max' must be explicitly set to 0. | `string` | `"AUTO"` | no |
```
Neelabh94: Please consider adding a brief section in this README describing the new `provisioning_engine = "MIG"` option, its benefits (GCE hardware auto-repair for static compute nodes), and the prerequisite that `node_count_dynamic_max` must be set to `0`.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/partition.tf
```diff
@@ -93,12 +106,104 @@ module "nodeset_cleanup" {
...
+        nodeset            = ns
+      }
+    }
+    if local.nodeset_resolved_engine[name] == "MIG" && !ns.dws_flex.enabled
+  ]...)
+
+  # Filter out null or empty strings from zone policy allow sets
+  nodeset_mig_zones = {
+    for k, mig in local.nodeset_migs : k => [
+      for z in try(coalesce(mig.zone_policy_allow, []), []) : z if z != null && z != ""
```
Neelabh94: Nit: `try(coalesce(mig.zone_policy_allow, []), [])` can be simplified with a ternary:
```hcl
for z in (mig.zone_policy_allow != null ? mig.zone_policy_allow : []) : z if z != null && z != ""
```
This is more idiomatic HCL and avoids using `try()` as a fallback for `coalesce()` when passed empty/null arguments.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1811,6 +1811,42 @@ def node_is_tpu(self, node_name=None):
     def nodeset_is_tpu(self, nodeset_name=None) -> bool:
         return self.cfg.nodeset_tpu.get(nodeset_name) is not None
 
+    def is_mig_engine(self) -> bool:
```
Neelabh94: `is_mig_engine()` appears to be vestigial / dead code. `provisioning_engine` is configured strictly per nodeset (`nodeset.provisioning_engine`) rather than at the cluster-level `cfg`.

Can we remove `is_mig_engine()` and have `is_nodeset_mig(nodeset_name)` check the nodeset's provisioning engine directly without falling back to `self.is_mig_engine()`?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_util.py
```diff
@@ -839,3 +839,303 @@ def test_compute_service_custom_universe_domain(mocker):
     assert kwargs.get("static_discovery") is True
     assert kwargs.get("client_options").api_endpoint == "https://compute.apis-sovereign.goog/compute/beta/"
     assert kwargs.get("client_options").universe_domain == "apis-sovereign.goog"
+
+
+def test_is_mig_engine():
```
Neelabh94: Please remove `test_is_mig_engine()` along with `is_mig_engine()` in `util.py`.

Neelabh94 (approved):

arpit974:
### CI Status: `PR-test-ml-g4-onspot-slurm` Flakiness & Root Cause

the failure on `PR-test-ml-g4-onspot-slurm` (Cloud Build `6acbae5e-c535-4662-96e3-f5f1b6bab8c2`) is an environmental spot capacity stockout unrelated to this pr:

1. **No Code / Blueprint Touch**: PR does not modify, the G4 daily test definitions, or any GPU hardware mappings.
2. **Strict Opt-In Backward Compatibility**: In PR, `provisioning_engine` defaults to `AUTO`, which resolves unconfigured nodesets to `BULK_INSERT` (`partition.tf:110-118`). `g4_nodeset` creates 0 MIG resources and executes the identical legacy `bulkInsert()` code path.
3. **All 10 Core Slurm Suites Passed**: `PR-test-slurm-gcp-v6-rocky8`, `debian`, `ubuntu`, `topology`, `reconfig-size`, `ssd`, `simple-job-completion`, `hpc-enterprise-slurm`, `ml-slurm`, and `htc-slurm` all passed with exit code 0.
4. **Transient Spot Capacity Stockout**: The test timed out after 72 minutes looping in `wait_for_available_zone.sh` due to `g4-standard-48` spot capacity constraints in the test project (a known condition addressed in recent CI commits #6205 and #6208).

PR #6135, opened 2026-08-12
fix(gke): enhance MTC CRD verification and storage ramdisk path rules

Neelabh94:
Depends on #6134. Enhances MTC CRD verification and storage ramdisk path validation.

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -648,7 +648,7 @@ func (g *GKEOrchestrator) verifyCheckpointConfigurationCR(job *orchestrator.JobD
 }
 
 func (g *GKEOrchestrator) validateMTCConfig(job *orchestrator.JobDefinition) error {
-	if !job.GKEMTCEnabled {
+	if job == nil || !job.GKEMTCEnabled {
 		return nil
 	}
 
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -603,6 +603,78 @@ func (g *GKEOrchestrator) isSystemPool(np gkeJobNodePool) bool {
 	return false
 }
 
+func (g *GKEOrchestrator) getTargetNamespace(job *orchestrator.JobDefinition) (string, error) {
+	if job == nil {
+		return "", fmt.Errorf("job definition cannot be nil")
+	}
+
+	if job.GKENamespace != "" {
+		return job.GKENamespace, nil
+	}
+
+	return g.getCurrentNamespace(job.ClusterName, job.ClusterLocation, job.ProjectID)
+}
+
+func (g *GKEOrchestrator) verifyCheckpointConfigurationCR(job *orchestrator.JobDefinition, docRemediationMsg string) error {
+	client, err := g.getDynamicClient()
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/orchestrator/gke/storage.go
```diff
@@ -86,14 +118,43 @@ func (sm *StorageManager) ValidateMounts(mounts []string) error {
 			return err
 		}
 
+		cleanDest := normalizeMountPath(dest)
```
agrawalkhushi18, line 121: In your strict new ValidateRamdiskDir function, you check that the path starts with a / (an absolute path) using strings.HasPrefix. This is a great safety measure.

But, we can add this same check for regular volume mounts here in ValidateMounts. If a user accidentally types a relative path like my-drive;data/folder (missing the leading /), it sneaks past your new system-folder security checks. The job will submit, but will crash asynchronously later because Kubernetes requires paths to start with /.

Should we add a quick strings.HasPrefix(cleanDest, "/") here to catch typos instantly?

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -603,6 +603,78 @@ func (g *GKEOrchestrator) isSystemPool(np gkeJobNodePool) bool {
...
+	}
+	if len(checkpointConfigList.Items) == 0 {
+		return fmt.Errorf("Multi-Tier Checkpointing (MTC) requires a CheckpointConfiguration resource to be deployed in the target namespace (%s). %s", targetNamespace, docRemediationMsg)
+	}
+
+	return nil
+}
+
+func (g *GKEOrchestrator) validateMTCConfig(job *orchestrator.JobDefinition) error {
+	if !job.GKEMTCEnabled {
```
agrawalkhushi18: Just a small defensive check here.

In validateTargetNamespaceExists function, we have a really safe if job == nil check at the top. But here, we immediately check !job.GKEMTCEnabled. If job ever happens to be passed as nil (like in a test or a future refactor), this will trigger a fatal crash/panic.
Should we add `if job == nil` guard clause to the top of this function too just to be safe?

pkg/orchestrator/gke/storage.go
Neelabh94: Thanks for the suggestion! 
I checked `pkg/orchestrator/gke/storage.go`, and `parseSingleVolume()` actually already enforces `!strings.HasPrefix(dest, "/")` on line 191:
```go
if src == "" || dest == "" || !strings.HasPrefix(dest, "/") {
    return "", "", "", false, missingDestOrFormatErr(vStr)
}
```

Because ValidateMounts() calls parseSingleVolume() for every mount string, any destination path missing a leading / (e.g. my-drive;data/folder) is already caught synchronously with a format error at submission time before reaching Kubernetes.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: * In the main workflow, `job` is guaranteed to be non-nil by `SubmitJob` / `initializeJobSubmission` before calling `validateMTCConfig`. 
* In contrast, getTargetNamespace has an explicit if `job == nil` check because it is a generic helper called across multiple disparate callers and standalone verification utilities.

However, adding a defensive `if job == nil { return nil }` guard clause at the top of `validateMTCConfig` is a harmless safety check. I'll add that quick guard clause!

agrawalkhushi18 (approved):

PR #6144, opened 2026-08-13
feat: add turnkey Slurm support for Google Cloud Dedicated environments

juntangc:
## Description
This PR introduces end-to-end turnkey support for deploying Slurm clusters on C3 instances within Google Cloud Dedicated environments.

### Key Changes:
1. **Cloud NAT Module (`modules/network/cloud-nat`):** Added a reusable Cloud NAT module for private subnets.
2. **Sovereign Cloud Endpoint Resolution (`util.py`):** Added custom discovery URLs for GCE and GCS in alternate universe domains.
3. **Automated SchedMD Setup (`setup.py`):**
   - Automated Slurm configuration distribution (`slurm.conf`, `cloud.conf`) via shared NFS mount (`/home/`).
   - Automated static compute node provisioning on controller boot (`resume.py`).
   - Automated SELinux and MariaDB state directory permission initialization.
   - Guarded optional services for Rocky Linux 8 / Slurm 20.11 compatibility.
4. **NFS Permissions (`install-nfs-server.sh.tpl`):** Exported `/exports` with `1777` sticky permissions for seamless user SSH key generation.
5. **Example Blueprint (`examples/hpc-slurm-c3-tpc/hpc-slurm-c3-tpc.yaml`):** Added a tested, reproducible turnkey blueprint.

### Verification:
- All unit tests pass (`go test ./pkg/...`).
- Deployed full cluster in `u-germany-northeast1-a` on C3 instances and verified multi-node MPI execution with `srun -N2 hostname`.

### DIFF
--- a/community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
+++ b/community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
@@ -1,0 +1,273 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+blueprint_name: hpc-c3-slurm-google-cloud-dedicated
+vars:
+  project_id: # <project-id>
+  deployment_name: hpc-slurm
+  region: us-central1
+  zone: us-central1-a
+  universe_domain: googleapis.com # apis-<location>.goog
+  service_account_email: default # <account-id>-compute@developer.gserviceaccount.com
+
+  # Image Build Settings
+  source_image_family: rocky-linux-8-optimized-gcp
+  source_image_project: rocky-linux-cloud
+  image_build_machine_type: c3-standard-4
+  build_slurm_from_git_ref: 6.12.1
+  built_instance_image:
+    family: slurm-c3-rocky8
+    project: $(vars.project_id)
+
+terraform_providers:
+  google:
+    source: hashicorp/google
+    version: '>= 5.45.0'
+    configuration:
+      project: $(vars.project_id)
+      region: $(vars.region)
+      zone: $(vars.zone)
+      universe_domain: $(vars.universe_domain)
+  google-beta:
+    source: hashicorp/google-beta
+    version: '>= 5.45.0'
+    configuration:
+      project: $(vars.project_id)
+      region: $(vars.region)
+      zone: $(vars.zone)
+      universe_domain: $(vars.universe_domain)
+
+deployment_groups:
+- group: setup
+  modules:
+  - id: network
+    source: modules/network/pre-existing-vpc
+    settings:
+      network_name: default
+      subnetwork_name: default
+
+  - id: slurm-build-script
+    source: modules/scripts/startup-script
+    settings:
+      install_ansible: false
+      runners:
+      - type: shell
+        destination: prep-for-slurm-build.sh
+        content: |
+          #!/bin/bash
+          set -e -o pipefail
+          # Ensure SELinux allows NFS home directories
+          setsebool -P use_nfs_home_dirs 1 || true
+          # Slurm build will install and set python3.12 as the default python3
+          dnf install -y python3.12 python3.12-pip perl-ExtUtils-MakeMaker perl-devel
+          python3.12 -m pip install pip --upgrade
+          python3.12 -m pip install ansible==8.7.0
+          export PATH=/usr/local/bin:$PATH
+          ansible --version
+          n=0
+          until [ "$n" -ge 3 ] || ( { dnf install -y --disablerepo=ciq-sigcloud-next nfs-utils 2>/dev/null || dnf install -y nfs-utils; } && ansible-galaxy role install googlecloudplatform.google_cloud_ops_agents ); do
+            ((n=n+1))
+            sleep 5
+          done
+      - type: data
+        destination: /var/tmp/slurm_vars.json
+        # Add or update slurm-gcp flags as required, https://github.com/GoogleCloudPlatform/slurm-gcp/blob/master/ansible/playbook.yml
+        content: |
+          {
+            "reboot": false,
+            "install_cuda": false,
+            "nvidia_version": "latest",
+            "install_ompi": true,
+            "install_lustre": false,
+            "install_gcsfuse": true,
+            "allow_kernel_upgrades": false,
+            "monitoring_agent": "cloud-ops",
+            "install_managed_lustre": false
+          }
+
+      - type: shell
+        destination: install_slurm.sh
+        content: |
+          #!/bin/bash
+          set -e -o pipefail
+          dnf update -y
+          dnf install -y epel-release git ansible-core
+          ansible-pull -U https://github.com/GoogleCloudPlatform/slurm-gcp -C $(vars.build_slurm_from_git_ref) \
+              -i localhost, --limit localhost --connection=local \
+              -e @/var/tmp/slurm_vars.json ansible/playbook.yml
+
+- group: build-slurm
+  modules:
+  - id: custom-slurm-image
+    source: modules/packer/custom-image
+    kind: packer
+    use: [network, slurm-build-script]
+    settings:
+      source_image_project_id: [$(vars.source_image_project)]
+      source_image_family: $(vars.source_image_family)
+      machine_type: $(vars.image_build_machine_type)
+      image_family: $(vars.built_instance_image.family)
+      disk_type: hyperdisk-balanced
+      zone: $(vars.zone)
+      omit_external_ip: false
+      image_licenses: []
+      state_timeout: 30m
+
+- group: cluster
+  modules:
+  - id: homefs
+    source: community/modules/file-system/nfs-server
+    use: [network]
+    settings:
+      local_mounts: [/home, /opt/apps]
+      machine_type: $(vars.image_build_machine_type)
+      boot_disk_type: hyperdisk-balanced
+      type: hyperdisk-balanced
+      disk_size: 100
+      instance_image:
+        family: $(vars.built_instance_image.family)
+        project: $(vars.project_id)
+
+  - id: ssh_config
+    source: modules/scripts/startup-script
+    settings:
+      install_ansible: false
+      runners:
+      - type: shell
+        destination: setup_non_oslogin_ssh.sh
+        content: |
+          #!/bin/bash
+          # 1. SELinux & Crypto Policy for Rocky Linux 8/9
+          setsebool -P use_nfs_home_dirs 1 || true
+          update-crypto-policies --set DEFAULT:SHA1 || true
+
+          # 2. Populate metadata keys and secure inter-node keypair on NFS /home
+          METADATA="http://metadata.google.internal/computeMetadata/v1"
+          PROJ_KEYS=`curl -s -f -H "Metadata-Flavor: Google" "$METADATA/project/attributes/ssh-keys" 2>/dev/null || true`
+          INST_KEYS=`curl -s -f -H "Metadata-Flavor: Google" "$METADATA/instance/attributes/ssh-keys" 2>/dev/null || true`
+          KEYS=`printf "%s\n%s\n" "$PROJ_KEYS" "$INST_KEYS" | grep -v '^[[:space:]]*$' | sort -u`
+
+          for user in `echo "$KEYS" | cut -d: -f1 | sort -u`; do
+            [ -z "$user" ] && continue
+
+            # Create group and user with deterministic UID/GID to ensure consistency across the cluster
+            if ! id "$user" &>/dev/null; then
+              UID_HASH=`printf "%s" "$user" | cksum | cut -d' ' -f1`
+              USER_UID=`expr 2000 + $UID_HASH % 58000`
+              groupadd -g "$USER_UID" "$user" || true
+              useradd -u "$USER_UID" -g "$USER_UID" -m -s /bin/bash "$user"
+            fi
+            mkdir -p "/home/$user/.ssh"
+            chown -R "$user:$user" "/home/$user"
+            chmod 700 "/home/$user/.ssh"
+
+            # Safely serialize key generation & authorized_keys rebuilding across the cluster
+            (
+              flock 9
+
+              # Generate ED25519 cluster key natively
+              if [ ! -f "/home/$user/.ssh/id_ed25519" ]; then
+                ssh-keygen -t ed25519 -N "" -f "/home/$user/.ssh/id_ed25519"
+              fi
+
+              # Rebuild authorized_keys atomically on the exact same NFS filesystem (prevents O_TRUNC zeroing)
+              TMP_KEYS=`mktemp "/home/$user/.ssh/authorized_keys.XXXXXX"`
+              if [ -n "$TMP_KEYS" ] && [ -f "$TMP_KEYS" ]; then
+                echo "$KEYS" | grep "^$user:" | cut -d: -f2- > "$TMP_KEYS"
+                [ -f "/home/$user/.ssh/id_ed25519.pub" ] && cat "/home/$user/.ssh/id_ed25519.pub" >> "$TMP_KEYS"
+                mv -f "$TMP_KEYS" "/home/$user/.ssh/authorized_keys"
+              fi
+
+            ) 9>"/home/$user/.ssh/keygen.lock"
+
+            # Secure remaining permissions to the user
+            chown -R "$user:$user" "/home/$user/.ssh"
+            chmod 600 "/home/$user/.ssh/"*
+          done
+
+          # 3. Disable StrictHostKeyChecking safely for internal network jumps only
+          if ! grep -q "Host 10.\*" /etc/ssh/ssh_config; then
+            echo -e "Host 10.*\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
+          fi
+
+          # 4. Restart sshd so it loads DEFAULT:SHA1 crypto policy for RSA keys
+          systemctl restart sshd
+
+  - id: c3_nodeset
+    source: community/modules/compute/schedmd-slurm-gcp-v6-nodeset
+    use: [network, ssh_config]
+    settings:
+      node_count_static: 2
+      node_count_dynamic_max: 0
+      machine_type: c3-standard-176
+      disk_type: hyperdisk-balanced
+      bandwidth_tier: gvnic_enabled
+      allow_automatic_updates: false
+      # reservation_name: c3-16node-reservation
+      # If not using reservation:
+      # enable_placement: true
+      # placement_max_distance: 2
+      instance_image:
+        family: $(vars.built_instance_image.family)
+        project: $(vars.project_id)
+      instance_image_custom: true
+      service_account_email: $(vars.service_account_email)
+      service_account_scopes: &id001
+      - https://www.googleapis.com/auth/cloud-platform
+      enable_oslogin: false
+
+  - id: c3_partition
+    source: community/modules/compute/schedmd-slurm-gcp-v6-partition
+    use: [c3_nodeset]
+    settings:
+      partition_name: compute
+      is_default: true
+      exclusive: false
+
+  - id: slurm_login
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-login
+    use: [network, ssh_config]
+    settings:
+      machine_type: $(vars.image_build_machine_type)
+      disk_type: hyperdisk-balanced
+      enable_login_public_ips: true
+      instance_image:
+        family: $(vars.built_instance_image.family)
+        project: $(vars.project_id)
+      instance_image_custom: true
+      service_account_email: $(vars.service_account_email)
+      service_account_scopes: *id001
+      enable_oslogin: false
+
+  - id: slurm_controller
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
+    use: [network, c3_partition, homefs, slurm_login]
+    settings:
+      controller_startup_script: $(ssh_config.controller_startup_script)
+      login_startup_script: $(ssh_config.startup_script)
+      machine_type: $(vars.image_build_machine_type)
+      disk_type: hyperdisk-balanced
+      controller_state_disk:
+        type: hyperdisk-balanced
+        size: 50
+      enable_controller_public_ips: false
+      instance_image:
+        family: $(vars.built_instance_image.family)
+        project: $(vars.project_id)
+      instance_image_custom: true
+      service_account_email: $(vars.service_account_email)
+      service_account_scopes: *id001
+      enable_slurm_auth: true
+      enable_oslogin: false
--- a/community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
+++ b/community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
@@ -170,7 +170,7 @@ deployment_groups:
               useradd -u "$USER_UID" -g "$USER_UID" -m -s /bin/bash "$user"
             fi
             mkdir -p "/home/$user/.ssh"
-            chown -R "$user:$user" "/home/$user"
+            chown "$user:$user" "/home/$user"
             chmod 700 "/home/$user/.ssh"
 
             # Safely serialize key generation & authorized_keys rebuilding across the cluster
@@ -199,7 +199,9 @@ deployment_groups:
 
           # 3. Disable StrictHostKeyChecking safely for internal network jumps only
           if ! grep -q "Host 10.\*" /etc/ssh/ssh_config; then
-            echo -e "Host 10.*\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
+            if ! grep -q "StrictHostKeyChecking no" /etc/ssh/ssh_config; then
+              echo -e "Host 10.* 172.16.* 192.168.* *.internal *\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
+            fi
           fi
 
           # 4. Restart sshd so it loads DEFAULT:SHA1 crypto policy for RSA keys
@@ -256,7 +258,6 @@ deployment_groups:
     use: [network, c3_partition, homefs, slurm_login]
     settings:
       controller_startup_script: $(ssh_config.controller_startup_script)
-      login_startup_script: $(ssh_config.startup_script)
       machine_type: $(vars.image_build_machine_type)
       disk_type: hyperdisk-balanced
       controller_state_disk:
--- a/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
+++ b/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
@@ -301,6 +301,9 @@ def setup_key(lkp: util.Lookup, is_primary: bool = True) -> None:
     if lkp.cfg.enable_slurm_auth:
         # Put key into shared volume for distribution
         if is_primary:
+            util.mkdirp(util.slurmdirs.key_distribution)
+            util.chown_slurm(util.slurmdirs.key_distribution, mode=0o755)
+
             distributed = util.slurmdirs.key_distribution / file_name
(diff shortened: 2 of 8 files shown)


aslam-quad:
Hi @juntangc 

Please address the pre-commit failures and review the Gemini comments. if they are valid, please address them as well.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -334,6 +334,96 @@ def setup_sudoers():
     sudoers_file.chmod(0o0440)
 
 
+def setup_user_ssh():
+    """Ensure SELinux allows NFS home dirs and sync metadata SSH keys into /home/$USER/.ssh/."""
+    try:
+        run("setsebool -P use_nfs_home_dirs 1", timeout=30)
+    except Exception:
+        pass
+
```
AdarshK15: Modifying the core `setup.py` script with automated SSH key generation introduces security risks and can cause race-conditions or conflicts for deployments utilizing central identity managers. 

**`Cluster Toolkit`** already supports `configure_ssh_host_patterns`, could you remove the `setup_user_ssh()` logic from `setup.py` entirely and update the blueprint to leverage this feature.

Could you please add something like this to the blueprint as per your requirement and test if it works:
```yaml
  - id: ssh_config
    settings:
      configure_ssh_host_patterns:
        - "10.*"
        - "hpc*"
      runners:
      - type: shell
        destination: fix_selinux.sh
        content: |
          #!/bin/bash
          setsebool -P use_nfs_home_dirs 1 || true
  - id: nodeset
    use: [network, ssh_config]
...
  - id: controller
    use: [network, partition, login]
    settings:
     ...
      controller_startup_script: $(ssh_config.controller_startup_script)
      login_startup_script: $(ssh_config.startup_script)
...
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -671,11 +672,16 @@ def build_request(http, *args, **kwargs):
             new_http = google_auth_httplib2.AuthorizedHttp(credentials, http=new_http)
         return googleapiclient.http.HttpRequest(new_http, *args, **kwargs)
 
+    ud = universe_domain()
     ver = endpoint_version(ApiEndpoint.COMPUTE)
     disc_url = googleapiclient.discovery.DISCOVERY_URI
-    if ver:
+
+    if ud and ud != DEFAULT_UNIVERSE_DOMAIN:
+        version = ver if ver else "v1"
```
AdarshK15: we should fallback to default `version` if `ver` is `null`.
```
version = ver if ver else version
```

examples/hpc-slurm-c3-tpc/hpc-slurm-c3-tpc.yaml
```diff
@@ -0,0 +1,200 @@
...
+  - id: slurm_controller
+    source: community/modules/scheduler/schedmd-slurm-gcp-v6-controller
+    use: [network, c3_partition, homefs, slurm_login]
+    settings:
+      machine_type: $(vars.image_build_machine_type)
+      disk_type: hyperdisk-balanced
+      controller_state_disk:
+        type: hyperdisk-balanced
+        size: 50
+      enable_controller_public_ips: true
```
AdarshK15: `enable_controller_public_ips` and `enable_login_public_ips` are both set to true. Since TPC environments typically restrict external IPs, shouldn't these be set to false?

Additionally, there are no references to the newly introduced `cloud-nat` module anywhere in this blueprint, nor elsewhere in the PR. For a TPC environment without public IPs, nodes will likely need a Cloud NAT to reach the internet.
Should the `cloud-nat` module be added to this blueprint's `setup` group? If it is not needed, we should remove the `cloud-nat` module entirely."

community/modules/file-system/nfs-server/scripts/install-nfs-server.sh.tpl
```diff
@@ -25,7 +25,7 @@ if [ ! -d "/exports" ]; then # first load, format and mount the disk
 
     %{ for mount in local_mounts ~}
     mkdir -p /exports${mount}
-    chmod 755 /exports${mount}
+    chmod 1777 /exports${mount}
```
AdarshK15: The `nfs-server` module is shared across the Cluster Toolkit for both Slurm and GKE. Setting the export permissions to `1777` introduces a significant security risk and could cause regressions for other environments.
Any custom permission requirement should be handled in the blueprint via startup scripts.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -433,8 +523,10 @@ def configure_dirs():
         (dirs.scripts / "etc", slurmdirs.etc),
         (dirs.scripts / "log", dirs.log),
     ):
-        if sl.exists() and sl.is_symlink():
+        if sl.is_symlink() or sl.is_file():
             sl.unlink()
+        elif sl.is_dir():
+            shutil.rmtree(sl)
```
AdarshK15: Are these `rmtree` changes still needed? Now that we have moved back to the standard `slurm-gcp` installation flow instead of the custom slurm installation, this directory conflict shouldn't happen anymore.

Unless there's another edge case I am missing, it's probably safer to revert this block so we are not running recursive deletes in the setup script.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -156,6 +156,7 @@ def mkdirp(path: Path) -> None:
     etc = Path("/usr/local/etc/slurm"),
     state = Path("/var/spool/slurm"),
     key_distribution = Path("/slurm/key_distribution"),
+    run = Path("/run/slurm"),
```
AdarshK15: I think this is also another remnant from the custom Slurm installation. Since we are moving back to the standard installation flow, we shouldn't need Python to manually recreate /run partitions on boot.

modules/network/cloud-nat/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  region                             = var.region
+  project                            = var.project_id
+  nat_ip_allocate_option             = "AUTO_ONLY"
+  source_subnetwork_ip_ranges_to_nat = "ALL_SUBNETWORKS_ALL_IP_RANGES"
+
+  log_config {
+    enable = false
+    filter = "ALL"
+  }
+}
```
AdarshK15: Setting this to `ALL_SUBNETWORKS_ALL_IP_RANGES` introduces a significant risk for users deploying into Shared VPCs. It will hijack outbound routing for all other subnets in the region and cause API conflicts if someone tries to deploy a second NAT.

You can scope it to only the required subnets used by this deployment by doing something like this: 
1. Introduce a new variable `subnetwork_self_link` to target the subnetwork to apply the NAT to. 
2. Modify the existing block to look like this:
```hcl
resource "google_compute_router_nat" "nat" {
  name                               = local.nat_name
  router                             = google_compute_router.router.name
  region                             = var.region
  project                            = var.project_id
  nat_ip_allocate_option             = "AUTO_ONLY"
  source_subnetwork_ip_ranges_to_nat = "LIST_OF_SUBNETWORKS"
  subnetwork {
    name                    = var.subnetwork_self_link
    source_ip_ranges_to_nat = ["ALL_IP_RANGES"]
  }
}
```

modules/network/cloud-nat/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+  type        = string
+}
+
+variable "region" {
+  description = "Region for Cloud Router and NAT."
+  type        = string
+}
+
+variable "network_name" {
+  description = "Name of VPC network to attach Cloud NAT to."
```
AdarshK15: The router uses `network_name` to attach to the VPC. In a Shared VPC topology, the network resides in a Host Project, so passing just the short name will cause Terraform to look in the Service Project and fail.

Could you rename this variable to `network_self_link` (or `network_id`) and pass it to the router instead? This ensures the Cloud Router can resolve the network across project boundaries using the full resource URI.

AdarshK15:
In the blueprint `enable_controller_public_ips` and `enable_login_public_ips` are both set to true. Since air-gapped or disconnected environments typically restrict external IPs, shouldn't these be set to false?

Additionally, there are no references to the newly introduced `cloud-nat` module anywhere in the blueprint, nor elsewhere in the PR. For an environment without public IPs, nodes will likely need a Cloud NAT to reach the internet. Should the `cloud-nat` module be added to the blueprint's `setup` group? If it is not needed, we should remove the `cloud-nat` module entirely.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -671,11 +671,17 @@ def build_request(http, *args, **kwargs):
...
     disc_url = googleapiclient.discovery.DISCOVERY_URI
+
     if ver:
         version = ver
-        disc_url = disc_url.replace(DEFAULT_UNIVERSE_DOMAIN, universe_domain())
+
+    if ud and ud != DEFAULT_UNIVERSE_DOMAIN:
+        disc_url = f"https://compute.{ud}/$discovery/rest?version={version}"
+    elif ver:
+        disc_url = disc_url.replace(DEFAULT_UNIVERSE_DOMAIN, ud)
```
AdarshK15: Can we simplify this block?
ex:
```python
    ud = universe_domain()
    ver = endpoint_version(ApiEndpoint.COMPUTE)
    
    if ver:
        version = ver

    if ud and ud != DEFAULT_UNIVERSE_DOMAIN:
        disc_url = f"https://compute.{ud}/$discovery/rest?version={version}"
    else:
        disc_url = googleapiclient.discovery.DISCOVERY_URI
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -693,11 +699,14 @@ def storage_client() -> storage.Client:
...
     """
     ud = universe_domain()
-    # Check if we need a custom universe domain, otherwise pass None
+    api_endpoint = f"https://storage.{ud}" if (ud and ud != DEFAULT_UNIVERSE_DOMAIN) else None
     universe_domain_val = ud if (ud and ud != DEFAULT_UNIVERSE_DOMAIN) else None
     
     return storage.Client(
-        client_options=ClientOptions(universe_domain=universe_domain_val)
+        client_options=ClientOptions(
+            api_endpoint=api_endpoint,
```
AdarshK15, line 714: The `google-cloud-storage` SDK automatically computes `api_endpoint` (e.g., storage.{ud}) under the hood if you just pass `universe_domain=universe_domain_val`. Specifying both is redundant. 
Can we drop `api_endpoint` and just rely on `universe_domain`?

community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
```diff
@@ -0,0 +1,235 @@
...
+
+---
+blueprint_name: hpc-c3-slurm-no-oslogin
+vars:
+  project_id: # <project-id>
+  deployment_name: # <deployment-name>
+  region: # <region>
+  zone: # <zone>
+  universe_domain: # apis-berlin-build0.goog
+  service_account_email: # <account-id>-compute@developer.gserviceaccount.com
```
AdarshK15: PR-validation test is failing as these fields are empty, Can we add some default values similar to any existing blueprints? example - https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/91195f25de6e4caeb7feb268dbacbf91489948ab/examples/hpc-slurm.yaml#L19-L23

examples/hpc-slurm-c3-no-oslogin/hpc-slurm-c3-no-oslogin.yaml
```diff
@@ -0,0 +1,235 @@
...
+          #!/bin/bash
+          set -e -o pipefail
+          dnf update -y
+          dnf install -y epel-release git ansible-core
+          ansible-pull -U https://github.com/GoogleCloudPlatform/slurm-gcp -C $(vars.build_slurm_from_git_ref) \
+              -i localhost, --limit localhost --connection=local \
+              -e @/var/tmp/slurm_vars.json ansible/playbook.yml
+
+  - id: ssh_config
+    source: modules/scripts/startup-script
```
AdarshK15: Can we move this to `cluster` group as this is not required for building image?

examples/hpc-slurm-c3-no-oslogin/hpc-slurm-c3-no-oslogin.yaml
```diff
@@ -0,0 +1,235 @@
...
+        type: hyperdisk-balanced
+        size: 50
+      enable_controller_public_ips: false
+      instance_image:
+        family: $(vars.built_instance_image.family)
+        project: $(vars.project_id)
+      instance_image_custom: true
+      service_account_email: $(vars.service_account_email)
+      service_account_scopes: *id001
+      enable_slurm_auth: false
```
AdarshK15: This flag will be deprecated soon, can we test the deployment with SAuth?
SAuth is enabled by default, we just need to remove this flag.

examples/hpc-slurm-c3-no-oslogin/hpc-slurm-c3-no-oslogin.yaml
```diff
@@ -0,0 +1,235 @@
...
+          dnf update -y
+          dnf install -y epel-release git ansible-core
+          ansible-pull -U https://github.com/GoogleCloudPlatform/slurm-gcp -C $(vars.build_slurm_from_git_ref) \
+              -i localhost, --limit localhost --connection=local \
+              -e @/var/tmp/slurm_vars.json ansible/playbook.yml
+
+  - id: ssh_config
+    source: modules/scripts/startup-script
+    settings:
+      install_ansible: false
```
AdarshK15: Thanks for simplifying this script, I have tested this SSH script with OSLogin set to false and found few issues:
1. Because `curl -s` lacks the `-f` flag, the script incorrectly parses `404 HTML` errors as usernames when project SSH keys are missing, causing `useradd` to generate unwanted accounts like `<!DOCTYPE` and `404`, just need to add -f param.
2. Since all cluster nodes share an NFS `/home` directory and boot concurrently, the script's `> authorized_keys` logic truncates and overwrites keys that earlier-booting nodes just generated, causing SSH lockouts. We can fix this race condition by wrapping the file rebuild inside a cluster-wide lock (`flock`) and staging the final edits via a temporary file before writing to the main file.
3. We can remove `StrictHostKeyChecking no` for `Host *` as it is a security risk and as `Host 10.*` is already present and sufficient.
4. Upgrade the internally generated key type from `rsa` to `ed25519` to ensure it is compatible with newer images.
5. Also few other things we can remove like `systemctl restart sshd`, modify `chown -R` to just `/home/$user/.ssh`.

I updated the script and tested it. Could you please try this version and check if it works for you?
```bash
          #!/bin/bash
          # 1. SELinux & Crypto Policy for Rocky Linux 8/9
          setsebool -P use_nfs_home_dirs 1 || true
          update-crypto-policies --set DEFAULT:SHA1 || true

          # 2. Populate metadata keys and secure inter-node keypair on NFS /home
          METADATA="http://metadata.google.internal/computeMetadata/v1"
          # Added -f to curl to fail silently preventing 404 HTML parsing bugs
          KEYS=`curl -s -f -H "Metadata-Flavor: Google" "$METADATA/project/attributes/ssh-keys" 2>/dev/null || true`

          for user in `echo "$KEYS" | cut -d: -f1 | sort -u`; do
            [ -z "$user" ] && continue
            
            # Create user if missing
            id "$user" &>/dev/null || useradd -m -s /bin/bash "$user"
            mkdir -p "/home/$user/.ssh"
            chown "$user:$user" "/home/$user/.ssh"
            chmod 700 "/home/$user/.ssh"

            # Safely serialize key generation & authorized_keys rebuilding across the cluster
            (
              flock 9
              
              # Generate ED25519 cluster key natively
              if [ ! -f "/home/$user/.ssh/id_ed25519" ]; then
                ssh-keygen -t ed25519 -N "" -f "/home/$user/.ssh/id_ed25519"
              fi
              
              # Rebuild authorized_keys atomically on the exact same NFS filesystem (prevents O_TRUNC zeroing)
              TMP_KEYS=`mktemp "/home/$user/.ssh/authorized_keys.XXXXXX"`
              echo "$KEYS" | grep "^$user:" | cut -d: -f2- > "$TMP_KEYS"
              [ -f "/home/$user/.ssh/id_ed25519.pub" ] && cat "/home/$user/.ssh/id_ed25519.pub" >> "$TMP_KEYS"
              
              cat "$TMP_KEYS" > "/home/$user/.ssh/authorized_keys"
              rm -f "$TMP_KEYS"
              
            ) 9>"/home/$user/.ssh/keygen.lock"
            
            # Secure remaining permissions to the user
            chown -R "$user:$user" "/home/$user/.ssh"
            chmod 600 "/home/$user/.ssh/"*
          done

          # 3. Disable StrictHostKeyChecking safely for internal network jumps only
          if ! grep -q "Host 10.\*" /etc/ssh/ssh_config; then
            echo -e "Host 10.*\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
          fi
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
juntangc: This is required. Because when api_endpoint was not passed, it sent the google-cloud-dedicated credentials to public GCP (storage.googleapis.com), resulting in 401 Unauthorized.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -301,9 +301,19 @@ def setup_key(lkp: util.Lookup, is_primary: bool = True) -> None:
...
+            os.chmod(util.slurmdirs.key_distribution, 0o755)
+            util.chown_slurm(util.slurmdirs.key_distribution)
+
             distributed = util.slurmdirs.key_distribution / file_name
             shutil.copyfile(dst, distributed)
             util.chown_slurm(distributed, mode=0o400)
+
+            # Start NFS server and export key share immediately so concurrent compute/login nodes don't wait
+            run("systemctl enable nfs-server", timeout=30)
+            run("systemctl start nfs-server", timeout=30)
```
AdarshK15: Can these be moved to node startup script or removed as node timeout was increased to 300s?

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
```diff
@@ -101,10 +101,17 @@ def resolve_network_storage() -> List[NSMount]:
...
     # NOTE: Valid Lustre server_ip can take the form of '<IP>@tcp'
-    server_ip = mount.server_ip.split("@")[0]
-    mount_addr = util.host_lookup(server_ip)
-    return mount_addr == lookup().control_host_addr
+    server_ip = str(mount.server_ip).split("@")[0]
+    mount_addr = util.host_lookup(server_ip) if server_ip else None
+    return (
+        mount_addr == lookup().control_host_addr
+        or server_ip == lookup().control_host
+        or server_ip == lookup().hostname
```
AdarshK15: This method is to check if it is a controller mount, but adding `or server_ip == lookup().hostname` makes it true even for compute nodes, which is invalid.

community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
```diff
@@ -0,0 +1,264 @@
...
+          KEYS=`printf "%s\n%s\n" "$PROJ_KEYS" "$INST_KEYS" | grep -v '^[[:space:]]*$' | sort -u`
+
+          for user in `echo "$KEYS" | cut -d: -f1 | sort -u`; do
+            [ -z "$user" ] && continue
+
+            # Create user if missing
+            id "$user" &>/dev/null || useradd -m -s /bin/bash "$user"
+            mkdir -p "/home/$user/.ssh"
+            chown "$user:$user" "/home/$user/.ssh"
+            chmod 700 "/home/$user/.ssh"
```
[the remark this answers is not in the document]
juntangc: Applied! Updated the startup script to compute a deterministic UID/GID from cksum mapped into the 2000–60000 range. This ensures all cluster nodes assign identical UIDs/GIDs for consistent NFS permissions across the cluster without relying on external LDAP.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
juntangc: Great catch! I've updated the logic so lookup().hostname and localhost aliases are strictly scoped inside if lookup().is_controller.

aslam-quad:

 Thank you!

arpit974 (dismissed):
LGTM, let's monitor the tests

community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
```diff
@@ -0,0 +1,272 @@
...
+
+---
+blueprint_name: hpc-c3-slurm-google-cloud-dedicated
+vars:
+  project_id: # <project-id>
+  deployment_name: hpc-slurm
+  region: us-central1
+  zone: us-central1-a
+  universe_domain: # apis-<location>.goog
+  service_account_email: # <account-id>-compute@developer.gserviceaccount.com
```
AdarshK15: PR-Validation test failed. Can you please add some default values for `universe_domain` and `service_account_email`.

arpit974: actually, 

To fix this, please add this blueprint to the `EXCLUDE_EXAMPLE` array in `tools/validate_configs/validate_configs.sh` (around line 131, right after `eda-hybrid-cloud.yaml`):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -2258,7 +2284,7 @@ def etc_dir(self) -> Path:
         return Path(self.cfg.output_dir or slurmdirs.etc)
 
     def controller_mount_server_ip(self) -> str:
-        return self.control_addr or self.control_host
+        return self.control_addr or self.control_host or (self.hostname if self.is_controller else "$controller")
```
AdarshK15: Should we change `$controller` to `None`?

AdarshK15 (dismissed):

juntangc:
@saara-tyagi27 @shivam222 please help with the review.

AdarshK15 (dismissed):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup_network_storage.py
```diff
@@ -101,10 +101,25 @@ def resolve_network_storage() -> List[NSMount]:
 
 
 def is_controller_mount(mount) -> bool:
+    if not mount or not getattr(mount, "server_ip", None):
```
arpit974: One minor check for the shared controller code:

For clusters mounting Google Cloud Storage buckets directly via `gcsfuse` (`get_slurm_bucket_mount()`), `mount.server_ip` is `""` (empty string). 

Because `not getattr(mount, "server_ip", None)` evaluates to `True`, `is_controller_mount()` will return `True` on the controller. `separate()` will then treat GCSFUSE as a controller export and omit it from `/etc/fstab` on the controller, while `setup_nfs_exports()` tries to export it.

Could we guard against `gcsfuse` so bucket mounts are never classified as controller exports? E.g.:

```python
def is_controller_mount(mount) -> bool:
    if not mount:
        return False
    if getattr(mount, "fs_type", None) == "gcsfuse":
        return False
    if not getattr(mount, "server_ip", None):
        return lookup().is_controller

File: community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py [hunk 21]
```diff
@@ -1641,11 +1674,11 @@ def control_addr(self) -> Optional[str]:
 
     @property
     def control_host(self):
-        return self.cfg.slurm_control_host
+        return self.cfg.slurm_control_host or (self.hostname if self.is_controller else None)
```
arpit974: On compute and login nodes (`is_controller == False`), when `slurm_control_host` is not explicitly set in the blueprint/config, `self.cfg.slurm_control_host or (self.hostname if self.is_controller else None)` evaluates to `None`.

This breaks compute node bootstrapping in `setup.py:setup_compute()`:
- `primary_control_host = lkp.control_host_addr or lkp.control_host` evaluates to `None`.
- `servers = [f"{primary_control_host}:{lkp.control_host_port}"]` evaluates to `["None:6818"]`.
- `slurmd` is launched with `--conf-server="None:6818"` and fails to join the cluster.
- Additionally, `controller_mount_server_ip()` (line 2293) falls back to `None`, causing internal NFS mounts to attempt mounting `None:/...`.

Please fall back to the standard cluster controller naming convention for non-controller instances:

```suggestion
        return (
            self.cfg.slurm_control_host
            or (self.hostname if self.is_controller else f"{self.cfg.slurm_cluster_name}-controller")
        )
```

community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
```diff
@@ -0,0 +1,273 @@
...
+            ) 9>"/home/$user/.ssh/keygen.lock"
+
+            # Secure remaining permissions to the user
+            chown -R "$user:$user" "/home/$user/.ssh"
+            chmod 600 "/home/$user/.ssh/"*
+          done
+
+          # 3. Disable StrictHostKeyChecking safely for internal network jumps only
+          if ! grep -q "Host 10.\*" /etc/ssh/ssh_config; then
+            echo -e "Host 10.*\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
```
arpit974: `Host 10.*` only disables strict host key checking when nodes connect using numeric IP addresses starting with `10.`.

When MPI runtimes (OpenMPI, Intel MPI) or Slurm (`srun`, `mpirun`) launch distributed tasks across nodes, processes address peer nodes by short hostname (`c3-nodeset-0`, `c3-nodeset-1`) or internal FQDN (`*.internal`). Because hostnames do not match `Host 10.*`, SSH prompts for interactive host verification (`The authenticity of host ... can't be established`), which causes non-interactive MPI jobs to immediately fail with `Host key verification failed`.

Please update the rule to cover internal hostnames and private network ranges:

```suggestion
          if ! grep -q "StrictHostKeyChecking no" /etc/ssh/ssh_config; then
            echo -e "Host 10.* 172.16.* 192.168.* *.internal *\n    StrictHostKeyChecking no\n    UserKnownHostsFile=/dev/null" >> /etc/ssh/ssh_config
          fi
```

community/examples/hpc-slurm-google-cloud-dedicated/hpc-slurm-google-cloud-dedicated.yaml
```diff
@@ -0,0 +1,273 @@
...
+
+            # Create group and user with deterministic UID/GID to ensure consistency across the cluster
+            if ! id "$user" &>/dev/null; then
+              UID_HASH=`printf "%s" "$user" | cksum | cut -d' ' -f1`
+              USER_UID=`expr 2000 + $UID_HASH % 58000`
+              groupadd -g "$USER_UID" "$user" || true
+              useradd -u "$USER_UID" -g "$USER_UID" -m -s /bin/bash "$user"
+            fi
+            mkdir -p "/home/$user/.ssh"
+            chown -R "$user:$user" "/home/$user"
```
arpit974: Because `/home` is a shared NFS volume mounted across all cluster nodes, every node boot/reboot will execute `chown -R` across the entire directory tree of every user in `/home`.

As users accumulate virtual environments, datasets, or build caches in their home directories, recursively traversing the entire NFS tree over the network on every node boot causes an NFS metadata RPC storm that can easily breach the 300-second startup script timeout and fail node provisioning.

The `.ssh` directory and key files are already specifically and recursively chowned on lines 196–197 (`chown -R "$user:$user" "/home/$user/.ssh"`). Line 173 should only chown the user directory root non-recursively:

```suggestion
            chown "$user:$user" "/home/$user"
```

arpit974 (changes requested):

arpit974 (approved):

AdarshK15 (approved):

PR #6145, opened 2026-08-14
feat(slurm-controller): add x11_parameters to cloud_parameters

qsogia:
Supersedes #5363, which the inactivity bot auto-closed on 2026-08-14 and which GitHub refuses to reopen. Same change, now merged up to current `develop`.

## Description

Adds `x11_parameters` to the `cloud_parameters` object in `schedmd-slurm-gcp-v6-controller` so that Slurm's `X11Parameters` directive can be set via the blueprint without forking `slurm.conf.tpl`.

The setting flows through `cloud.conf` (generated by `conf.py`), which is already included by `slurm.conf` via the existing `include cloud.conf` directive — no template changes required.

## Changes

- `modules/slurm_files/scripts/conf.py` — map `x11_parameters` → `X11Parameters` in `SlurmConfigGenerator.get_conf_options()`
- `modules/slurm_files/variables.tf` — add `x11_parameters = optional(string)` to `cloud_parameters`
- `variables.tf` — same addition at the top-level controller module

Since #5363 was opened, `conflines()` was refactored into `SlurmConfigGenerator`, so the mapping now lives in the shared `get_conf_options()`. `SlurmConfigGeneratorV2511` extends it via `super()` and `SlurmConfigGeneratorV2411` does not override it, so 24.11, 25.05 and 25.11 all pick it up. Verified all three emit `X11Parameters` when the setting is present, and `tests/test_conf.py` passes unchanged (unset values are filtered out by `dict_to_conf`, so default output is unaffected).

## Usage

```yaml
- id: slurm_controller
  settings:
    cloud_parameters:
      prolog_flags: X11
      x11_parameters: enable_send_credentials
```

## Why

`PrologFlags=X11` was already supported via `cloud_parameters.prolog_flags`. `X11Parameters` was not. Both are required for `srun --x11` to work:

- `PrologFlags=X11` — enables X11 forwarding in `slurmctld` at allocation time
- `X11Parameters=enable_send_credentials` — forwards the MIT-MAGIC-COOKIE from the login node to the compute node so the display tunnel can authenticate


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/conf.py
Neelabh94, line 1: Please add / update unit tests in community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/tests/test_conf.py for this change

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/variables.tf
```diff
@@ -391,6 +391,7 @@ variable "cloud_parameters" {
     tree_width              = optional(number)
     prolog_flags            = optional(string)
     switch_type             = optional(string)
+    x11_parameters          = optional(string)
```
Neelabh94, line 394: A comment explaining that that `srun --x11` requires `xorg-x11-xauth` (or equivalent `xauth` package) installed on both login and compute VM images would be helpful here.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf
```diff
@@ -549,6 +549,7 @@ variable "cloud_parameters" {
     tree_width              = optional(number)
     prolog_flags            = optional(string)
     switch_type             = optional(string)
+    x11_parameters          = optional(string)
```
Neelabh94, line 552: We should add a validation block for variable "cloud_parameters" in 
`community/modules/scheduler/schedmd-slurm-gcp-v6-controller/variables.tf` that catches blueprint misconfigurations at terraform plan time. Something like this:

```hcl
variable "cloud_parameters" {
  description = "cloud.conf options. Default behavior defined in scripts/conf.py"
  type = object({
   .... <Existing>
    switch_type             = optional(string)
    x11_parameters          = optional(string)
  })
  default  = {}
  nullable = false
  validation {
    condition = (
      var.cloud_parameters.x11_parameters == null ||
      (var.cloud_parameters.prolog_flags != null && can(regex("(^|,)X11(,|$)", var.cloud_parameters.prolog_flags)))
    )
    error_message = "When 'x11_parameters' is specified in cloud_parameters, 'prolog_flags' must also be specified and include 'X11' (e.g. prolog_flags = \"X11\")."
  }
}

Neelabh94 (changes requested):

aslam-quad:

Thank you!

PR #6167, opened 2026-08-19
fix(job): consolidate Kueue lifecycle and queue management into kueue.go

Neelabh94:
## Description
Consolidates 100% of Kueue-related lifecycle management, queue resolution, priority class installation, and flavor capacity calculations from `infra_manager.go` and `gke_job_orchestrator.go` into `pkg/orchestrator/gke/kueue.go`.

## Key Changes
1. **Consolidated Kueue Logic (`kueue.go`):** Centralized constants, queue resolution, lifecycle management, priority classes, and resource flavor calculations into 5 logical sections.
2. **Kueue Test Migration (`kueue_test.go`):** Ported all Kueue unit tests from `infra_manager_test.go` into `kueue_test.go`.
3. **Queue Fallback & Multi-Tenant Namespace Naming Bug Fix:** Removed static `defaultClusterQueue = "default-queue"` constant; dynamic `ClusterQueue` naming now prevents multi-tenant namespace collisions.
4. **Scoped `ResourceFlavor` Creation:** `EnsureResourceFlavors` is now scoped to initial queue creation and checks for pre-existing flavors before applying manifests, respecting admin-configured clusters and skipping redundant `kubectl apply` calls.

## Testing
- Passed 100% of unit tests (`go test -v ./pkg/orchestrator/gke/...`).
- Pre-commit hooks passed (`go fmt`, `go vet`, `go imports`, `go-cyclo`, `golangci-lint`).
- Tested live job submission and log streaming on GKE cluster `bodaborg-tpu7x-nap`.


pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1029 @@
...
+			Subsets []struct {
+				Addresses []struct {
+					Ip string `json:"ip"`
+				} `json:"addresses"`
+			} `json:"subsets"`
+		}
+		if err := json.Unmarshal([]byte(cmdEndpoints.Stdout), &eps); err != nil {
+			return false, fmt.Errorf("failed to unmarshal endpoints json: %w", err)
+		}
+		for _, subset := range eps.Subsets {
```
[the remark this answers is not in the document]
Neelabh94: This is a contradictory feedback to the earlier feedbacks. 
It is okay to fail, there should be no scenario for the endpoint to return incomplete data

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+	for _, c := range kueuePermissionChecks {
+		res := g.executor.ExecuteCommand("kubectl", "auth", "can-i", c.verb, c.resource)
+		if res.ExitCode != 0 || strings.TrimSpace(res.Stdout) != "yes" {
+			return fmt.Errorf("unable to re-install kueue to version %s, this could be a permission issue. Please contact your cluster administrator for updating KUEUE settings", version)
+		}
+	}
+	return nil
+}
+
+func (g *GKEOrchestrator) isKueueInstalled() (bool, error) {
```
shubpal07, line 417: Nice work consolidating the Kueue logic! 

One quick note on RBAC permissions: in shared or multi-tenant GKE clusters, namespace-scoped users (non-admins) will likely get a `403 Forbidden` error when querying cluster-scoped CRDs here, which would block their job submissions.

Could we catch `403 Forbidden` errors here, log a warning, and assume Kueue is installed? We can rely on subsequent namespace-scoped checks (like verifying if the `LocalQueue` exists in the target namespace) to confirm if Kueue is actually usable for the user.

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+	}
+	errStr := strings.ToLower(res.Stderr + " " + res.Stdout)
+	if strings.Contains(errStr, "notfound") || strings.Contains(errStr, "not found") {
+		logging.Info("Kueue CRD not found.")
+		return false, nil
+	}
+	return false, fmt.Errorf("failed to check for Kueue CRD: %s\n%s", res.Stderr, res.Stdout)
+}
+
+func (g *GKEOrchestrator) isKueueDeploymentInstalled() (bool, error) {
```
shubpal07, line 435: Similar to the CRD check above, querying resources in the system namespace (`kueue-system`) might return a `403 Forbidden` for users without cluster-wide permissions.

Would it make sense to catch `403` here as well, log a warning, and proceed?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+		// If the user lacks RBAC permissions to read ClusterQueue, assume it exists in shared clusters
+		// so we skip cluster-scoped creation (which would also fail with 403) and proceed to create LocalQueue.
+		logging.Warn("Insufficient RBAC permissions to read ClusterQueue %q (403 Forbidden). Assuming it exists in shared cluster.", name)
+		return true, nil
+	}
+	return false, fmt.Errorf("failed to check clusterqueue status: %s", res.Stderr)
+}
+
+// createDefaultQueues creates a LocalQueue in the target namespace and a corresponding ClusterQueue (named after the target namespace if non-default, or defaultClusterQueue).
+func (g *GKEOrchestrator) createDefaultQueues(localQueueName, ns string) error {
```
shubpal07, line 161: Using the namespace name as the ClusterQueue name (`cqName := ns`) is a neat way to isolate queues! However, it might introduce a couple of edge cases in multi-tenant environments:

1. **403 Fallback:** If a non-admin gets `403` on the ClusterQueue check, we assume it exists and point the LocalQueue to a ClusterQueue named `<namespace>`. If the cluster admin hasn't pre-created a ClusterQueue with that exact name, the workloads will hang silently in Kueue.
2. **Resource Overcommit:** For admin users, auto-creating a ClusterQueue per namespace over the same physical nodes could lead to resource overcommit (bypassing Kueue's scheduling controls).

Would it be safer to fall back to a standard shared ClusterQueue (like `"default"`) when a `403` is encountered, and perhaps allow users to explicitly configure the target ClusterQueue name if they need custom mappings?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+}
+
+// DeleteAllKueueResources deletes all Kueue custom resources and CRDs from the cluster.
+func (g *GKEOrchestrator) DeleteAllKueueResources() error {
+	resourceList := strings.Join(kueueCRDs, ",")
+	logging.Info("Deleting all Kueue resources...")
+	g.executor.ExecuteCommand("kubectl", "delete", resourceList, "--all", "--ignore-not-found", "--wait=false")
+
+	logging.Info("Deleting Kueue CRDs...")
+	args := append([]string{"delete", "crd", "--ignore-not-found"}, kueueCRDs...)
```
shubpal07: Using `--wait=false` when deleting Kueue resources, followed immediately by deleting the CRDs (line 504), might cause the CRDs or the namespace to get stuck in a `Terminating` state if there are active finalizers on the resources.

Could we remove `--wait=false` or wait for the resource deletion to complete before cleaning up the CRDs?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+		"apiVersion": "kueue.x-k8s.io/" + kueueAPIVersion,
+		"kind":       "ResourceFlavor",
+		"metadata": map[string]interface{}{
+			"name": name,
+		},
+	}
+	if len(nodeLabels) > 0 {
+		filteredLabels := make(map[string]string)
+		for k, v := range nodeLabels {
+			if k != tpuTopologyLabel &&
```
shubpal07, line 830: Could we add a brief inline comment explaining *why* we filter out these specific TPU topology, slice, and partition labels? It would help other developers maintaining this in the future.

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+	_, hasPathways := g.capacity.Flavors["pathways-flavor"]
+	for _, resName := range resList {
+		quota := g.getNominalQuota(resName, fc, fname)
+		if hasPathways && isTPUFlavor(fname) {
+			switch resName {
+			case "cpu":
+				quota = pathwaysCPUNominalQuota
+			case "memory":
+				quota = pathwaysMemoryNominalQuota
+			}
```
shubpal07, line 1058: Nit: Could we add a comment explaining the rationale behind setting the nominal quotas to `999999` when Pathways is enabled?

Neelabh94:
Thanks for the thorough review @shubpal07 , I have made the changes in the new commit [abd13832e](https://github.com/GoogleCloudPlatform/cluster-toolkit/commit/abd13832e)

pkg/orchestrator/gke/kueue.go
Neelabh94: Done

pkg/orchestrator/gke/kueue.go
Neelabh94: Done.

pkg/orchestrator/gke/kueue.go
Neelabh94: Done, replaced this with a timeout to prevent infinite hang during CRD deletion

pkg/orchestrator/gke/kueue.go
Neelabh94: Thanks for pointing this case out. Have modified the logic. PTAL

pkg/orchestrator/gke/kueue.go
Neelabh94: Done,

pkg/orchestrator/gke/kueue.go
Neelabh94: Done

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+	if version == "" {
+		version = defaultKueueVersion
+	} else if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
+
+	needReinstall, reinstallReason := g.evaluateKueueReinstall(kueueCRDInstalled, kueueDeploymentInstalled, currentVersion, version)
+	if needReinstall {
+		if g.checkDynamicSlicingViaGKE() {
+			return fmt.Errorf("automatic Kueue installation blocked: we detected that cluster %s is set up for Dynamic-slicing (found 'PROVISION_ONLY' in the node pool's placementPolicy). Wiping Kueue would corrupt your custom topology configurations. Please install Kueue and the required custom CRDs manually", clusterName)
```
shubpal07: Nit: The warning message `"Wiping Kueue would corrupt your custom topology..."` might be a bit confusing to a user if Kueue was never installed on the cluster in the first place.

We may think to rephrase it to something more generic, like: `"Automatic Kueue installation/modification is blocked when dynamic slicing is enabled..."`?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+			}
+		}
+	}
+	return yaml.Marshal(rfMap)
+}
+
+func (g *GKEOrchestrator) installKueueResources(cqName string, lqName string) error {
+	logging.Info("Installing Kueue resources (ClusterQueue, LocalQueue)...")
+
+	if err := g.EnsureResourceFlavors(); err != nil {
```
shubpal07, line 848: Currently, if `EnsureResourceFlavors()` fails due to a non-RBAC error (like a network timeout or invalid YAML), we log a warning but continue execution. This might leave the cluster in an unhealthy state (e.g., ClusterQueues referencing missing flavors) without failing the command.

Would it be safer to only ignore verified RBAC `Forbidden` errors and treat other unexpected errors as fatal?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+		return true, nil
+	}
+	return false, fmt.Errorf("failed to check clusterqueue status: %s", res.Stderr)
+}
+
+// createDefaultQueues creates a LocalQueue in the target namespace and a corresponding ClusterQueue (named after the target namespace if non-default, or defaultClusterQueue).
+func (g *GKEOrchestrator) createDefaultQueues(localQueueName, ns string) error {
+	logging.Info("Creating default ClusterQueue and LocalQueue...")
+
+	if err := g.EnsureResourceFlavors(); err != nil {
```
shubpal07, line 164: Same [here](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/6167/changes#r3822361819)

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+
+	return g.installKueueResources(defaultClusterQueue, defaultLocalQueue)
+}
+
+func (g *GKEOrchestrator) checkKueueInstallPermissions(version string) error {
+	logging.Info("Verifying cluster permissions for Kueue installation...")
+	for _, c := range kueuePermissionChecks {
+		res := g.executor.ExecuteCommand("kubectl", "auth", "can-i", c.verb, c.resource)
+		if res.ExitCode != 0 || strings.TrimSpace(res.Stdout) != "yes" {
+			return fmt.Errorf("unable to re-install kueue to version %s, this could be a permission issue. Please contact your cluster administrator for updating KUEUE settings", version)
```
shubpal07: This error message is clear, but since we check multiple permissions in a loop, it might be hard for the user/admin to know which specific permission is missing.

Could we include the failed verb and resource (e.g. `c.verb` and `c.resource`) in the error message to make it more actionable?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
...
+		time.Sleep(3 * time.Second)
+	}
+	return fmt.Errorf("timed out waiting for kueue-webhook-service endpoints to be available")
+}
+
+func (g *GKEOrchestrator) probeKueueWebhookReadiness() error {
+	logging.Info("Probing Kueue webhook readiness...")
+	probeName := fmt.Sprintf("gcluster-webhook-probe-%d", time.Now().UnixNano())
+	probeManifest := fmt.Sprintf("apiVersion: kueue.x-k8s.io/%s\nkind: ResourceFlavor\nmetadata:\n  name: %s\n", kueueAPIVersion, probeName)
+	f, err := os.CreateTemp("", "gcluster-webhook-probe-*.yaml")
```
shubpal07, line 573: Nit: 
Writing a temp file to disk and running `kubectl apply -f` in a loop works fine, but it might leak files in `/tmp` if the execution is interrupted.

Would it be cleaner to pipe the manifest content directly to `kubectl apply -f -` via stdin to avoid writing to disk?

pkg/orchestrator/gke/infra_manager.go
```diff
@@ -657,168 +129,6 @@ func (g *GKEOrchestrator) waitForJobSetWebhook() error {
...
-				if len(subset.Addresses) > 0 {
-					return true, nil
-				}
-			}
-		}
-	}
-	return false, nil
-}
-
 func (g *GKEOrchestrator) isJobSetCRDInstalled() (bool, error) {
```
shubpal07, line 135: Similar to the Kueue checks, namespace-scoped users might get a `403 Forbidden` here when querying cluster-scoped CRDs. 

Would it make sense to catch `403` here as well, log a warning, and assume JobSet is installed if the user lacks cluster-level read permissions?

pkg/orchestrator/gke/kueue_test.go
```diff
@@ -0,0 +1,1036 @@
...
+	orc := &GKEOrchestrator{executor: mock}
+
+	covered, isNew, err := orc.checkClusterQueueCoverage("default")
+	if err != nil {
+		t.Fatalf("checkClusterQueueCoverage should succeed and skip on 403 Forbidden, got: %v", err)
+	}
+	if !covered || isNew {
+		t.Errorf("expected covered=true, isNew=false on 403 Forbidden, got covered=%v, isNew=%v", covered, isNew)
+	}
+}
```
shubpal07, line 1097: The test suite in `kueue_test.go` is very thorough! One area we might want to cover is the Pathways queue coverage flow (`ensureClusterQueueCoverage`).

Could we add a test case in `kueue_test.go` that mocks a Pathways job (`IsPathwaysJob: true`) to verify that the orchestration flow in `configureClusterEnvironment` executes the coverage checks correctly?

pkg/orchestrator/gke/kueue.go
```diff
@@ -0,0 +1,1051 @@
+// Copyright 2026 Google LLC
```
shubpal07, line 1: (Optional):
*Note: This [template](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/pkg/orchestrator/gke/templates/priority_classes.tmpl#L36) is pre-existing and not modified in this PR, but worth noting for a follow-up:*
Setting `globalDefault: true` on the default `medium` PriorityClass (value 500) will cause all pods on the cluster without an explicit priority class to default to 500. 

**Risk:** This can allow default pods to preempt workloads explicitly set to lower priorities (e.g., `low` at 250), potentially disrupting the GKE cluster's eviction and scheduling logic in shared/multi-tenant environments.

We should consider removing `globalDefault: true` in a future PR, or at least warning the user in the logs during installation.

shubpal07 (changes requested):
Thanks for this PR @Neelabh94 ! The Kueue logic consolidation is a great cleanup, and the test suite is very thorough. 
However, I've found a few improvements which can be discussed. Please Feel free to defer them in case you feel there are out of scope of this PR.
I've left detailed inline comments on these and a few major testing/deletion improvements.

pkg/orchestrator/gke/kueue.go
Neelabh94: `g.executor.ExecuteCommand` does not support stdin streaming, and the defer os.Remove(probeFile) + kubectl delete resourceflavor <probeName> block guarantees immediate cleanup upon function exit.

Neelabh94:
thanks again @shubpal07 . Have made the required changes around handling the 403 errors. PTAL

pkg/orchestrator/gke/kueue.go
Neelabh94: Thanks for the suggestion Shubham. I'll keep it for a separate discussion later.

shubpal07 (approved):
Thanks @Neelabh94 for addressing all the improvements!!
LGTM

PR #6184, opened 2026-08-20
NetApp Volumes: mount large-capacity volumes via Cloud DNS

okrause:
Large-capacity volumes and FlexCache expose multiple NFS endpoint IPs. Add volume dns_config and private-zone support so clients, including Slurm, mount by FQDN and use every endpoint. Update EDA and Slurm examples plus docs.


okrause:
PRed from main instead of commit. Closing and will open correct PR.

PR #6186, opened 2026-08-20
NetApp Volumes: mount large-capacity volumes via Cloud DNS

okrause:
Large-capacity volumes and FlexCache expose multiple NFS endpoint IPs. Add volume dns_config and private-zone support so clients, including Slurm, mount by FQDN and use every endpoint. Update EDA and Slurm examples plus docs.

Goal is to have a larger populations of clients spread their clients evenly (statistically) over all endpoint IPs to even out load on storage backend.


modules/file-system/netapp-volume/dns.tf
```diff
@@ -0,0 +1,59 @@
...
+}
+
+data "google_dns_managed_zone" "mount" {
+  count = var.dns_config != null ? 1 : 0
+
+  project = var.project_id
+  name    = var.dns_config.managed_zone_name
+
+  depends_on = [google_project_service.dns_api]
+}
```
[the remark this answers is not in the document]
okrause: Fair. Pushing the requested change.

aslam-quad:

Thank you!

examples/netapp-volumes-slurm.yaml
```diff
@@ -0,0 +1,132 @@
...
+    use: [netapp_pool]
+    settings:
+      volume_name: homefs
+      # 4800 GiB is the minimum Flex Unified large-capacity volume size.
+      capacity_gib: 4800
+      large_capacity_config:
+        constituent_count: 48
+      local_mount: /home
+      protocols: [NFSV3]
+      unix_permissions: "0777"
```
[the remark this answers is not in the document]
okrause: While this is generally true, changing it to "0755" will break deployments. When using a volume as home directory (with USE directive), VMs will create a home directory for any user on first login. The VM images seem to do this operation using the user permissions and not root. Since the root inode is owned by root:root, this home directory creations will fail if unix permissions are set to 0755 and the user will not have a proper home dir. Unless the compute images are not changed to do the home directory creation as root user, we need 0777 for VMs (e.g. SLURM) to be able to create user home dirs.

okrause:
@agrawalkhushi18 This is ready to review.

modules/file-system/netapp-volume/dns.tf
```diff
@@ -0,0 +1,51 @@
...
+data "google_dns_managed_zone" "mount" {
+  count = var.dns_config != null ? 1 : 0
+
+  project = var.project_id
+  name    = var.dns_config.managed_zone_name
+}
+
+locals {
+  dns_record_name = var.dns_config != null ? coalesce(var.dns_config.record_name, var.volume_name) : null
+  mount_fqdn = var.dns_config != null ? (
```
agrawalkhushi18: If a user explicitly passes a trailing dot (e.g., record_name = "vol."), the ternary evaluates to true, returns "vol." directly, and drops the zone append entirely. When Terraform pushes this to the Cloud DNS API, it will be rejected as an invalid record because it lacks the assigned Managed Zone suffix.
We can make this robust and fail-safe by dropping the endswith() ternary entirely and trimming preemptively:
`  mount_fqdn = var.dns_config != null ? "${trimsuffix(local.dns_record_name, ".")}.${data.google_dns_managed_zone.mount[0].dns_name}" : null`

modules/file-system/netapp-volume/variables.tf
```diff
@@ -184,6 +184,39 @@ variable "deletion_policy" {
...
+  default = null
+  validation {
+    condition     = var.dns_config == null || var.dns_config.managed_zone_name != ""
+    error_message = "dns_config.managed_zone_name must be non-empty when dns_config is set."
+  }
+  validation {
+    condition     = var.dns_config == null || var.dns_config.ttl == null || var.dns_config.ttl > 0
+    error_message = "dns_config.ttl must be greater than 0 when set."
+  }
+  validation {
```
agrawalkhushi18, line 210: The two validation blocks shielding against underscores execute checks strictly against != null. If a user passes an explicit empty string (record_name = ""), it bypasses both validation errors. coalesce then correctly falls back to var.volume_name in dns.tf. If var.volume_name contains an underscore, the invalid string hits the DNS record payload and fails at the API layer.

We can consolidate both of these blocks into a single bulletproof check that leverages your fallback logic:
```
  validation {
    condition     = var.dns_config == null || !strcontains(coalesce(var.dns_config.record_name, var.volume_name), "_")
    error_message = "The DNS record_name (or volume_name if record_name is not provided) cannot contain underscores. DNS hostnames allow only letters, digits, and hyphens. Use hyphens (-) instead."
  }
```

modules/file-system/netapp-volume/dns.tf
okrause: Great catch. I will treat record_name as relative name only.

modules/file-system/netapp-volume/variables.tf
okrause: Good catch. Thank you!

okrause:
I accidentally pushed other code on this branch. I reverted it back to good commit. My apologies.

vikramvs-gg (dismissed):
LGTM

agrawalkhushi18 (dismissed):

okrause:
@vikramvs-gg @agrawalkhushi18 Thank you for your approval. Very appreciated. Can you tell me why the pipeline it not moving forward? We have two approvals.

okrause:
Thank you for triggering the run. Turns our region and zone where missing from the new examples/netapp-volumes-slurm.yaml blueprint. I added them. Should work now.

agrawalkhushi18:
> Thank you for triggering the run. Turns our region and zone where missing from the new examples/netapp-volumes-slurm.yaml blueprint. I added them. Should work now.

Thanks @okrause . But why are we hardcoding the region and zone values in the sample blueprint?

okrause:
> > Thank you for triggering the run. Turns our region and zone where missing from the new examples/netapp-volumes-slurm.yaml blueprint. I added them. Should work now.
> 
> Thanks @okrause . But why are we hardcoding the region and zone values in the sample blueprint?

@agrawalkhushi18 One of the PR workflows (IIRC it is PR-validation) validates all the example blueprints. There seem to be two types of tests. The smaller ones are ran in `tools/validate_configs/validate_configs.sh` and need hard-coded values. There are many other blueprints in `examples/` which take that approach. Since I did not pass region and zone, this happend:
```
Step #1 - "make-tests": 2026-09-14T08:53:41Z: Error: deployment variable "region" was not set
Step #1 - "make-tests": 27:   region:     ## Set GCP Region Here ##
Step #1 - "make-tests":       ^
Step #1 - "make-tests": Error: deployment variable "zone" was not set
Step #1 - "make-tests": 28:   zone:       ## Set GCP Zone Here ##
Step #1 - "make-tests":       ^
Step #1 - "make-tests": *** ERROR: error creating deployment with gcluster for netapp-volumes-slurm.yaml
```

The other category of tests are bigger test cases. examples/netapp-volumes.yaml uses that approach. They have a dedicated test setup in `tools/cloud-build/daily-tests/tests/` and are excluded from running in https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/6f755b4fe2cf780c0f82b24e7b05bb0ea3448c72/tools/validate_configs/validate_configs.sh#L133. Since NetApp Volumes functionality is already tested by netapp-volumes.yaml, netapp-volumes-slurm.yaml is not re-testing it again, but shows how it is configured using DNS. Having a basic verification through validate_configs.sh in a good thing, but I don't think it requires a big test with full SLURM attachment verification.

Does that make sense?

agrawalkhushi18:
Got it, thanks for the explaination @okrause .

agrawalkhushi18 (approved):

vikramvs-gg (approved):

okrause:
Thank you @vikramvs-gg and @agrawalkhushi18 for your support. Very appreciated!

PR #6201, opened 2026-08-24
Fix Spot Zones Filter Job

AdarshK15:
The declaration for `OPTIONS_GCS_PATH` has been updated for tests migrated to Kueue, causing the dynamic `filter-zone-options` job to silently miss them. The filter job skipped pruning unwanted zones that lack support or capacity for secondary machine types or auxiliary resources (like Filestore), causing `zone is not supported` or `zone doesn't have capacity` failures in Tests.

### DIFF
--- a/tools/cloud-build/filter-zone-options.yaml
+++ b/tools/cloud-build/filter-zone-options.yaml
@@ -60,11 +60,9 @@ steps:
         # Extract the GCS options file path using the standard Kueue format
         GCS_PATH=$$(grep -A 1 'name: OPTIONS_GCS_PATH' "$$yaml" | grep -o 'value: "gs://[^"]*' | cut -d'"' -f2 || true)
 
-        # Fallback to the legacy if Kueue format is missing
+        # If build file doesn't use spot options, silently skip it.
         if [[ -z "$$GCS_PATH" ]]; then
-          GCS_PATH=$$(grep -o 'OPTIONS_GCS_PATH=gs://[^"]*' "$$yaml" | cut -d= -f2 || true)
-          # If build file doesn't use spot options, silently skip it.
-          [[ -z "$$GCS_PATH" ]] && continue
+          continue
         fi
 
         FILE_NAME=$$(basename "$$GCS_PATH")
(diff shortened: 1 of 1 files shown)


AdarshK15:
Triggered the `filter-spot-zones` job with these changes and it passed (can be seen under PR checks). It is now correctly filtering for all machine types.

tools/cloud-build/filter-zone-options.yaml
```diff
@@ -59,6 +59,10 @@ steps:
 
         # Extract the GCS options file path
         GCS_PATH=$$(grep -o 'OPTIONS_GCS_PATH=gs://[^"]*' "$$yaml" | cut -d= -f2 || true)
+        if [[ -z "$$GCS_PATH" ]]; then
+          # Fallback for Kueue job style env vars
+          GCS_PATH=$$(grep -A 1 'name: OPTIONS_GCS_PATH' "$$yaml" | grep -o 'value: "gs://[^"]*' | cut -d'"' -f2 || true)
+        fi
         if [[ -z "$$GCS_PATH" ]]; then
```
Neelabh94, line 64: Should this be removed?

tools/cloud-build/filter-zone-options.yaml
```diff
@@ -59,6 +59,10 @@ steps:
 
         # Extract the GCS options file path
         GCS_PATH=$$(grep -o 'OPTIONS_GCS_PATH=gs://[^"]*' "$$yaml" | cut -d= -f2 || true)
+        if [[ -z "$$GCS_PATH" ]]; then
```
Neelabh94: KUEUE is the default architecture for all test runs now, can we simplify the logic to directly use the "fallback"?

Neelabh94 (changes requested):

tools/cloud-build/filter-zone-options.yaml
AdarshK15: Yes, we migrated to Kueue but `slurm-gcp-v6-tpu.yaml` was not updated. Anyway, I removed the legacy grep as this test is not active now due to capacity. Even if we run it in the future, updating that test file is better than adding a separate check here just for it.

tools/cloud-build/filter-zone-options.yaml
AdarshK15: As this job checks for all test files in the build folder, we do need to silently skip build files that don't use a spot options file.

Neelabh94 (approved):

PR #6220, opened 2026-08-27
fix(job): resolve race condition in --await-job-completion workload discovery

Neelabh94:
## Description
Fixes a race condition in `gcluster job submit --await-job-completion`.

### Root Cause
1. `findTargetWorkload` performed a single immediate lookup for the Kueue `Workload` resource labeled `kueue.x-k8s.io/job-uid=<uid>`. Because Kueue creates the `Workload` resource asynchronously after `kubectl apply` submits the JobSet, the lookup returned empty, immediately throwing `failed to find Kueue workload for jobset <name>` and crashing scripts with `set -euo pipefail`.
2. `getJobSetStatus` previously selected conditions purely based on `LastTransitionTime` without checking `cond.Status == "True"`, causing newer inactive conditions (such as `Suspended: False`) to overwrite terminal statuses, and did not recognize `JobSetCompleted` / `JobSetFailed` condition types.

### Changes
- **Polling with Timeout in `findTargetWorkload`**: Refactored `findTargetWorkload` to accept a timeout parameter and poll until the Kueue `Workload` is minted by the controller.
- **Parametrized Timeouts**:
  - `awaitJobCompletion`: calls `findTargetWorkload` with a 60s timeout.
  - `inspectWorkload` (`inspector.go`): calls `findTargetWorkload` with a 2s timeout for fast point-in-time diagnostics.
- **Robust Condition Evaluation**:
  - `getTerminalJobSetCondition`: Checks active (`Status: True`) terminal conditions (`Completed`, `JobSetCompleted`, `Succeeded`, `Failed`, `JobSetFailed`).
  - `getLatestActiveJobSetCondition`: Only considers active conditions (`Status: True`).
  - Added fallback checking for `Spec.Suspend` (`Suspended` vs `Running`).
- **Defensive Go Idioms & Safety**:
  - Comma-ok type assertion in `parseConditions`.
  - Nil guard on `kubeClient` in `findTargetWorkload`.
  - Non-blocking path when `timeout <= 0`.
- **Unit & Live Tests**:
  - Added test cases covering retry discovery, timeout, API error, `JobSetCompleted`, `JobSetFailed`, inactive condition handling, and nil clients.
  - Verified end-to-end on live GKE cluster `test-kueue-cluster`.

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -1466,26 +1466,29 @@ func parseJobStatus(obj map[string]interface{}) (statusStr, completionTime strin
 
 func parseConditions(conditions []interface{}, statusStr *string, completionTime *string) {
 	for _, c := range conditions {
-		cond := c.(map[string]interface{})
+		cond, ok := c.(map[string]interface{})
+		if !ok {
+			continue
+		}
 		condType, _ := cond["type"].(string)
 		condStatus, _ := cond["status"].(string)
 		if condStatus == "True" {
 			switch condType {
-			case "Completed", "Succeeded":
+			case "Completed", "JobSetCompleted", "Succeeded":
 				*statusStr = "Succeeded"
 				if *completionTime == "" {
 					if transitionTime, ok := cond["lastTransitionTime"].(string); ok {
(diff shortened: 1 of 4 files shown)


SwarnaBharathiMantena (approved):
LGTM

PR #6223, opened 2026-08-27
Automate Kueue lock generation, resource validation for integration tests

rahimkhan19:
## Overview
This PR standardizes how test locks are requested, ensures consistent Kubernetes resource requests, auto-generates cluster configurations, and seamlessly integrates with our Terraform provisioning.

### 1. Pre-commit Automation & Validation
* **`validate_kueue_tests.py`**: Added a new pre-commit hook that parses the embedded bash heredocs inside Cloud Build YAML files. It enforces that any test using Kueue explicitly requests a `test-lock` and automatically fixes missing or non-standard `cpu` values in the job manifests.
* **`generate_kueue_locks.py`**: Added a pre-commit hook that dynamically extracts `test-locks/` requests from build definitions and injects them seamlessly into `dummy-device-plugin.yaml` and `kueue-setup.yaml`.

### 2. Terraform Integration
* **`apply_kueue_locks.sh`**: Created a deployment script to apply updated Kueue locks to gke clusters.
* **`kueue-locks.tf`**: Added a `null_resource` to `tools/cloud-build/provision/` so that running `terraform apply` automatically detects config modifications and executes the shell script to push the updated locks to the clusters.

### 3. Blueprint & Documentation Updates
* Added `test-infra-kueue.yaml`  blueprint and `kueue-reaper-pod.yaml` into the standard daily-tests blueprint directory.
* Updated `tools/cloud-build/README.md` and `tools/cloud-build/provision/README.md` to document the unified Kueue automation workflow and provide standardized instructions for testing new locks locally via `pre-commit`.

### DIFF
--- a/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
+++ b/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
@@ -99,7 +99,7 @@ spec:
       serviceAccountName: kueue-reaper-ksa
       containers:
       - name: reaper
-        image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
+        image: gcr.io/google.com/cloudsdktool/cloud-sdk:stable
         command: ["/bin/bash", "/scripts/reaper.sh"]
         volumeMounts:
         - name: script-volume
--- a/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
+++ b/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
@@ -55,7 +55,7 @@ data:
 
     while true; do
       # Fetch all jobs, checking their Complete and Failed conditions
-      jobs_data=$(kubectl get jobs -l build-id -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.labels.build-id}{"|"}{.status.conditions[?(@.type=="Complete")].status}{"|"}{.status.conditions[?(@.type=="Failed")].status}{"\n"}{end}' 2>/dev/null)
+      jobs_data=$(kubectl get jobs -l build-id -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.labels.build-id}{"|"}{.status.conditions[?(@.type=="Complete")].status}{"|"}{.status.conditions[?(@.type=="Failed")].status}{"\n"}{end}')
 
       if [ -n "$jobs_data" ]; then
         active_jobs=()
@@ -86,7 +86,7 @@ data:
             if [ -n "$bid" ]; then
               build_statuses["$bid"]="$bstatus"
             fi
-          done < <(gcloud builds list --filter="$filter_str" --format="value(id,status)" 2>/dev/null)
+          done < <(gcloud builds list --filter="$filter_str" --format="value(id,status)")
 
           # Check and delete jobs for cancelled/failed builds
           for item in "${active_jobs[@]}"; do
--- a/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
+++ b/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
@@ -1,0 +1,85 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kueue-reaper-role
+  namespace: default
+rules:
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["get", "list", "watch", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: kueue-reaper-rolebinding
+  namespace: default
+subjects:
+- kind: ServiceAccount
+  name: kueue-reaper-ksa
+  namespace: default
+roleRef:
+  kind: Role
+  name: kueue-reaper-role
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kueue-reaper-script
+  namespace: default
+data:
+  reaper.sh: |
+    #!/bin/bash
+    echo "Starting Centralized K8s Reaper Pod..."
+    echo "This pod monitors active jobs and cancels them if their Cloud Build is CANCELLED."
+    while true; do
+      # Fetch all jobs that have a build-id label
+      kubectl get jobs -l build-id -o custom-columns=NAME:.metadata.name,BUILD_ID:.metadata.labels.build-id --no-headers | while read job_name build_id; do
+        if [ -z "$build_id" ] || [ "$build_id" == "<none>" ]; then
+          continue
+        fi
+
+        # Check the Cloud Build status. Suppress stderr so we don't spam logs
+        status=$(gcloud builds describe "$build_id" --format="value(status)" 2>/dev/null)
+
+        if [ "$status" == "CANCELLED" ] || [ "$status" == "FAILURE" ] || [ "$status" == "TIMEOUT" ]; then
+          echo "WARNING: Cloud Build $build_id is $status!"
+          echo "Deleting Kubernetes Job $job_name to trigger graceful cleanup or remove it from the Kueue..."
+          kubectl delete job "$job_name"
+        fi
+      done
+
+      # Poll every 30 seconds
+      sleep 30
+    done
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kueue-reaper
+  namespace: default
+  labels:
+    app: kueue-reaper
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: kueue-reaper
+  template:
+    metadata:
+      labels:
+        app: kueue-reaper
+    spec:
+      serviceAccountName: kueue-reaper-ksa
+      containers:
+      - name: reaper
+        image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
+        command: ["/bin/bash", "/scripts/reaper.sh"]
+        volumeMounts:
+        - name: script-volume
+          mountPath: /scripts
+      volumes:
+      - name: script-volume
+        configMap:
+          name: kueue-reaper-script
+          defaultMode: 0777
--- a/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
+++ b/tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
@@ -55,25 +55,51 @@ data:
 
     while true; do
       # Fetch all jobs, checking their Complete and Failed conditions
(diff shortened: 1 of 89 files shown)


tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+        # Check the Cloud Build status. Suppress stderr so we don't spam logs
+        status=$(gcloud builds describe "$build_id" --format="value(status)" 2>/dev/null)
+
+        if [ "$status" == "CANCELLED" ] || [ "$status" == "FAILURE" ] || [ "$status" == "TIMEOUT" ]; then
+          echo "WARNING: Cloud Build $build_id is $status!"
+          echo "Deleting Kubernetes Job $job_name to trigger graceful cleanup or remove it from the Kueue..."
+          kubectl delete job "$job_name"
+        fi
+      done
```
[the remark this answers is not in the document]
rahimkhan19: Great catch! We absolutely need to preserve the logs, but checking .status.active won't work because suspended jobs also have no active pods and would be accidentally skipped by the reaper. I've updated the script to achieve this by explicitly checking for Complete and Failed conditions instead so we can safely preserve finished jobs while still aggressively clearing out the queue!

tools/cloud-build/provision/generate_kueue_locks.py
```diff
@@ -0,0 +1,176 @@
...
+        last_group['coveredResources'].append(f"test-locks/{lock}")
+        last_group['flavors'][0]['resources'].append({
+            "name": f"test-locks/{lock}",
+            "nominalQuota": 1
+        })
+        added += 1
+
+    if added > 0:
+        with open(filepath, 'w') as f:
+            yaml.dump_all(docs, f)
```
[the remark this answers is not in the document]
rahimkhan19: Invalid, ruamel.yaml.YAML does have a dump_all method, and it is required here to correctly format the output as multiple distinct YAML documents separated by ---. Using yaml.dump() would incorrectly serialize the resources as a single YAML array, which would break kubectl apply.

tools/cloud-build/provision/kueue-locks.tf
```diff
@@ -0,0 +1,25 @@
...
+  triggers = {
+    dummy_hash  = filemd5("${path.module}/../daily-tests/blueprints/test-infra-kueue/configs/dummy-device-plugin.yaml")
+    setup_hash  = filemd5("${path.module}/../daily-tests/blueprints/test-infra-kueue/configs/kueue-setup.yaml")
+    script_hash = filemd5("${path.module}/apply_kueue_locks.sh")
+  }
+
+  provisioner "local-exec" {
+    command = "${path.module}/apply_kueue_locks.sh"
+  }
+}
```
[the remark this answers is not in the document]
rahimkhan19: Invalid,  the apply_kueue_locks.sh script relies on a hardcoded array of 3 projects (including hpc-toolkit-gsc, which is missing from your list) and does not read a PROJECTS_LIST environment variable. Additionally, null_resource is working fine and already defined in our provider dependencies, so there is no need for this refactor.

tools/cloud-build/provision/kueue-locks.tf
```diff
@@ -0,0 +1,25 @@
...
+resource "null_resource" "apply_kueue_locks" {
+  triggers = {
+    dummy_hash  = filemd5("${path.module}/../daily-tests/blueprints/test-infra-kueue/configs/dummy-device-plugin.yaml")
+    setup_hash  = filemd5("${path.module}/../daily-tests/blueprints/test-infra-kueue/configs/kueue-setup.yaml")
+    script_hash = filemd5("${path.module}/apply_kueue_locks.sh")
+  }
+
+  provisioner "local-exec" {
+    command = "${path.module}/apply_kueue_locks.sh"
+  }
```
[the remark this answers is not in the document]
rahimkhan19: added the bash prefix to prevent permission errors, which is a great catch! However, I omitted the environment block because local-exec automatically inherits environment variables, and passing only those two Terraform variables would erroneously exclude our third test project

tools/cloud-build/provision/apply_kueue_locks.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "Applying Kueue Locks to Clusters..."
+
+ERRORS=0
+for PROJECT in "${PROJECTS[@]}"; do
+	echo "==========================================="
+	echo "Project: $PROJECT"
+	echo "==========================================="
+
+	# Check if cluster exists in this project
+	if gcloud container clusters describe "$CLUSTER_NAME" --region "$CLUSTER_REGION" --project "$PROJECT" >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
rahimkhan19: Instead of just removing the stderr suppression (which would cause ugly 404 log spam on expected skips), I switched the command to gcloud container clusters list, which silently returns empty on misses but still loudly surfaces genuine auth or network errors.

tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
```diff
@@ -0,0 +1,111 @@
...
+      app: kueue-reaper
+  template:
+    metadata:
+      labels:
+        app: kueue-reaper
+    spec:
+      serviceAccountName: kueue-reaper-ksa
+      containers:
+      - name: reaper
+        image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
```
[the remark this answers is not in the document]
rahimkhan19: We cannot use the :stable tag because it is a stripped-down variant that excludes the kubectl component, which causes the Reaper pod to immediately crash loop with kubectl: command not found.

tools/cloud-build/provision/apply_kueue_locks.sh
```diff
@@ -0,0 +1,68 @@
...
+	echo "==========================================="
+
+	# Check if cluster exists in this project without throwing a 404 error if it's missing
+	if gcloud container clusters list --region "$CLUSTER_REGION" --project "$PROJECT" --filter="name=$CLUSTER_NAME" --format="value(name)" | grep -q "^${CLUSTER_NAME}$"; then
+		echo "Getting credentials for $CLUSTER_NAME in $PROJECT..."
+		if gcloud container clusters get-credentials "$CLUSTER_NAME" --region "$CLUSTER_REGION" --project "$PROJECT"; then
+			echo "Applying dummy-device-plugin..."
+			kubectl apply -f "$CONFIGS_DIR/dummy-device-plugin.yaml" || ERRORS=1
+
+			echo "Applying kueue-setup..."
```
Neelabh94, line 55: Under `set -eo pipefail`, combining a command pipeline with `grep -q` can trigger a `SIGPIPE` (exit code 141) on `gcloud` if grep exits immediately upon matching, causing the pipeline check to spuriously fail.

Capturing the output directly into a variable is cleaner and avoids pipefail race conditions:
```bash
cluster_found=$(gcloud container clusters list --region "$CLUSTER_REGION" --project "$PROJECT" --filter="name=$CLUSTER_NAME" --format="value(name)")
if [ "$cluster_found" == "$CLUSTER_NAME" ]; then
```

tools/cloud-build/provision/apply_kueue_locks.sh
```diff
@@ -0,0 +1,68 @@
...
+	fi
+done
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+CONFIGS_DIR="$SCRIPT_DIR/../daily-tests/blueprints/test-infra-kueue/configs"
+CLUSTER_NAME="test-kueue-cluster"
+CLUSTER_REGION="us-central1"
+if [ -n "$TEST_INFRA_KUEUE_PROJECTS" ]; then
+	# Read space-separated projects into an array
+	read -r -a PROJECTS <<<"$TEST_INFRA_KUEUE_PROJECTS"
```
Neelabh94, line 35: When executed locally by a developer or via `terraform apply`, `gcloud container clusters get-credentials` modifies the operator's default `~/.kube/config` and alters their active kubectl context.

Consider isolating the credentials to a temporary kubeconfig:
```bash
TMP_KUBECONFIG=$(mktemp)
export KUBECONFIG="$TMP_KUBECONFIG"
trap 'rm -f "$TMP_KUBECONFIG"' EXIT
```

tools/cloud-build/daily-tests/validate_kueue_tests.py
```diff
@@ -0,0 +1,173 @@
...
+    for step in data.get('steps', []):
+        if 'args' in step:
+            for i, arg in enumerate(step['args']):
+                if isinstance(arg, str) and 'cat <<' in arg and 'job.yaml' in arg:
+                    lines = arg.split('\n')
+                    start_idx = -1
+                    end_idx = -1
+                    eof_marker = None
+                    for j, line in enumerate(lines):
+                        match = re.search(r'cat <<\s*[\'"]?([A-Z_]+)[\'"]?\s*>.*job\.yaml', line)
```
Neelabh94: Using `'cat <<' in arg` requires an exact single space between `cat` and `<<`. This could miss valid bash heredocs formatted without spaces (e.g. `cat<<EOF`) or tab-stripping heredocs (`cat <<-EOF`).

Consider using regex to match variations:
```python
if isinstance(arg, str) and re.search(r'cat\s*<<-?', arg) and 'job.yaml' in arg:
```

tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
```diff
@@ -0,0 +1,138 @@
...
+        command: ["/bin/bash", "/scripts/reaper.sh"]
+        volumeMounts:
+        - name: script-volume
+          mountPath: /scripts
+      volumes:
+      - name: script-volume
```
Neelabh94, line 135: `gcr.io` is deprecated across Google Cloud in favor of Artifact Registry, and unpinned `:latest` can introduce unexpected breakages on upstream updates.

Since the GKE cluster is located in `us-central1` and the node pool already has `artifactregistry.reader` assigned, consider pointing to Artifact Registry:
* Use the repository's existing regional test runner image: `us-central1-docker.pkg.dev/hpc-toolkit-dev/hpc-toolkit-repo/test-runner:latest` (or pinned version) which already bundles `gcloud` and `kubectl`.
* Or use the public Artifact Registry mirror: `us-docker.pkg.dev/google.com/cloudsdktool/google-cloud-cli:stable`.

Neelabh94 (commented):
The overall design and migration to `terraform_data` look very clean. I've left a few minor suggestions regarding bash pipeline robustness, kubeconfig isolation, and container image registry alignment.

tools/cloud-build/provision/apply_kueue_locks.sh
rahimkhan19: done

tools/cloud-build/provision/apply_kueue_locks.sh
rahimkhan19: Makes total sense, done

tools/cloud-build/daily-tests/validate_kueue_tests.py
rahimkhan19: done

tools/cloud-build/daily-tests/blueprints/test-infra-kueue/configs/kueue-reaper-pod.yaml
rahimkhan19: agreed on moving away from gcr.io. Updated the Reaper pod to use the repository's regional test runner image (us-central1-docker.pkg.dev/hpc-toolkit-dev/hpc-toolkit-repo/test-runner:latest)

Neelabh94 (approved):

PR #6252, opened 2026-09-02
Fix transient webhook timeouts on kueue job submission

sudheer-quad:
This PR prevents the Cloud Build pipeline from terminating  immediately during kueue job submission due to `set -e` when the GKE control plane or Kueue mutating webhook (mjob.kb.io) experiences temporary latency.

### Submission Checklist

### DIFF
--- a/tools/cloud-build/submit_and_monitor_kueue_job.sh
+++ b/tools/cloud-build/submit_and_monitor_kueue_job.sh
@@ -57,7 +57,26 @@ ATTEMPT=1
 
 while true; do
 	echo "=== ATTEMPT $ATTEMPT: Submitting Kueue Job ==="
-	kubectl apply -f /workspace/job.yaml
+
+	SUBMIT_RETRIES=3
+	SUBMIT_ATTEMPT=1
+
+	while true; do
+		echo "Executing job submission (attempt ${SUBMIT_ATTEMPT}/${SUBMIT_RETRIES})..."
+
+		if kubectl apply -f /workspace/job.yaml; then
+			break
+		fi
+
+		if ((SUBMIT_ATTEMPT >= SUBMIT_RETRIES)); then
+			echo "ERROR: Failed to apply job manifest after ${SUBMIT_RETRIES} attempts." >&2
+			exit 1
+		fi
+
+		echo "WARNING: kubectl apply failed. Retrying in 5s..." >&2
+		sleep 5
+		SUBMIT_ATTEMPT=$((SUBMIT_ATTEMPT + 1))
+	done
 
 	set +e
 	(


tools/cloud-build/submit_and_monitor_kueue_job.sh
```diff
@@ -57,7 +57,26 @@ ATTEMPT=1
 
 while true; do
 	echo "=== ATTEMPT $ATTEMPT: Submitting Kueue Job ==="
-	kubectl apply -f /workspace/job.yaml
+
+	SUBMIT_RETRIES=5
```
Neelabh94: Let's reduce the retries to 3

tools/cloud-build/submit_and_monitor_kueue_job.sh
```diff
@@ -57,7 +57,26 @@ ATTEMPT=1
...
+			break
+		fi
+
+		if ((SUBMIT_ATTEMPT >= SUBMIT_RETRIES)); then
+			echo "ERROR: Failed to apply job manifest after ${SUBMIT_RETRIES} attempts." >&2
+			exit 1
+		fi
+
+		echo "WARNING: kubectl apply failed. Retrying in 10s..." >&2
+		sleep 10
```
Neelabh94: 10 seconds is too long of a wait, I believe 5 second should be enough here, wdyt?

Neelabh94 (changes requested):

tools/cloud-build/submit_and_monitor_kueue_job.sh
```diff
@@ -57,7 +57,26 @@ ATTEMPT=1
...
 	echo "=== ATTEMPT $ATTEMPT: Submitting Kueue Job ==="
-	kubectl apply -f /workspace/job.yaml
+
+	SUBMIT_RETRIES=5
+	SUBMIT_ATTEMPT=1
+
+	while true; do
+		echo "Executing job submission (attempt ${SUBMIT_ATTEMPT}/${SUBMIT_RETRIES})..."
+
+		if kubectl apply -f /workspace/job.yaml; then
```
Neelabh94: Currently, this retries on any non-zero exit code from `kubectl apply`. If there is an invalid field or syntax error in `/workspace/job.yaml`, `kubectl apply` fails with a permanent schema error, but the loop will still sleep and retry across all attempts before failing—unnecessarily stalling CI feedback.

Consider capturing the output so we fail fast on permanent errors and only retry on transient webhook/connection timeouts:
```bash
		if APPLY_OUT=$(kubectl apply -f /workspace/job.yaml 2>&1); then
			echo "$APPLY_OUT"
			break
		fi
		echo "$APPLY_OUT"

		if ((SUBMIT_ATTEMPT >= SUBMIT_RETRIES)); then
			echo "ERROR: Failed to apply job manifest after ${SUBMIT_RETRIES} attempts." >&2
			exit 1
		fi

		# Fail fast on non-transient schema or validation errors
		if ! echo "$APPLY_OUT" | grep -iqE "(webhook|timeout|context deadline exceeded|connection refused)"; then
			echo "ERROR: Non-retriable apply failure detected. Failing immediately." >&2
			exit 1
		fi
```

Neelabh94 (dismissed):

tools/cloud-build/submit_and_monitor_kueue_job.sh
```diff
@@ -57,7 +57,34 @@ ATTEMPT=1
...
+		fi
+		echo "$APPLY_OUT"
+
+		if ((SUBMIT_ATTEMPT >= SUBMIT_RETRIES)); then
+			echo "ERROR: Failed to apply job manifest after ${SUBMIT_RETRIES} attempts." >&2
+			exit 1
+		fi
+
+		# Fail fast on non-transient schema or validation errors
+		if ! echo "$APPLY_OUT" | grep -iqE "(webhook|timeout|context deadline exceeded|connection refused)"; then
```
rahimkhan19: The current regex is too restrictive and causes the pipeline to prematurely fail on the very first attempt during transient errors. For example, if kubectl apply hits connect: connection timed out, etcdserver: request timed out, or The connection to the server ... was refused, none of these match the regex keywords (timed out != timeout).

Because this loop calls exit 1, it kills the entire script immediately on Attempt 1, before the job is even submitted and completely bypassing our outer retry loop (check_retriable_error.sh).

Let's remove the regex check entirely and retry unconditionally up to 3 times

rahimkhan19 (approved):
LGTM

Neelabh94 (approved):

PR #5859, opened 2026-06-26
"feat(gke): Add G4 Confidential VM support with Blackwell GPUs and CMEK storage in cluster toolkit""

shubpal07:
The main feature PR GoogleCloudPlatform/cluster-toolkit#5817 was reverted by  GoogleCloudPlatform/cluster-toolkit#5857 due to merge conflicts while merging main back to develop.
Re-created the feature PR.

PR #6254, opened 2026-09-02
fix(gke-cluster): emit confidential_nodes only when enabled

lasse-ii:
The `gke-cluster` module sends `confidential_nodes { enabled = var.enable_confidential_nodes }` unconditionally (since #5859). The google provider marks `confidential_nodes` ForceNew, so for any cluster created before that block existed, the diff `absent → { enabled = false }` plans a **full cluster replacement** — with every node pool replaced behind it — even though nothing changes semantically.

Observed on a 15-pool GKE cluster when moving from a pre-#5859 toolkit to v1.98.0+:

```
# module.a4x-cluster.google_container_cluster.gke_cluster must be replaced
      + confidential_nodes { # forces replacement
          + enabled = false # forces replacement
```

This change wraps the block in `dynamic "confidential_nodes"` so it is only emitted when `enable_confidential_nodes = true`. Clusters that enable confidential nodes are unaffected; clusters that do not no longer see a diff.

**Tested:** `terraform plan` against the live cluster above with this change shows no cluster or node-pool changes; without it, 36 to add / 16 to destroy. `terraform fmt -check` passes.

### DIFF
--- a/modules/scheduler/gke-cluster/main.tf
+++ b/modules/scheduler/gke-cluster/main.tf
@@ -402,9 +402,15 @@ resource "google_container_cluster" "gke_cluster" {
     }
   }
 
-  confidential_nodes {
-    enabled                    = var.enable_confidential_nodes
-    confidential_instance_type = var.confidential_instance_type
+  # Emit the block only when enabled. The provider marks confidential_nodes
+  # ForceNew, so sending `enabled = false` to a cluster created without the
+  # block plans a full cluster replacement on every pre-existing cluster.
+  dynamic "confidential_nodes" {
+    for_each = var.enable_confidential_nodes ? [1] : []
+    content {
+      enabled                    = true
+      confidential_instance_type = var.confidential_instance_type
+    }
   }
 
 


Neelabh94:
@lasse-ii could you please also sign the CLA so we can take this PR forward? 

Thanks

shubpal07:
Triggered PR-test-gke-g4-confidential

lasse-ii:
Done, thanks

shubpal07 (approved):
LGTM. Thanks for catching this upgrade regression - wrapping `confidential_nodes` in a dynamic block correctly prevents `ForceNew` cluster replacements and aligns `google_container_cluster` with the existing pattern in our node pool modules.

SwarnaBharathiMantena (approved):
LGTM. 

@shubpal07 - Can we please clean up the comment in the main.tf in a followup PR?
I agree it adds context but the purpose of dynamic block / variable is well understood and this way, the files are cleaner. Thanks!

PR #6274, opened 2026-09-05
fix(gke): resolve TPU 7x NAP workload policies and improve GKE auth endpoint selection

Neelabh94:
### Problem
1. **TPU 7x NAP Placement Policy**: When submitting a TPU 7x workload to a GKE cluster with Node Auto-Provisioning (NAP) where no matching node pool exists (scale up from 0):
   - GKE NAP's autoscaling controller attempts to create a GCE Managed Instance Group (MIG) with a default `groupPlacementPolicy` (`collocation: COLLOCATED`).
   - GCE rejects the MIG creation request with `HTTP 400 (UNSUPPORTED_OPERATION): Creation of a managed instance group with tpu7x-standard-4t machine type with placement policy is not supported. Use workload policy instead.`
   - GKE NAP requires incoming Pods to specify `cloud.google.com/placement-policy-name: <workload-policy-name>` in `nodeSelector`. Cluster Toolkit previously injected `cloud.google.com/gke-placement-group` instead and did not auto-discover or create the required GCE Workload Policy (`tpu7x-{cores}-{topology}-placement-policy`).
2. **GKE Auth DNS Endpoint Selection**: When refreshing cluster credentials via `gcloud container clusters get-credentials`, passing `--dns-endpoint` when public IP endpoints are active causes HTTP 431 ("Request Header Fields Too Large") failures on corporate networks and proxies due to large authorization header sizes.

### Solution
1. **Workload Policy Resolution & Provisioning (`pkg/orchestrator/gke/nap.go`)**:
   - Implemented `resolveTPUWorkloadPolicy` to discover existing policies from current cluster node pools or via `gcloud compute resource-policies describe/list`.
   - Provisions canonical workload policies (`tpu7x-{chips}-{topology}-placement-policy`) via `gcloud compute resource-policies create workload-policy` if missing (skipped on dry-runs).
   - Unified resource policy caching via `resourcePolicyCache` to share query results between dynamic slicing and NAP policy resolution.
   - Handled non-fatal IAM permission errors with clear warnings when querying or creating policies.
2. **Clean Label Assignment (`pkg/orchestrator/gke/scheduling.go`)**:
   - Added `IsTPU` to `SchedulingOptions` and directly assigned `cloud.google.com/placement-policy-name` for TPUs and `cloud.google.com/gke-placement-group` for non-TPUs in `getNodeSelector`, removing post-hoc map mutations.
3. **Hardware Guard (`pkg/orchestrator/gke/resource_resolver.go`)**:
   - Strictly matches GA `tpu7x` hardware and enforces workload policies exclusively for multi-host TPU 7x workloads (`NodesPerSlice > 1`).
4. **Control Plane Endpoint Selection (`pkg/orchestrator/gke/gke_job_orchestrator.go`, `types.go`)**:
   - Updated `refreshGKEAuth` via `shouldUseDNSEndpoint()` to only pass `--dns-endpoint` when external DNS traffic is enabled AND public IP endpoint is disabled or absent, allowing clients to cleanly connect via the public IP endpoint without HTTP 431 errors.

### Testing
- Added unit tests in `nap_test.go` covering policy preservation, node pool discovery, canonical describe, regional list fallback, dry-run safety, and auto-creation.
- Added unit tests in `scheduling_test.go` verifying placement policy label assignment for TPU vs non-TPU workloads.
- Added unit tests in `gke_job_orchestrator_test.go` for `shouldUseDNSEndpoint` and `refreshGKEAuth` covering all endpoint configurations.
- All pre-commit presubmits and tests pass with exit code 0.

### DIFF
--- a/pkg/orchestrator/gke/nap.go
+++ b/pkg/orchestrator/gke/nap.go
@@ -329,12 +329,21 @@ func (g *GKEOrchestrator) resolveTPUWorkloadPolicy(machineType, topology, cluste
 	return canonicalPolicyName, nil
 }
 
+// isStaticWorkloadPolicyMode checks whether the accelerator topology mode provides a
+// static, pre-connected accelerator topology suitable for non-dynamic workloads.
+// In GCE, standard static workload policies omit this field (defaulting to auto-connected)
+// or explicitly specify "AUTO_CONNECT".
+func isStaticWorkloadPolicyMode(mode string) bool {
+	trimmed := strings.TrimSpace(mode)
+	return trimmed == "" || strings.EqualFold(trimmed, "AUTO_CONNECT")
+}
+
 func (g *GKEOrchestrator) getCachedTPUWorkloadPolicy(canonicalPolicyName, topology string) string {
 	if g.resourcePolicyCache == nil {
 		return ""
 	}
 	cached := g.resourcePolicyCache[canonicalPolicyName]
-	if cached != nil && strings.EqualFold(cached.Type, "HIGH_THROUGHPUT") && cached.AcceleratorTopology == topology {
+	if cached != nil && strings.EqualFold(cached.Type, "HIGH_THROUGHPUT") && cached.AcceleratorTopology == topology && isStaticWorkloadPolicyMode(cached.AcceleratorTopologyMode) {
 		return canonicalPolicyName
 	}
 	return ""
@@ -342,7 +351,7 @@ func (g *GKEOrchestrator) getCachedTPUWorkloadPolicy(canonicalPolicyName, topolo
 
 // findPolicyInClusterNodePools inspects the cluster's existing node pools for an active
 // placement policy matching the requested machine type and TPU topology.
-// Dynamic slicing policies (PROVISION_ONLY) are skipped for static workloads.
+// Non-static policies (such as dynamic slicing PROVISION_ONLY) are skipped for static workloads.
 func (g *GKEOrchestrator) findPolicyInClusterNodePools(machineType, topology string) string {
 	for _, np := range g.clusterDesc.NodePools {
 		if strings.EqualFold(np.Config.MachineType, machineType) && np.PlacementPolicy != nil && np.PlacementPolicy.PolicyName != "" {
@@ -352,8 +361,8 @@ func (g *GKEOrchestrator) findPolicyInClusterNodePools(machineType, topology str
 					mode = cached.AcceleratorTopologyMode
 				}
 			}
-			if strings.EqualFold(mode, "PROVISION_ONLY") {
-				continue // Skip dynamic slicing policies for static workloads
+			if !isStaticWorkloadPolicyMode(mode) {
+				continue // Skip non-static (e.g. PROVISION_ONLY) policies for static workloads
 			}
 			topo := np.PlacementPolicy.TpuTopology
 			if topo == "" && np.Config.Labels != nil {
@@ -397,36 +406,48 @@ func (g *GKEOrchestrator) resolvePolicyRegionAndProject(clusterLocation, project
 // discoverRegionalWorkloadPolicy searches for an existing HIGH_THROUGHPUT workload policy
 // in the specified GCP region and project. It first checks for the canonical policy name,
 // and if not found or incompatible, falls back to querying regional policies matching the topology.
+// Only policies with static auto-connected topology modes (empty or AUTO_CONNECT) are matched.
 func (g *GKEOrchestrator) discoverRegionalWorkloadPolicy(canonicalPolicyName, region, projectID, topology string) (string, error) {
 	policy, err := g.describeResourcePolicyCached(canonicalPolicyName, region, projectID)
 	if err != nil {
 		return "", err
 	}
 	if policy != nil {
-		if strings.EqualFold(policy.Type, "HIGH_THROUGHPUT") && policy.AcceleratorTopology == topology {
+		if strings.EqualFold(policy.Type, "HIGH_THROUGHPUT") && policy.AcceleratorTopology == topology && isStaticWorkloadPolicyMode(policy.AcceleratorTopologyMode) {
 			logging.Info("Discovered existing workload policy %q in region %s", canonicalPolicyName, region)
 			return canonicalPolicyName, nil
 		}
-		logging.Warn("Found policy %q in region %s, but its type (%q) or topology (%q) does not match expected (HIGH_THROUGHPUT, %s). Skipping it.",
-			canonicalPolicyName, region, policy.Type, policy.AcceleratorTopology, topology)
+		logging.Warn("Found policy %q in region %s, but its type (%q), mode (%q), or topology (%q) does not match expected (HIGH_THROUGHPUT, static auto-connected, %s). Skipping it.",
+			canonicalPolicyName, region, policy.Type, policy.AcceleratorTopologyMode, policy.AcceleratorTopology, topology)
 	}
 
 	filter := fmt.Sprintf("region:( %s ) AND workloadPolicy.acceleratorTopology=%s AND workloadPolicy.type=HIGH_THROUGHPUT", region, topology)
-	listRes := g.executor.ExecuteCommand("gcloud", "compute", "resource-policies", "list", "--project="+projectID, "--filter="+filter, "--format=value(name)")
+	listRes := g.executor.ExecuteCommand("gcloud", "compute", "resource-policies", "list", "--project="+projectID, "--filter="+filter, "--format=value(name,workloadPolicy.acceleratorTopologyMode)")
 	if listRes.ExitCode == 0 && strings.TrimSpace(listRes.Stdout) != "" {
-		names := strings.Fields(listRes.Stdout)
-		if len(names) > 0 {
-			if g.resourcePolicyCache == nil {
-				g.resourcePolicyCache = make(map[string]*GCEWorkloadPolicy)
+		for _, line := range strings.Split(strings.TrimSpace(listRes.Stdout), "\n") {
+			fields := strings.Fields(line)
+			if len(fields) == 0 {
+				continue
 			}
-			g.resourcePolicyCache[names[0]] = &GCEWorkloadPolicy{
-				Name:                names[0],
-				Region:              region,
-				AcceleratorTopology: topology,
-				Type:                "HIGH_THROUGHPUT",
+			name := fields[0]
+			mode := ""
+			if len(fields) > 1 {
+				mode = fields[1]
 			}
-			logging.Info("Discovered matching workload policy %q for topology %s in region %s", names[0], topology, region)
-			return names[0], nil
+			if isStaticWorkloadPolicyMode(mode) {
+				if g.resourcePolicyCache == nil {
+					g.resourcePolicyCache = make(map[string]*GCEWorkloadPolicy)
+				}
+				g.resourcePolicyCache[name] = &GCEWorkloadPolicy{
+					Name:                    name,
+					Region:                  region,
+					AcceleratorTopology:     topology,
+					AcceleratorTopologyMode: mode,
+					Type:                    "HIGH_THROUGHPUT",
+				}
+				logging.Info("Discovered matching workload policy %q for topology %s in region %s", name, topology, region)
+				return name, nil
+			}
 		}
 	}
 	return "", nil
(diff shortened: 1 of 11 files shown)


pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -265,44 +271,63 @@ func (g *GKEOrchestrator) checkNodePoolsDynamicSlicing(requestedMachineName stri
...
 
-	g.policyCache[shortPolicyName] = mode
-	return mode, nil
-}
-
-func (g *GKEOrchestrator) fetchComputeResourcePolicyTopologyMode(shortPolicyName, location, projectID string) (string, error) {
 	region := shell.ExtractRegion(location)
-
-	res := g.executor.ExecuteCommand("gcloud", "compute", "resource-policies", "describe", shortPolicyName, "--region="+region, "--project="+projectID, "--format=value(workloadPolicy.acceleratorTopologyMode)")
+	res := g.executor.ExecuteCommand("gcloud", "compute", "resource-policies", "describe", shortPolicyName, "--region="+region, "--project="+projectID, "--format=json")
```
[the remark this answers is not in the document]
Neelabh94: It is impossible for the project and or location to be empty at this stage since this is checked at the beginning with every execution of the job submit command.

pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -265,44 +271,63 @@ func (g *GKEOrchestrator) checkNodePoolsDynamicSlicing(requestedMachineName stri
...
+func (g *GKEOrchestrator) describeResourcePolicyCached(policyName, location, projectID string) (*GCEWorkloadPolicy, error) {
+	if g.resourcePolicyCache == nil {
+		g.resourcePolicyCache = make(map[string]*GCEWorkloadPolicy)
 	}
 
-	mode, err := g.fetchComputeResourcePolicyTopologyMode(shortPolicyName, location, projectID)
-	if err != nil {
-		return "", err
+	shortPolicyName := path.Base(policyName)
+	if policy, found := g.resourcePolicyCache[shortPolicyName]; found {
```
agrawalkhushi18, line 295: In describeResourcePolicyCached, shortPolicyName := path.Base(policyName) is used directly as the cache map key. If an orchestrator instance evaluates clusters across multiple regions or projects that share standard policy names, cache collisions will return incorrect metadata from the first project/region queried.
Suggested fix :
```
shortPolicyName := path.Base(policyName)
cacheKey := fmt.Sprintf("%s/%s/%s", projectID, region, shortPolicyName)
if policy, found := g.resourcePolicyCache[cacheKey]; found {
    return policy, nil
}
...
g.resourcePolicyCache[cacheKey] = policy
```

pkg/orchestrator/gke/nap.go
```diff
@@ -272,3 +277,160 @@ func (g *GKEOrchestrator) populateNAPFlavors(flavors map[string]FlavorCapacity)
...
+	return ""
+}
+
+func getCanonicalTPUWorkloadPolicyName(topology string) string {
+	chips := calculateChipsFromTopology(topology)
+	tensorcores := chips * 2
+	return fmt.Sprintf("tpu7x-%d-%s-placement-policy", tensorcores, topology)
+}
+
+func (g *GKEOrchestrator) resolvePolicyRegionAndProject(clusterLocation, projectID string) (string, string) {
```
agrawalkhushi18, line 391: When a user invokes gcluster job submit --name my-job --cluster my-cluster without an explicit --location flag, clusterLocation is empty.

resolvePolicyRegionAndProject checks g.clusterDesc.Locations (slice) and g.clusterZones, but misses g.clusterDesc.Location (the standard singular location string populated from gcloud container clusters describe). On zonal clusters where Locations is empty, region resolves to "", causing resolveTPUWorkloadPolicy to silently skip policy discovery and auto-creation in non-dry-run mode.

Suggested fix :
```
func (g *GKEOrchestrator) resolvePolicyRegionAndProject(clusterLocation, projectID string) (string, string) {
    loc := clusterLocation
    if loc == "" {
        loc = g.clusterDesc.Location
    }
    if loc == "" && len(g.clusterDesc.Locations) > 0 {
        loc = g.clusterDesc.Locations[0]
    }
    if loc == "" && len(g.clusterZones) > 0 {
        loc = g.clusterZones[0]
    }
    region := shell.ExtractRegion(loc)
    if projectID == "" {
        projectID = g.projectID
    }
    return region, projectID
}
```

pkg/orchestrator/gke/nap.go
```diff
@@ -272,3 +277,160 @@ func (g *GKEOrchestrator) populateNAPFlavors(flavors map[string]FlavorCapacity)
...
+		region = shell.ExtractRegion(g.clusterZones[0])
+	}
+	if projectID == "" {
+		projectID = g.projectID
+	}
+	return region, projectID
+}
+
+func (g *GKEOrchestrator) discoverRegionalWorkloadPolicy(canonicalPolicyName, region, projectID, topology string) (string, error) {
+	policy, err := g.describeResourcePolicyCached(canonicalPolicyName, region, projectID)
```
agrawalkhushi18, line 411: In discoverRegionalWorkloadPolicy, if a policy with canonicalPolicyName already exists in the project/region, it is returned immediately without verifying that policy.Type == "HIGH_THROUGHPUT" and policy.AcceleratorTopology == topology.

If an incompatible policy with that canonical name already exists (e.g. created manually as a standard GROUP_PLACEMENT or with a different topology like 2x2x4), gcluster will accept it, and the failure will only surface later during GKE pod scheduling or NAP MIG scale-up with an obscure GCE 400 rejection.
Can we check this once?

pkg/orchestrator/gke/scheduling.go
```diff
@@ -31,13 +31,18 @@ type SchedulingOptions struct {
...
 	IsStaticSlicing    bool
+	IsTPU              bool
 }
 
 func getNodeSelector(opts SchedulingOptions) (map[string]string, error) {
 	nodeSelector := make(map[string]string)
 
 	if opts.PlacementPolicy != "" {
-		nodeSelector["cloud.google.com/gke-placement-group"] = opts.PlacementPolicy
+		if opts.IsTPU {
```
agrawalkhushi18, line 46: Can we recheck the logic once here , I think the distinction between these labels should not exactly be based on whether it is a TPU or not.

For the `!opts.IsTPU` , we are injecting opts.PlacementPolicy into `cloud.google.com/gke-placement-group`.
gke-placement-group expects an arbitrary user-defined string (e.g., my-job-group-1) so NAP can dynamically generate a compact placement policy on the fly. If `opts.PlacementPolicy` actually contains the name of a pre-created GCE Compact Placement Policy, the correct label to use is `cloud.google.com/placement-policy-name`

We can separate the concept of a "Placement Group Identifier" (dynamic) from a "Placement Policy Name" (pre-created GCE resource). Furthermore, the TPU logic must check the accelerator type/version before injecting the policy name label. There might be the case where older tpu versions dont support workload-policy and might use a different label.

pkg/orchestrator/gke/nap.go
Neelabh94: A couple of quick notes on why we can keep the current implementation:

* CLI Flag Validation: location is a required flag enforced at the root CLI level in cmd/job/job.go (PersistentPreRunE). If omitted without a config default, the command errors out before reaching the orchestrator, so clusterLocation is guaranteed to be non-empty during job submission.
* Struct Definition: g.clusterDesc is typed as gkeCluster (in pkg/orchestrator/gke/types.go), which defines Locations []string rather than a singular Location string field. Adding g.clusterDesc.Location would result in a compile error.
* Zonal Handling: In GKE describe output, zonal clusters populate locations with [ "<zone>" ] (e.g. ["us-central1-a"]). If clusterLocation were ever empty, shell.ExtractRegion(g.clusterDesc.Locations[0]) correctly extracts the region (us-central1).

Given this, the existing fallback chain already covers all scenarios safely.

pkg/orchestrator/gke/resource_resolver.go
Neelabh94: We can keep the simple shortPolicyName key here for two reasons:

* CLI Execution Lifecycle: gcluster runs as a short-lived CLI command where the orchestrator is bound to a single cluster, region, and project per invocation (orc.Initialize). There is no multi-region or cross-project state sharing within a single process.
* Cross-Function Cache Sharing: g.resourcePolicyCache is shared across checkNodePoolsDynamicSlicing and resolveTPUWorkloadPolicy in nap.go (which performs lookups using canonicalPolicyName and path.Base(np.PlacementPolicy.PolicyName)). Changing the key format here would cause a key mismatch and prevent cache hits across those functions without refactoring all call sites.

Additionally, with the validation check added based on the previous comment, any policy retrieved from cache is strictly verified for HIGH_THROUGHPUT and matching topology before being used.

pkg/orchestrator/gke/nap.go
Neelabh94: This makes sense. For manually created policies it is a (small) possibility, have made the changes, thanks for pointing this out.

pkg/orchestrator/gke/scheduling.go
Neelabh94: Here is the context on why the IsTPU check is structured this way:

* Non-TPU Backward Compatibility: The code currently maps `--placement-policy` to `cloud.google.com/gke-placement-group` for existing workloads. Retaining this for `!opts.IsTPU` preserves full backward compatibility for current GPU/CPU users without introducing breaking changes or new CLI flags.

* Older TPU Generations (v4, v5e, v5p, v6e): Older TPUs do not use GCE workload policies or placement policies; their slice co-locality and topology are handled natively by GKE via TPU topology labels (`cloud.google.com/gke-tpu-topology`). In our code, automatic policy resolution is strictly guarded by if `isTPU7x` in `resolveTPU7xWorkloadPolicyIfRequired`. For older TPUs, `opts.PlacementPolicy` is left empty, so no placement label is injected at all.

* TPU Label Compatibility: If a placement policy is attached to a TPU workload in GKE (as required for TPU 7x multi-host NAP), the GKE API only recognizes cloud.google.com/placement-policy-name. cloud.google.com/gke-placement-group is exclusively for GCE compact placement groups on standard VM/GPU nodes and is not supported for TPUs. (This also matches the label convention used in XPK).

Because of this, distinguishing by `opts.IsTPU` correctly routes TPU workloads to `placement-policy-name` while preserving existing behavior for non-TPU workloads.

To ensure this is clear for future maintenance I have added some inline comments throughout the code to make it easy to follow. 

Thanks

agrawalkhushi18: Thanks for the detailed context .
The `opts.IsTPU` routing makes sense for the scope of this PR to unblock TPU 7x NAP while preserving backward compatibility for existing GPU/CPU compact placement groups.
As a note for future work: modern GPU node pools (e.g., A4X, A3 Ultra) that utilize GCE Workload Policies also rely on `cloud.google.com/placement-policy-name` rather than `cloud.google.com/gke-placement-group`. We can verify it once from GPU job submission perspective.

pkg/orchestrator/gke/scheduling.go
```diff
@@ -31,13 +31,23 @@ type SchedulingOptions struct {
...
 
 	if opts.PlacementPolicy != "" {
-		nodeSelector["cloud.google.com/gke-placement-group"] = opts.PlacementPolicy
+		// In GKE, placement policy labels differ by accelerator architecture:
+		// 1. TPUs (specifically TPU 7x multi-host NAP): GKE requires referencing a pre-created
+		//    GCE Workload Policy (HIGH_THROUGHPUT) via "cloud.google.com/placement-policy-name".
+		// 2. Non-TPUs (GPUs/CPUs): GKE NAP uses "cloud.google.com/gke-placement-group" to dynamically
+		//    generate compact placement groups on the fly for co-located VM scheduling.
+		if opts.IsTPU {
+			nodeSelector["cloud.google.com/placement-policy-name"] = opts.PlacementPolicy
```
agrawalkhushi18, line 47: If a user specifies a full GCP resource URI via `--placement-policy` (e.g. `projects/my-project/regions/us-central1/resourcePolicies/my-policy`), passing `opts.PlacementPolicy` directly into `nodeSelector` causes Kubernetes API validation to fail on `kubectl apply` because label values cannot contain forward slashes (`/`).

Can we sanitize `opts.PlacementPolicy` with `path.Base` here (similar to how it's handled in `findPolicyInClusterNodePools` and `describeResourcePolicyCached`)?

```suggestion
	if opts.PlacementPolicy != "" {
		shortPolicy := path.Base(opts.PlacementPolicy)
		if opts.IsTPU {
			nodeSelector["cloud.google.com/placement-policy-name"] = shortPolicy
		} else {
			nodeSelector["cloud.google.com/gke-placement-group"] = shortPolicy
		}
	}

**Review by agrawalkhushi18 — DISMISSED** [review]
One nit comment. Otherwise, LGTM

File: pkg/orchestrator/gke/nap.go [hunk 7]
```diff
@@ -272,3 +277,198 @@ func (g *GKEOrchestrator) populateNAPFlavors(flavors map[string]FlavorCapacity)
...
+		}
+		logging.Warn("Found policy %q in region %s, but its type (%q) or topology (%q) does not match expected (HIGH_THROUGHPUT, %s). Skipping it.",
+			canonicalPolicyName, region, policy.Type, policy.AcceleratorTopology, topology)
+	}
+
+	filter := fmt.Sprintf("region:( %s ) AND workloadPolicy.acceleratorTopology=%s AND workloadPolicy.type=HIGH_THROUGHPUT", region, topology)
+	listRes := g.executor.ExecuteCommand("gcloud", "compute", "resource-policies", "list", "--project="+projectID, "--filter="+filter, "--format=value(name)")
+	if listRes.ExitCode == 0 && strings.TrimSpace(listRes.Stdout) != "" {
+		names := strings.Fields(listRes.Stdout)
+		if len(names) > 0 {
```
SwarnaBharathiMantena: Similar to how the PR specifically adds the if strings.EqualFold(mode, "PROVISION_ONLY") { continue } when scanning existing cluster node pools, 
I think we need filtering out PROVISION_ONLY in the gcloud compute resource-policies list regional fallback query as well.

If we do not filter out here, then a static NAP workload can accidentally receive a dynamic slicing policy (PROVISION_ONLY):

1. GKE NAP successfully creates the node pool and Compute Engine boots the tpu7x-standard-4t VMs.
2. However, because the policy is PROVISION_ONLY, Compute Engine does not connect the TPU inter-chassis links.
3. Because the job is running via standard NAP (not through the GKE Slice Admission Controller / OCS orchestrator), nothing is listening to connect the chips on demand.
4. When the multi-node distributed workload starts (e.g., PyTorch / JAX running jax.distributed.initialize()), it crashes or hangs indefinitely because the TPU chips cannot communicate with each other across VMs.

SwarnaBharathiMantena: Again, this logic I suggested seems to focus on NAP usecase. How about non-NAP usecase. We need to cross verify that scenario as well.

Neelabh94: Good point Swarna! I think instead of putting a negative check (!= "PROVISION_ONLY"), I'll update the code to check for the allowed values ("" and "AUTO_CONNECT"), this way if any more `mode` values are introduced in the future too, our code will continue to work unaffected, wdyt?
Will update the same for the check for policies on existing node pools

pkg/orchestrator/gke/scheduling.go
Neelabh94: This is a good point Khushi. I also realize that there are pros and cons associated with silent trimming, especially if the user provides a policy of a different project that may or may not exist in this specific project.
Since we currently do not validate the existence of the policy passed by the user, this will go unnoticed and can cause the pod to remain stuck or PENDING with the error:` Failed to scale up: resourcePolicy not found`.

It is an actual problem that we should solve, but to keep this PR focused, I prefer to do it in a follow-up PR, does that work for you?

SwarnaBharathiMantena (dismissed):
LGTM

SwarnaBharathiMantena (dismissed):

pkg/orchestrator/gke/scheduling.go
agrawalkhushi18: Sure, thanks!

agrawalkhushi18 (dismissed):

agrawalkhushi18 (approved):

PR #6287, opened 2026-09-08
fix: resolve MTC CSI driver race condition and Kueue flavor-default pollution

Neelabh94:
### Description
This PR resolves two related issues affecting GKE job execution, specifically causing timeouts in Multi-Tier Checkpointing (MTC) validation:

1. **MTC Driver Restart Race Condition**:
   - GKE's HighScaleCheckpointing addon dynamically names the MTC driver DaemonSet `multitier-driver-<uuid>`.
   - `gcluster` hardcoded the DaemonSet name as `"multitier-driver"`. When updating Workload Identity credentials, it failed to patch the DaemonSet and fell back to deleting driver pods across the cluster. It then immediately checked for `"multitier-driver"`, received `NotFound`, and exited without waiting.
   - Job pods scheduled while MTC driver pods were down failed volume mounting (`FailedMount: driver name multitier-checkpoint.csi.storage.gke.io not found`), triggering Kubelet backoffs.
   - **Fix**: Updated `getMTCDaemonSet` to dynamically discover `multitier-driver-*` DaemonSets in the namespace. Updated `restartMTCDriverPods` to perform a rolling restart via patch before pod deletion fallback, and `waitForMTCDriverDaemonSetReady` to poll the actual DaemonSet until 100% rollout completion.

2. **Kueue `flavor-default` Node Pool Pollution**:
   - In clusters with multiple generic CPU node pools (`debug`, `local-ssd`, `hp-pool`), all CPU pools map to `flavor-default`.
   - During capacity calculation, `processNodePoolCapacity` set `cloud.google.com/gke-nodepool = np.Name`, overwriting `flavor-default`'s nodeLabels with the last pool iterated.
   - Because GKE node pool iteration order is non-deterministic, `flavor-default` was arbitrarily pinned to a specific node pool (such as `hp-pool`), bypassing other valid CPU pools and preventing Kueue from scheduling across available CPU capacity.
   - **Fix**: Restricted `cloud.google.com/gke-nodepool` labeling in `processNodePoolCapacity` strictly to `pathways-flavor`, keeping `flavor-default` unconstrained.

### Testing
- Unit tests added:
  - `TestGetMTCDaemonSet`: verifies static name resolution, dynamic name resolution via list, and 403 Forbidden handling.
  - `TestRestartMTCDriverPods_DynamicName`: verifies rolling restart patch is directed to dynamic DaemonSet name.
  - `TestCalculateClusterCapacity_MultipleCPUPools`: verifies cluster with multiple CPU pools aggregates capacity without setting `cloud.google.com/gke-nodepool` on `flavor-default`.
- Ran all orchestrator unit tests: `go test -v ./pkg/orchestrator/gke/...` (all passed).
- Ran pre-commit hooks (all passed 100% cleanly).

### DIFF
--- a/pkg/orchestrator/gke/gke_job_orchestrator.go
+++ b/pkg/orchestrator/gke/gke_job_orchestrator.go
@@ -689,13 +689,16 @@ func (g *GKEOrchestrator) getNodeServiceAccount() string {
 }
 
 func isDaemonSetRolloutComplete(dsObj *unstructured.Unstructured) bool {
+	if dsObj == nil || dsObj.Object == nil {
+		return false
+	}
 	gen, _, _ := unstructured.NestedInt64(dsObj.Object, "metadata", "generation")
 	obsGen, _, _ := unstructured.NestedInt64(dsObj.Object, "status", "observedGeneration")
 	desired, _, _ := unstructured.NestedInt64(dsObj.Object, "status", "desiredNumberScheduled")
 	ready, _, _ := unstructured.NestedInt64(dsObj.Object, "status", "numberReady")
 	updated, _, _ := unstructured.NestedInt64(dsObj.Object, "status", "updatedNumberScheduled")
 
-	if obsGen >= gen && ready >= desired && updated >= desired {
+	if desired > 0 && obsGen >= gen && ready >= desired && updated >= desired {
 		logging.Info("MTC multitier-driver DaemonSet is ready (%d/%d nodes ready).", ready, desired)
 		return true
 	}
@@ -709,7 +712,7 @@ func getMTCDaemonSet(ctx context.Context, client dynamic.Interface, namespace st
 			dsObj.SetName("multitier-driver")
 		}
 		return dsObj, nil
-	} else if isForbiddenError(err) {
+	} else if !apierrors.IsNotFound(err) {
 		return nil, err
 	}
 
@@ -730,30 +733,10 @@ func getMTCDaemonSet(ctx context.Context, client dynamic.Interface, namespace st
 
 var daemonSetPollInterval = 2 * time.Second
 
-// waitForMTCDriverDaemonSetReady waits for the multitier-driver DaemonSet to finish rollout and become ready.
-func waitForMTCDriverDaemonSetReady(ctx context.Context, client dynamic.Interface, namespace string) {
-	logging.Info("Waiting for MTC multitier-driver DaemonSet in %s to be ready...", namespace)
+// waitForDaemonSetRollout polls the DaemonSet until its rollout is complete or context times out.
+func waitForDaemonSetRollout(ctx context.Context, client dynamic.Interface, namespace, dsName string) {
 	waitCtx, cancel := context.WithTimeout(ctx, 90*time.Second)
 	defer cancel()
-
-	dsObj, err := getMTCDaemonSet(waitCtx, client, namespace)
-	if err != nil {
-		if isForbiddenError(err) {
-			logging.Warn("Insufficient RBAC permissions to get multitier-driver DaemonSet status (403 Forbidden). Proceeding with job submission.")
-			return
-		}
-		if apierrors.IsNotFound(err) {
-			logging.Warn("MTC multitier-driver DaemonSet not found in %s. Proceeding with job submission.", namespace)
-			return
-		}
-		logging.Warn("Error finding MTC multitier-driver DaemonSet in %s: %v. Proceeding with job submission.", namespace, err)
-		return
-	}
-
-	dsName := dsObj.GetName()
-	if isDaemonSetRolloutComplete(dsObj) {
-		return
-	}
 
 	ticker := time.NewTicker(daemonSetPollInterval)
 	defer ticker.Stop()
@@ -767,7 +750,7 @@ func waitForMTCDriverDaemonSetReady(ctx context.Context, client dynamic.Interfac
 			currentObj, err := client.Resource(daemonsetGVR).Namespace(namespace).Get(waitCtx, dsName, metav1.GetOptions{})
 			if err != nil {
 				if isForbiddenError(err) {
-					logging.Warn("Insufficient RBAC permissions to get multitier-driver DaemonSet status (403 Forbidden). Proceeding with job submission.")
+					logging.Warn("Insufficient RBAC permissions to get multitier-driver DaemonSet %s in %s status (403 Forbidden). Proceeding with job submission.", dsName, namespace)
 					return
 				}
 				logging.Warn("Retrying get of multitier-driver DaemonSet %s: %v", dsName, err)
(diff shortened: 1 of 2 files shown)


pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -726,7 +736,22 @@ func waitForMTCDriverDaemonSetReady(ctx context.Context, client dynamic.Interfac
 	waitCtx, cancel := context.WithTimeout(ctx, 90*time.Second)
 	defer cancel()
 
-	if ready, err := checkDaemonSetReady(waitCtx, client, namespace); ready && err == nil {
+	dsObj, err := getMTCDaemonSet(waitCtx, client, namespace)
```
SwarnaBharathiMantena: This seems to be a duplicate discovery resulting in redundant discovery calls.

Suggestion: Accept dsName string (or make it optional / use a helper waitForDaemonSetReady(ctx, client, namespace, dsName)). If dsName is already known from the patch step, pass it directly to avoid duplicate discovery round-trips:

```
func waitForMTCDriverDaemonSetReady(ctx context.Context, client dynamic.Interface, namespace string, dsNameOpt ...string) {
    var dsName string
    if len(dsNameOpt) > 0 && dsNameOpt[0] != "" {
        dsName = dsNameOpt[0]
    } else {
        dsObj, err := getMTCDaemonSet(waitCtx, client, namespace)
        if err != nil { ... }
        dsName = dsObj.GetName()
    }
    // Poll dsName ...
}
```

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -702,20 +702,30 @@ func isDaemonSetRolloutComplete(dsObj *unstructured.Unstructured) bool {
...
 
-func checkDaemonSetReady(ctx context.Context, client dynamic.Interface, namespace string) (bool, error) {
-	dsObj, err := client.Resource(daemonsetGVR).Namespace(namespace).Get(ctx, "multitier-driver", metav1.GetOptions{})
-	if err != nil {
-		if isForbiddenError(err) {
-			logging.Warn("Insufficient RBAC permissions to get multitier-driver DaemonSet status (403 Forbidden). Proceeding with job submission.")
-			return true, nil
+func getMTCDaemonSet(ctx context.Context, client dynamic.Interface, namespace string) (*unstructured.Unstructured, error) {
+	// First attempt direct get of "multitier-driver" in case it exists with static name.
+	if dsObj, err := client.Resource(daemonsetGVR).Namespace(namespace).Get(ctx, "multitier-driver", metav1.GetOptions{}); err == nil {
```
SwarnaBharathiMantena, line 710: Should we consider a few other failure scenarios: context timeout / network failure ?

pkg/orchestrator/gke/gke_job_orchestrator.go
```diff
@@ -736,30 +761,53 @@ func waitForMTCDriverDaemonSetReady(ctx context.Context, client dynamic.Interfac
...
+		dsName := dsObj.GetName()
+		patchData := fmt.Appendf(nil, `{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"%s"}}}}}`, time.Now().UTC().Format(time.RFC3339))
+		_, patchErr := client.Resource(daemonsetGVR).Namespace(namespace).Patch(ctx, dsName, types.StrategicMergePatchType, patchData, metav1.PatchOptions{})
+		if patchErr == nil {
+			logging.Info("Triggered rolling restart of %s DaemonSet in %s", dsName, namespace)
+			waitForMTCDriverDaemonSetReady(ctx, client, namespace)
+			return
+		} else if isForbiddenError(patchErr) {
+			logging.Warn("Insufficient RBAC permissions to restart %s DaemonSet in %s (403 Forbidden). Skipping driver restart.", dsName, namespace)
+			return
```
SwarnaBharathiMantena, line 776: If getMTCDaemonSet encounters an unexpected error (e.g. network timeout or internal API error), none of the else if branches match, and execution falls through into deleteMTCDriverPodsFallback without any log message explaining why.

Is it better to add an explicit else branch logging the unexpected error and returning, rather than attempting pod deletion when discovery failed unexpectedly?

pkg/orchestrator/gke/gke_job_orchestrator_test.go
```diff
@@ -4474,3 +4470,339 @@ func TestValidateNamespaceExists(t *testing.T) {
...
+	if !exists {
+		t.Fatalf("expected flavor-default to exist in calculated flavors")
+	}
+	if defaultFlavor.CPUs != 10 {
+		t.Errorf("expected flavor-default to have 10 CPUs, got %d", defaultFlavor.CPUs)
+	}
+	// Verify flavor-default has NO cloud.google.com/gke-nodepool label!
+	if npLabel, ok := defaultFlavor.NodeLabels["cloud.google.com/gke-nodepool"]; ok {
+		t.Errorf("expected flavor-default not to have cloud.google.com/gke-nodepool label, but got %q", npLabel)
+	}
```
SwarnaBharathiMantena, line 5069: Can we add a unit test verifying waitForMTCDriverDaemonSetReady polling a dynamic multitier-driver-<uuid> DaemonSet? (if it is not already covered)

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: Good point. Flattened the error handling in `restartMTCDriverPods` with early returns. If `getMTCDaemonSet` fails with an unexpected error or 404 NotFound, it logs a warning and returns early without attempting pod deletion. 

Pod deletion fallback only triggers if `getMTCDaemonSet` succeeded but Patch failed with a non-403 error. Added unit test Unexpected error on getMTCDaemonSet skips pod deletion fallback to verify this.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: Updated getMTCDaemonSet to check else if !apierrors.IsNotFound(err) instead of only filtering for 403 Forbidden.

Now, if Get fails due to context timeout, cancellation, network failure, or 403 Forbidden, it immediately returns nil, err without issuing a redundant namespace-wide List call. Only genuine 404 NotFound falls through to dynamic discovery.

Added dedicated test cases in TestGetMTCDaemonSet_EdgeCases verifying that context.Canceled and transient network failures fail fast without invoking List.

pkg/orchestrator/gke/gke_job_orchestrator_test.go
Neelabh94: Dynamic UUID name discovery and rollout are verified end-to-end in TestRestartMTCDriverPods_DynamicName (which mocks discovery of a UUID suffix multitier-driver-<uuid>, patches it, and verifies polling until ready). 

Also added TestWaitForDaemonSetRollout (covering immediate readiness, multi-tick polling, and transient network retry) and table-driven tests in TestIsDaemonSetRolloutComplete.

pkg/orchestrator/gke/gke_job_orchestrator.go
Neelabh94: extracted the polling logic into `waitForDaemonSetRollout(ctx, client, namespace, dsName)` which takes the already-discovered `dsName` directly. `restartMTCDriverPods` now passes `dsName` into `waitForDaemonSetRollout` across both the patch restart and pod deletion fallback flows, and deleted the unused `waitForMTCDriverDaemonSetReady` wrapper entirely since it was now dead code.

SwarnaBharathiMantena (approved):
LGTM

PR #6301, opened 2026-09-10
optional opt in mig Support in slurm for advanced GPU machinne type(A4x) for static nodes

arpit974:
# feat(slurm): MIG Phase 2 — Multi-Slice NVL72/A4X GPU Topologies & Strict MIG Opt-In

## Overview
This PR implements **MIG Phase 2**, adding native support for multi-slice GPU network topologies (such as NVIDIA GB200 NVL72 / `a4x-highgpu-4g` on GCP) within Slurm on Google Cloud. It introduces automated slice-partitioned Managed Instance Groups (MIGs), per-slice Workload Placement Policies, coordinated resume/suspend lifecycle guards, and strict opt-in engine enforcement across the Cluster Toolkit.

---

### 1. Multi-Slice Topology Sharding & Workload Policies
- **Topology Slicing**: Automatically partitions topologized GPU nodesets into discrete 18-node chassis slices (e.g., 36 nodes -> 2 discrete MIGs; 1188 nodes -> 66 discrete MIGs).
- **Workload Policy Generation**: Creates and binds dedicated `google_compute_resource_policy` (`type = HIGH_THROUGHPUT`, `accelerator_topology = "1x72"`) to each individual slice MIG.
- **Divisibility Guardrails**: Enforces via HCL preconditions that `node_count_static` must be an integer multiple of 18 (e.g. rejecting 1180 nodes).

### 2. Strict MIG Opt-In & Non-Regression Safeguards
- **Zero Silent Engine Promotion**: Nodesets under `AUTO` or omitted `provisioning_engine` strictly default to `BULK_INSERT` for standard static/dynamic compute.
- **Fail-Fast Opt-In Validation**: Configured `accelerator_topology` requires explicit `provisioning_engine = "MIG"`. Rejection is enforced at both:
  - **Compile-time**: `ConditionalValidator` in `metadata.yaml` fails `gcluster create`.
  - **Plan-time**: HCL output precondition in `schedmd-slurm-gcp-v6-nodeset` fails `terraform plan`.

### 3. Slurm Controller & Lifecycle Scripts (`resume.py`, `suspend.py`, `util.py`)
- **Atomic Rollback on Stockout**: Partially provisioned slice MIGs undergoing quota or capacity exhaustion are automatically rolled back to prevent stranded, unjoinable nodes.
- **In-Flight Deletion Guard**: Implements a 20-second safety window before creating instances if instances with the same names are pending deletion, preventing GCE naming collisions.
- **Cloud Autoscaler Bypass**: Bypasses `slurm_cloud_autoscalerd` for exclusive multi-slice jobs to eliminate scheduler/autoscaler race conditions.
- **Exclusive Multi-Shard Job Tracking**: Adds `ExclusiveJobTracker` to synchronize multi-shard job lifecycles and prevent orphan billing leaks.

### 4. Core Blueprint Validators (`pkg/validators`)
- **Compound Trigger Support**: Enhanced `ConditionalValidator` to accept multi-key trigger maps (`triggers: {engine: MIG, enable_placement: true}`).
- **Dynamic Deferral**: Added `!known` token deferral to `ConditionalRegexValidator` for computed Terraform expressions.
- **Complexity Refactoring**: Decomposed `Validate` helper methods to comply with repository `go-cyclo <= 13` standards.

---

### 1. Static Pre-Flight Analysis
- **Pytest**: `210 passed in 0.94s` (`tests/test_resume.py`, `test_util.py`).
- **Mypy**: `0 issues found across 5 source files` (`resume.py`, `suspend.py`, `util.py`).
- **Go Unit Tests**: `34/34 passed in 0.142s` (`./pkg/validators/...`).
- **Pre-Commit Hooks**: Clean runs for `go-cyclo`, `go fmt`, `go vet`, `go imports`, `golangci-lint`, `yamllint`, `PyMarkdown`.

### 2. Comprehensive 28-Scenario Permutation Matrix
- **18 Table 1 Engine Scenarios (`ENG-01`..`18`)**:
  - `MIG` (Static, Spot, DWS Flex): Supported and verified.
  - `MIG` (Dynamic, Hybrid): Blocked by preconditions.
  - `BULK_INSERT` (Static, Dynamic, Hybrid, Spot): Supported and verified.
  - `AUTO` / Omitted (Static, Dynamic, Hybrid): Verified strict opt-in (0 MIGs created; resolves to `BULK_INSERT`).
- **8 Table 2 A4X Scenarios (`A4X-01`..`08`)**:
  - `A4X-01` (18 nodes, 1x72): 1 MIG, 1 Workload Policy created.
  - `A4X-02` (36 nodes, 1x72): 2 Slice MIGs, 2 Workload Policies created.
  - `A4X-03` (Specific Reservation): Verified `SPECIFIC_RESERVATION` affinity generation.
  - `A4X-08` (Topologized A4X with AUTO): Verified strict compile-time rejection.
- **Scale Scenarios**:
  - `SCALE-1188` (1188 nodes, 66 slices): Verified plan generation of 66 MIGs + 66 Workload Policies.
  - `SCALE-1180` (Non-divisible): Rejected by HCL divisibility precondition.

### 3. Live GCP Control-Plane Deployment & Teardown Audit
- **Deployment**: `test-p2-a4x` deployed in live project and zone (`us-central1`).
- **GCE REST API Verification**:
  - Verified MIGs: `testp2a4x-a4xnodeset-mig-0`, `testp2a4x-a4xnodeset-mig-1` (`targetSize = 0`).
  - Verified Policies: Attached `HIGH_THROUGHPUT` policies with `acceleratorTopology: 1x72`.
- **Teardown & Leak Audit**:
  - Executed `terraform destroy --auto-approve` (15 resources destroyed).
  - Queried Compute REST API post-destroy: **0 residual MIGs, 0 policies, 0 instances, 0 disks, 0 networks**.

---

## Impact & Backward Compatibility
- **Non-MIG Workloads**: Zero regression. Non-topologized and standard CPU/GPU nodesets continue using `BULK_INSERT` by default.
- **Existing MIG Workloads**: Phase 1 single-MIG static clusters remain fully supported without syntax changes.


community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -89,6 +89,10 @@ def suspend_mig_nodes(nodes: List[str], lkp: util.Lookup) -> None:
     if not nodes:
         return
 
+    # Invalidate cached managed instances to ensure fresh live state from GCE
+    lkp.get_mig_instances.cache_clear()
```
Neelabh94, line 93: When `resume.py` aborts a multi-slice exclusive job (e.g., after a zonal stockout on slice 1) and invokes `suspend_mig_nodes()` to roll back slice 0, `lkp.instances` (decorated with `@lru_cache(maxsize=1)`) still holds pre-provisioning state cached at the start of `resume.py`. If `listManagedInstances` encounters a transient API error during rollback (`mig_list_success == False`), the fallback `lkp.instance(short_name)` will return `None` for newly created instances unless `lkp.instances.cache_clear()` is also called here:
```python
    # Invalidate cached managed instances to ensure fresh live state from GCE
    lkp.get_mig_instances.cache_clear()
    lkp.get_mig_repairing_instances.cache_clear()
    lkp.instances.cache_clear()
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
```diff
@@ -1836,11 +1843,54 @@ def mig_name(self, nodeset_name: str, index: int = 0) -> str:
...
+        or reads slice_size from config. Defaults to 1000 for standard MIGs.
+        """
+        nodeset = self.cfg.nodeset.get(nodeset_name)
+        if not nodeset:
+            return 1000
+        slice_val = nodeset.get("slice_size") if isinstance(nodeset, dict) else getattr(nodeset, "slice_size", None)
+        if slice_val:
+            try:
+                return max(1, int(slice_val))
+            except (ValueError, TypeError):
```
Neelabh94, line 1859: Because Terraform (`partition.tf:273`) already computes `slice_size = local.nodeset_slice_size[ns.nodeset_name]` and writes it directly to `config.yaml` for every nodeset, lines 1859–1874 in `util.py` (as well as `calculate_hosts_per_topo` in `resume.py:927-954`) are never executed in production.

More importantly, keeping Terraform as the single source of truth for `slice_size` avoids any split-brain risk where Python's runtime `template_info()` fallback could compute a different slice size than the MIGs created by Terraform (which would cause `node_mig_name` `idx // slice_size` to route nodes to the wrong MIG index).

We can simplify `nodeset_slice_size` to strictly read `slice_size` from `config.yaml` (and remove `calculate_hosts_per_topo` from `resume.py`):
```python
    def nodeset_slice_size(self, nodeset_name: str) -> int:
        """Returns the slice size (hosts per slice) configured by Terraform for a NodeSet."""
        nodeset = self.cfg.nodeset.get(nodeset_name)
        if not nodeset:
            return 1000
        slice_val = nodeset.get("slice_size") if isinstance(nodeset, dict) else getattr(nodeset, "slice_size", None)
        if slice_val:
            return max(1, int(slice_val))
        return 1000
```

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
arpit974: done

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/util.py
arpit974: For static MIGs, Terraform is already the single source of truth: `partition.tf:303` writes `slice_size` to `config.yaml`, and both `util.py:nodeset_slice_size` (L1852) and `resume.py:calculate_chunk_size` (L981) return `slice_val` immediately without calling `template_info()`.

We kept `calculate_hosts_per_topo` in `resume.py` (and the fallback in `util.py`) because `partition.tf:303` sets `slice_size = null` for non-MIG nodesets (`BULK_INSERT` / `DWS Flex`). Those paths still rely on `calculate_hosts_per_topo` at runtime to derive chunk size from live `template_info()`, and the `util.py` fallback supports unit tests / synthetic configs where `slice_size` isn't pre-rendered by Terraform.

Neelabh94 (commented):
PR Review Comments

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -451,14 +472,48 @@ def resume_mig_nodes(nodes: List[str], excl_job_id: Optional[int], lkp: util.Loo
...
+                lkp.get_mig_repairing_instances.cache_clear()
+                successful_creates.extend(chunk_nodes)
             except Exception as e:
                 log.error(f"Failed createInstances for MIG {mig_name} on nodes {to_hostlist(chunk_nodes)}: {e}")
+                nodes_to_cleanup = (
+                    list(dict.fromkeys(successful_creates + chunk_nodes))
+                    if excl_job_id is not None
+                    else list(chunk_nodes)
+                )
+                deferred_reclaim = bool(nodes_to_cleanup) and not observed_terminal_failure
```
Neelabh94, line 493: Evaluating `deferred_reclaim` before classifying the error might lead to incorrect handling of transient errors. Consider evaluating it after `classify_gcp_error`. If `action` is `REQUEUE` or outcome is unknown, then defer reclaim. Permanent errors should not be deferred.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -89,6 +89,11 @@ def suspend_mig_nodes(nodes: List[str], lkp: util.Lookup) -> None:
...
```
Neelabh94, line 93: Note: During early provisioning, `listManagedInstances` can return `{"instance": None}` for instances in `CREATING` state. In `suspend_mig_nodes` (line 107 in this file), skipping these might lead to leaked VMs during rollbacks. Consider handling `None` values or constructing fallback URLs.

pkg/validators/metadata_validators_test.go
```diff
@@ -1129,6 +1129,108 @@ func TestConditionalValidator_Triggers(t *testing.T) {
...
+		if !strings.Contains(err.Error(), expectedMsg) {
+			t.Fatalf("expected error message %q, got %q", expectedMsg, err.Error())
+		}
+	})
+}
+
+func TestConditionalValidator_TriggersNilExpectedValue(t *testing.T) {
+	// A nil trigger value resolves through the module defaults, so the module must be
+	// registered; without this the lookup of an unset trigger panics in InfoOrDie.
+	modulereader.SetModuleInfo("test/module", "terraform", modulereader.ModuleInfo{})
```
Neelabh94: Mutating global state via `modulereader.SetModuleInfo` can cause test pollution and flakiness. Use unique module names or `t.Cleanup()` to restore state.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/suspend.py
```diff
@@ -133,15 +141,34 @@ def suspend_mig_nodes(nodes: List[str], lkp: util.Lookup) -> None:
                     if link not in seen_links:
                         links.append(link)
                         seen_links.add(link)
+                elif short_name in mig_unplaced_names:
+                    nodeset = lkp.node_nodeset(node)
+                    zones = list(getattr(nodeset, "zone_policy_allow", None) or [])
+                    if not zones:
+                        try:
+                            reg_info = util.ensure_execute(
+                                lkp.compute.regions().get(project=lkp.project, region=region)
```
Neelabh94, line 150: **[Context]** Because we introduced the `mig_unplaced_names` fallback logic in the last commit, this performance issue was introduced.

Inside the `for node in mig_nodes:` loop, if `zone_policy_allow` is empty, the script makes a synchronous `compute.regions().get` API call. This introduces severe performance degradation during suspension (O(N) API calls) and quota exhaustion risks if many nodes hit this path.

**Suggested Fix**: Move the region API lookup outside the loop or simplify the fallback logic to use a single default zone.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
```diff
@@ -389,6 +395,18 @@ def resume_mig_nodes(nodes: List[str], excl_job_id: Optional[int], lkp: util.Loo
...
                 log.error(f"Instances {still_deleting} still deleting in GCE after timeout; failing resume to trigger immediate requeue.")
                 if excl_job_id is not None:
+                    if successful_creates:
+                        log.warning(
+                            f"Aborting multi-slice exclusive job {excl_job_id}: cleaning up {len(successful_creates)} previously created instances: {to_hostlist(successful_creates)}"
+                        )
+                        try:
+                            # Lazy import to avoid circular dependency between resume and suspend
+                            import suspend
+                            lkp.get_mig_instances.cache_clear()
```
Neelabh94, line 405: **[Context]** Even though we updated `suspend.py` to clear the instance cache in the previous cycle, we missed out on updating `resume.py` rollback paths to do the same.

When rolling back after a partial failure in `resume_mig_nodes`, the script clears `lkp.get_mig_instances` but omits `lkp.instances.cache_clear()`. If `listManagedInstances` fails during rollback, the script falls back to stale cached data, potentially skipping deletion of orphaned VMs.

**Suggested Fix**: Add `lkp.instances.cache_clear()` before invoking `suspend.suspend_mig_nodes` in the rollback blocks.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
```diff
@@ -343,7 +343,7 @@ variable "guest_accelerator" {
 
 variable "accelerator_topology" {
   type        = string
-  description = "Specifies the shape of the Accelerator (GPU/TPU) slice."
+  description = "Specifies the shape of the Accelerator (GPU/TPU) slice. Note: When set, 'enable_placement' must be set to true (and 'node_count_dynamic_max' must be explicitly set to 0 when using 'provisioning_engine = MIG'). Warning: on a deployed nodeset using 'provisioning_engine = MIG', ANY change to this value - including changing one topology for another, and removing it - forces replacement of the managed instance group and destroys its running VMs. Drain the nodeset before changing it."
   nullable    = true
```
Neelabh94, line 347: **[Context]** Nit.

Users are forced to explicitly set `enable_placement: true` when `accelerator_topology` is defined. Since `accelerator_topology` logically implies placement, we can improve Developer Experience (DX) by automatically inferring it in locals and removing the strict validation blocking the user.

**Suggested Fix**: 
```terraform
locals {
  enable_placement = var.enable_placement || (var.accelerator_topology != null && var.accelerator_topology != "")
}
```

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/outputs.tf
```diff
@@ -111,12 +145,9 @@ output "nodeset" {
...
   precondition {
-    condition     = !(var.node_count_dynamic_max > 0 && !var.dws_flex.enabled && var.provisioning_engine == "MIG")
-    error_message = "Dynamic compute NodeSets with provisioning_engine = 'MIG' are currently not supported. When using provisioning_engine = 'MIG', please explicitly set node_count_dynamic_max = 0."
-  }
-
-  precondition {
-    condition     = !(var.provisioning_engine == "MIG" && var.enable_placement && !var.dws_flex.enabled)
-    error_message = "MIG engine currently does not support runtime dynamic compact placement policies (enable_placement = true). Please set enable_placement = false when using provisioning_engine = 'MIG'. Compact placement and Workload Policies for MIGs will be supported in a future release."
+    condition = !(
+      var.node_count_dynamic_max > 0 && var.provisioning_engine == "MIG" && !var.dws_flex.enabled
```
Neelabh94, line 149: **[Context]** Nit.

Users must explicitly set this to 0 for MIGs. Similar to the placement flag, we can improve ergonomics by silently overriding or defaulting it properly in locals instead of forcing the user to type boilerplate variables.

arpit974: node_count_dynamic_max: 0 isn't MIG boilerplate — it's already the convention across the repo. 15 blueprints set it explicitly, including a3high, a3mega, a3ultra, a4high, ml-slurm-g4 and both A4X ones, and none of those are MIG. So MIG users aren't typing anything extra compared to any other static GPU nodeset.

Defaulting it in locals also means flipping the variable default from 10 to null, which silently drops every nodeset that omits the setting out of the telemetry dynamic-node-count metric (parseDefaultValue guards on inputDefault != nil) — MIG or not.

Prefer to keep it explicit and fail loudly.

community/modules/compute/schedmd-slurm-gcp-v6-nodeset/variables.tf
arpit974: enable_placement already defaults to true (variables.tf:445), so users aren't forced to set it. The precondition only trips when someone explicitly passes false and a topology, which is a real contradiction worth failing on.

Inferring it in locals would also make enable_placement: false unreachable whenever a topology is set, and it would flip the future_reservation and placement_max_distance preconditions from pass to fail for existing BULK_INSERT A4X nodesets that set accelerator_topology without opting into MIG. That breaks the strict opt-in boundary this PR is built on, so I'd rather leave it.

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/resume.py
arpit974: this one is already covered. suspend_mig_nodes() clears it itself at suspend.py:95, before any lookup happens, so both rollback paths already get a clean cache. Adding it in resume.py would call the same thing one line earlier.

The early if not nodes: return at suspend.py:89 isn't a hole either, since both call sites are guarded — if successful_creates: (resume.py:398) and if nodes_to_cleanup: (resume.py:504).

Happy to add it if you'd prefer it explicit at the call site, but as it stands it's a no-op.

Neelabh94 (approved):
LGTM

arpit974:
### CI Note: The 3 non-green GPU On-Spot Slurm checks are unrelated to this PR

- `PR-test-ml-a3-highgpu-onspot-slurm` (`failure`)
- `PR-test-ml-a4-highgpu-onspot-slurm` (`failure`)
- `PR-test-ml-a3-ultragpu-onspot-slurm` (`action_required` / Cloud Build error)

**Why these are unrelated to our changes:**
1. **Failing identically on clean `develop`:** All three corresponding daily jobs are currently failing on `develop` (`09ef70a5f`) without any changes from this PR:
   - `DAILY-test-ml-a3-highgpu-onspot-slurm` — failed (`2026-09-21T23:59Z`)
   - `DAILY-test-ml-a4-highgpu-onspot-slurm` — failed (`2026-09-22T01:47Z`)
   - `DAILY-test-ml-a3-ultragpu-onspot-slurm` — failed (`2026-09-22T09:51Z`)
2. **Passed on the previous commit (`3836fc35b`) of this PR:** Both `PR-test-ml-a3-highgpu-onspot-slurm` and `PR-test-ml-a3-ultragpu-onspot-slurm` passed on `3836fc35b`, and the only delta between `3836fc35b` and `922adc3df` is inside `suspend_mig_nodes()` in `suspend.py`.
3. **Zero code-path overlap (`BULK_INSERT` vs. `MIG`):** None of `a3high-`, `a3ultra-`, or `a4high-slurm-blueprint.yaml` set `provisioning_engine: "MIG"` (all default to `"BULK_INSERT"`). In Terraform, `local.nodeset_migs` evaluates to `{}` (`0` MIG/workload-policy resources created), and at runtime `lkp.is_node_mig()` evaluates to `False`, so neither `resume_mig_nodes()` nor `suspend_mig_nodes()` is ever invoked.

PR #6303, opened 2026-09-10
feat(artifact-registry): add repository_name override and fix registry URL output

jacobw125:
- Allow overriding the Artifact Registry repository name via the `repository_name` variable
- Default repository name to `${var.deployment_name}-${random_suffix}`, substituting dashes for underscores, which aren't supported in artifact repo names
- Fix `registry_url` output to point to the actual created repository ID instead of `var.deployment_name`
- Add `repo_url`, `repository_id`, and `repository_name` outputs for downstream modules
- Update documentation to reflect new inputs and outputs

**BREAKING CHANGE:**
Changing the default naming convention for repository_name from the previous format/mode/host-based name to the new `${var.deployment_name}-${random_id.resource_name_suffix.hex}` scheme is a breaking change. For any existing deployments upgrading to this version, the change in repository_id will force the recreation of the google_artifact_registry_repository resource, resulting in the deletion of all stored container images and artifacts. Users can avoid this by 'pinning' their existing repository name as described in the  README.md.

### Submission Checklist

- [x] Add or modify unit tests to cover code changes

### DIFF
--- a/community/modules/container/artifact-registry/main.tf
+++ b/community/modules/container/artifact-registry/main.tf
@@ -21,36 +21,17 @@ locals {
   # Auto (i.e., empty) vs user-managed replication
   auto = length(var.user_managed_replication) == 0 ? true : false
 
-  # For remote custom repositories, parse out host to create a base_component name
-  mirror_url_no_proto = var.repo_mirror_url != null ? replace(replace(var.repo_mirror_url, "https://", ""), "http://", "") : ""
-  mirror_host         = local.mirror_url_no_proto != "" ? split("/", local.mirror_url_no_proto)[0] : ""
-
-  base_component = replace(
-    replace(
-      replace(
-        lower(
-          local.mirror_host != ""
-          ? "${var.format}-${var.repo_mode}-${local.mirror_host}"
-          : "${var.format}-${var.repo_mode}-nohost"
-        ),
-        "\\.", "-"
-      ),
-      "/", "-"
-    ),
-    "_", "-"
+  # The final name for the artifact registry repository
+  repository_name = var.repository_name != null ? var.repository_name : replace(
+    lower("${var.deployment_name}-${random_id.resource_name_suffix.hex}"),
+    "/[_.]/",
+    "-"
   )
 
-  repository_suffix = random_id.resource_name_suffix.hex
-
-  # The final name for the artifact registry repository
-  repository_name = replace(
-    replace(
-      lower(
-        format("%s-%s", local.base_component, local.repository_suffix)
-      ),
-      ".", "-"
-    ),
-    "/", "-"
+  repo_url = (
+    contains(["APT", "YUM"], upper(var.format)) ?
+    "${var.region}-${lower(var.format)}.pkg.dev/projects/${var.project_id}/${local.repository_name}" :
+    "${var.region}-${lower(var.format)}.pkg.dev/${var.project_id}/${local.repository_name}"
   )
 
   # The secret name is derived from the repository name
--- a/community/modules/container/artifact-registry/outputs.tf
+++ b/community/modules/container/artifact-registry/outputs.tf
@@ -12,7 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+output "repo_url" {
+  description = "The URL of the artifact registry repo."
+  value       = local.repo_url
+}
+
 output "registry_url" {
-  description = "The URL of the created artifact registry."
-  value       = "${var.region}-docker.pkg.dev/${var.project_id}/${var.deployment_name}"
+  description = "The URL of the artifact registry repo."
+  # Technically, this is the repo URL, not the registry URL
+  # It is preserved for backward compatibility
+  value = local.repo_url
+}
+
+output "repository_id" {
+  description = "The ID of the created artifact registry repository."
+  value       = local.repository_name
+}
+
+output "repository_name" {
+  description = "The name (ID) of the created artifact registry repository."
+  value       = local.repository_name
+}
+
+output "repository_resource_name" {
+  description = "The full resource name of the repository."
+  value       = google_artifact_registry_repository.artifact_registry.name
 }
--- a/community/modules/container/artifact-registry/variables.tf
+++ b/community/modules/container/artifact-registry/variables.tf
@@ -27,6 +27,17 @@ variable "deployment_name" {
   type        = string
 }
 
+variable "repository_name" {
+  description = "The repository name (ID) for the Artifact Registry repository. If null, deployment_name (lowercased, with dots and underscores replaced with hyphens) with a random suffix is used."
+  type        = string
+  default     = null
+
+  validation {
+    condition     = var.repository_name == null ? true : can(regex("^[a-z][-a-z0-9]{0,61}[a-z0-9]$", var.repository_name))
+    error_message = "The repository_name must be between 2 and 63 characters, start with a lowercase letter, end with a lowercase letter or number, and contain only lowercase letters, numbers, and hyphens."
+  }
+}
+
 variable "labels" {
   description = "Labels to add to the artifact registry. Key-value pairs."
   type        = map(string)
--- a/community/modules/container/artifact-registry/README.md
+++ b/community/modules/container/artifact-registry/README.md
@@ -47,6 +47,17 @@ Create a standard Docker repository.
     format: DOCKER
 ```
 
+Create a standard Docker repository with a custom name.
+
+```yaml
+- id: custom_registry
+  source: community/modules/container/artifact-registry
+  settings:
+    repository_name: my-custom-repo
+    repo_mode: STANDARD_REPOSITORY
+    format: DOCKER
+```
+
 Mirror of public Docker Hub repository.
 
 ```yaml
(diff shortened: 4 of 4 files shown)


community/modules/container/artifact-registry/main.tf
```diff
@@ -21,36 +21,15 @@ locals {
...
-      ),
-      ".", "-"
+      lower("${var.deployment_name}-${random_id.resource_name_suffix.hex}"),
+      "_",
+      "-"
     ),
-    "/", "-"
+    ".",
+    "-"
   )
```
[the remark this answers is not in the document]
jacobw125: I had considered this as well, but the existing output is broken today, so I doubt this is module is used much. I think the breaking change is worth moving to a more sensible default repository name, which likely matched the author's original intent based on [this line](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/6f755b4fe2cf780c0f82b24e7b05bb0ea3448c72/community/modules/container/artifact-registry/outputs.tf#L17).

community/modules/container/artifact-registry/main.tf
```diff
@@ -21,36 +21,17 @@ locals {
...
-      ),
-      "/", "-"
-    ),
-    "_", "-"
+  # The final name for the artifact registry repository
+  repository_name = var.repository_name != null ? var.repository_name : replace(
+    lower("${var.deployment_name}-${random_id.resource_name_suffix.hex}"),
+    "/[_.]/",
+    "-"
   )
```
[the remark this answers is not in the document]
jacobw125: This goes against the established convention, e.g. cloud-storage-bucket's naming strategy.

community/modules/container/artifact-registry/main.tf
```diff
@@ -21,36 +21,17 @@ locals {
...
-        "\\.", "-"
-      ),
-      "/", "-"
-    ),
-    "_", "-"
+  # The final name for the artifact registry repository
+  repository_name = var.repository_name != null ? var.repository_name : replace(
+    lower("${var.deployment_name}-${random_id.resource_name_suffix.hex}"),
+    "/[_.]/",
+    "-"
```
shubpal07: The regex validation on `var.repository_name` is great, but when var.repository_name` == `null`, this fallback bypasses validation and can cause 400 Bad Request during `terraform apply` under a few scenarios:

1. **Leading character**: GCP Artifact Registry requires repository IDs to start with a letter (^[a-z]). If a user's `var.deployment_name` starts with a digit (e.g., `100k-cluster`), the fallback will start with a digit.
2. **Length limit & suffix preservation**: The GCP limit is 63 characters. Appending the 5-character suffix `(-${random_id...})` means deployment names longer than 58 characters will fail. A naive end-truncation (substr(..., 0, 63)) would chop off the` random_id` suffix (destroying cross-deployment uniqueness) or could slice mid-hyphen leaving an illegal trailing -.
3. **Lowercasing order:** Applying regex before lower() would replace uppercase letters with hyphens instead of converting them.

We can budget string lengths deterministically (3 chars for "ar-", 5 chars for the suffix, leaving 55 chars for the deployment slug):

```terraform
  # 1. Lowercase first, then collapse sequences of non-alphanumeric chars into a single hyphen
  raw_dep_slug = replace(lower(var.deployment_name), "/[^a-z0-9]+/", "-")

  # 2. Strip edge hyphens, slice to 55 chars, and strip any trailing hyphen introduced by slicing
  clean_dep_name = trimsuffix(substr(trim(local.raw_dep_slug, "-"), 0, 55), "-")

  # 3. Assemble: always starts with 'ar-', ends with '[0-9a-f]', length <= 63, preserves uniqueness
  default_repo_name = (
    local.clean_dep_name != ""
    ? "ar-${local.clean_dep_name}-${random_id.resource_name_suffix.hex}"
    : "ar-${random_id.resource_name_suffix.hex}"
  )

  repository_name = var.repository_name != null ? var.repository_name : local.default_repo_name
```

community/modules/container/artifact-registry/main.tf
```diff
@@ -21,36 +21,17 @@ locals {
...
-      lower(
-        format("%s-%s", local.base_component, local.repository_suffix)
-      ),
-      ".", "-"
-    ),
-    "/", "-"
+  repo_url = (
+    contains(["APT", "YUM"], upper(var.format)) ?
+    "${var.region}-${lower(var.format)}.pkg.dev/projects/${var.project_id}/${local.repository_name}" :
+    "${var.region}-${lower(var.format)}.pkg.dev/${var.project_id}/${local.repository_name}"
```
shubpal07: If `repo_url` references `local.repository_name`, Terraform computes it in memory before the GCP resource is actually created. In Cluster Toolkit blueprints, downstream modules consuming `module.<registry>.repo_url` (_such as container image builders or startup scripts_) will attempt to run immediately, causing intermittent 404s on fresh deployments.

By referencing `google_artifact_registry_repository.artifact_registry.repository_id` instead of `local.repository_name`, we build a native `Terraform DAG dependency edge` without needing `depends_on` in outputs (which avoids plan-time unknown value cascades):


```suggestion
    repo_url = (
    contains(["APT", "YUM"], upper(var.format)) ?
    "${var.region}-${lower(var.format)}.pkg.dev/projects/${var.project_id}/${google_artifact_registry_repository.artifact_registry.repository_id}" :
    "${var.region}-${lower(var.format)}.pkg.dev/${var.project_id}/${google_artifact_registry_repository.artifact_registry.repository_id}"
  )
```

community/modules/container/artifact-registry/outputs.tf
```diff
@@ -12,7 +12,29 @@
...
+}
+
 output "registry_url" {
-  description = "The URL of the created artifact registry."
-  value       = "${var.region}-docker.pkg.dev/${var.project_id}/${var.deployment_name}"
+  description = "The URL of the artifact registry repo."
+  # Technically, this is the repo URL, not the registry URL
+  # It is preserved for backward compatibility
+  value = local.repo_url
+}
```
shubpal07, line 24: Keeping `registry_url` as an alias for backwards compatibility looks good! Can we update the description to explicitly mark it as deprecated so blueprint authors know to prefer `repo_url`.

```terraform
output "registry_url" {
  description = "The URL of the artifact registry repo. (Deprecated: Use 'repo_url' instead)"
  # Technically, this is the repo URL, not the registry URL
  # It is preserved for backward compatibility
  value = local.repo_url
}
```

community/modules/container/artifact-registry/outputs.tf
```diff
@@ -12,7 +12,29 @@
...
+
+output "repository_id" {
+  description = "The ID of the created artifact registry repository."
+  value       = local.repository_name
+}
+
+output "repository_name" {
+  description = "The name (ID) of the created artifact registry repository."
+  value       = local.repository_name
+}
```
shubpal07, line 34: Similar to `repo_url`, referencing `google_artifact_registry_repository.artifact_registry.repository_id` here ensures that any downstream module depending on the repository ID waits for the GCP resource creation to complete:


```suggestion
output "repository_id" {
  description = "The ID of the created artifact registry repository."
  value       = google_artifact_registry_repository.artifact_registry.repository_id
}

output "repository_name" {
  description = "The name (ID) of the created artifact registry repository."
  value       = google_artifact_registry_repository.artifact_registry.repository_id
}
```

community/modules/container/artifact-registry/variables.tf
```diff
@@ -27,6 +27,17 @@ variable "deployment_name" {
...
+variable "repository_name" {
+  description = "The repository name (ID) for the Artifact Registry repository. If null, deployment_name (lowercased, with dots and underscores replaced with hyphens) with a random suffix is used."
+  type        = string
+  default     = null
+
+  validation {
+    condition     = var.repository_name == null ? true : can(regex("^[a-z][-a-z0-9]{0,61}[a-z0-9]$", var.repository_name))
+    error_message = "The repository_name must be between 2 and 63 characters, start with a lowercase letter, end with a lowercase letter or number, and contain only lowercase letters, numbers, and hyphens."
+  }
+}
```
shubpal07, line 39: The regex `^[a-z][-a-z0-9]{0,61}[a-z0-9]$` is good for GCP Artifact Registry's 2–63 character constraints!

**One important nuance:** because variable `validation` blocks only run on the input variable itself, when `var.repository_name` is `null` (the default), this check passes unconditionally and does not validate the fallback ID generated in `main.tf`.

To catch any invalid fallback names during `terraform plan` rather than `terraform apply`, could we also mirror this check as a `precondition` on `google_artifact_registry_repository.artifact_registry` in `main.tf` (or in `validation.tf`)?

```terraform
lifecycle {
  precondition {
    condition     = can(regex("^[a-z][-a-z0-9]{0,61}[a-z0-9]$", local.repository_name))
    error_message = "The resolved repository_name '${local.repository_name}' is invalid for GCP Artifact Registry (must be 2-63 chars, start with a lowercase letter, end with a letter/number, and contain only lowercase alphanumeric and hyphens)."
  }
}
```

Also, a minor documentation note: can we update the description field slightly so it doesn't hardcode the exact regex replacement steps of the fallback logic.

shubpal07 (changes requested):
Thanks for working on this @jacobw125! Adding the `repository_name` override and aligning the default naming convention with `deployment_name` brings this module in line with the rest of Cluster Toolkit (e.g. `cloud-storage-bucket`, `vpc`).

I've left inline comments on `main.tf`, `variables.tf`, and `outputs.tf` regarding DAG dependency anchoring and GCP character/length boundaries.

### Breaking Change & Release Notes
Because Google Cloud Artifact Registry repository IDs and Secret Manager secret IDs are immutable, updating the default naming from the legacy format (`docker-standard-repository-nohost-...`) to `ar-<deployment_name>-...` will cause Terraform to plan a resource replacement (`forces replacement`) on existing clusters, which will **permanently delete all stored container images and Secret Manager mirror credentials**.

To handle this cleanly:

1. Please include a `BREAKING CHANGE:` section in the PR description and add a migration guide to `README.md`.

To make this as easy as possible, here is a sample **Migration Guide** that you can drop into your PR description and `README.md` (_Please review it again as per your latest commits_):


### Breaking Change & Migration Guide: Upgrading Existing Deployments

In this PR, the default repository naming convention changes from the legacy format (`<format>-<mode>-nohost-<suffix>`) to `ar-<deployment_name>-<suffix>`.

Because Google Cloud Artifact Registry repository IDs and Secret Manager secret IDs are immutable, changing this default forces Terraform to plan a resource replacement (`-/+`). **Applying this replacement will permanently delete your existing Artifact Registry repository, all stored container images, and any upstream mirror secrets.**

Existing deployments can upgrade with **zero downtime and zero image deletion** by pinning `repository_name` to their current repository ID.

---

#### Step 1: Retrieve Your Existing Repository ID
From your deployment directory, inspect your Terraform state:

```bash
# Navigate to your deployment group folder (default: primary)
cd <deployment-folder>/primary

# Extract repository IDs mapped to their module addresses
terraform show -json | jq -r '.. | objects | select(.type? == "google_artifact_registry_repository") | "\(.address): \(.values.repository_id)"'
```

*(Alternative via `gcloud` CLI):*
```bash
gcloud artifacts repositories list \
  --project="<PROJECT_ID>" \
  --location="<REGION>" \
  --filter='description="<DEPLOYMENT_NAME>"' \
  --format='value(name.basename())'
```

*Example output:* `docker-standard-repository-nohost-4b1a`

---

#### Step 2: Pin `repository_name` in Your Blueprint YAML
Update your Cluster Toolkit blueprint YAML by explicitly setting `repository_name` to your existing repository ID under the module's `settings`:

```yaml
  - id: registry
    source: community/modules/container/artifact-registry
    settings:
      repository_name: "docker-standard-repository-nohost-4b1a" # Paste exact ID from Step 1
      repo_mode: STANDARD_REPOSITORY
      format: DOCKER
```

> **Note:** If your blueprint defines multiple repository modules (e.g. an internal repository and a public mirror), repeat this step for each module using its corresponding address from Step 1.

---

#### Step 3: Regenerate Deployment Files
Regenerate your deployment configuration using `./gcluster create` with the `--overwrite-deployment` (`-w`) flag:

```bash
./gcluster create <your-blueprint.yaml> -o <deployment-folder> -w
```

---

#### Step 4: Verify Plan (DO NOT Skip)
Navigate to your deployment group folder, initialize modules, and generate a plan:

```bash
cd <deployment-folder>/primary
terraform init
terraform plan
```

Inspect the plan output carefully:
* **Safe to proceed:** 
  - `google_artifact_registry_repository.artifact_registry` reports `No changes` or `in-place update (~)` for labels/outputs.
  - Downstream modules referencing `registry_url` or `repo_url` may show in-place updates (`~`) as the URL output is updated to reflect the real repository name.
* **STOP / DO NOT APPLY:** If you see replacement requested:
  `# module.registry.google_artifact_registry_repository.artifact_registry must be replaced (-/+)`
  Verify that `repository_name` in your YAML matches your state's `repository_id` character-for-character with no leading/trailing spaces.

---

#### Step 5: Safely Apply
Once verified that no destructive replacement is planned:

```bash
# Using Terraform directly:
terraform apply

# OR using the Cluster Toolkit orchestrator:
./gcluster deploy <deployment-folder>
```

jacobw125:
@shubpal07, thank you for your detailed and helpful review! I have applied the changes you suggested and added the breaking change/migration section to the README.md.

shubpal07 (approved):
Thanks for addressing all the comments @jacobw125 !!
LGTM

SwarnaBharathiMantena (approved):

PR #6305, opened 2026-09-11
feat(gke): validate placement policy, reservation, and queue at job submission

Neelabh94:
### Description
This pull request adds preflight validation and locality resolution across `--placement-policy`, `--gke-nap-reservation`, and `--queue` during GKE job submission, failing fast with actionable errors rather than letting workloads sit pending or fail late during cluster deployment.

### Key Changes
1. **Workload Placement Policy Validation (`--placement-policy`)**:
   - Validates existence, project, and region locality against GCE Compute REST API contracts before writing to Kubernetes manifests.
   - For static multi-host TPU 7x workloads, validates that policies are of type `HIGH_THROUGHPUT` and use non-dynamic topology modes (`AUTO_CONNECT` or default static placement), skipping dynamic slicing `PROVISION_ONLY` policies.
   - Sanitizes full HTTPS URLs and relative resource URIs down to bare names for Kubernetes node selectors (`cloud.google.com/placement-policy-name` for TPU, `cloud.google.com/gke-placement-group` for non-TPU).
2. **GKE NAP Reservation Validation (`--gke-nap-reservation`)**:
   - Preflight checks GCE reservations (zonal describe and regional list) against requested machine type and cluster locality.
   - Normalizes partial URI machine types using `path.Base` to prevent false mismatches.
   - Supports cross-project shared reservations while enforcing regional compatibility.
3. **Kueue LocalQueue Validation (`--queue`)**:
   - Enforces Kubernetes RFC 1123 DNS subdomain naming rules via `validation.IsDNS1123Subdomain` to prevent HTTP 422 API server rejections.
   - Strips resource path prefixes (`namespaces/.../localQueues/...`) if provided.
4. **Resilience & Enterprise IAM**:
   - Gracefully handles `403 Forbidden` / permission denied errors by logging an informational warning and proceeding, ensuring least-privilege enterprise users are never blocked.
   - Bypasses all network calls when `--dry-run-out` is specified.
5. **Architecture & Clean Code**:
   - Consolidated GCE API DTOs into `types.go`.
   - Introduced single-pass `parseURIParts` dictionary parser to avoid repetitive URI token scanning.
   - Inlined single-caller thin wrappers (`checkNAPFlagsSupported`, `resolveTPU7xWorkloadPolicyIfRequired`).

### Testing
- Unit tests added covering all new validation branches, edge cases, and error paths.
- Executed `go test -v ./pkg/orchestrator/gke/...` with 100% pass rate.
- Verified all pre-commit hooks (`pre-commit run`) passed cleanly (`go fmt`, `go vet`, `go imports`, `go-cyclo`, `golangci-lint`, `addlicense`, `codespell`).

### DIFF
--- a/pkg/orchestrator/gke/resource_resolver.go
+++ b/pkg/orchestrator/gke/resource_resolver.go
@@ -274,16 +274,6 @@ func (g *GKEOrchestrator) checkNodePoolsDynamicSlicing(requestedMachineName stri
 // ErrResourcePolicyPermissionDenied indicates the caller lacks GCP IAM permissions to read the compute resource policy.
 var ErrResourcePolicyPermissionDenied = errors.New("permission denied querying compute resource policy")
 
-type gceResourcePolicyRaw struct {
-	Name           string `json:"name"`
-	Region         string `json:"region"`
-	WorkloadPolicy struct {
-		AcceleratorTopology     string `json:"acceleratorTopology"`
-		AcceleratorTopologyMode string `json:"acceleratorTopologyMode"`
-		Type                    string `json:"type"`
-	} `json:"workloadPolicy"`
-}
-
 // describeResourcePolicyCached retrieves GCE resource policy metadata using gcloud compute resource-policies describe,
 // utilizing a local cache to prevent redundant invocations across dynamic slicing and NAP policy resolution.
 func (g *GKEOrchestrator) describeResourcePolicyCached(policyName, location, projectID string) (*GCEWorkloadPolicy, error) {
@@ -303,7 +293,7 @@ func (g *GKEOrchestrator) describeResourcePolicyCached(policyName, location, pro
 		if strings.Contains(stderrLower, "not found") || strings.Contains(stderrLower, "404") {
 			return nil, nil // Policy does not exist
 		}
-		if strings.Contains(stderrLower, "permission") || strings.Contains(stderrLower, "forbidden") || strings.Contains(stderrLower, "403") {
+		if isPermissionDenied(res.Stderr) {
 			return nil, fmt.Errorf("%w: %s", ErrResourcePolicyPermissionDenied, res.Stderr)
 		}
 		return nil, fmt.Errorf("gcloud compute resource-policies describe failed: %s\n"+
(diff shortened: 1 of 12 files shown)


pkg/orchestrator/gke/resource_resolver.go
```diff
@@ -575,6 +558,88 @@ func (g *GKEOrchestrator) resolveTPU7xWorkloadPolicyIfRequired(job *orchestrator
...
+// - "my-policy" -> Name: "my-policy"
+// - "projects/my-proj/regions/us-central1/resourcePolicies/my-policy" -> Project: "my-proj", Region: "us-central1", Name: "my-policy"
+// - "https://www.googleapis.com/compute/v1/projects/my-proj/regions/us-central1/resourcePolicies/my-policy" -> Project: "my-proj", Region: "us-central1", Name: "my-policy"
+func parseResourcePolicyURI(policyStr string) parsedResourcePolicy {
+	policyStr = strings.TrimSuffix(strings.TrimSpace(policyStr), "/")
+	if !strings.Contains(policyStr, "/") {
+		return parsedResourcePolicy{Name: policyStr}
+	}
+
+	name := extractURIPart(policyStr, "resourcepolicies")
```
agrawalkhushi18, line 572: If a user passes a resource policy URI with a trailing slash (e.g., `projects/my-proj/regions/us-central1/resourcePolicies/`), `extractURIPart` for `"resourcepolicies"` returns `""`. It then falls back to `path.Base(policyStr)`, which evaluates to `"resourcePolicies"` (the collection name) rather than the actual resource. 

Can we return a proper validation error here instead of falling back to `path.Base`?

```go
	name := extractURIPart(policyStr, "resourcepolicies")
	if name == "" {
		return parsedResourcePolicy{} // or return an error upstream
	}

File: pkg/orchestrator/gke/types.go [hunk 2]
```diff
@@ -516,7 +516,61 @@ type kueueWorkloadList struct {
...
+		if !found {
+			break
+		}
+		uri = rest
+	}
+	return ""
+}
+
+// isPermissionDenied returns true if the error text indicates a GCP IAM permission denial (403).
+func isPermissionDenied(errStr string) bool {
```
agrawalkhushi18, line 573: This substring check for `"permission"`, `"forbidden"`, or `"403"` is extremely broad. If a container build or artifact pull encounters an unrelated HTTP 403 error, `isPermissionDenied` will misclassify it as an IAM permission failure on Compute Engine resource policies/reservations. This causes the orchestrator to silently swallow the error (`logging.Warn`) and bypass validation.

Consider restricting this match to explicit gcloud CLI permission error codes or expected resource-policy/reservation permission messages.

pkg/orchestrator/gke/resource_resolver.go
Neelabh94: Agreed. We removed the `path.Base` fallback entirely, strictly requiring canonical GCP URI/URL formats (`/resourcePolicies/`) or bare names, and returning an upfront error when the name cannot be determined.
Added unit tests verifying that trailing slashes on collection URIs return an empty name/error, while canonical URIs (including policies named `resourcepolicies-xyz`) continue to resolve cleanly.

pkg/orchestrator/gke/types.go
Neelabh94: 404 is checked first, this helper is strictly scoped to preflight GCE gcloud commands (never used for containers or image pulls), and maintaining individual permission strings would create high maintenance burden.
We have tightened the matching logic to explicit multi-word phrases (`permission denied`, `permission_denied`, `required 'compute.`, and `403` with `forbidden`) to prevent false positives.

agrawalkhushi18:
One nit suggestion : `GKEOrchestrator` declares multiple internal cache maps here (`machineCapCache`, `dynamicSlicingCache`, `staticSlicingCache`, `topologyCache`, and `resourcePolicyCache`) without any synchronization primitives. 

As we expand to support parallel job validation and pre-fetching, concurrent reads and writes across multiple goroutines will trigger a fatal Go runtime panic (`fatal error: concurrent map write`). 
Please embed a `sync.RWMutex` in this struct and guard all cache lookups and assignments across `resource_resolver.go` and `nap.go` using `.RLock()` / `.Lock()` blocks.

agrawalkhushi18 (approved):
LGTM

PR #6309, opened 2026-09-11
feat(agent-platform): add stage-to-gcs community module for GCS asset…

hamadfadi:
Add reusable community module to stage and synchronize files and directories into Google Cloud Storage buckets from local filesystem paths or remote Git repositories.

### Submission Checklist


community/modules/agent-platform/stage-to-gcs/main.tf
```diff
@@ -0,0 +1,130 @@
...
+      repo_ref = try(coalesce(f.repo_ref, var.repo_ref), "")
+      content_hash = try(coalesce(f.repo_url, var.repo_url, "local"), "local") == "local" ? (
+        try(
+          filesha256(f.source_path),
+          try(
+            filesha256("${path.root}/${f.source_path}"),
+            try(filesha256("${path.root}/../../${f.source_path}"), "")
+          )
+        )
+      ) : ""
```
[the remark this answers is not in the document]
hamadfadi: Supported deeper directory traversal.
Fixed.

community/modules/agent-platform/stage-to-gcs/main.tf
```diff
@@ -0,0 +1,130 @@
...
+            try(
+              sha256(jsonencode({
+                for f in fileset("${path.root}/../../${d.source_path}", "**") : f => filesha256("${path.root}/../../${d.source_path}/${f}")
+                if length([for ex in d.exclude : ex if can(regex(ex, f))]) == 0
+              })),
+              ""
+            )
+          )
+        )
+      ) : ""
```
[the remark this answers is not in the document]
hamadfadi: Supported deeper directory traversal.
Fixed.

community/modules/agent-platform/stage-to-gcs/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "bucket_name" {
+  description = "Name of the target Google Cloud Storage bucket."
+  type        = string
+}
```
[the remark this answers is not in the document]
hamadfadi: Added validation.
Fixed.

hamadfadi:
Why this is needed: 
- Provides a unified, generic staging mechanism to synchronize arbitrary directories and upload individual files into Google Cloud Storage (GCS) buckets.
- Dual-Source Ingestion: Dynamically supports sourcing files and directories from either the local filesystem or remote Git repositories (via repo_url and repo_ref).
- Directory Synchronization: Recursively syncs entire directories (directories) to target GCS prefixes using gcloud storage rsync.
- Configurable Exclusion Filters: Supports regex exclusion rules (exclude, defaulting to [".*\\.Dockerfile$"]), preventing container build specifications from being staged into runtime storage mounts.
- Individual File Uploads: Uploads specific files to custom target object paths in GCS using gcloud storage cp.

community/modules/agent-platform/stage-to-gcs/main.tf
```diff
@@ -0,0 +1,139 @@
...
+    each.value.source_path,
+    each.value.repo_url,
+    each.value.repo_ref,
+    each.value.destination_uri,
+    each.value.exclude_regex,
+    each.value.content_hash,
+  ]
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash"]
```
SwarnaBharathiMantena, line 99: `/bin/bash` might not be present on every image. Please drop `interpreter = ["/bin/bash"]` or document the dependency.

community/modules/agent-platform/stage-to-gcs/main.tf
```diff
@@ -0,0 +1,139 @@
...
+    var.bucket_name,
+    each.value.source_path,
+    each.value.repo_url,
+    each.value.repo_ref,
+    each.value.destination_uri,
+    each.value.content_hash,
+  ]
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash"]
```
SwarnaBharathiMantena, line 128: `/bin/bash` might not be present on every image. Please drop `interpreter = ["/bin/bash"]` or document the dependency.

community/modules/file-system/gcs-sync/outputs.tf
```diff
@@ -0,0 +1,43 @@
...
+  description = "List of destination GCS URIs (gs://...) for staged directories."
+  value       = [for d in local.staged_directories : d.destination_uri]
+}
+
+output "staged_file_uris" {
+  description = "List of destination GCS URIs (gs://...) for staged files."
+  value       = [for f in local.staged_files : f.destination_uri]
+}
+
+output "all_staged_uris" {
```
SwarnaBharathiMantena, line 50: Outputs need to create dependency edges

   Currently only `id` does. `bucket_name`, `staged_directory_uris`, `staged_file_uris` and `all_staged_uris` all derive from `var.*` / `local.*`, so Terraform sees no ordering constraint. `all_staged_uris` is the dangerous one — it's the output a blueprint author would naturally wire into a Cloud Batch job, and it'll resolve happily before anything has been uploaded, producing a job that starts against an empty prefix.

   Please derive them through the `terraform_data` resources so `use` actually orders things.

SwarnaBharathiMantena (commented):
Thanks for this PR. There's a real capability here that #6311 can't provide — pulling content from a Git repo at apply time. I suggest to narrow it, so that the two modules split cleanly on **whether Terraform owns the object** rather than overlapping.

The proposed boundary and suggestions:

- **#6311 (`gcs-objects`)** — inline content, local files, local directories. All `google_storage_bucket_object`: in state, drift-detected, removed on destroy, written with the provider's credentials.
- **This module** — Git-sourced content only. `local-exec` + `gcloud storage rsync`, explicitly not managed by Terraform.

The reason for drawing it there rather than routing internally on `repo_url`: a single module whose destroy behaviour silently depends on which field you set is a footgun. A user who sets `source_path` gets cleanup; one who sets `repo_url` doesn't; nothing in the interface signals that. Putting the difference at the module boundary means it shows up in the name and the docs, where someone will actually see it.

1. Rename to `gcs-sync`, and move to `community/modules/file-system/`

   `stage-to-gcs` doesn't convey "unmanaged", and now that it's Git-only it doesn't convey Git either. `gcs-sync` carries the one-way-sync semantics — paired with `gcs-objects` for the managed side, the two names should let someone pick correctly without reading either README.

   Also please move out of `community/modules/agent-platform/` (which is then empty and can be deleted) into `community/modules/file-system/`, and make necessary edits to the READMEs.

2. Drop local entries entirely

   Every entry should require `repo_url`:

   ```hcl
   validation {
     condition     = alltrue([for f in var.files : f.repo_url != null])
     error_message = "files entries must set repo_url; use the gcs-objects module for local files."
   }
   ```

   with the same for `directories`. Failing loudly with a pointer is much better than silently giving someone unmanaged objects when they wanted managed ones.

   This deletes a good chunk of complexity as a side effect — in particular the four-rung `try()` ladder, including `try(filesha256("${path.root}/../../../${f.source_path}"), ...)`. That's worth flagging on its own, because it's currently a live bug: `try()` swallows *every* error rather than just "file not found", so a path that fails to resolve silently lands on `""`, `triggers_replace` then never changes, and local edits stop syncing while Terraform reports "No changes". Removing local support removes the ladder.

   Keep `files` and `directories` as two separate lists. Nothing has been cloned at `terraform plan` time, so `fileexists()` can't tell you whether a repo path is one file or a tree — auto-detection genuinely isn't available here.

3. State the cleanup limitation; don't add a destroy provisioner

   Objects staged by this module aren't in Terraform state, so `terraform destroy` won't remove them. Please state that in the README and recommend a bucket lifecycle rule as the mitigation.

   I want to state that the fix: a `when = destroy` provisioner **will not work here**. Destroy provisioners don't run on a tainted resource, and a failed create-time provisioner is what taints it — so they're absent in exactly the case where cleanup matters. They also don't run at all if the module block is removed from config, and `create_before_destroy` disables them entirely. See [Provisioners are a Last Resort](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax). An honest README beats a mechanism that looks like cleanup and isn't.

4. `repo_ref` must be pinned

   `triggers_replace` hashes the ref string, so `repo_ref = "main"` never changes and the repo is **never re-fetched after the first apply**. Please document that `repo_ref` should be a commit SHA or tag, and consider warning on obvious branch names. This is the kind of thing that works in testing and quietly stops working months later.

5. `gcloud storage rsync`

   - Add `--delete-unmatched-destination-objects` so files removed upstream don't accumulate.
   - Please gate it so it can only ever apply beneath this module's own `destination_path` prefix. Pointed at a bucket root it would delete other modules' objects, and this module writes into buckets it doesn't own.


### Question

Could you note in the PR description why this needs to be a Terraform module rather than a CI step? I think the answer is DAG ordering — staging has to complete before a runtime or job starts, and `use` gives you that for free — but it'd be good to have it stated, since it's the main thing justifying a `local-exec` module in the first place.

hamadfadi:
@SwarnaBharathiMantena  Thanks for the thorough review.  I have updated both PRs according to the proposed split:
    
1. Renamed and Relocated: Renamed module to `gcs-sync`, relocated to `community/modules/file-system/gcs-sync/`, and removed the empty `agent-platform` folder.
2. Git-Only Scope: Dropped local path support and the 4-level `try()` ladder. Added validation requiring `repo_url` across all `files` and `directories` entries with a clear error pointing users to `gcs-objects` for local content.
3. DAG Ordering in Outputs: Derived `staged_directory_uris`, `staged_file_uris`, `all_staged_uris`, and `bucket_name` through `depends_on` to enforce strict DAG ordering for downstream resources.
4. gcloud storage rsync: Added `--delete-unmatched-destination-objects` and added validation gating `destination_path` so it cannot target bucket root (`/`).
5. Portability: Dropped `interpreter = ["/bin/bash"]` and ensured `stage_to_gcs.sh` is portable POSIX `/bin/sh`.
6. Documentation: Documented DAG ordering as the primary justification for the module, documented cleanup limitations (and why destroy provisioners are avoided), and noted the recommendation to pin `repo_ref`.

community/modules/file-system/gcs-sync/outputs.tf
hamadfadi: Done as suggested.

community/modules/agent-platform/stage-to-gcs/main.tf
hamadfadi: Done as suggested.

community/modules/agent-platform/stage-to-gcs/main.tf
hamadfadi: Done as suggested.

SwarnaBharathiMantena (approved):
LGTM

agrawalkhushi18 (approved):

PR #6311, opened 2026-09-11
feat(files): add gcs-file community module

hamadfadi:
Add reusable community module to create and upload a single file to a Google Cloud Storage (GCS) bucket from inline content or a local file source.

### Submission Checklist


community/modules/files/gcs-file/main.tf
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+resource "google_storage_bucket_object" "file" {
+  name         = var.object_path
+  bucket       = var.bucket_name
+  content      = var.content
+  source       = var.source_path
+  content_type = var.content_type
+}
```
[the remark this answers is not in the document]
hamadfadi: Fixed added:
detect_md5hash = var.source_path != null ? filemd5(substr(var.source_path, 0, 1) == "/" ? var.source_path : "${path.root}/${var.source_path}") : null

community/modules/files/gcs-file/variables.tf
```diff
@@ -0,0 +1,51 @@
...
+    condition     = can(regex("^[a-z0-9][a-z0-9._-]{1,61}[a-z0-9]$", var.bucket_name))
+    error_message = "bucket_name must be a valid GCS bucket name."
+  }
+}
+
+variable "object_path" {
+  description = "GCS object path (e.g., 'config/variables.env' or 'scripts/setup.sh')."
+  type        = string
+  validation {
+    condition     = length(trimspace(var.object_path)) > 0 && !startswith(var.object_path, "/")
```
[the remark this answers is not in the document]
hamadfadi: Fixed, changed to:
condition     = length(trimspace(var.object_path)) > 0 && substr(var.object_path, 0, 1) != "/"

community/modules/files/gcs-file/versions.tf
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+terraform {
+  required_version = ">= 1.0"
```
agrawalkhushi18: Repo standard is required_version = ">= 1.12.2". Both modules/file-system/cloud-storage-bucket/versions.tf and the sibling community/modules/files/fsi-montecarlo-on-batch/versions.tf pin this. Please match.

community/modules/files/gcs-file/versions.tf
```diff
@@ -0,0 +1,25 @@
...
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 4.42"
+    }
```
agrawalkhushi18, line 23: This module is missing its provider_meta block.

Please add:

```
  provider_meta "google" {
    module_name = "blueprints/terraform/hpc-toolkit:gcs-file/v1.103.0"
  }
```
The version must be exactly v1.103.0. tools/create-release-candidate.sh bumps versions via a literal git sed of the current release string, so any other value will never be rewritten and drifts permanently.

community/modules/files/gcs-file/metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+
+spec:
+  requirements:
+    services:
```
agrawalkhushi18, line 18: Consider adding the toolkit-specific ghpc section with the exclusive module-level validator for content / source_path:

```
  ghpc:
    validators:
    - validator: exclusive
      inputs: [content, source_path]
```
Context: the provider already declares ExactlyOneOf: [content, source], so both "neither set" and "both set" fail cleanly at terraform plan — a lifecycle { precondition } would be redundant. But tools/validate_configs/validate_configs.sh only runs terraform init -backend=false + terraform validate, never plan, and validate does not evaluate variables. So a broken example blueprint would sail through CI. This validator closes that gap at gcluster create time.

community/modules/files/gcs-file/variables.tf
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "bucket_name" {
+  description = "Target GCS bucket name where the file will be stored."
+  type        = string
+  validation {
+    condition     = can(regex("^[a-z0-9][a-z0-9._-]{1,61}[a-z0-9]$", var.bucket_name))
```
agrawalkhushi18: This rejects valid input while catching very little:
- It caps length at 63. Dot-separated GCS bucket names are legal up to 222 characters, so those are wrongly rejected.
- It does not catch consecutive dots, the reserved goog prefix, names containing google, or IP-address-formatted names — i.e. the things that actually fail at the API.
can we loosen the sanity check and recheck what is recommended ?

community/modules/files/gcs-file/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+```yaml
+- group: files
+  modules:
+  - id: script-file
+    source: community/modules/files/gcs-file
+    settings:
+      bucket_name: my-storage-bucket
+      object_path: scripts/setup.sh
+      source_path: ./scripts/setup.sh
```
agrawalkhushi18: This example may not work as written. main.tf resolves relative paths against path.root (the deployment-group directory), not the blueprint directory, and filemd5() on a missing file fails at plan time with a raw error.

Please change to:
`      source_path: $(ghpc_stage("./scripts/setup.sh"))`
and add a short note documenting that source_path must be staged, matching modules/scripts/startup-script/README.md.

community/modules/files/gcs-file/metadata.yaml
hamadfadi: Thank you for this suggestion. Added, done.

community/modules/files/gcs-file/versions.tf
hamadfadi: Thank you for this info. Added, done.

community/modules/files/gcs-file/versions.tf
hamadfadi: Thank you for this info. Added, done.

community/modules/files/gcs-file/variables.tf
hamadfadi: Similar to other modules, we removed this condition check. Thank you for noticing this.

community/modules/files/gcs-file/README.md
hamadfadi: Updated accordingly.

hamadfadi (commented):
@agrawalkhushi18  thank you for your helpful review and feedback. All your suggestions and the issues you raised are fixed.

agrawalkhushi18 (approved):

SwarnaBharathiMantena:
Thanks for this PR, Fadi — the declarative approach here is right, and I'd like to make this the primary module for getting things into GCS. A few changes to widen its scope and settle where it lives.

I observed that this PR and PR #6309 are raised to do related tasks - not the same. So, I think they should stay separate and split on **whether Terraform owns the object**. This module owns its objects — they're in state, they show drift, and `terraform destroy` removes them. #6309 can't offer that for Git-sourced content. Concretely that means this module takes everything local and inline, and #6309 narrows to Git only. More detail on #6309.

I captured all most comments here so that it is easy to carry context between different suggestions:

1. Rename to `gcs-objects`, move to `community/modules/file-system/`

    There is a suggestion on this comment (number 2 below) to extend the scope from files to all objects. So, I think `gcs-objects` fits better.

    Both this PR and #6309 fit better in the `community/modules/file-system/` folder. Can we move then there and update the associated README docs?

    `community/modules/files/gcs-file/` → `community/modules/file-system/gcs-objects/`

2. Widen from one object to lists of files and directories

    Right now this handles a single object. To be the general answer for local content it needs to cover directories too — otherwise every user with a directory is pushed toward #6309 and the split stops working. It also confuses an end user on which module to use.

    Suggested interface, deliberately parallel to #6309's so the two read alike:
    ```hcl
    variable "files" {
      description = "Individual files to upload, from inline content or a local path."
      type = list(object({
        destination_path = string
        source_path      = optional(string) # local path, via ghpc_stage
        content          = optional(string) # inline; mutually exclusive with source_path
        content_type     = optional(string)
      }))
      default = []
      validation {
        condition = alltrue([
          for f in var.files :
          (f.content != null) != (f.source_path != null)
        ])
        error_message = "each entry in files must set exactly one of content or source_path."
      }
    }
    variable "directories" {
      description = "Local directories to upload recursively."
      type = list(object({
        destination_path = string
        source_path      = string # local path, via ghpc_stage
        exclude          = optional(list(string), [])
      }))
      default = []
    }
    ```

    Directories expand with `fileset()`:
    ```hcl
    locals {
      directory_objects = merge([
        for d in var.directories : {
          for rel in fileset("${path.root}/${d.source_path}", "**") :
          "${d.destination_path}/${rel}" => "${path.root}/${d.source_path}/${rel}"
          if !anytrue([for pattern in d.exclude : can(regex(pattern, rel))])
        }
      ]...)
    }
    ```

    Please document whether `exclude` is regex or glob. The snippet above assumes regex to match #6309; either is fine, but a user reading `*.Dockerfile` will assume glob, so it needs to be explicit.

    The `exclusive` validator in `metadata.yaml` is worth a look too — it only fires when *more than one* input is set, so it won't catch an entry with neither `content` nor `source_path`. The `alltrue` block above covers both directions.

3. `detect_md5hash` → `source_md5hash`, and fix the `content` case

     There's a real bug in the current expression:

     ```hcl
     detect_md5hash = var.source_path != null ? filemd5(...) : null
     ```

     When `content` is set, `source_path` is null, so the hash is `null` and **editing inline content never triggers an update**. Please cover both branches, and switch to `source_md5hash` — the provider steers you to it for content that changes, and it's what `modules/scripts/startup-script/main.tf:304` uses:

     ```hcl
     source_md5hash = each.value.content != null ? md5(each.value.content) : filemd5(each.value.source_path)
     ```

4. Require `ghpc_stage` for local paths

    Right now `source` gets the raw `var.source_path` while the hash resolves `${path.root}/${source_path}` — two different rules for one input. They happen to agree when Terraform's working directory is the root module, which is the normal `gcluster` case, so this isn't urgent; it's just a trap for anyone who runs Terraform from elsewhere.

    Standardising on `$(ghpc_stage("..."))` and resolving consistently against `${path.root}` fixes it and matches @agrawalkhushi18's earlier point and `modules/scripts/startup-script`. Worth a line in the README showing the blueprint-side usage, since that's the part users get wrong.

5. `versions.tf`

    - `google` floor `>= 4.42` → `>= 6.41`, required for `source_md5hash` (same floor as `modules/scripts/startup-script/versions.tf`).
    - Add:

    ```hcl
    provider_meta "google" {
      module_name = "blueprints/terraform/hpc-toolkit:gcs-objects/v1.104.0"
    }
    ```

hamadfadi:
@SwarnaBharathiMantena Thanks for the thorough review. I have updated the PR according to the proposed split:

1. Renamed to `gcs-objects` and relocated to `community/modules/file-system/gcs-objects/`.
2. Expanded interface to support both `files` (inline content or local file) and `directories` (recursive upload with RE2 regex exclusions).
3. Fixed MD5 hashing by adopting `source_md5hash = each.value.content != null ? md5(each.value.content) : filemd5(each.value.source_path)` and standardizing on `${path.root}` via `$(ghpc_stage(...))`.
4. Bumped Google provider to `>= 6.41` and added the `provider_meta` block.

SwarnaBharathiMantena (approved):
LGTM

agrawalkhushi18 (approved):

PR #6295, opened 2026-09-09
feat(project): add pre-existing-project community module

hamadfadi:
Introduce a reusable community Terraform module to query and expose Google Cloud project attributes (project number, ID, and name) via the google_project data source.

### Submission Checklist

NOTE: Community submissions can take up to 2 weeks to be reviewed.

Please take the following actions before submitting this pull request.

* Fork your PR branch from the Toolkit "develop" branch (not main)
* Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
* Confirm that "make tests" passes all tests
* Add or modify unit tests to cover code changes
* Ensure that unit test coverage remains above 80%
* Update all applicable documentation
* Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)

PR #6312, opened 2026-09-11
feat(project): add project-metadata community module

hamadfadi:
Add reusable community module to set project compute metadata items on a Google Cloud project.

### Submission Checklist


hamadfadi:
Why this is needed:

- Sets one or more Google Cloud Compute Engine project metadata items (google_compute_project_metadata_item).
- Accepts either a single key and value pair or a map of metadata dictionary items.

community/modules/project/project-metadata/main.tf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  combined_metadata = var.key != "" && var.value != "" ? merge(var.metadata, { (var.key) = var.value }) : var.metadata
```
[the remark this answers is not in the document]
hamadfadi: Fixed as suggested:
 combined_metadata = var.key != "" ? merge(var.metadata, { (var.key) = var.value }) : var.metadata

SwarnaBharathiMantena:
Hi @hamadfadi , the existing modules on the repo currently use the data type `data "google_project"`. As this is a sinlge data type, I think it is simpler to place it in the individual module where required rather than create a new module.

Example: https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/6f755b4fe2cf780c0f82b24e7b05bb0ea3448c72/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf#L80

Please share your thoughts.

hamadfadi:
> Hi @hamadfadi , the existing modules on the repo currently use the data type `data "google_project"`. As this is a sinlge data type, I think it is simpler to place it in the individual module where required rather than create a new module.
> 
> Example:
> 
> https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/6f755b4fe2cf780c0f82b24e7b05bb0ea3448c72/community/modules/scheduler/schedmd-slurm-gcp-v6-controller/controller.tf#L80
> 
> Please share your thoughts.

Hi @SwarnaBharathiMantena thank you for the feedback. I am not sure if the comment is for this PR or for the other PR for the project-info module: https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/6295.

SwarnaBharathiMantena:
Hi @hamadfadi , yes you are right. My comment was supposed to target the PR #6295. I added the comment on that PR.

community/modules/project/project-metadata/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+  type        = string
+}
+
+variable "metadata" {
+  description = "A map of key/value pairs to set as project compute metadata items."
+  type        = map(string)
+  default     = {}
+}
+
+variable "key" {
```
SwarnaBharathiMantena: As the module accepts a single key-value pair or metadata map, these variables become redundant. Can we remove the key and value input variables, and have metadata as the single input entry for key-value pairs? 

In order to pass the variables key and value using the `$(...)` parser format, we can update metadata to a list object and update resource this way:

```hcl
variable "metadata" {
  description = "Project metadata items to set."
  type = list(object({
    key   = string
    value = string
  }))
  # no default -> gcluster rejects a module that omits it
  validation {
    condition     = length(distinct([for m in var.metadata : m.key])) == length(var.metadata)
    error_message = "metadata keys must be unique."
  }
}
```
```hcl
resource "google_compute_project_metadata_item" "items" {
  for_each = { for m in var.metadata : m.key => m.value }
  project  = var.project_id
  key      = each.key
  value    = each.value
}
```

Three things this buys:
- **Any number of computed keys.** `key`/`value` gives exactly one per module instance.
- **Duplicate keys stop being silent.** I checked all three ways a duplicate can arise:
  | | Today | With a list |
  |---|---|---|
  | Same key twice in the YAML map | caught at parse: `mapping key "FOO" already defined` | n/a |
  | `key` scalar names something already in `metadata` | **silent** — scalar wins, map entry discarded | n/a |
  | Same key twice in the list | n/a | caught — `metadata keys must be unique.` |
  
  The middle row is the one to fix. `-var 'metadata={"FOO":"from-map"}' -var 'key=FOO' -var 'value=from-scalar'` produces `{ FOO = "from-scalar" }` — the map entry is gone, with no error or warning. And it gets harder to spot once keys are computed, since a collision between `"$(vars.a)_NAME"` and a literal entry only exists after interpolation, where the YAML parser can't see it.
- **No silent no-op.** Today `metadata = {}` with `key = ""` — both defaults — creates zero resources and reports success, which is a confusing thing to debug. With one required input, `gcluster` catches it at expand time instead: `Error: a required setting "metadata" is missing from a module "<your-module-id>"`, with the offending line. That's earlier and clearer than anything Terraform would say.

community/modules/project/project-metadata/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+  description = "GCP project ID where compute metadata will be set."
+  type        = string
+}
+
+variable "metadata" {
+  description = "A map of key/value pairs to set as project compute metadata items."
+  type        = map(string)
+  default     = {}
+}
+
```
SwarnaBharathiMantena, line 28: Please expose `deletion_policy`. It can be module-wide.

I noticed the resource supports `deletion_policy`, defaulting to `"DELETE"`; I confirmed `"ABANDON"` is accepted. On destroy the key is removed rather than restored to any previous value, so for keys other things depend on, leaving them in place is often what you want. An optional input keeping `"DELETE"` as the default would be a no-op for current behaviour and a useful escape hatch.

community/modules/project/project-metadata/versions.tf
```diff
@@ -0,0 +1,28 @@
...
+terraform {
+  required_version = ">= 1.12.2"
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 4.42"
+    }
+  }
+  provider_meta "google" {
+    module_name = "blueprints/terraform/hpc-toolkit:project-metadata/v1.103.0"
```
SwarnaBharathiMantena: Please update this to `v1.104.0`.

community/modules/project/project-metadata/README.md
```diff
@@ -0,0 +1,76 @@
+# Project Metadata Module
+
+Sets one or more Google Cloud Compute Engine project metadata items (`google_compute_project_metadata_item`). Accepts either a single `key` and `value` pair or a map of `metadata` dictionary items.
+
```
SwarnaBharathiMantena, line 4: Two points worth mentioning in the README: 
- Project metadata applies to **every VM in the project**, including instances this blueprint didn't create — `enable-oslogin`, `ssh-keys` and `startup-script` all change behaviour project-wide. Worth stating given how small the module looks.
- Two deployments in the same project setting the same key will collide — per the note in my top-level comment, the second one fails to apply rather than overwriting. That's the safe outcome, but a confusing one if the key name isn't obviously deployment-scoped.

SwarnaBharathiMantena (commented):
Hi Fadi, thanks for raising this PR.

In addition to the comments at specific lines / files, this is one of my observations based on my exploration on day-2 operations with help from Gemini:

#### Applying to a metadata key that already exists fails
| Scenario | Result |
|---|---|
| Key absent | Created fine |
| Re-apply, state intact | Clean no-op — properly idempotent |
| Key pre-set via `gcloud`, not in state | **Fails** |
| Key created by TF then dropped from state | **Fails** |
```
Error: key "<key>" already present in metadata for project "<project>".
Use `terraform import` to manage it with Terraform
```
This is deliberate provider behaviour rather than a transient issue — `resourceComputeProjectMetadataItemCreate` passes the `failIfPresent` constant.

The **`plan` succeeds.** Terraform reported `Plan: 1 to add, 0 to change, 0 to destroy` and only failed during apply. So `gcluster deploy` gets through planning and dies partway through the apply, after earlier resources in the group already exist.
**It isn't a drop-in for what it replaces.** The design doc lists this module as replacing `gcloud compute project-info add-metadata`. That command *upserts* — it overwrites an existing key silently. This resource errors instead. So on any project where the previous flow has already run, the first `gcluster deploy` with this module fails on keys that flow created. That's the migration path, so it's the normal case rather than an edge case.
There's no clean in-module fix. I spent a while trying to move this to `plan` time before concluding that, since failing at plan would clearly beat dying mid-apply. Four approaches, all actually run:
| Approach | Catches it at plan | Re-plan stays clean | Allows a legit value change | Visible via `gcluster` |
|---|---|---|---|---|
| `precondition` on key alone | yes | **no — errors forever** | no | yes |
| `precondition` on key **+ value** | yes | yes | **no — blocked** | yes |
| `check` block on key + value | warning only | yes | yes | **no — swallowed** |
| `gcluster` validator | yes, pre-Terraform | **no — same blindness** | no | yes |

Details, in case you want to retrace any of it:
- **There's no data source.** `data "google_compute_project_metadata"` doesn't exist, and nothing in the provider returns `commonInstanceMetadata` — I went through `terraform providers schema -json` to be sure. A `data "http"` probe against `compute/v1/projects/<project>` with a `google_client_config` token does work and reads the real key list, so the data is reachable; the difficulty is the comparison.
- **Comparing on the key alone breaks re-plans.** Once the module has applied its own key, that key is in project metadata, so every subsequent plan fails the precondition — a first-run failure traded for a permanent one.
- **Comparing on key *and value* fixes that, but blocks updates.** "Key exists and holds a different value" gives a correct plan-time error for a foreign key and a clean `No changes` on re-plan — but changing a value you already manage (`v1` → `v2`) also trips it, so you could never update a metadata value. That rules it out.
- **A `check` block behaves correctly but never reaches the user.** It warns for a foreign key, stays silent on an unchanged re-plan, and lets updates through. However `gcluster`'s `planModule` captures `PlanJSON` into a buffer it only reads on error, `ShowPlanFileRaw` doesn't render check results (I grepped — zero occurrences of the message), and `--auto-approve` never renders a plan at all. The warning is invisible in every path.
- **A `gcluster` validator would be visible, but has the same flaw.** `validators.Execute` receives the blueprint and the validator's inputs and has no access to Terraform state, so it can't tell a foreign key from one this deployment already manages — it would fire on every re-deploy, same as the precondition. It's also not something a module can opt into today; validators are registered centrally and keyed off blueprint vars.

The root cause is that the guard needs to know "is this key already in my state", and Terraform doesn't expose state membership to HCL — and `gcluster` doesn't expose it to validators either. Every proxy for it is unknown at plan on the first run, which is exactly the run that matters.

Worth saying explicitly, because it makes this less alarming than it reads: nothing is destroyed. `failIfPresent` means the module *cannot* overwrite a pre-existing key, and since the create fails the key never enters state, so `destroy` can't remove it either. The gap is purely *when* you find out and *what the error tells you to do* — `terraform import` isn't reachable from a blueprint, since `gcluster` owns the state. So I think this is a docs-and-expectations matter:
- State in the README that the module manages only keys it creates, and that applying to a pre-existing key fails with the above error.
- Recommend deployment- or experiment-scoped key names. The design doc's `"$(vars.user_experiment_name)_USER_EXPERIMENT_NAME"` pattern does this well and is worth promoting into the README.
- Give the recovery step: `gcloud compute project-info remove-metadata --keys <key>`, then re-run.
Please do reproduce on your side — happy to be corrected.

hamadfadi:
@SwarnaBharathiMantena Thanks for the thorough review and insights. I have addressed all the feedback:

1. Consolidated to `metadata` list: Dropped the redundant `key` and `value` scalar variables. `metadata` is now a required `list(object({ key = string, value = string }))` with uniqueness validation across keys.
2. Exposed `deletion_policy`: Added `deletion_policy` defaulting to `"DELETE"`, with `"ABANDON"` supported as an escape hatch for keys that should persist post-destroy.
3. Bumped Provider Meta: Updated `module_name` in `versions.tf` to `v1.104.0`.
4. Documentation:
       - Documented project-wide scope and VM impact.
       - Documented the `failIfPresent` collision behavior when applying against pre-existing keys.
       - Recommended scoping keys with deployment/experiment prefixes.
       - Documented the recovery step (`gcloud compute project-info remove-metadata`).

community/modules/project/project-metadata/README.md
hamadfadi: Done as suggested.

community/modules/project/project-metadata/versions.tf
hamadfadi: Done as suggested.

community/modules/project/project-metadata/variables.tf
hamadfadi: Done as suggested.

community/modules/project/project-metadata/variables.tf
hamadfadi: Done as suggested.

SwarnaBharathiMantena (approved):
LGTM

ishitachail (approved):

PR #6313, opened 2026-09-11
Add missing "parallelism" variable to CTK Batch job template module.

jacobw125:
- Allows users to configure parallelism of their cloud batch jobs ([docs](https://docs.cloud.google.com/python/docs/reference/batch/latest/google.cloud.batch_v1.types.TaskGroup)). This is especially useful for array jobs.

### Submission Checklist

- [X] Fork your PR branch from the Toolkit "develop" branch (not main)
- [X] Test all changes with pre-commit in a local branch [#](https://goo.gle/hpc-toolkit#development)
- [X] Confirm that "make tests" passes all tests
- [X] Add or modify unit tests to cover code changes
- [X] Ensure that unit test coverage remains above 80%
- [X] Update all applicable documentation
- [X] Follow Cluster Toolkit Contribution guidelines [#](https://goo.gle/hpc-toolkit-contributing)


modules/scheduler/batch-job-template/main.tf
```diff
@@ -33,6 +33,7 @@ locals {
       synchronized       = var.mpi_mode
       runnables          = local.runnables
       task_count         = var.task_count
+      parallelism        = var.parallelism
```
Neelabh94: When `mpi_mode = true`, the template injects synchronization barriers. Verified against the Cloud Batch API: Batch rejects any job containing barriers where `parallelism != task_count` (`INVALID_ARGUMENT: Barriers require task_count = parallelism`).

Because `var.submit` defaults to `false`, `terraform apply` succeeds and writes an invalid job file, only failing when submitted. Consider adding a plan-time precondition to `local_file.job_template` to catch this early:

```hcl
    precondition {
      condition     = !var.mpi_mode || var.parallelism == null || var.parallelism == var.task_count
      error_message = "When mpi_mode is true, Batch requires parallelism == task_count (barriers require task_count = parallelism)."
    }
```

modules/scheduler/batch-job-template/variables.tf
```diff
@@ -68,6 +68,17 @@ variable "task_count_per_node" {
   default     = null
 }
 
+variable "parallelism" {
```
Neelabh94, line 71: None of the test blueprints under `tools/validate_configs/` currently set `parallelism`, so `make validate_configs` will only ever exercise the `null` default and the new template branch remains uncovered.

Please add test coverage for this new variable in an existing test blueprint like `tools/validate_configs/test_configs/multiple-batch-jobs.yaml`. Because that blueprint defines a standard (non-MPI) array job, setting `parallelism < task_count` will properly exercise concurrency throttling in CI:

```yaml
      task_count: 4
      parallelism: 2
```

modules/scheduler/batch-job-template/main.tf
shubpal07: +1 to Neelabh's point on catching explicit barrier mismatches. 

To take this a step further and follow the existing pattern in `main.tf` (like `tasks_per_node` on [line 25](https://github.com/jacobw125/cluster-toolkit/blob/ad6e66e26083c6da07a4dca31070b312600bbef7/modules/scheduler/batch-job-template/main.tf#L25)), we can make the module auto-default `parallelism` to `var.task_count` when `mpi_mode = true` and `var.parallelism` is omitted:

```hcl
locals {
  # ...
  job_template_contents = templatefile(
    "${path.module}/templates/batch-job-base.yaml.tftpl",
    {
      # ...
      task_count  = var.task_count
      parallelism = var.parallelism != null ? var.parallelism : (var.mpi_mode ? var.task_count : null)
      # ...

File: modules/scheduler/batch-job-template/variables.tf [hunk 3]
```diff
@@ -68,6 +68,17 @@ variable "task_count_per_node" {
...
+  description = "Max number of tasks that can run in parallel across the job."
+  type        = number
+  default     = null
+
+  validation {
+    condition     = var.parallelism == null ? true : (var.parallelism > 0 && floor(var.parallelism) == var.parallelism)
+    error_message = "The parallelism value must be an integer greater than 0."
+  }
+}
```
shubpal07: Could we update the variable description to clarify the default behavior and MPI semantics?
```hcl
variable "parallelism" {
  description = "Max number of tasks that can run in parallel across the job. If null, defaults to min(task_count, 1000) for standard jobs, or task_count if mpi_mode is true. When mpi_mode is true, Batch requires parallelism == task_count."
  type        = number
  default     = null
  validation {
    condition     = var.parallelism == null ? true : (var.parallelism > 0 && floor(var.parallelism) == var.parallelism)
    error_message = "The parallelism value must be an integer greater than 0."
  }
}

shubpal07 (changes requested):
+1 to @Neelabh94's comments regarding the plan-time barrier precondition and adding CI coverage in `tools/validate_configs/test_configs/multiple-batch-jobs.yaml`.
To complement that:
1. **Auto-fallback in `main.tf`**: Recommend setting `parallelism = var.parallelism != null ? var.parallelism : (var.mpi_mode ? var.task_count : null)` (mirroring `tasks_per_node` on line 25). This automatically handles the omission case in MPI jobs cleanly.
2. **Precondition**: Keep Neelabh's precondition in `local_file.job_template` to guard against explicit mismatches (`parallelism != task_count` when `mpi_mode = true`).
3. **Docs**: Update the description in `variables.tf` and run `make terraform-format`.
4. **CI**: Verify tests pass with `make validate_configs` after adding `parallelism: 2` to `multiple-batch-jobs.yaml`.

modules/scheduler/batch-job-template/variables.tf
jacobw125: FYI, I don't see an existing array job in `tools/validate_configs/test_configs/multiple-\batch-jobs.yaml`, but I went ahead and added this, and tests are passing.

jacobw125:
@shubpal07 and @Neelabh94, thanks for your reviews. I have applied your suggestions.

Neelabh94 (approved):

shubpal07 (approved):

PR #6262, opened 2026-09-03
feat(examples): add Slurm hybrid cloud burst blueprints and instructions for GCD

juntangc:
## Description

Adds turnkey Cluster Toolkit blueprints and documentation for **Multi-Cluster Slurm with elastic cloud bursting across two autonomous projects in Google Cloud Dedicated (GCD)** / sovereign cloud environments.

## What's added

New directory `community/examples/hpc-slurm-google-cloud-dedicated/hybrid-slurm-cluster/`:

* **`primary-cluster.yaml`** — Primary cluster (Project A). Deployment groups:
  * `setup`: VPC `primary-cluster-net` (`10.128.0.0/20`) + cross-cluster peering firewall rules.
  * `build-slurm`: Packer `modules/packer/custom-image` build of a Rocky Linux 8 + Slurm `6.12.1` image into family `slurm-c3-rocky8` (one-time per project).
  * `cluster`: standalone NFS server (`community/modules/file-system/nfs-server`, Hyperdisk Balanced) exporting `/home` and `/opt/apps`, Slurm v6 controller/login/nodeset (`compute` partition), and a startup script that exports the master `slurm.key` and controller IP to shared `/home`.
* **`burst-cluster.yaml`** — Burst target (Project B). `setup` provisions a non-overlapping VPC `burst-cluster-net` (`10.130.0.0/20`); `cluster` mounts the primary cluster's NFS `/home` and `/opt/apps` via `pre-existing-network-storage`, reuses the pre-baked Slurm image, and defines an elastic `burst` partition (`node_count_static: 0`, `node_count_dynamic_max: 16`, `c3-standard-176`) that scales `0 -> N` on demand. Its startup script imports the shared `slurm.key`, discovers the primary controller IP from `/home/.primary_cluster_ctrl_ip`, and appends `AccountingStorageExternalHost=<primary>:6819` so the burst cluster auto-registers in the primary SlurmDBD — no manual `sacctmgr` cross-registration.
* **`README.md`** — Step-by-step deployment guide: architecture diagram, GCD-vs-commercial-GCP differences table, WIF login and `GHPC_MOCK_MACHINE_CONFIG` setup, one-time Packer image build, strict deploy order, bidirectional VPC peering commands, and verification of cross-cluster dispatch (`sinfo -M burst`, `sbatch -M burst --wrap="hostname"`, `squeue -M burst`).

Supporting updates:

[...]

PR #6323, opened 2026-09-14
fix: skip state disk mount when already mounted

juntangc:
`mount_save_state_disk()` calls `mount {mount_point}` unconditionally. `util.run()`
defaults to `check=True` and mount(8) exits non-zero when the target is already
mounted, so controller setup raises `CalledProcessError` on reboot.

The function itself appends a persistent `/etc/fstab` entry, so systemd mounts the
disk automatically on subsequent boots before `setup.py` runs. First boot is
unaffected; every reboot after that fails.

Guard the call with `mountpoint -q` and log when the mount is skipped.

Requested by @AdarshK15 in #6262, where this was carried as a runtime regex
patch against `/slurm/scripts/setup.py`. That workaround was removed from that PR;
this lands the permanent fix so standard GCP deployments get it too.

The check uses `mountpoint -q` rather than `os.path.ismount()`:
`posixpath.ismount` decides via `s1.st_dev != s2.st_dev or s1.st_ino == s2.st_ino`,
so it misses bind mounts within a single filesystem, whereas `mountpoint` consults
`/proc/self/mountinfo` via libmount. 
This also matches the existing predicate pattern in this file in setup_cloud_ops() (run("systemctl is-active --quiet ...", check=False).returncode).

No unit test here: there's no `test_setup.py` in the repo yet, and covering
`mount_save_state_disk()` requires mocking `lookup()`, `util.slurmdirs`, `util.run`,
`util.chown_slurm`, `os.makedirs` and `/etc/fstab` read+append — a lot of scaffolding
for a three-line guard. Happy to add one if you'd prefer; just wanted to keep the
change minimal.


agrawalkhushi18:
Nice @juntangc , clean check using os.path.ismount to avoid shelling out to mountpoint -q. Approved.

agrawalkhushi18 (dismissed):

community/modules/scheduler/schedmd-slurm-gcp-v6-controller/modules/slurm_files/scripts/setup.py
```diff
@@ -201,7 +201,10 @@ def mount_save_state_disk():
     util.run(f"systemctl daemon-reload")
 
     os.makedirs(mount_point, exist_ok=True)
-    util.run(f"mount {mount_point}")
+    if os.path.ismount(mount_point):
```
AdarshK15: `os.path.ismount` may fail in some scenarios like bind mounts.
Can we change this to
```
if util.run(f"mountpoint -q {mount_point}", check=False).returncode == 0:
    log.info(f"Mount point {mount_point} is already mounted. Skipping mount.")
else:
    util.run(f"mount {mount_point}")
```

juntangc: good catch.  updated.

AdarshK15 (approved):
LGTM

agrawalkhushi18 (approved):

PR #6330, opened 2026-09-15
fix(job): initialize dynamic client in gcluster job inspect to prevent nil pointer panic

Neelabh94:
### Summary
Fixes a deterministic nil pointer dereference (`SIGSEGV`) in `gcluster job inspect --name <workload>` when `--gke-namespace` is omitted

### Root Cause
1. In `InspectCluster`, `g.configureKubectl(...)` calls `g.getCurrentNamespace(...)`, which invokes `g.getKubeClient()`.
2. Because `InspectCluster` had not yet called `g.getDynamicClient()`, `g.getKubeClient()` cached `g.kubeClient = &DefaultKubeClient{dynClient: nil}`.
3. Later in `inspectWorkload(...)`, `if g.kubeClient != nil` evaluated to `true` and called `g.findTargetWorkload(...)` -> `DefaultKubeClient.ListWorkloads(...)`, which dereferenced `d.dynClient` (`nil`) and crashed.
4. Additionally, when `--gke-namespace` was explicitly provided, `g.kubeClient` remained `nil` in `inspectWorkload(...)`, causing Kueue workload discovery (`jobset-<name>-<hash>`) to be silently skipped.

### Solution
- Centralize synchronization between `g.dynClient` and `g.kubeClient` in `syncKubeClient()` across `SetDynamicClient()`, `getDynamicClient()`, and `getKubeClient()`.
- Ensure `inspectWorkload()` and `findTargetWorkload()` lazily initialize `g.dynClient` via `g.getDynamicClient()` when `g.kubeClient` is `nil` or holds a `nil` `dynClient`.
- Add defensive nil checks in `DefaultKubeClient` (`ListWorkloads`, `DeleteJobSet`, `ListJobSets`) so they return descriptive errors rather than panicking if `dynClient` is `nil`.
- Add comprehensive unit tests covering nil dynamic client guards and Kueue workload discovery during cluster inspection.

### Testing
- `go test -v ./pkg/orchestrator/gke/...`
- `pre-commit run`


pkg/orchestrator/gke/inspector.go
agrawalkhushi18: InspectCluster (the caller) already resolves the exact target namespace by calling g.getCurrentNamespace (around line 78). Calling it a second time here inside inspectWorkload creates unnecessary API overhead.

Suggestion: Update the inspectWorkload function signature to accept the pre-resolved targetNamespace string from its caller. You can then safely delete this entire redundant resolution block (lines from ns, err := g.getCurrentNamespace down through the logging.Warn() fallback).

agrawalkhushi18: If you refactor inspectWorkload to natively accept the workloadNamespace string from its caller (as suggested above), this initial workloadNamespace := "default" assignment becomes obsolete. Relying strictly on the dynamically resolved namespace from InspectCluster eliminates the risk of accidentally running queries against namespace="default" if logic errors are introduced later.

Neelabh94: Thanks for the suggestion! We've refactored  to accept the pre-resolved  directly from . This eliminates the redundant resolution logic and the fallback to 'default', making the inspection pipeline cleaner and more robust.

Neelabh94: Agreed. The fallback to 'default' has been removed along with the redundant namespace resolution block.

agrawalkhushi18 (approved):

PR #6335, opened 2026-09-16
feat(job): add GCSFuse storage profiles and volume attributes

agrawalkhushi18:
This PR introduces native support for GKE GCSFuse Storage Profiles and CSI volume attributes to `gcluster job submit`, granting users granular control over FUSE caching behavior and mount performance tuning on a per-mount basis.

By extending the `--mount` flag syntax to support `profile=` and `attributes=`, we achieve parity with XPK's caching mechanisms while automating the lifecycle of dynamic PVs and PVCs cleanly behind the scenes. 

### Key Changes
* **Storage Profiles Integration (`profile=`)**:
  * Users can pass `profile=<training|checkpointing|serving>` to provision GCS FUSE volumes using pre-installed `gcsfusecsi-<profile>` StorageClasses.
  * Automatically handles the dynamic rendering of collision-free, pre-bound PVs and PVCs (named `gcluster-gcsfuse-<bucket>-<profile>`), falling back cleanly to ephemeral inline CSI volumes if no profile is specified.
* **CSI Volume Attributes (`attributes=`)**:
  * Users can pass comma-separated advanced CSI tuning parameters overriding the profile's defaults (e.g. `attributes=fileCacheCapacity=100Gi,gcsfuseLoggingSeverity=debug`).
  * Seamlessly applies attributes to either the dynamic PersistentVolume (`PV.Spec.CSI.VolumeAttributes`) or Inline CSI (`Pod.Volumes.CSI.VolumeAttributes`).
* **SubPath & AccessMode Delegation**: 
  * Prevents locking the PV to a specific sub-folder or read-only mode, correctly delegating `subPath` and read-only properties to the Pod's `volumeMounts` for re-usability across jobs.
* **De-duplication**: 
  * Correctly coalesces identical Bucket + Profile requirements so jobs mounting multiple sub-directories from the same bucket only provision one shared gateway PV/PVC, reducing cluster bloat.
* **Annotations Injection**:
  * Ensures `.NeedsGCSFuseSidecar` is surfaced properly through templates so `gke-gcsfuse/volumes: "true"` gets appropriately injected when GCS mounts are present.
* **Refactored Test Completeness**:
  * Added extensive unit-tests for parsing validation and yaml verification, actively eliminating cyclomatic bloat (`go-cyclo`) in `storage_test.go` and `storage.go`.

### Submission Checklist


docs/gcluster_job_guide.md
```diff
@@ -215,6 +221,194 @@ Mounting an existing PVC named `lustre-pvc` (read-only):
...
+> (`anywhereCacheZones`, see below) is enabled - on any profile. It is not a
+> performance recommendation you can trade away: a cross-region bucket is
+> unsupported in these configurations.
+>
+> `gcluster` does **not** validate this today. A mismatch is not rejected at
+> submit time and surfaces later as degraded or failing mounts. Check the bucket
+> location yourself before submitting:
+>
+> ```bash
+> gcloud storage buckets describe gs://<YOUR_BUCKET_NAME> --format="value(location)"
```
Neelabh94, line 248: Why can we not do it internally? Like we should document this hard requirement but then do the check internally as well, since we already know the cluster's region this should be super simple, wdyt?

docs/gcluster_job_guide.md
```diff
@@ -215,6 +221,194 @@ Mounting an existing PVC named `lustre-pvc` (read-only):
...
+> ```bash
+> kubectl get pvc,pv -l gcluster.google.com/managed-by=cluster-toolkit
+> kubectl delete pvc,pv -l gcluster.google.com/managed-by=cluster-toolkit
+> ```
+>
+> The `gcluster.google.com/managed-by: cluster-toolkit` label is written onto
+> every GCS FUSE storage-profile gateway `gcluster` creates, so the selector
+> above will not touch storage you created yourself. Note it is *only* on those
+> gateways: the PV/PVC pair generated for a `filestore://` mount carries no
+> labels, so it is not matched by the selector and is left alone.
```
Neelabh94: We should make the experience consistent here, is it possible to add the same labels for filestore:// PVC that we are generating?

docs/gcluster_job_guide.md
```diff
@@ -215,6 +221,194 @@ Mounting an existing PVC named `lustre-pvc` (read-only):
...
+> [!IMPORTANT]
+> **Known limitation: storage gateways are not reclaimed automatically.**
+> Nothing in `gcluster` deletes a gateway PV/PVC today - not job completion, not
+> `ttlSecondsAfterFinished`, and not `gcluster job cancel`. This is not a leak
+> that breaks anything: the gateway is deterministic per (bucket, profile,
+> namespace) and is *reused* by the next job that mounts the same bucket with the
+> same profile, so resubmitting works and no bucket data is affected. The cost is
+> a leftover PV/PVC object pair per distinct gateway.
+>
+> Remove them by hand when you no longer need them:
```
Neelabh94, line 295: nit/ for future: can we do it as part of gcluster job cancel (could be an additional flag --clear-storage??)

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
```diff
@@ -0,0 +1,49 @@
+apiVersion: v1
```
Neelabh94, line 1: Just a nit, we do provide the capability to our users to override any template with their own, please ensure this feature is also compatible and can used overridden templates.

pkg/orchestrator/gke/storage.go
```diff
@@ -33,49 +37,77 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// mountBuildState maps generated PV name -> Pod volume name to de-duplicate gateways within a job.
+type mountBuildState struct {
```
Neelabh94: all type definitions should live in types.go

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
```diff
@@ -0,0 +1,49 @@
...
+  name: {{.PVName}}
+  labels:
+    gcluster.google.com/managed-by: cluster-toolkit
+    gcluster.google.com/storage-type: gcsfuse
+spec:
+  storageClassName: {{.StorageClassName}}
+  capacity:
+    storage: {{.Capacity}}
+  accessModes:
+  - ReadWriteMany
```
Neelabh94, line 13: Parameterize the accessModes in the template to support ReadOnlyMany. To make this work without clashes, you would need to include the requested access mode (ro or rw) in the hash used to generate the PV/PVC names, thereby creating isolated gatew

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
```diff
@@ -0,0 +1,49 @@
...
+    name: {{.PVCName}}
+{{- if .MountOptions }}
+  mountOptions:
+{{- range .MountOptions }}
+  - {{ printf "%q" . }}
+{{- end }}
+{{- end }}
+  csi:
+    driver: gcsfuse.csi.storage.gke.io
+    volumeHandle: {{ printf "%q" .VolumeHandle }}
```
Neelabh94, line 26: Quote all string injections, especially map keys, using {{ printf "%q" .Var }}.

```YAML
# In the PV/PVC metadata:
name: {{ printf "%q" .PVName }}
namespace: {{ printf "%q" .Namespace }}

# In the CSI volumeAttributes:
{{- range $key, $value := .VolumeAttributes }}
  {{ printf "%q" $key }}: {{ printf "%q" $value }}
{{- end }}
```

pkg/orchestrator/gke/storage.go
```diff
@@ -240,7 +467,36 @@ func validateSrcScheme(src string, vStr string) error {
...
+		ns, err = sm.orchestrator.getCurrentNamespace(job.ClusterName, job.ClusterLocation, job.ProjectID)
+		if err != nil {
+			logging.Warn("failed to get current namespace: %v. Defaulting to 'default' for PV name.", err)
+		}
+	}
+	if ns == "" {
+		ns = "default"
+	}
+	return ns
+}
```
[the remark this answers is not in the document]
Neelabh94: I agree, unless it is an RBAC permission issue, we should avoid silent fallbacks and fail loudly

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +562,155 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
 	}
 	pvYAML := buf.String()
 
+	return info, pvYAML, nil
+}
+
+const (
```
Neelabh94: nit: Let's declare all constants together at the top of the file so that they are easy to maintain

docs/gcluster_job_guide.md
Neelabh94, line 1: We should try to reduce the text here by two methods:
1> gcluster should do the validation that we are currently not doing and then in the README we simply mention a list of pre-requistes
2> Details around GKE behaviors and how gcluster implements/handles them can be added to docs/advanced_gke_features.md and we cna add a reference symlink to them here. 

Necessary details & examples should be kept in the file but we want to avoid bloat so that the file is readable by new users of the tool without causing an information overload

Neelabh94 (changes requested):

docs/gcluster_job_guide.md
agrawalkhushi18: Yes, we would be handling this in `gcluster job cancel` itself. I have planned to take that up in the very next PR as a follow up along with pre-flight IAM validation for storage profiles.

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
agrawalkhushi18: Done, this was previously done with reference to the filestore template file. Do we anticipate these changes in the filestore tmpl as well?

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
agrawalkhushi18: Good point, I checked this specifically. The new template goes through the same loader as the existing ones, so overrides already work , `generateGCSFuseProfileResources` calls parseGKETextTemplate("gcs_fuse_pv_pvc.tmpl"), which is exactly what filestore.tmpl does today. It looks in `--gke-custom-templates-path` first and only falls back to the embedded copy if the file isn't there.

Extraction is covered too, since `ExtractDefaultTemplates` just walks the embedded templates/ directory without any allowlist, so gcluster job gke-template-extract picks up `gcs_fuse_pv_pvc.tmpl` along with everything else.

I've added TestGCSFuseProfile_HonorsCustomTemplateOverride. It covers three cases: an override genuinely being rendered, falling back to the embedded default when the file is missing, and a broken override failing with an error that names the template rather than doing something silently odd. Thanks for flagging it.

docs/gcluster_job_guide.md
agrawalkhushi18: Agreed, and this is already marked for the follow-up PR that adds the pre-flight checks (alongside the bucket IAM validation) — I kept this PR to parsing and manifest generation deliberately so the runtime-check surface lands together and can be reasoned about as one thing. I have added it as a requirement for now and will automate it in nxt PR.

docs/gcluster_job_guide.md
agrawalkhushi18: Good call, done. Both the PV and PVC for filestore:// mounts now carry gcluster.google.com/managed-by: cluster-toolkit and gcluster.google.com/storage-type: filestore. This is basically for the cleanup mechansim we will be adding in the next PR.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done

pkg/orchestrator/gke/storage.go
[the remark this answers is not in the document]
agrawalkhushi18: Done, resolveNamespace now returns an error and both call sites propagate it.

On the RBAC carve-out: doesn't apply here. getCurrentNamespace only reads the local kubeconfig, it never calls the API server, so the only two failures are "kubeconfig won't load" and "no context matching this cluster" both already point at --gke-namespace.

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
Neelabh94: yes, we should update the filestore template as well if needed!

agrawalkhushi18: Sure, keeping this PR scope to GCSfuse, will do it in a separate PR.

docs/gcluster_job_guide.md
agrawalkhushi18: Thanks Neelabh, I have refined the job_submit_guide. PTAL

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
agrawalkhushi18: I looked at how XPK handles this. It's a bit different there: readonly is a property of a named storage object you create up front with xpk storage create, and each one of those owns its own PV and PVC. So there's no cost to putting readonly on it, because the user already decided how many storage objects they wanted.
We don't have that. There's no named storage object in gcluster, the gateway just gets derived from the mount string, and it's shared between mounts and between jobs. So if we put ro/rw into the PV, we automatically stop sharing it. That's really the whole problem in one line: the PV is a shared object, but ro/rw is a per-mount thing, so you can't put one into the other without splitting the PV.
On honouring what the user typed, I think we already do, just not in that field. ro sets readOnly: true on the volume mount, and that's the one the kubelet actually enforces. accessModes isn't an enforcement field, it's only used to match a PVC to a PV when Kubernetes is picking one dynamically. We pre-bind the pair ourselves with claimRef and volumeName, so that matching never runs. Putting ReadOnlyMany there wouldn't stop a write, it would just look like it does, which feels worse to me than leaving it off. GKE's own storage profile examples use ReadWriteMany on these for the same reason.
The other thing is cost. options= and attributes= are opt-in, so hashing them only affects people who use them. But ro/rw is on every mount, so hashing it splits gateways for everyone. The two-mount example in the guide would go from one PV and PVC to two, for the same bucket and the same profile, and that's more objects for the cleanup logic to keep track of later.
If what we want is a mount that genuinely can't be written to, that already works today: options=o=ro. gcsfuse enforces it, and since options= is part of the name hash it gets its own gateway anyway.

docs/gcluster_job_guide.md
```diff
@@ -173,9 +173,15 @@ If you want to run a job across multiple groups of GPU nodes (e.g., 2 groups of
...
 * `options` is optional and allows passing custom mount options (currently only supported for GCS fuse volumes).
+* `profile` is optional, GCS-only, and selects a [GKE Cloud Storage FUSE storage profile](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/gcsfuse-profiles) (see below).
+* `attributes` is optional, GCS-only and sets key-value pairs for CSI volume settings (`key=val,...`).
+  * **Without `profile=`**: Applies standard [GCSFuse CSI volume attributes](https://docs.cloud.google.com/kubernetes-engine/docs/reference/cloud-storage-fuse-csi-driver/volume-attr) directly to the inline mount.
+  * **With `profile=`**: Overrides default [storage profile StorageClass parameters](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/gcsfuse-profiles#storageclass_configuration_reference) on the generated PVs.
+
+  NOTE: Use `options=` rather than `attributes=mountOptions=...` for gcsfuse mount flags; `gcluster` routes `options=` into the correct field on both paths and rejects `mountOptions` as an attribute to keep the two from silently overwriting each other.
```
Neelabh94: Do we throw an error if this is not followed?

docs/gcluster_job_guide.md
```diff
@@ -200,9 +206,26 @@ Mounting a GCS bucket (read-write):
...
+```bash
+./gcluster job submit \
+  --name my-training-job \
+  --command "python app.py" \
+  --compute-type n2-standard-32 \
+  --base-image python:3.9-slim \
+  --build-context job_details \
+  --service-account "workload-identity-k8s-sa" \
+  --mount "gs://<DATASET_BUCKET>/imagenet;/data;ro;profile=training" \
+  --mount "gs://<CKPT_BUCKET>/run-42;/checkpoints;rw;profile=checkpointing"
```
Neelabh94: Did we explain or export these variables?

docs/gcluster_job_guide.md
Neelabh94: Please make sure we update the README again in the follow-up PR. Ideally it would be better if both PRs go in the same release to avoid flip flop type of an experience for the user

pkg/orchestrator/gke/storage.go
```diff
@@ -163,44 +205,218 @@ func missingDestOrFormatErr(vStr string) error {
...
+		}
+	}
+
+	return out, nil
+}
+
+func validateGCSOnlySegments(pm parsedMount, profileSet bool) error {
+	if strings.HasPrefix(pm.Src, "gs://") {
+		return nil
+	}
```
Neelabh94, line 378: Consider a practical use case where the user is unaware of this restriction and tries to pass options, profile and attributes for a filestore.
When they try for the first time, they get an error "options= is currently...."
so they remove the options= and try again, they get a new error "profile= is only...." and then the third time is "attributes= is only.." This will be bad UX.

I see two workarounds, either we check for all three together in an OR condition and print a single generic message, something like: "options, profile and attributes are only supported for GCS ...."

Or we could check for all three, whichever is present, we add it to a list of incorrect params and then print them, somthing like this: "<My incorrect params list> is only supported ...."

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +548,151 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+	// Append a digest if sanitization collapsed '.' or '_' so distinct buckets never collide.
+	lossy := name != bucket
+	if budget > 0 && (lossy || len(name) > budget) {
+		name = shortenWithDigest(name, bucket, budget)
+	}
+
+	return sanitizePVCName(prefix + name + "-" + suffix)
+}
+
+func gcsFuseGatewayPVName(claim, namespace string) string {
```
Neelabh94: This can be inline, the thin wrapper is not needed

pkg/orchestrator/gke/storage.go
```diff
@@ -339,11 +718,15 @@ func (sm *StorageManager) AddVolumeOptions(opts *ManifestOptions, vols []MountIn
...
 
 	for _, v := range vols {
 		mountSpecs = append(mountSpecs, buildVolumeMountSpec(v))
-		volSpecs = append(volSpecs, buildVolumeSpec(v))
-		if v.Type == "gcsfuse" {
+		if !seenVolumes[v.Name] {
+			seenVolumes[v.Name] = true
+			volSpecs = append(volSpecs, buildVolumeSpec(v))
+		}
+		if v.Type == "gcsfuse" || v.NeedsGCSFuseSidecar {
```
Neelabh94, line 744: In what scenario will we need the OR condition? Shouldn't chekcing v.Type=="gcsfuse" be enough here?

Neelabh94 (changes requested):

pkg/orchestrator/gke/storage_test.go
```diff
@@ -1137,3 +1471,934 @@ func TestProcessMounts_Filestore_IPv6(t *testing.T) {
...
+	}
+
+	assertProfilePVSpec(t, findDoc(t, docs, "PersistentVolume"), wantPV, wantPVC)
+	assertProfilePVCSpec(t, findDoc(t, docs, "PersistentVolumeClaim"), wantPV, wantPVC)
+}
+
+func renderGCSFuseGateway(t *testing.T, params GCSFusePVPVCTemplateParams) string {
+	t.Helper()
+	var g *GKEOrchestrator
+	tmpl, err := g.parseGKETextTemplate("gcs_fuse_pv_pvc.tmpl")
```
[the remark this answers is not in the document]
agrawalkhushi18: No panic is possible.

Proof chain:
parseGKETextTemplate never touches a field on g. Its only use of the receiver is g.getTemplatePath() on line 63.
getTemplatePath opens with an explicit nil guard: if `g != nil && g.gkeCustomTemplatesPath != ""`. On a nil receiver, short-circuit evaluation skips the field access and it falls straight to return `"", nil`.
Back in the caller, `customPath == ""`, so the whole custom-path block (67–77) is skipped and line 78 returns template.ParseFS(templatesFS, ...).

docs/gcluster_job_guide.md
agrawalkhushi18: Yes. It's a hard error at validation time, not a warning.
parseVolumeAttributes rejects reserved keys, and that runs inside ValidateMounts, which is the first thing SubmitJob calls before initializeJobSubmission, so it fails .

pkg/orchestrator/gke/storage.go
```diff
@@ -163,44 +205,218 @@ func missingDestOrFormatErr(vStr string) error {
...
+		if len(pair) != 2 || strings.TrimSpace(pair[0]) == "" {
+			return nil, fmt.Errorf("invalid volume attribute %q. Expected format: attributes=<key>=<value>[,<key>=<value>...]", kv)
+		}
+		key := strings.TrimSpace(pair[0])
+		if !volumeAttributeKeyPattern.MatchString(key) {
+			return nil, fmt.Errorf("invalid volume attribute key %q. Keys may only contain letters, digits, '.', '_' and '-'", key)
+		}
+		if key == "mountOptions" {
+			return nil, fmt.Errorf("volume attribute %q is not accepted; use options=<opt1>,<opt2> instead, which gcluster renders into the correct field for both inline and storage-profile mounts", key)
+		}
```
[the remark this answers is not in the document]
agrawalkhushi18: good suggestion, added.

docs/gcluster_job_guide.md
agrawalkhushi18: Modified the note for more clarity.

docs/gcluster_job_guide.md
agrawalkhushi18: Yes, both PRs would go in the same release.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done

pkg/orchestrator/gke/storage.go
agrawalkhushi18: The OR is needed because `profile= mounts` are Type: "pvc", not "gcsfuse",  the pod uses the gateway PVC and the gcsfuse driver sits on the PV behind it. Checking only Type == "gcsfuse" would skip them, so the `gke-gcsfuse/volumes: "true"` annotation wouldn't be added and the sidecar wouldn't be injected. Checking only Type == "pvc" doesn't work either, because a user's own PVC is also "pvc" and must not get a sidecar. NeedsGCSFuseSidecar is what separates those two.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Agreed, Went with your second suggestion — a single generic message would lose track of which segment the user actually passed. It now collects every offending segment and reports them together, plus the source that failed (handy when several --mount flags are in play). Test added for the multi-segment case.

docs/gcluster_job_guide.md
agrawalkhushi18: Added a note for these placeholders.

pkg/orchestrator/gke/storage.go
```diff
@@ -240,7 +456,30 @@ func validateSrcScheme(src string, vStr string) error {
...
+	if len(name) > maxLen {
+		name = strings.TrimRight(name[:maxLen], "-")
+	}
+	return name
+}
+
+func (sm *StorageManager) resolveNamespace(job orchestrator.JobDefinition) (string, error) {
+	if sm.orchestrator == nil {
+		return "default", nil
+	}
```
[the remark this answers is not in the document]
agrawalkhushi18: Not reachable in practice — all three production call sites pass the orchestrator, so this only ever triggers in unit tests that construct a bare &StorageManager{}. No real submission can hit it.

pkg/orchestrator/gke/storage.go
```diff
@@ -163,44 +208,234 @@ func missingDestOrFormatErr(vStr string) error {
...
 
-	if options != "" && !strings.HasPrefix(src, "gs://") {
-		return "", "", "", false, fmt.Errorf("options= is currently only supported for GCS fuse volumes (gs://...)")
+	if segments.ProfileSet {
+		sc, err := normalizeProfileName(segments.RawProfile)
+		if err != nil {
+			return parsedMount{}, err
+		}
+		pm.Profile = sc
+		if _, _, err := splitGCSSource(pm.Src); err != nil {
```
Neelabh94: `splitGCSSource` is only called when `profile=` is set, and its results are discarded — it's acting purely as a validator here.

For `--mount "gs://my-bucket/subdir;/data"` (no profile), `buildVolumeSpec` falls through to `strings.TrimPrefix(v.Source, "gs://")` at line 801 and emits `bucketName: "my-bucket/subdir"`. That isn't a valid bucket, so the mount fails at pod start. The same source *with* `profile=` works, because the subpath is split out and delegated to `volumeMounts.subPath`.

Pre-existing behaviour, but the guide you're adding uses `gs://<YOUR_DATASET_BUCKET>/imagenet`, so dropping `profile=` from a documented example now silently breaks.

Suggest splitting for every `gs://` mount:

```go
if strings.HasPrefix(pm.Src, "gs://") {
	bucket, subPath, err := splitGCSSource(pm.Src)
	if err != nil {
		return parsedMount{}, err
	}
	pm.Src = "gs://" + bucket
	pm.SubPath = subPath
}
```

and setting `info.SubPath = pm.SubPath` in `buildPodLevelMount`, so inline mounts delegate the subpath the same way.

pkg/orchestrator/gke/storage.go
```diff
@@ -113,23 +153,28 @@ func (sm *StorageManager) ValidateMounts(mounts []string) error {
...
 		}
 
-		if seenSources[src] {
-			return fmt.Errorf("duplicate volume source: %s", src)
+		sourceKey := pm.Src
+		if pm.Profile != "" {
+			sourceKey = pm.Src + ";profile=" + pm.Profile
+		}
+
+		if seenSources[sourceKey] {
```
Neelabh94, line 188: `sourceKey` keys on `Src` + `Profile`, but `gcsFuseGatewayPVCName` also hashes `options=`/`attributes=` specifically so those mounts get their own gateway. The validator is stricter than the feature it guards.

This is rejected today with `duplicate volume source`, though it's exactly the two-gateway case the hash exists for:

```bash
--mount "gs://bkt;/a;profile=training;options=implicit-dirs" \
--mount "gs://bkt;/c;profile=training;options=file-cache:max-size-mb:2000"
```

Keying on the same inputs the gateway name derives from keeps the two in sync:

```go
sourceKey := pm.Src
if pm.Profile != "" {
	sourceKey = pm.Src + ";profile=" + pm.Profile
	if hash := customGatewayOptionsHash(pm.Options, pm.Attributes); hash != "" {
		sourceKey += ";" + hash
	}
}
```

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +567,147 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+		return "", "", fmt.Errorf("invalid GCS bucket name %q in mount %q", bucket, src)
+	}
+	if len(parts) > 1 {
+		subPath = strings.Trim(parts[1], "/")
+	}
+	return bucket, subPath, nil
+}
+
+// customGatewayOptionsHash returns a deterministic suffix for custom options/attributes, or "" when none are supplied.
+func customGatewayOptionsHash(options string, attrs map[string]string) string {
```
Neelabh94, line 612: `customGatewayOptionsHash` covers only user-supplied `options=`/`attributes=`. Everything gcluster derives — `volumeHandle`, derived `volumeAttributes`, `mountOptions`, `capacity`, `storageClassName` — sits outside the hash, so changing any of them keeps the gateway name while changing the body. `spec.persistentVolumeSource` is immutable, so `kubectl apply` is rejected with `Forbidden: is immutable after creation`.

Nothing has to be deleted for this to happen — the gateway is long-lived by design, so a gcluster upgrade that touches the template, or a modified `--gke-custom-templates-path`, is enough. `job submit` then dies on a raw kubectl error naming a PV the user never knowingly created.

Folding a template version into the hash would self-heal but silently orphans the old gateway. Pre-flighting the existing PV is probably the better trade, and covers the stale-`claimRef` case too — a PV left `Released` by an externally deleted PVC won't rebind, leaving the new PVC `Pending` indefinitely:

```go
return MountInfo{}, "", fmt.Errorf(
	"gateway PV %q already exists with different settings (created by an older gcluster "+
		"version or a custom template). Delete it with `kubectl delete pv %s` and resubmit",
	pvName, pvName)
```

pkg/orchestrator/gke/storage_test.go
```diff
@@ -1137,3 +1477,938 @@ func TestProcessMounts_Filestore_IPv6(t *testing.T) {
...
+		t.Fatalf("unexpected error: %v", err)
+	}
+	opts := &ManifestOptions{}
+	sm.AddVolumeOptions(opts, infos)
+	if !opts.GCSFuseEnabled {
+		t.Error("GCSFuseEnabled must be true for profile-backed PVC mounts")
+	}
+}
+
+func TestGCSFuseProfile_SharedGatewayDeduplication(t *testing.T) {
```
Neelabh94, line 2146: Every test constructs `&StorageManager{}`, so `sm.orchestrator` is nil and `resolveNamespace` short-circuits at the nil branch. The error propagation you asked for last round is never exercised — 75% coverage with both error branches uncovered.

Worth injecting a resolver that errors, and one returning `""`, asserting `ProcessMounts` surfaces `failed to resolve namespace for storage gateway` and `Specify one explicitly with --gke-namespace`.

Three smaller gaps:
- Mixed `ro`/`rw` on one shared gateway — delegating read-only to `volumeMounts` is the crux of the shared-PV design, but this test only uses `ro`. A regression pushing `readOnly` back onto the shared volume would pass today.
- Two mounts differing only by `options=`/`attributes=` must produce two gateways — no negative de-dup test.
- `customGatewayOptionsHash` sorts keys so `a=1,b=2` and `b=2,a=1` hash identically; nothing asserts that.

pkg/orchestrator/gke/storage.go
```diff
@@ -240,7 +475,30 @@ func validateSrcScheme(src string, vStr string) error {
...
+// truncatePVCName enforces maxLen without leaving a trailing '-'.
+func truncatePVCName(name string, maxLen int) string {
+	if len(name) > maxLen {
+		name = strings.TrimRight(name[:maxLen], "-")
+	}
+	return name
+}
+
+func (sm *StorageManager) resolveNamespace(job orchestrator.JobDefinition) (string, error) {
+	if sm.orchestrator == nil {
```
Neelabh94, line 511: `orchestrator` is set at all three production sites (`gke_job_orchestrator.go:95`, `:687`, `manifest_generator.go:186`), so this branch is test-only — and there it hands back `"default"`, masking the error path below it.

Dropping it and injecting a resolver in tests would remove the last silent fallback. If it stays as scaffolding, a comment saying so would help the next reader. (Also covers the open bot comment on these lines.)

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +567,147 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+		Source:              pvcName,
+		MountPath:           pm.Dest,
+		Type:                "pvc",
+		ReadOnly:            pm.ReadOnly,
+		SubPath:             subPath,
+		NeedsGCSFuseSidecar: true,
 	}
 
-	return info, pvYAML, nil
+	if existing, ok := state.gatewayVolumeNames[pvName]; ok {
```
Neelabh94: Identical to the get-or-assign block in `generateFilestoreResources` at line 542 — a small method on `mountBuildState` keeps them from drifting:

```go
func (s *mountBuildState) volumeNameFor(pvName string, idx int) (string, bool) {
	if existing, ok := s.gatewayVolumeNames[pvName]; ok {
		return existing, true
	}
	name := fmt.Sprintf("vol-%d", idx)
	s.gatewayVolumeNames[pvName] = name
	return name, false
}
```

docs/gcluster_job_guide.md
```diff
@@ -215,6 +240,43 @@ Mounting an existing PVC named `lustre-pvc` (read-only):
   --mount "lustre-pvc;/data"
 ```
 
+#### GCS FUSE storage profiles (`profile=`)
```
Neelabh94, line 243: Following up on the doc-bloat point from last round: this section may fit better in `docs/gke-advanced-features.md`, which is organised one section per advanced GKE feature. Storage profiles map onto that shape — blueprint prerequisites (GKE version, `enable_gcsfuse_csi`), cluster requirements (Service Agent IAM, region co-location), `gcluster job submit` usage, example, GKE doc link. Section 4 (Multi-Tier Checkpointing) is the closest precedent.

`--mount` is a core flag unlike `--gke-mtc-enabled`, so a split rather than a move:
- **Keep here:** the format string, one example, the flag-table row.
- **Move there as section 5:** prerequisites, the IAM/region `IMPORTANT` block, gateway sharing and naming behaviour, attributes reference — with a link back.

Fine as follow-up work alongside the cleanup PR; not blocking this one.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Fixed — parseSingleVolume now splits every gs:// source once into pm.Src = "gs://" + bucket and pm.SubPath, and buildPodLevelMount sets info.SubPath = pm.SubPath. Also updated generateGCSFuseProfileResources to reuse pm.SubPath (dropping the duplicate splitGCSSource call) and included pm.SubPath in ValidateMounts's sourceKey so mounting multiple subfolders of the same bucket continues to work. Verified with unit tests and live on GKE.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Fixed, ValidateMounts now appends customGatewayOptionsHash(pm.Options, pm.Attributes) to sourceKey so two mounts on the same bucket and profile with distinct options= or attributes= pass validation and get their separate gateways, while identical options=/attributes= are still rejected. Added unit tests in TestGCSFuseProfile_SameBucketTwoProfilesPassesValidation.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done — added checkExistingGatewayPV in generateGCSFuseProfileResources. Before returning the rendered manifest (when not a dry run), it inspects kubectl get pv <pvName> --ignore-not-found -o json against the rendered PV spec (reflect.DeepEqual(existing.Spec, rendered.Spec)) and returns the kubectl delete pv <pvName> error both when the existing PV's spec differs (e.g., template changes) and when the PV is stuck in Released/Failed state from an externally deleted PVC. Added unit tests in TestGCSFuseProfile_ExistingGatewayPVCheck .

pkg/orchestrator/gke/storage_test.go
agrawalkhushi18: Done, added TestGCSFuseProfile_ResolveNamespaceErrors (using MockKubeClient to cover both the lookup-error and empty-namespace branches of resolveNamespace) and updated TestGCSFuseProfile_SharedGatewayDeduplication to cover mixed ro/rw mounts on a shared gateway, attribute key-order invariance (a=1,b=2 vs b=2,a=1), and negative deduplication when options= differ.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Added a comment noting sm.orchestrator == nil is only for unit tests (since existing tests in storage_test.go create &StorageManager{} directly), and added TestGCSFuseProfile_ResolveNamespaceErrors to test both error paths below it.

docs/gcluster_job_guide.md
agrawalkhushi18: Sure, will take it up in the next PR. Thanks for the suggestion.

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +586,179 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+	if existing.Status.Phase == "Released" || existing.Status.Phase == "Failed" {
+		return fmt.Errorf(
+			"gateway PV %q already exists in %s state (left behind after its PVC was deleted) and cannot rebind. "+
+				"Delete it with `kubectl delete pv %s` and resubmit",
+			pvName, existing.Status.Phase, pvName)
+	}
+
+	if !reflect.DeepEqual(existing.Spec, rendered.Spec) {
+		return fmt.Errorf(
+			"gateway PV %q already exists with different settings (created by an older gcluster "+
```
Neelabh94: The fix this message suggests doesn't work in the common case. It shows up when the gateway's PVC is still `Bound`, often while another job is using it. [PV protection](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#storage-object-in-use-protection) postpones PV deletion until the PV is no longer bound, so:

1. `kubectl delete pv <name>` hangs in the user's terminal (`--wait` defaults to true) and the PV sits in `Terminating`.
2. If they Ctrl-C and resubmit, `kubectl get pv` still returns the old spec, so they get the same message again, with no hint that deletion is already pending.

The message should give the full, safe sequence and say what it disrupts. It should also handle `Terminating` separately (add `metadata.deletionTimestamp` to `existingGatewayPV`):

```go
	if existing.Metadata.DeletionTimestamp != "" {
		return fmt.Errorf("gateway PV %q is being deleted and is waiting for PVC %s/%s to be released; "+
			"cancel the jobs that mount it, then run `kubectl delete pvc %s -n %s` and resubmit", pvName, ns, pvcName, pvcName, ns)
	}
	if !reflect.DeepEqual(existing.Spec, rendered.Spec) {
		return fmt.Errorf("gateway PV %q already exists with different settings. It is shared by every job in namespace %q "+
			"that mounts this bucket/profile. To replace it: cancel those jobs, then run "+
			"`kubectl delete pvc %s -n %s && kubectl delete pv %s`, and resubmit. "+
			"Bucket data is not affected (reclaimPolicy: Retain)", pvName, ns, pvcName, ns, pvName)
	}
```
(`ns`/`pvcName` need to be passed in.) See the naming comment below for a way to make this branch mostly unreachable.

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +586,179 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+		return "", "", fmt.Errorf("invalid GCS bucket name %q in mount %q", bucket, src)
+	}
+	if len(parts) > 1 {
+		subPath = strings.Trim(parts[1], "/")
+	}
+	return bucket, subPath, nil
+}
+
+// customGatewayOptionsHash returns a deterministic suffix for custom options/attributes, or "" when none are supplied.
+func customGatewayOptionsHash(options string, attrs map[string]string) string {
```
Neelabh94: Upgrade concern: the gateway name only hashes the user's `options=`/`attributes=`. Anything else that ends up in the immutable PV spec isn't in the name:

- the embedded template
- `gcsFuseGatewayCapacity`
- any default attribute we add later
- a user's `--gke-custom-templates-path` override

So the first release that changes any of these gives every existing user the same PV name with a different spec. Every resubmit that touches an existing gateway then fails `checkExistingGatewayPV`. The only way out is the teardown above, which means stopping running jobs.

Suggestion: derive the suffix from the **rendered PV spec** instead of the raw inputs. Same inputs still give the same name, so sharing is unchanged. But any change in the spec (release, default or custom template) produces a new gateway: running jobs keep the old one, new jobs get the new one, and nothing has to be torn down by hand. The old gateway then becomes a labelled orphan for the planned cleanup in `job cancel` to sweep. Roughly:

```go
// render first, then:
specSum := sha256.Sum256([]byte(renderedPVSpec)) // or of the params struct + template bytes
suffix := sanitizePVCName(profileShortName) + "-" + hex.EncodeToString(specSum[:])[:6]
```

If you'd rather keep the current scheme, a version segment in the prefix (`gcluster-gcsfuse-v1-...`) is the minimum, so the next spec change has a clean way forward. Either way, please add a table test that pins exact names for a few inputs (canonical, options, attributes, dotted bucket), e.g. `gcsFuseGatewayPVCName("imagenet-dataset", "training", "", nil) == "gcluster-gcsfuse-imagenet-dataset-training"`. Today's tests only check relative properties, so a refactor that silently renames every gateway on upgrade wouldn't be caught.

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +586,179 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+	if res.ExitCode != 0 || strings.TrimSpace(res.Stdout) == "" {
+		return nil
+	}
+
+	var existing, rendered existingGatewayPV
+	if yaml.Unmarshal([]byte(res.Stdout), &existing) != nil || yaml.Unmarshal([]byte(renderedYAML), &rendered) != nil {
+		return nil
+	}
+
+	if existing.Status.Phase == "Released" || existing.Status.Phase == "Failed" {
```
Neelabh94: Detecting `Released`/`Failed` is a good catch; otherwise the new PVC would sit `Pending` forever. But gcluster could fix this itself instead of pushing it onto the user. A `Released` PV has no bound claim, so no pod can be using it, and with `reclaimPolicy: Retain` deleting the PV object never touches bucket data. When the PV carries our `gcluster.google.com/managed-by: cluster-toolkit` label, deleting it and letting the apply recreate it is safe:

```go
	if (existing.Status.Phase == "Released" || existing.Status.Phase == "Failed") &&
		existing.Metadata.Labels["gcluster.google.com/managed-by"] == "cluster-toolkit" {
		logging.Info("Recreating stale gateway PV %q (%s)", pvName, existing.Status.Phase)
		if res := sm.orchestrator.executor.ExecuteCommand("kubectl", "delete", "pv", pvName, "--wait=true"); res.ExitCode != 0 {
			return fmt.Errorf("failed to delete stale gateway PV %q: %s", pvName, res.Stderr)
		}
		return nil
	}
```

Once the follow-up PR cleans up PVCs on `job cancel`, every cancel-then-resubmit will leave a `Released` PV behind. Without this, the error will become the normal second-submit experience.

pkg/orchestrator/gke/storage.go
```diff
@@ -305,15 +586,179 @@ func (sm *StorageManager) handleFilestoreMount(src, dest string, readOnly bool,
...
+	return info, buf.String(), nil
+}
+
+func (sm *StorageManager) checkExistingGatewayPV(pvName, renderedYAML string, dryRun bool) error {
+	if dryRun || sm.orchestrator == nil || sm.orchestrator.executor == nil {
+		return nil
+	}
+
+	res := sm.orchestrator.executor.ExecuteCommand("kubectl", "get", "pv", pvName, "--ignore-not-found", "-o", "json")
+	if res.ExitCode != 0 || strings.TrimSpace(res.Stdout) == "" {
```
Neelabh94: With `--ignore-not-found`, NotFound already comes back as exit 0 with empty stdout. So a non-zero exit here is a real failure: expired credentials, wrong context, no cluster-scoped `get pv` permission, or API throttling. Returning `nil` skips the Released/mismatch checks, and the user later gets the raw `spec.persistentvolumesource: Forbidden: ... immutable` error from `apply`, which is exactly what this function is meant to prevent. The same applies to the swallowed `Unmarshal` errors on line 744. I'd fail fast:

```go
	if res.ExitCode != 0 {
		return fmt.Errorf("failed to inspect existing gateway PV %q (needs cluster-scoped `get pv`): %s", pvName, strings.TrimSpace(res.Stderr))
	}
	if strings.TrimSpace(res.Stdout) == "" {
		return nil
	}
```

docs/gcluster_job_guide.md
```diff
@@ -215,6 +240,43 @@ Mounting an existing PVC named `lustre-pvc` (read-only):
...
+> requirement, though same-region buckets remain the better choice for
+> throughput and egress cost.
+
+Behaviour worth knowing:
+
+* The generated claim is named `gcluster-gcsfuse-<bucket>-<profile>` and the
+  PersistentVolume `gcluster-gcsfuse-<bucket>-<profile>-<namespace>`. The names
+  are deterministic, so several jobs that use the same bucket and profile in the
+  same namespace **share one gateway** rather than each creating their own.
+* Supplying custom `options=` or `attributes=` gives that mount its own gateway
```
Neelabh94: This section explains how gateways are created, but not how to manage them afterwards. The most common Day-2 loop is tuning: submit with `attributes=fileCacheCapacity=50Gi`, then `100Gi`, then `200Gi`. Each iteration leaves another hashed PV/PVC behind, and users have no documented way to see or clean them up. Suggest adding:

```markdown
* Gateways outlive the job. List them with
  `kubectl get pv,pvc -A -l gcluster.google.com/managed-by=cluster-toolkit,gcluster.google.com/storage-type=gcsfuse`.
* To remove one, first make sure no running job mounts it, then delete the claim before the volume:
  `kubectl delete pvc <claim> -n <namespace> && kubectl delete pv <claim>-<namespace>`.
  Bucket contents are not affected.
```

The follow-up cleanup-on-cancel PR can then replace the manual step with the flag.

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
```diff
@@ -0,0 +1,49 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: {{ printf "%q" .PVName }}
+  labels:
+    gcluster.google.com/managed-by: cluster-toolkit
```
Neelabh94: The planned cleanup, and the auto-recreate suggested above, will select on these labels. Right now they are literals in two templates. Templates can be overridden with `--gke-custom-templates-path`, so an override written before this PR, or one that just drops the block, produces gateways the cleanup will never find. Consider defining the keys and values as Go constants, passing them in (`{{ .ManagedByLabel }}`), and having `checkExistingGatewayPV` log a warning when an existing gateway PV is missing them.

Neelabh94 (changes requested):
Requesting changes for the gateway lifecycle on resubmits and gcluster upgrades (storage.go:757 and :612); details inline.

pkg/orchestrator/gke/templates/gcs_fuse_pv_pvc.tmpl
agrawalkhushi18: Done: the label keys and values are now constants, passed to both templates as .ManagedByLabel/.ManagedByValue/.StorageTypeLabel/.StorageType. We warn at render time if a gateway PV is missing managed-by, so an override that drops the block is flagged on its first submit, for both backends. I tried the warning in checkExistingGatewayPV too, but tested live it was a false alarm. I stripped the labels from an existing gateway PV/PVC and resubmitted, and the submit's kubectl apply re-added them. So an existing unlabelled gateway fixes itself on the next submit whenever the rendered manifest has the labels. When it doesn't, the render-time warning already fires.

docs/gcluster_job_guide.md
agrawalkhushi18: Done, added section.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done: a non-zero exit from kubectl get pv now fails fast with the kubectl stderr, and only empty stdout is treated as "no PV". Both parse failures now return an error too. Added test cases for the kubectl-failure and bad-JSON paths.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done: a Released/Failed gateway PV carrying our managed-by label is now deleted and recreated by the apply; unlabelled ones keep the error. I added --timeout=60s so --wait can't hang, and --ignore-not-found for concurrent submits. Verified live: deleting only the PVC, then resubmitting, logs "Recreating stale gateway PV" and both the new PV and the PVC bind. One note: the follow-up cancel cleanup deletes both PVC and PV, so it won't create Released PVs itself. The main trigger is a manual PVC-only delete.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Done. Added deletionTimestamp handling, checked first, with a message pointing at the PVC that holds it. The mismatch message now gives the full sequence (cancel jobs → delete PVC → delete PV), says the gateway is shared across the namespace, and notes bucket data is safe. I kept the "older gcluster version or custom template" cause and added kubectl describe pvc (Used By) so users can find the jobs to cancel. Verified live: a --wait=false delete on a bound gateway gives the new message, and following its steps unblocks the next submit.

pkg/orchestrator/gke/storage.go
agrawalkhushi18: Went with the spec hash . The name suffix is now a short hash of the rendered PV spec: the same fields checkExistingGatewayPV compares (storage class, mount options, capacity, volume handle, volume attributes). claimRef is left out because it contains the name. The template is rendered once to get the hash, then again with the final names.
Added a test that pins the names, and a test where two custom templates that differ only in capacity get different names.

PR #6336, opened 2026-09-16
Make LNet runner non-fatal and conditional on Lustre

vikramvs-gg:
**Issue being fixed:**
Deploying the A4X-Max slurm blueprint with NFS/Filestore instead of Managed Lustre aborts compute node setup. Nodes then appear idle to Slurm while being unusable. The blueprint hardcodes a `configure_lnet_early.sh` runner as the first entry in the A4X-Max nodeset startup bundle. On a node with no Lustre, `lnetctl lnet` configure exits 254.

**Fixes made:**
- New `lustre_in_use()` runs before any LNet work. Lustre in use → configure LNet exactly as before. Not in use → log the reason and exit 0, without loading kernel modules or calling lnetctl at all.
- `$IP` escaping fixed: The eth0 wait loop now tests the real IP value instead of a literal string, so it genuinely waits for an address and its empty-IP guard is reachable.
- Relevant warning logging added when runner fails at certain point.
- Runner moved from first to last. LNet is optional; nothing should sit behind it.

**Testing:**
The runner was validated offline against the exact script `gcluster create` generates:
- **Blueprint expansion:** expands cleanly with `homefs` set to `managed-lustre` and with `homefs` swapped to `filestore` (no runner edits required).
- **Static checks:** `bash -n` and `shellcheck -S warning` clean on both generated variants; `pre-commit` passes.
- **Simulated node runs:** extracted the generated `configure_lnet_early.sh` and ran it against stubbed node states — NFS-only node (the reported failure), healthy Lustre node, `modprobe` failure, `lnetctl` returning 254, and missing `eth0` IP. All exit 0; `lnetctl` is never called on NFS. Running the *currently shipped* script through the same harness fails all cases and reproduces the exit 254 abort.


tools/validate_configs/golden_copies/configs/text_escape.yaml
```diff
@@ -31,6 +31,10 @@ deployment_groups:
     settings:
       labels:
         brown: \$(fox)
+        # A backslash escapes only "$(". Before a plain variable, it is not
+        # consumed: "\$IP" is emitted as "\$IP", not as "$IP".
+        escaped_var: \$IP
```
Neelabh94, line 36: The Google Cloud REST API strictly enforces that label keys and values may only contain lowercase letters, numeric characters, underscores, and dashes. The values `\$IP` and `\${HOME}` contain invalid characters (backslash, dollar sign, curly braces) and uppercase letters, which will cause Terraform to fail at apply-time with a 400 Bad Request. Consider moving these test cases to `metadata` instead of `labels`.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
```diff
@@ -783,6 +745,80 @@ deployment_groups:
                 name: nvidia-persistenced.service
                 state: started
                 enabled: true
+      # Placed LAST: LNet is optional, so nothing else should wait on it.
```
Neelabh94: Moving this runner to the end of the list causes `homefs.mount_runner` to execute before LNet is configured on `eth0`. Mounting Lustre implicitly starts LNet using default settings, making it 'busy' when this script later attempts to unconfigure or reconfigure it. This can lead to hanging mounts or sub-optimal storage networking. It should be placed between `homefs.client_install_runner` and `homefs.mount_runner`.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
```diff
@@ -783,6 +745,80 @@ deployment_groups:
...
+          export PATH=$PATH:/usr/sbin:/sbin
+
+          log() { echo "configure_lnet: $*"; }
+
+          # Both managed-lustre and filestore export network_storage.fs_type,
+          # so switching storage backends needs no edits to this runner.
+          BLUEPRINT_FS_TYPE="$(homefs.network_storage.fs_type)"
+          log "storage mode (blueprint): $BLUEPRINT_FS_TYPE"
+
+          lustre_in_use() {
```
Neelabh94: The `lustre_in_use()` function introduces speculative system-level checks (like checking `/proc/filesystems`) that can lead to false positives if the Lustre kernel module is pre-loaded on the image, even if no Lustre filesystem is actually mounted in this deployment. Since the blueprint explicitly knows the storage type via `$(homefs.network_storage.fs_type)`, this check can be simplified to rely strictly on that variable to reduce complexity and potential errors.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
```diff
@@ -783,6 +745,80 @@ deployment_groups:
...
+              log "No Lustre filesystem configured for this node; skipping LNet configuration."
+              exit 0
+          fi
+          log "Lustre in use. Configuring LNet on eth0."
+
+          if [ ! -e /dev/lnet ]; then
+              log "/dev/lnet absent; loading lnet module."
+              sudo modprobe lnet || {
+                  log "WARNING: 'modprobe lnet' failed; skipping LNet configuration."
+                  exit 0
```
Neelabh94: Exiting with `0` (silent bypass) here obscures critical failures when Lustre is confirmed to be in use. If `modprobe lnet` fails, `lnetctl` is missing, or the IP times out, the deployment will fail much later and more cryptically during the mount phase. It is better to fail fast with `exit 1` so the root cause is clear in the logs.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
```diff
@@ -783,6 +745,80 @@ deployment_groups:
...
+          fi
+
+          if ! command -v lnetctl >/dev/null 2>&1; then
+              log "WARNING: lnetctl not found; skipping LNet configuration."
+              exit 0
+          fi
+
+          IP=""
+          for _ in {1..30}; do
+              IP=\$(ip -4 addr show eth0 | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
```
Neelabh94: If `eth0` has multiple IPv4 addresses or secondary CIDRs assigned early in boot, `grep -oP` may emit multiple lines. While the variable check might handle this, logging it can mangle output, and downstream scripts expecting a single IP might break. Consider defensively extracting only the primary IP by appending `| head -n 1`.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
```diff
@@ -783,6 +745,80 @@ deployment_groups:
...
+              log "Waiting for eth0 to get an IP..."
+              sleep 1
+          done
+          if [ -z "$IP" ]; then
+              log "WARNING: eth0 did not get an IP in 30s; skipping LNet configuration."
+              ip addr
+              exit 0
+          fi
+
+          if sudo lnetctl net show 2>/dev/null | grep -q "tcp"; then
```
Neelabh94: Using `grep` without word boundaries (`-w`) here and for `eth0` below performs an unanchored substring match. This can lead to false matches (e.g., matching `veth0` instead of `eth0`), causing the script to falsely claim configuration is done. Hardening this with `grep -w` prevents potential idempotency issues.

Neelabh94 (commented):
Thank you for the PR. We have reviewed the changes and have some feedback to improve robustness and compatibility.

SwarnaBharathiMantena (commented):
LGTM - Please address Neelabh's comments

tools/validate_configs/golden_copies/configs/text_escape.yaml
vikramvs-gg: `text_escape.yaml` is a golden-copy test blueprint, `tools/validate_configs/golden_copies/validate.sh` only runs gcluster create and diffs the generated output. It never runs terraform init/plan/apply, so the label values never reach the API. The file intentionally contains invalid values for this reason: `brown: \$(fox)` in the same block and `ñred: ñblue` in vars.labels.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
vikramvs-gg: Moved back to its original position 1.
Note: `homefs.mount_runner` isn't in this list — homefs is attached to slurm_controller, not the nodeset and slurm-gcp mounts network storage in `setup_network_storage()` before `run_custom_scripts()`, so the mount precedes every runner either way. Position wasn't affecting this but I've updated it so that the diff reads better in place.

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
vikramvs-gg: Done

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
vikramvs-gg: The Lustre mount happens before this runner. slurm-gcp mounts network storage first, then runs the custom scripts, and then enables slurmd. Lustre mounts over LNet - so if `modprobe lnet` had failed, or `lnetctl` were missing, or eth0 had no address, the mount would already have failed and setup would have stopped there. There is no later mount phase left to fail.

Setup runs the scripts one by one and checks each exit code. If a script exits non-zero, `systemctl enable slurmd` never runs. The node still boots and Slurm still reports it `idle`, so jobs get sent to a node with no slurmd on it. That was the original failure we're fixing.For visibility I've made log() also write to syslog and switched these four messages from `WARNING:` to `ERROR:`

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
vikramvs-gg: Done

examples/machine-learning/a4x-maxgpu-4g-metal/a4xmax-bm-slurm-blueprint.yaml
vikramvs-gg: Done

Neelabh94 (approved):
LGTM

PR #6340, opened 2026-09-17
fix(kubectl-apply): use setunion instead of union for directory manifest filesets

SwarnaBharathiMantena:
## Summary
Fixes an invalid Terraform function call in [`modules/management/kubectl-apply/main.tf`](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/develop/modules/management/kubectl-apply/main.tf) when processing directory-based manifests (`apply_manifests`).

## Root Cause
When iterating over files discovered in a manifest directory, the module called `union(...)` to combine the set outputs of `fileset(manifest.source, "*.yaml")`, `fileset(manifest.source, "*.yml")`, and `fileset(manifest.source, "*.tftpl")`. However, `union()` is not a valid built-in Terraform function (`Call to unknown function: There is no function named "union"`). Terraform's built-in function for computing the union of sets is `setunion()`.

## Changes
- Replaced `union(` with `setunion(` in `local.processed_apply_manifests_map` (`modules/management/kubectl-apply/main.tf`).

## Testing & Verification
- Verified `terraform fmt`, `terraform validate`, and `tflint` pass on `modules/management/kubectl-apply`.
- Verified all repository pre-commit hooks (`terraform-readme`, `addlicense`, `codespell`, etc.) pass.


agrawalkhushi18 (approved):
LGTM

